Issue #18, opened 2019-10-17
datamodel: create datacite model based on zenodo


ppanero:
# InvenioRDM data model

- [DataCite JSON Schema](https://github.com/datacite/schema/blob/master/source/json/kernel-4.3/datacite_4.3_schema.json) and [PDF version](https://schema.datacite.org/meta/kernel-4.3/doc/DataCite-MetadataKernel_v4.3.pdf).
- Translate all camel to snake case.
- Let's shorten common structures like schemeURI/valueURI -> scheme/value

## Metadata fields 
- [x] Resource Type (resourceTypeGeneral)
    - ``{'resourceTypeGeneral': ..., 'resourceType': ...}`` -> ``{'type': ..., 'subtype': ...}``
    - Cannot use the CV from DataCite.
    - We should have a mapping from internal types to DataCite types
    - Two level type: *Type* and *Subtype* (see Zenodo implementation).
- [ ] Identifiers:
    -  wait
- [ ] Creators: See https://github.com/inveniosoftware/invenio-rdm-records/issues/19
    - Use DataCite
- [x] Titles
    - ``title`` (string field)
    - ``additional_titles`` (like DataCite ``titles``)
- [ ] Publisher:
    - Use DataCite
    - (to be discussed in detail)
- [x] Publication Year:
    - Use ``publication_date`` instead. Must be a valid date. In the future we can support extended time format or similar to represent only years.
- [ ] Subjects
    - Use DataCite
    - To be investigated if we can simplify
- [ ] Contributors: See https://github.com/inveniosoftware/invenio-rdm-records/issues/19
    - Use DataCite
- [x] Dates
    - Use DataCite (and see Zenodo implementation)
    - Special dates: ``publication_date``, ``embargo_date``.
    - How to deal with Embargo Date?
    - Check Zenodo for description text
- [x] Language
    - Use DataCite
- [ ] Alternate and Related Identifiers:
    - Use DataCite
    - Issues with naming and metadata schema
    - Issue with resourceTypeGeneral
- [ ] Sizes
    - Use DataCite
    - sizes. Not present in Zenodo, followed DataCite.
       **Questions**: 
       - Shall we separate the unit (e.g. MB) from the number? This would be helpful for sorting/filtering.
       - Shall it be only one size in the file? Do we want metadata sizes too?
- [ ] Formats
    - Use DataCite
       **Questions**: 
[...]

Issue #5, opened 2019-10-01
customization: resource types


ppanero:
Support custom resource types. There are different approaches:

- Add the list as a configuration variable in `invenio.cfg`.
- Create a command in the CLI (in *rdm-records* or in the *scritps*) in order to allow: `load`, `search`, `removal`, and `edit` of the types.

Note: We must consider the maximum size we expect this list to grow, as in the second case we need to store them in DB. Consider the use of controlled vocabularies.

ppanero:
Note the comments about different types, check [hierarchical](https://github.com/galterlibrary/InvenioRDM-at-NU/blob/master/cd2h_repo_project/modules/records) of Galter Lib.

PR #21, opened 2019-10-25
datamodel: add non-issue fields based on zenodo/datacite

ppanero:
**Fields**:
- Resource type
- additional titles
- additional descriptions
- dates (Following Zenodo implementation, not DataCite)
- publication date
- embargo date
- language: Using ISO 639-3 (as Zenodo) instead of 639-1 (as DataCite). It's newer and uses 3 letters instead of 2.
- version
- rights_list

**Validation**:
- All *lang* attributes be validated and follow the ISO 639-3
- All *enums* are validated against their list of values in Marshmallow instead of the JSON Schema. This allows to define configurable control vocabularies.

Addreses part of https://github.com/inveniosoftware/invenio-rdm-records/issues/18

### DIFF
--- a/invenio_rdm_records/data/objecttypes.json
+++ b/invenio_rdm_records/data/objecttypes.json
@@ -1,0 +1,678 @@
+[{
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "publication",
+    "id": "https://zenodo.org/objecttypes/publication#",
+    "title": {
+      "en": "Publication"
+    },
+    "title_plural": {
+      "en": "Publications"
+    },
+    "schema.org": "https://schema.org/ScholarlyArticle",
+    "datacite": {"general": "Text"},
+    "eurepo": "info:eu-repo/semantics/other",
+    "children": [
+      {"$ref": "https://zenodo.org/objecttypes/publication/book"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/section"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/conferencepaper"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/article"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/patent"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/preprint"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/report"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/softwaredocumentation"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/thesis"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/technicalnote"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/workingpaper"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/other"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/datamanagementplan"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/annotationcollection"},
+      {"$ref": "https://zenodo.org/objecttypes/publication/taxonomictreatment"}
+    ],
+    "openaire": {
+      "resourceType": "0001",
+      "type": "publication"
+    },
+    "csl": "article"
+  },
+  {
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "poster",
+    "id": "https://zenodo.org/objecttypes/poster#",
+    "title": {
+      "en": "Poster"
+    },
+    "title_plural": {
+      "en": "Posters"
+    },
+    "datacite": {"general": "Text", "type": "Poster"},
+    "eurepo": "info:eu-repo/semantics/conferencePoster",
+    "schema.org": "https://schema.org/CreativeWork",
+    "openaire": {
+      "resourceType": "0004",
+      "type": "publication"
+    },
+    "csl": "graphic"
+  },
+  {
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "presentation",
+    "id": "https://zenodo.org/objecttypes/presentation#",
+    "title": {
+      "en": "Presentation"
+    },
+    "title_plural": {
+      "en": "Presentations"
+    },
+    "datacite": {"general": "Text", "type": "Presentation"},
+    "eurepo": "info:eu-repo/semantics/lecture",
+    "schema.org": "https://schema.org/PresentationDigitalDocument",
+    "openaire": {
+      "resourceType": "0004",
+      "type": "publication"
+    },
+    "csl": "speech"
+  },
+  {
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "dataset",
+    "id": "https://zenodo.org/objecttypes/dataset#",
+    "title": {
+      "en": "Dataset"
+    },
+    "title_plural": {
+      "en": "Datasets"
+    },
+    "datacite": {"general": "Dataset"},
+    "eurepo": "info:eu-repo/semantics/other",
+    "schema.org": "https://schema.org/Dataset",
+    "openaire": {
+      "resourceType": "0021",
+      "type": "dataset"
+    },
+    "csl": "dataset"
+  },
+  {
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "image",
+    "id": "https://zenodo.org/objecttypes/image#",
+    "title": {
+      "en": "Image"
+    },
+    "title_plural": {
+      "en": "Images"
+    },
+    "schema.org": "https://schema.org/ImageObject",
+    "datacite": {"general": "Image"},
+    "eurepo": "info:eu-repo/semantics/other",
+    "children": [
+      {"$ref": "https://zenodo.org/objecttypes/image/figure"},
+      {"$ref": "https://zenodo.org/objecttypes/image/plot"},
+      {"$ref": "https://zenodo.org/objecttypes/image/drawing"},
+      {"$ref": "https://zenodo.org/objecttypes/image/diagram"},
+      {"$ref": "https://zenodo.org/objecttypes/image/photo"},
+      {"$ref": "https://zenodo.org/objecttypes/image/other"}
+    ],
+    "openaire": {
+      "resourceType": "0025",
+      "type": "dataset"
+    },
+    "csl": "graphic"
+  },
+  {
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "video",
+    "id": "https://zenodo.org/objecttypes/video",
+    "title": {
+      "en": "Video/Audio"
+    },
+    "title_plural": {
+      "en": "Videos/Audio"
+    },
+    "datacite": {"general": "Audiovisual"},
+    "eurepo": "info:eu-repo/semantics/other",
+    "schema.org": "https://schema.org/MediaObject",
+    "openaire": {
+      "resourceType": "0033",
+      "type": "dataset"
+    },
+    "csl": "motion_picture"
+  },
+  {
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "software",
+    "id": "https://zenodo.org/objecttypes/software#",
+    "title": {
+      "en": "Software"
+    },
+    "title_plural": {
+      "en": "Software"
+    },
+    "datacite": {"general": "Software"},
+    "eurepo": "info:eu-repo/semantics/other",
+    "openaire": {
+      "resourceType": "0029",
+      "type": "software"
+    },
+    "schema.org": "https://schema.org/SoftwareSourceCode",
+    "csl": "article"
+  },
+  {
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "lesson",
+    "id": "https://zenodo.org/objecttypes/lesson#",
+    "title": {
+      "en": "Lesson"
+    },
+    "title_plural": {
+      "en": "Lessons"
+    },
+    "datacite": {"general": "InteractiveResource"},
+    "eurepo": "info:eu-repo/semantics/lecture",
+    "schema.org": "https://schema.org/CreativeWork",
+    "openaire": {
+      "resourceType": "0010",
+      "type": "other"
+    },
+    "csl": "speech"
+  },
+  {
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "other",
+    "id": "https://zenodo.org/objecttypes/other#",
+    "title": {
+      "en": "Other"
+    },
+    "title_plural": {
+      "en": "Other"
+    },
+    "datacite": {"general": "Other"},
+    "eurepo": "info:eu-repo/semantics/other",
+    "schema.org": "https://schema.org/CreativeWork",
+    "openaire": {
+      "resourceType": "0020",
+      "type": "other"
+    },
+    "csl": "article"
+  },
+  {
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "publication-book",
+    "id": "https://zenodo.org/objecttypes/publication/book#",
+    "title": {
+      "en": "Book"
+    },
+    "title_plural": {
+      "en": "Books"
+    },
+    "schema.org": "https://schema.org/Book",
+    "datacite": {"general": "Text", "type": "Book"},
+    "eurepo": "info:eu-repo/semantics/book",
+    "parent": {"$ref": "https://zenodo.org/objecttypes/publication"},
+    "openaire": {
+      "resourceType": "0002",
+      "type": "publication"
+    },
+    "csl": "book"
+  },
+  {
+    "$schema": "https://zenodo.org/schemas/records/objecttype-v1.0.0.json",
+    "internal_id": "publication-section",
+    "id": "https://zenodo.org/objecttypes/publication/section#",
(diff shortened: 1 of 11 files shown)


invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -123,19 +123,152 @@
...
+        },
+        "required": [
+          "type"
+        ],
+        "type": "object"
+      },
+      "type": "array"
+    },
+    "description": {
+      "description": "Description/abstract for record.",
```
fenekku: Librarians on our side favour just "Description for record." here since "abstract" will go in the `"additional_descriptions"` section.

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -123,19 +123,152 @@
...
+      "type": "string"
+    },
+    "embargo_date": {
+      "description": "Embargo date of record (ISO8601 formatted date).",
+      "title": "Embargo Date",
```
fenekku: remove "title"

Do we keep string with format "date-time" or switch to `"type": "datetime"` ?

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -123,19 +123,152 @@
...
+      "type": "string",
+      "format": "date-time"
+    },
+    "keywords": {
+      "description": "Free text keywords.",
+      "items": {
+        "type": "string"
+      },
+      "type": "array"
+    },
```
fenekku: comment: We will revisit later.

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -123,19 +123,152 @@
...
+    "language": {
+      "description": "Primary language of the resource. ISO 639-3 language code.",
+      "type": "string"
+    },
```
fenekku: minor but helpful: `"maxLength": 3`

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -123,19 +123,152 @@
...
+      "description": "List of user IDs that are owners of the record.",
+      "items": {
+        "type": "number"
+      },
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true
+    },
+    "publication_date": {
+      "description": "Record publication date (IS8601-formatted). EDTF support to be added for field.",
```
fenekku: minor: EDTF for others too or just this one?

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -123,19 +123,152 @@
...
+      "type": "string",
+      "format": "date-time"
+    },
+    "recid": {
+      "description": "Invenio record identifier (integer).",
+      "type": "number"
```
fenekku: normal: `"type": "string"` now since we will be having alphanumeric identifiers.

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -123,19 +123,152 @@
...
+    },
+    "resource_type": {
+      "additionalProperties": false,
+      "description": "Record resource type.",
+      "properties": {
+        "openaire_subtype": {
+          "description": "OpenAIRE-specific resource type.",
+          "type": "string"
+        },
```
fenekku: normal: I think `openaire` should fit inside a generic resource type which we can deal with in November with the customization work.

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -123,19 +123,152 @@
...
+        "type": {
+          "default": "publication",
+          "description": "General resource type.",
+          "type": "string"
+        }
+      },
+      "required": [
+        "type"
```
fenekku: minor: might as well require "subtype" too!

invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
```diff
@@ -21,32 +21,34 @@
...
-        "description": {
-          "type": "text"
-        },
-        "owners": {
-          "type": "integer"
-        },
-        "keywords": {
-          "type": "keyword"
+        "additional_descriptions": {
+          "type": "object",
```
fenekku: normal: It will be important to think if we want "object" or "nested" for these types.

invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
```diff
@@ -81,11 +83,100 @@
...
+          "copy_to": "suggest_title"
+        },
+        "version": {
+          "type": "keyword"
+        },
         "_created": {
           "type": "date"
         },
         "_updated": {
           "type": "date"
```
fenekku: These were deleted from the jsonschema and should be deleted here too, right?

invenio_rdm_records/marshmallow/json.py
```diff
@@ -36,34 +42,235 @@ class PersonIdsSchemaV1(StrictKeysMixin):
...
+
+    @validates('title_type')
+    def validate_title_type(self, value):
+        """Validate that the title type is one of the allowed ones."""
+        if value not in self.TITLE_TYPES:
+            raise ValidationError(
+                _('Invalid title type. Not one of {allowed}.'.format(
+                                                    allowed=self.TITLE_TYPES)),
+                field_names=['additional_titles']
+            )
```
fenekku: Would this pattern be enough for "enums": 

`title_type = SanitizedUnicode(validate=validate.oneOf(TITLE_TYPES))` ?

invenio_rdm_records/marshmallow/json.py
```diff
@@ -36,34 +42,235 @@ class PersonIdsSchemaV1(StrictKeysMixin):
...
 
 class MetadataSchemaV1(StrictKeysMixin):
     """Schema for the record metadata."""
 
     # TODO: Check enumeration (i.e. only open/embargoed/... accepted)
     access_right = SanitizedUnicode(required=True)
     access = fields.Nested(AccessSchemaV1)
-    recid = PersistentIdentifier()
-    title = SanitizedUnicode(required=True, validate=validate.Length(min=3))
+    additional_descriptions = fields.List(fields.Nested(DescriptionSchemaV1))
```
fenekku: Would that be enough for what we need:

```suggestion
    additional_descriptions = fields.Nested(DescriptionSchemaV1, many=True)
```

invenio_rdm_records/marshmallow/json.py
```diff
@@ -36,34 +42,235 @@ class PersonIdsSchemaV1(StrictKeysMixin):
...
+    additional_titles = fields.List(fields.Nested(TitleSchemaV1))
+    contributors = Nested(ContributorSchemaV1, many=True, required=True)
+    dates = fields.List(
+        fields.Nested(DateSchemaV1), validate=validate.Length(min=1))
     description = SanitizedUnicode()
+    embargo_date = DateString()
+    keywords = fields.List(SanitizedUnicode(), many=True)
+    language = SanitizedUnicode()
```
fenekku:
```suggestion
    language = SanitizedUnicode(validate=validate_iso639_3)
```

invenio_rdm_records/marshmallow/json.py
```diff
@@ -36,34 +42,235 @@ class PersonIdsSchemaV1(StrictKeysMixin):
...
+    def validate_language(self, value):
+        """Validate that language is ISO 639-3 value."""
+        validate_iso639_3(value)
 
 
-# TODO: Use `RecordMetadataSchemaJSONV1` to inject PID in PUT/PATCH/...
 class RecordSchemaV1(StrictKeysMixin):
     """Record schema."""
 
+    # TODO: Use `RecordMetadataSchemaJSONV1` to inject PID in PUT/PATCH/...
```
fenekku: What does this mean? (maybe just cut / paste in which case, we can get rid of it)

invenio_rdm_records/marshmallow/utils.py
```diff
@@ -0,0 +1,23 @@
...
+"""Marshmallow utility functions."""
+
+import pycountry
+from flask_babelex import lazy_gettext as _
+from marshmallow import ValidationError
+
+
+def validate_iso639_3(value):
+    """Validate that language is ISO 639-3 value."""
+    if not pycountry.languages.get(alpha_3=value):
```
fenekku: nice!

invenio_rdm_records/models.py
```diff
@@ -0,0 +1,126 @@
...
+"""Models for Invenio RDM Records."""
+
+import json
+from os.path import dirname, join
+
+from elasticsearch_dsl.utils import AttrDict
+from jsonref import JsonRef
+
+
+class AccessRight(object):
```
fenekku: can be done later: This seems like a Jinja2 filter because it is really all about the frontend. Could be moved to `views.py` and adapted.

invenio_rdm_records/models.py
```diff
@@ -0,0 +1,126 @@
...
+        if 'subtype' in value:
+            if isinstance(value, AttrDict):
+                value = value.to_dict()
+            internal_id = "{0}-{1}".format(
+                value.get('type', ''),
+                value.get('subtype', '')
+            )
+        else:
+            internal_id = value['type']
+        return cls.get(internal_id)
```
fenekku: ... I am sorry ;). This is what it looks like in our instance: https://github.com/galterlibrary/InvenioRDM-at-NU/blob/master/cd2h_repo_project/modules/records/resource_type.py#L170 . We will come back to this (shelved)

invenio_rdm_records/data/objecttypes.json
```diff
@@ -0,0 +1,678 @@
...
+    "datacite": {"general": "Text", "type": "Taxonomic treatment"},
+    "eurepo": "info:eu-repo/semantics/other",
+    "parent": {"$ref": "https://zenodo.org/objecttypes/publication"},
+    "openaire": {
+      "resourceType": "0020",
+      "type": "publication"
+    },
+    "csl": "article"
+  }]
+
```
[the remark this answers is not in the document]
ppanero: I personally thing this might get into a configuration option later on. This ones are taken from Zenodo, they are here merely for testing purposes. I will try to move some of them just to the tests.

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -24,46 +24,51 @@
...
-        "type": "number"
+          "type": "object",
+          "properties": {
+              "description": {
+                "description": "Description/abstract for record.",
+                "type": "string"
+              },
+              "description_type": {
+                "description": "Type of description.",
+                "type": "string"
```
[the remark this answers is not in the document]
ppanero: There will be a further discussion on how to validate for all entry points in the same way. AFAIK ILS has similar problems. Would you prefer to add enums in the JSONschema until this has been solved, and we then proceed to remove them? We can live now with two validations, in the end, they are the same...

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
ppanero: I do not understand this part :)

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
ppanero: Do they always have description + abstract? I think there is the chance that some use the abstract as description, no?

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
ppanero: I do not find the JSONSchema `datetime` type. I am looking [here](https://json-schema.org/understanding-json-schema/reference/type.html) could you point me to the source?

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
ppanero: Taking into account this was created for bibliographic purposes, I only see the need to support it for the **publication**, not for the others that are more meta field for operational purposes (created, updated, embargo, etc.).

What do you think?

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
ppanero: Makes sense, DataCite requires both.

invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
ppanero: As object they are a bit more performant AFAIK, however is true that values are flatten and might produce false positives. I see `type` and `lang` will potentially be part of the filters (classic checkbox aggregators for example), however I dont see how a title might have different types for different languages. Does this makes sense in your use cases?

Since we have the mayor `description` field which will be shown these `additional_descriptions:description` might just be there for search purposes.

I would prefer to have `object` since its easier to work with (for queries, for example, Invenio query parser does not support nested queries yet, see [issue](https://github.com/inveniosoftware/invenio-search/issues/155))

invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
ppanero: Those are *system/operational* metadata, that does is not sent by the user (when creating a record). Therefore, is not in the JSONSchema that gets validated). However, it is added and sent to Elasticsearch. 

See Zenodo for reference:

- JSONschema: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/jsonschemas/records/record-v1.0.0.json
- ES7 mapping: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/mappings/v7/records/record-v1.0.0.json#L634

invenio_rdm_records/marshmallow/json.py
ppanero: Yes, I added also `error` to be able to use the custom message with i18 babel (`_`) that was there and might be helpful later on. 

Nice catch! thanks!

invenio_rdm_records/marshmallow/json.py
ppanero: I do not think is needed to be tackled now since identifiers are still in progress. However, I implemented [this](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/schemas/json.py#L102) because there was a case where:

1- User creates record
2- Somehow through the API the user specified a different `recid` in the body than the one in the URL, or does not specify it in the body.
3- It gets modified/deleted and ends in an inconsistent state.

The fix basically injects the URL pid in the body.

I am also unsure this might happen here since in the end we will always pass by *deposits*, that's why the `TODO`.

Sorry for the lack of context.

invenio_rdm_records/models.py
ppanero: Created a TODO :+1:

invenio_rdm_records/models.py
ppanero: I see, there's also the difference between JSON/CSV. I need to tackle this in customizations next month. Shall we treat this here https://github.com/inveniosoftware/invenio-rdm-records/issues/5 ?

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
ppanero: :+1:

invenio_rdm_records/marshmallow/json.py
ppanero: I think it is. However, I saw this issue https://github.com/marshmallow-code/marshmallow/issues/779 and I am quite reluctant to do any big change now until we fully migrate to Marshmallow 3 and that RFC gets accepted/rejected.

What do you think?

invenio_rdm_records/models.py
ppanero: Added a TODO

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
ppanero: You are right, removing...

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
ppanero: Talked IRL, removing "abstract"

ppanero: Talked IRL, removing "abstract"

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
ppanero: Discussed IRL, it's ok as it is.

fenekku: my bad, was thinking about `format`

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -24,46 +24,53 @@
...
-      "description": "Record title.",
       "type": "string"
     },
-    "description": {
-      "description": "Description/abstract for record.",
-      "type": "string"
-    },
-    "owners": {
-      "description": "List of user IDs that are owners of the record.",
+    "additional_descriptions": {
```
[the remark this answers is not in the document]
fenekku: It's because of the "convenient search constraint" coupled with the way we index documents:

1- convenient search: it is apparently known (or at least Zenodo and others want) that `description=mySearchTerm` can be entered in the search box to search on description. If a complex object is in that field that can't be done because...
2- naive indexing: as much as possible we naively index records from their jsonschema to their ElasticSearch mapping.

That said, this is "fixable" to achieve storing descriptions/titles in a single field and yet have convenient search keywords: building custom indexing hook for those, having a query parser module to map a passed query into our internal representation... We won't be getting to those by the end of the month so we will tackle it later. IMHO, the custom indexing hook seems like the simplest approach to achieve this. New issue for this: #24

PR #27, opened 2019-11-01
cli: create fake records

### LINKED ISSUE #26: demo: create and load records
As a hoster/developer I want to be able to test my instance with demo data.
- [x] Create demo dataset.
- [x] Create a script to load the data in the repository. (Note: A fixture approach might be easier)

fenekku:
Closes #26

### DIFF
--- a/invenio_rdm_records/cli.py
+++ b/invenio_rdm_records/cli.py
@@ -1,0 +1,72 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+# Copyright (C) 2019 Northwestern University,
+#                    Galter Health Sciences Library & Learning Center.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Command-line tools for demo module."""
+
+import uuid
+
+import click
+from faker import Faker
+from flask.cli import with_appcontext
+from invenio_db import db
+from invenio_indexer.api import RecordIndexer
+from invenio_pidstore import current_pidstore
+from invenio_records_files.api import Record
+from invenio_search import current_search
+
+
+def create_fake_record():
+    """Create records for demo purposes."""
+    fake = Faker()
+    data_to_use = {
+        "_access": {
+            "metadata_restricted": False,
+            "files_restricted": False
+        },
+        "access_right": "open",
+        "contributors": [{"name": fake.name()}],
+        "description": fake.bs(),
+        "owners": [1],
+        "publication_date": fake.iso8601(tzinfo=None, end_datetime=None),
+        "resource_type": {
+            "subtype": "dataset",
+            "type": "Dataset"
+        },
+        "title": fake.company() + "'s dataset"
+    }
+
+    # Create and index record
+    rec_uuid = uuid.uuid4()
+    current_pidstore.minters['recid'](rec_uuid, data_to_use)
+    record = Record.create(data_to_use)
+    RecordIndexer().index(record)
+
+    # Flush to index and database
+    current_search.flush_and_refresh(index='records')
+    db.session.commit()
+
+    return record
+
+
+@click.group()
+def rdm_records():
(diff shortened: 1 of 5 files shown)


invenio_rdm_records/cli.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+# Copyright (C) 2019 Northwestern University,
+#                    Galter Health Sciences Library & Learning Center.
+#
+# Invenio App RDM is free software; you can redistribute it and/or modify it
```
ppanero:
```suggestion
# Invenio RDM Records is free software; you can redistribute it and/or modify it
```

invenio_rdm_records/cli.py
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    # Create and index record
+    rec_uuid = uuid.uuid4()
+    current_pidstore.minters['recid'](rec_uuid, data_to_use)
+    record = Record.create(data_to_use)
+    RecordIndexer().index(record)
+
+    # Flush to index and database
+    current_search.flush_and_refresh(index='*')
```
ppanero: Minor: Here https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/config.py#L37 we have `records`, would it be safer to just flush and refresh the desired index? Or `*` takes the configured one?

setup.py
```diff
@@ -101,6 +102,9 @@
     include_package_data=True,
     platforms='any',
     entry_points={
+        'flask.commands': [
```
ppanero: Moderate: isn't it `console-scripts`? For example: https://github.com/inveniosoftware/invenio-cli/blob/master/setup.py#L77

Not sure if we want it under `invenio` or by itself

tests/conftest.py
```diff
@@ -29,3 +33,25 @@ def celery_config():
 def create_app():
     """Create app fixture."""
     return create_app
+
+
+@pytest.fixture
+def locations(db):
```
ppanero: Minor, since is for demo: I think this fixture is already available in pytest-invenio 1.2.0?  I do not see it documented but I see it [here](https://github.com/inveniosoftware/pytest-invenio/blob/master/pytest_invenio/fixtures.py#L560).

tho you have some extra stuff in here.

invenio_rdm_records/cli.py
fenekku: Thanks, used `Invenio-RDM-Records` like the others.

invenio_rdm_records/cli.py
fenekku: :heavy_check_mark: Switched to `'records'`. `*` takes all of them. `index='records'` may be a little quicker and safer like you say.

tests/conftest.py
fenekku: Thanks! I didn't know about the fixture.

setup.py
fenekku: `console-scripts` is for the name of the script by itself and `flask.commands` for the sub-commands under `invenio`. I think having it under `invenio` makes sense: we don't want to create multiple command-line tools (2 is enough :) ). With the current `setup.py` above, it works:

```
$ pipenv run invenio rdm-records fake
Creating demo records...
Created records!
$ curl -k -XGET https://localhost:5000/api/records/
{"aggregations":{"keywords":{"buckets":[],"doc_count_error_upper_bound":0,"sum_other_doc_count":0}},"hits":{"hits":[{"created":"2019-11-01T19:18:49.070162+00:00","id":"10","links":{"self":"https://localhost:5000/api/records/10"},"metadata":{"access_right":"open","contributors":[{"name":"Jason Collins"}],"description":"optimize leading-edge markets","owners":[1],"publication_date":"1978-03-30","recid":"10","resource_type":{"subtype":"dataset","type":"Dataset"},"title":"Rodriguez-Morgan's dataset"}...}
```

fenekku:
No problem. Changed to:

```console
pipenv run invenio rdm-records demo
```

PR #32, opened 2019-11-08
cli: demo records fix

lnielsen:


invenio_rdm_records/cli.py
```diff
@@ -44,7 +44,7 @@ def create_fake_record():
     # Create and index record
     rec_uuid = uuid.uuid4()
     current_pidstore.minters['recid'](rec_uuid, data_to_use)
-    record = Record.create(data_to_use)
+    record = Record.create(data_to_use, id_=rec_uuid)
```
fenekku: What was the issue here?

lnielsen: The impact was that you cannot access records individually via /api/records/1 or /records/1 (when the record is being resolved via pidstore in the database).

The underlying issue is that the PID maps to the ``rec_uuid``, but the ``Record.create`` will not use this ``rec_uuid`` as primary key for the record, but instead create a new random uuid. This causes the PID to point to non-existing uuid.

lnielsen:
I've already fast-tracked the PyPI release manually, so I'm merging this one.

PR #34, opened 2019-11-21
release: v1.0.0a5

fenekku:

PR #35, opened 2019-11-21
config: mint/fetch with 'cool DOI' ids

fenekku:
I guess I'd forgotten the actual integration of the new cool doi like id in the API :sweat_smile:


fenekku:
Merged via #34

PR #36, opened 2019-11-25
Release 1.0.0a6

fenekku:


invenio_rdm_records/marshmallow/json.py
```diff
@@ -148,9 +148,18 @@ class RightSchemaV1(StrictKeysMixin):
 class MetadataSchemaV1(StrictKeysMixin):
     """Schema for the record metadata."""
 
-    # TODO: Check enumeration (i.e. only open/embargoed/... accepted)
-    access_right = SanitizedUnicode(required=True)
-    access = fields.Nested(AccessSchemaV1)
+    ACCESS_RIGHT_CHOICES = [
```
ppanero: Thanks :)

tests/unit/test_schemas_json_load.py
```diff
@@ -238,3 +238,41 @@ def test_valid_rights(val, expected):
...
+        assert data['access_right'] == access_right
+    else:
+        assert 'access_right' in errors
+
+
+@pytest.mark.parametrize(('access', 'valid'), [
+    ({}, False),
+    ({'metadata_restricted': True}, False),
+    ({'files_restricted': False}, False),
+    ({'metadata_restricted': True, 'files_restricted': False}, True),
```
ppanero: Moderate:

``` python
 ({'metadata_restricted': True, 'files_restricted': False}, True)
```

Isn't this an invalid case? how can the record be restricted but the files not?

fenekku: Haha, this was just a random selection of True/False! I wasn't thinking about the meaning but just about the rules of the schema. Indeed this scenario should not occur like we talked about. Since this will be changed anyway, I am not too worried. I will change both to True though to be clear.

PR #39, opened 2019-12-05
config: use ext and only set defaults

fenekku:
Step in the right direction but ~~may be~~ is not enough yet. If it runs after invenio-records-rest config, that wouldn't do: I think we need to have invenio-app-rdm hard set the config from this module to close that loophole.

### DIFF
--- a/invenio_rdm_records/config.py
+++ b/invenio_rdm_records/config.py
@@ -130,16 +130,4 @@ RECORDS_FILES_REST_ENDPOINTS = {
 }
 """Set default files rest endpoints."""
 
-RECORDS_UI_ENDPOINTS = {
-    'recid': {
-        'pid_type': 'recid',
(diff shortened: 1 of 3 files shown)


fenekku:
Yup, like I said, on its own this is not enough (although I should not have said "may" in that first sentence, but rather "will" :) )

> we need to have invenio-app-rdm hard set the config from this module to close that loophole

That's the approach that I think could work best. We have invenio-rdm-records *define* the configuration it needs in its own `config.py` and `set_default`s it in its own `ext.py`. Then, invenio-app-rdm would *set *invenio-rdm-records'* `config.py` in its `config.py` (via config entrypoint). We split definition and setting for 2 reasons:

1- so we can still keep coupled things together and, in particular, 
2- test invenio-rdm-records with the correct configuration.

What do you think about that?

PR #41, opened 2019-12-10
dependencies: bump dependencies to comply with Invenio 3.2 

ppanero:
Changes:

- invenio-search
- invenio-db
- invenio-records-files
- invenio-records-rest

Requires https://github.com/inveniosoftware/invenio-records-permissions/pull/30
Ready for merge if all good (merge and release records-permissions).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,9 @@ tests_require = [
     'invenio_app>=1.2.3,<1.3.0'
 ]
 
-invenio_search_version = '1.2.0'
+# Should follow inveniosoftware/invenio versions
+invenio_db_version = '>=1.0.4,<1.1.0'
+invenio_search_version = '>=1.2.0,<1.3.0'
 
 extras_require = {
     'docs': [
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -69,10 +71,10 @@
     'arrow>=0.13.0',
```
fenekku: Do I recall correctly that we are pinning dependencies like this temporarily until we can rely on invenio>=3.2,<4.0 directly because we need specific packages at specific versions?

Another wacky idea: nothing prevents us from passing Pipfile.lock in cookiecutter, right? And then use that `Pipfile.lock`. t won't be as up-to-date necessarily, but it will freeze and only use known working dependencies so we won't get into problem of new versions breaking.

But for now, this is fine.

setup.py
```diff
@@ -69,10 +71,10 @@
...
     'Flask-BabelEx>=0.9.3',
     'invenio-jsonschemas>=1.0.0,<1.1.0',
-    'invenio-records-rest>=1.5.0,<1.6.0',
+    'invenio-records-rest>=1.6.0,<1.7.0',
     'invenio-records>=1.3.0,<1.4.0',
-    'invenio-records-files>=1.1.1,<1.2.0',
-    'invenio-records-permissions>=1.0.0a3',
+    'invenio-records-files>=1.2.0,<1.3.0',
+    'invenio-records-permissions>=1.0.0a5',
```
fenekku:
```suggestion
    'invenio-records-permissions>=1.0.0a4',
```

I will change and merge.

setup.py
ppanero: > Do I recall correctly that we are pinning dependencies like this temporarily until we can rely on invenio>=3.2,<4.0 directly because we need specific packages at specific versions?

I was actually thinking about this. I think these deps we are pinning might only be needed for tests?

>Another wacky idea: nothing prevents us from passing Pipfile.lock in cookiecutter, right? And then use that Pipfile.lock. It won't be as up-to-date necessarily, but it will freeze and only use known working dependencies so we won't get into problem of new versions breaking.

The main reason we cannot do that is because we would need to "parametrize" the lock file and matain like 6 lock files depending on ES and DB version/technology. Which is kind of hard due to hashes and so on.

PR #73, opened 2020-02-28
ui: fix if no identifiers + default icon

### LINKED ISSUE #69: Fix assumption that creators/contrbutors' identifiers are present in templates
If no `identifiers` field for creators/contributors, the record page breaks because they are expected.

fenekku:
closes #69

### DIFF
--- a/invenio_rdm_records/marshmallow/json.py
+++ b/invenio_rdm_records/marshmallow/json.py
@@ -377,12 +377,12 @@ class MetadataSchemaV1(BaseSchema):
 
     # Metadata fields
     identifiers = fields.List(fields.Nested(IdentifierSchemaV1),
-                              required=True)
+                              required=True)  # TODO: not required
     creators = fields.List(Nested(CreatorSchemaV1), required=True)
     titles = fields.List(fields.Nested(TitleSchemaV1), required=True)
     resource_type = fields.Nested(ResourceTypeSchemaV1, required=True)
     recid = PersistentIdentifier()
-    # No need for require since it's assign today's if not existent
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/theme/templates/invenio_rdm_records/details/list_people.html
```diff
@@ -0,0 +1,27 @@
+{#
+    Copyright (C) 2020 CERN.
+    Copyright (C) 2020 Northwestern University.
+
+    Invenio RDM Records is free software; you can redistribute it and/or modify
+    it under the terms of the MIT License; see LICENSE file for more details.
+#}
+
+{% macro list_creators_like(creators_like) %}
```
zzacharo: Silly question, why `creators_like`?

fenekku: Because it applies to list of creators *and* contributors (any object that conforms to the CreatorSchema interface)

zzacharo: I am not sure though that the `like` suffix would be enough to remember why...I would suggest to leave a comment there maybe? @ppanero WDYT?

fenekku: Would there be a different name that is more informative?  What about:

```jinja
{% macro list_creators_or_contributors(creators_or_contributors) %}
``` 

?

fenekku: Ended up going for `list_entities` : shorter

PR #75, opened 2020-03-05
extensions: allow metadata schema extension

fenekku:
- [x] marshmallow schema
- [x] jsonschema
- [x] elasticsearch 

Now that we are getting more and more into the data model, highlighting certain decisions will be helpful.

I have followed a similar pattern to Zenodo when it comes to the metadata extensions (what Zenodo calls 'custom metadata'). I've chosen to use `extensions` to be clearer I hope. 

Numbers, strings (and dates) are accepted and array of them as well. At the jsonschema level, I am differing from Zenodo by allowing arrays or straight up values to be stored. Zenodo only accepts arrays of Numbers/strings from the start. My thought was to be more expressive / map intent more clearly. (honestly don't know if this might make life harder in the long run).


ntarocco:
Also interested to this because we will have to do the same in [ILS](https://github.com/inveniosoftware/invenio-app-ils), ping @kprzerwa

Herrner:
Because I think this relates, I'm just bringing this back into attention:

https://invenio-talk.web.cern.ch/t/harvesting-custom-metadata-through-oai-pmh/43

invenio_rdm_records/ext.py
```diff
@@ -24,6 +25,13 @@ def __init__(self, app=None):
     def init_app(self, app):
         """Flask application initialization."""
         self.init_config(app)
+        self.metadata_extensions = MetadataExtensions(
+            app.config['RDM_RECORDS_METADATA_EXTENSIONS']
+        )
+        before_record_index.connect(
```
lnielsen: The receiver will receive all calls for indexing, so right now it will work but as soon as we have other records type this function will be called for them as well. The ``dynamic_conncet()`` will connect based on the ``$schema``

invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -6,6 +6,10 @@
   "additionalProperties": false,
 
   "properties": {
+    "$schema": {
+      "description": "This record's jsonschema. Needed?",
```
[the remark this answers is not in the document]
fenekku: Indeed, it's needed.

invenio_rdm_records/metadata_extensions.py
```diff
@@ -0,0 +1,149 @@
...
+                .get(_type)
+        )
+
+
+def add_es_metadata_extensions(record_dict):
+    """Add 'extensions_X' fields to record_dict prior to Elasticsearch index.
+
+    :param record_dict: dumped Record dict
+
+    QUESTION: Should we even keep indexing 'extensions'?
```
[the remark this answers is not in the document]
fenekku: Yes.

invenio_rdm_records/ext.py
fenekku: Ah, very nice!

fenekku:
Thanks for those!

**Namespacing**: The need for a JSON-LD url for the group is interesting... Our biomedical extension, for instance, doesn't have a full URL since it's the amalgamation of the library's experience of running their repository for many years. In other cases, like a potential future need for Clinical Trial records, I think there is a URL for it. So questions:
- what happens if there is no such URL? (would it have to be created by the institution somehow?)
- Is doing it on a field by field basis an option, especially if section is created from amalgamation of official fields of some kind (which is not always the case)? In this case:
```
{"<grouping>": {"<field>": {"types": {...}, "url": "https://example.com/field/definition"}}
```
seems like a possibility.
- Finally, we could also wait for the JSON-LD serialization task to consider this.

**Naming**: Funnily enough, I went for `extensions` because `custom` caused confusion in conversations (clear one in mind with @ppanero): custom resource types and custom controlled vocabularies are a different feature than "custom" metadata fields. Customization (`custom`) typically implies the ability to override/remove/add fields. Extension typically implies only addition of fields. So I went for `extensions` (`metadata_extensions` was too long I thought). The latter also seems more understandable to the librarians I checked with...

**user experience**: with `invenio.cfg` being a Python file (of sorts) that dict can be defined in another file and imported if the hoster wants to... I think there is some discussion about how to structure the file system layout for these configurations that might need more room (authentication group was discussing this). We should definitely communicate these recommended practices for when the extension is extensive (pun intended).

lnielsen:
- Naming: 👍 
- UX: Ok, let's postpone until we know exactly what are the customisation workflows.

lnielsen:
**Namespacing**

JSON-LD serialisation is pretty important (and also needed by Zenodo) and I'd rather get it in straight away because it deals with core definition of metadata extensions. 

There's a couple of reasons: 
- Encourage adoption of community agreed standard vocabularies over self-invented.
- Stability: Once in these fields exists in records they should not be changed (because then you have to update all records including it).
- Communities: Each community may want to define their own fields, which should be selected from the instance's possible new fields. With standard vocabularies, you're more likely to be able to reuse fields between communities.
- Allow serialisation of custom metadata as first-class citizen into the REST API (i.e it's not hidden in a "extended" section.

In case you don't have a standard vocabulary you invent one for your instance (see ``z`` vocabulary in the example below). 

Another thing is that we also need titles for fields and namespaces. All in all I would go with a config like method 1 and 2 below.

```python

# METHOD 1
RDM_RECORDS_METADATA_EXTENSIONS = {
    'dwc': {
        '@context': 'http://rs.tdwg.org/dwc/terms/',
        'title': _('Darwin Core'),
        'fields': {
            'family': {
                'title': _('Family'),
                'marshmallow': SanitizedUnicode(),
                'elasticsearch': 'keyword',
            }
        }
    },
    'z':{
       '@context': 'https://zenodo.org/terms',
       'title': _('Zenodo Metadata'),
       'fields': {
           'department': {
                'title': _('Department'),
                'marshmallow': SanitizedUnicode(),
                'elasticsearch': 'text',
            },
       }
    },
}

# METHOD 2
RDM_RECORDS_METADATA_NAMESPACES = {
   'dwc': {
       '@context': 'http://rs.tdwg.org/dwc/terms/',
       'title': _('Darwin Core'),
   },
   'z':{
       '@context': 'https://zenodo.org/terms',
       'title': _('Zenodo Metadata'),
   },
}

RDM_RECORDS_METADATA_EXTENSIONS = {
    'dwc:family': {
        'title': _('Family'),
        'marshmallow': SanitizedUnicode(),
        'elasticsearch': 'keyword',
    },
    'z:department': {
        'title': _('Department'),
        'marshmallow': SanitizedUnicode(),
        'elasticsearch': 'text',
    },
    # ...
}
```

Personally I would go for method 2 because it's less compact. Also, I could imagine that in the deposit form, you might want to group several fields from different vocabularies in the same field set.


**Question**
If we were to add more complex types, how would that look like?

```
??
'dwc:complexStuff': {
        'title': _('Complex Stuff'),
        'marshmallow_schema': MyComplexSchema(),
        'elasticsearch': '????',
},
```

lnielsen:
Now after writing above thing, I was in need of a names for "custom field". I.e. what's the translations:

- Custom metadata -> Metadata extension
- Custom field -> ??

fenekku:
> Encourage adoption of community agreed standard vocabularies over self-invented.
  
-> The extensions section precisely seemed a way to finally let self-invented fields be placed. My take is we will never please everyone and a committee will never come up with new fields that are valuable: individual on-the-ground instances will and we have to give them the ability to add these fields so they can try them out. 

> Allow serialisation of custom metadata as first-class citizen into the REST API (i.e it's not hidden in a "extended" section.
  
-> ...hmmm this is what we do now :sweat_smile:

> Personally I would go for method 2 because it's less compact

-> definitely would have gone for more compact :)

> group several fields from different vocabularies in the same field set

-> this is why the per field url aspect appealed to me: 
    because there isn't a URL for a group, there is more flexibility in how the fields are placed in the UI
    I also understood groups to be arbitrary as opposed to schema-tied

> title part: 
  
-> I started off with the title, but then realized that `_()` can be (ab)used 
  to generate a human readable label for other languages *and* English: `_('<id>')` 
  so I thought might as well save us a field.

> Complex types

  -> The constraint about only accepting certain fields works
  to our advantage here... I don't think we should allow deeply complex additional fields.
  If we were, then we should more drastically change our approach to the schema definition:
  make its configuration for us developers of the module the exact same as the one for 
  hosters somehow: given the complexity of the dataflow, allowing such flexibility seems too fragile...  

- custom field -> additional field

Question:
  - Is the exact term "@context" necessary? Because it seems it's JSON-LD tied, I 
    am trying to think of a way to make it more general to these other serializations
    that might pop-up.  

We can continue this discussion via video chat perhaps :)

fenekku:
So another throw agains the configuration wall:

```python
# METHOD 3
RDM_RECORDS_METADATA_EXTENSIONS = {
    'dwc:family': {
        'json-ld': { 
          '@context': 'http://rs.tdwg.org/dwc/terms/'
          # if nothing else could be placed as value for json-ld
        },
        'marshmallow': SanitizedUnicode(),
        'elasticsearch': 'keyword',
    },
    'dwc:behavior': {
        'json-ld': { 
          '@context': 'http://rs.tdwg.org/dwc/terms/'  # same as above 
          # if nothing else could be placed as value for json-ld
        },
        'marshmallow': SanitizedUnicode(),
        'elasticsearch': 'text',
    },
   ...
}
```

fenekku:
@lnielsen @zzacharo @ppanero Now it's ready to be (re)-reviewed.

invenio_rdm_records/config.py
```diff
@@ -226,3 +226,85 @@ def _(x):
...
+"""Fields added to the metadata schema.
+
+Of the shape:
+
+.. code-block:: python
+
+    {
+        '<prefix1>:<field1>': {
+            'elasticsearch': '<allowed elasticsearch type>'
+            'marshmallow': '<allowed marshmallow type>'
```
lnielsen:
```suggestion
            'marshmallow': <allowed marshmallow type>
```

invenio_rdm_records/config.py
```diff
@@ -226,3 +226,85 @@ def _(x):
...
+        },
+        # ...
+        '<prefixN>:<fieldN>': {
+            'elasticsearch': '<allowed elasticsearch type>'
+            'marshmallow': '<allowed marshmallow type>'
```
lnielsen:
```suggestion
            'marshmallow': <allowed marshmallow type>
```

invenio_rdm_records/config.py
lnielsen: Since it's a class and not a string?

invenio_rdm_records/ext.py
```diff
@@ -24,6 +25,15 @@ def __init__(self, app=None):
     def init_app(self, app):
         """Flask application initialization."""
         self.init_config(app)
+        self.metadata_extensions = MetadataExtensions(
```
lnielsen: Shelved: Should it be plural? There are multiple fields, but it's it only one extension 😇 (trivial question)

invenio_rdm_records/metadata_extensions.py
```diff
@@ -0,0 +1,156 @@
...
+        self._validate()
+
+    def _validate_marshmallow_type(self, field_cfg):
+        """Make sure the Marshmallow type is one we support."""
+        def validate_basic_marshmallow_type(_type):
+            allowed_types = [
+                Bool, DateString, Integer, SanitizedUnicode
+            ]
+            assert any([
+                isinstance(_type, allowed_type) for allowed_type
```
lnielsen: Shelved: What about subclasses of any of the types?

invenio_rdm_records/config.py
fenekku: Yes! Thanks for catching this.

invenio_rdm_records/ext.py
fenekku: Good question :thinking: ... I will ask around. (will probably merge anyway...)

invenio_rdm_records/metadata_extensions.py
fenekku: I was thinking we should be conservative at first: only accept very clear fields and over time see if the needs for more refined fields does come about. To revisit when the need does show.

Issue #51, opened 2020-02-10
Validate marshmallow identifiers via idutils


fenekku:
We could use [idutils](https://github.com/inveniosoftware/idutils) to validate identifier scheme in the marshmallow schema.

Make sure that references to identifiers are made to their lowercase form.

There was talk about distinguishing identifiers for different uses / targets (people / orgs / objects)
    Should we (invenio-rdm-records) check for those appropriately / is that idutils jobs ?

fenekku:
Closing this in favour of #97.

Also assuming that it is invenio-rdm-records' responsibility to distinguish between people, organizations, and object identifiers (not idutils for now).

PR #77, opened 2020-03-12
datamodel: test/display organizational entity

### LINKED ISSUE #52: marshmallow: validate Organizational creators/contributors
If the contributor/creator is of type "Organizational" it should be checked that "affiliation", "given_name" and "family_name" are not filled in.

### LINKED ISSUE #61: ui: creator/contributor icon
In `contributors.html`:

``` jinja
<span class="text-muted" {% if creator.affiliations and creator.affiliations[0] %}data-toggle="tooltip" title="{{creator.affiliations[0].name}}"{% endif %}>{{creator.name}}</span>{% if not loop.last %}; {% endif %}
```

Could eventually get away with no ; and always have an icon (even if generic) next to a creator

### LINKED ISSUE #78: identifiers field (and subfields) as dict
The `identifiers` field is an array of scheme and identifier. 

Pros: 
- matches up well with a potentially reusable UI component that adds an element to a list.
- allows for preference among different identifiers (1 one is the preferred one to be displayed)  

Cons: (in contrast with `identifiers` as a dict of scheme key and identifier value)
- no structural enforcement of unique scheme / identifier
- no reason could not create a reusable UI component that could do the trick for dict approach
- no reason could not add preference as a field (but would be finiky to be honest)   

Perhaps we should switch it to a dict.

**After In video life (IVL, I am coining this :) ) conversation**

Use dict for identifiers in top level `identifiers` and creators/contributors `identifiers`. Other cases have more data. To be seen on a case by case basis.

fenekku:
closes #52
closes #78 
closes #61

### DIFF
--- a/invenio_rdm_records/marshmallow/json.py
+++ b/invenio_rdm_records/marshmallow/json.py
@@ -28,11 +28,32 @@ class CommunitySchemaV1(BaseSchema):
     secondary = fields.List(SanitizedUnicode())
 
 
-class IdentifierSchemaV1(BaseSchema):
-    """Extra/Alternate identifiers of the record."""
+# 'Fake' Identifiers Field
+def _not_blank(error_msg):
+    """Returns a non-blank validation rule with custom error message."""
+    return validate.Length(min=1, error=error_msg)
 
-    identifier = SanitizedUnicode(required=True)
-    scheme = SanitizedUnicode(required=True)
+
+def Identifiers():
+    """Returns a "fake" Identifiers field.
+
+    Field expects:
+
+        "<scheme1>": "<identifier1>",
+        ...
+        "<schemeN>": "<identifierN>"
+    """
+    return fields.Dict(
+        # scheme
+        keys=SanitizedUnicode(
+            required=True, validate=_not_blank(_('Scheme cannot be blank.'))
+        ),
+        # identifier
+        values=SanitizedUnicode(
+            required=True,
+            validate=_not_blank(_('Identifier cannot be blank.'))
+        )
+    )
 
 
 class AffiliationSchemaV1(BaseSchema):
@@ -61,7 +82,7 @@ class CreatorSchemaV1(BaseSchema):
             ))
     given_name = SanitizedUnicode()
     family_name = SanitizedUnicode()
-    identifiers = fields.List(fields.Nested(IdentifierSchemaV1))
+    identifiers = Identifiers()
     affiliations = fields.List(fields.Nested(AffiliationSchemaV1))
 
 
@@ -379,8 +400,7 @@ class MetadataSchemaV1(BaseSchema):
     _contact = SanitizedUnicode(data_key="contact", attribute="contact")
 
     # Metadata fields
-    identifiers = fields.List(fields.Nested(IdentifierSchemaV1),
-                              required=True)  # TODO: not required
+    identifiers = Identifiers()
     creators = fields.List(Nested(CreatorSchemaV1), required=True)
     titles = fields.List(fields.Nested(TitleSchemaV1), required=True)
     resource_type = fields.Nested(ResourceTypeSchemaV1, required=True)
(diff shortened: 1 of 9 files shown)


invenio_rdm_records/theme/templates/invenio_rdm_records/details/list_entities.html
```diff
@@ -8,20 +8,20 @@
...
-      {%- if main_identifier.scheme == "Orcid" %}
-      <a href="{{ main_identifier.identifier|pid_url('orcid') }}">
-        <img class="inline-orcid" src="{{ url_for('static', filename='images/orcid.svg') }}" />
-      </a>
-      {# TODO: Other potential main identifiers #}
-      {%- endif %}
-    {% else %}
-      {# TODO: Change for an organization icon when organizations supported #}
-      <i class="fa fa-user" aria-hidden="true"></i>
+    {# TODO: Is the first identifier given precedence or is "Orcid" given precedence? #}
```
lnielsen: Shouldn't identifiers be a dict instead of list? Meaning we cannot select the first one? For now until this is fixed, selecting the first makes sense.

```
{
'orcid': …
'gnd': …
}
```

fenekku: I thought `identifiers` were an array of scheme and identifier (for every place identifiers appear in the schema) to match up with the deposit page where one adds another identifier to a list (reusable front-end component). But they could be adding to a dict in the case where each entry must be unique per scheme... @ppanero do you remember if we had a specific reason in mind for going with array over dict? I made an issue to revisit this: https://github.com/inveniosoftware/invenio-rdm-records/issues/78  

Schema as array:
https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/marshmallow/json.py#L61

fenekku: After talk: Going for dict and doing #78 with this one.

fenekku:
@lnielsen this can be reviewed again.

Top-level `identifiers` field and the `identifiers` field in `creators` or `contributors` have been changed. Other instances where identifiers are listed have been left alone until we get to them (they have more fields than just scheme + identifier).

fenekku:
@lnielsen @zzacharo @ppanero **Now** it's ready to be (re)-reviewed.

invenio_rdm_records/marshmallow/json.py
```diff
@@ -28,11 +28,32 @@ class CommunitySchemaV1(BaseSchema):
...
+
+    Field expects:
+
+        "<scheme1>": "<identifier1>",
+        ...
+        "<schemeN>": "<identifierN>"
+    """
+    return fields.Dict(
+        # scheme
+        keys=SanitizedUnicode(
```
lnielsen: Minor: We should later (i.e. not blocking) ensure that the scheme is auto-detected via idutils library.

invenio_rdm_records/theme/templates/invenio_rdm_records/details/list_entities.html
```diff
@@ -8,20 +8,20 @@
...
+    {# Assume first identifier is given precedence #}
+    {% set main_identifier = entity.identifiers[0] if entity.identifiers and entity.identifiers[0] else {"scheme": ""} %}
+    {%- if main_identifier.scheme == "Orcid" %}
```
lnielsen: idutils deliver scheme names in all lowercase, so I would stay with all lowercase scheme names. Also means we can  just call ``scheme.lower()`` to ensure that the API eats whatever you might give it (``ROR``/``ror``/``Orcid``/``ORCID``/``orcid``).

invenio_rdm_records/marshmallow/json.py
fenekku: Yes and we have an issue for that: #51

invenio_rdm_records/theme/templates/invenio_rdm_records/details/list_entities.html
fenekku: Agreed. I am thinking of dealing with this in #51 .

PR #79, opened 2020-03-18
23 publication date edtf lvl0

### LINKED ISSUE #23: publication_date: support for EDTF lvl 0
~~Currently only `publication_date` is specified to support EDTF.~~

~~However, all the rest of fields have `date-time` as format. Should they only be of format `date`? `date` was only introduced in JSONSchema draft 7 (more info [here](https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times)).~~

Redefinition of this task ;)

As a depositor, I may be unsure about the `publication_date` of a record (e.g. sometime in World War II), but I want to convey this range as the publication date, because a publication date is needed to mint a DOI (DataCite) and it is more informative than nothing.

**technical implications**
Allow `publication_date` field to be an EDTF of lvl 0

- [x] marhsmallow
- [x] jsonschema
- [x] elasticsearch + create a lower end date for sorting purposes

fenekku:
closes #23

### DIFF
--- a/invenio_rdm_records/ext.py
+++ b/invenio_rdm_records/ext.py
@@ -7,7 +7,9 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 """DataCite-based data model for Invenio."""
+import time
 
+from edtf.parser.grammar import level0Expression
 from invenio_indexer.signals import before_record_index
 
 from . import config
@@ -54,14 +56,36 @@ class InvenioRDMRecords(object):
                 app.config.setdefault(k, getattr(config, k))
 
 
+def prepare_publication_date(record_dict):
+    """
+    Add search compatible publication date to record_dict prior to ES index.
+
+    This date is the lowest year-month-day date from the interval or (partial)
+    date. The regular publication_date is not in a format ES can use for
+    powerful date queries.
+
+    :param record_dict: dumped Record dict
+    """
+    print("record_dict", record_dict)
+    parser = level0Expression("level0")
+    date_or_interval = parser.parseString(record_dict['publication_date'])[0]
+    # lower_strict() is available for EDTF Interval AND Date objects
+    date_tuple = date_or_interval.lower_strict()
+    record_dict['_publication_date_search'] = time.strftime(
+        "%Y-%m-%d", date_tuple
+    )
+
+
 def before_record_index_hook(
         sender, json=None, record=None, index=None, **kwargs):
     """Hook to transform Deposits before indexing in ES.
 
     :param sender: The entity sending the signal.
     :param json: The dumped Record dict which will be indexed.
-    :param record: The correspondng Record object.
+    :param record: The corresponding Record object.
     :param index: The index in which the json will be indexed.
     :param kwargs: Any other parameters.
     """
-    add_es_metadata_extensions(json)  # mutates json
+    # All thee operations mutate the json
+    add_es_metadata_extensions(json)  # TODO: Change for prepare
+    prepare_publication_date(json)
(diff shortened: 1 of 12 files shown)


invenio_rdm_records/ext.py
```diff
@@ -54,14 +56,36 @@ def init_config(self, app):
...
+    """
+    Add search compatible publication date to record_dict prior to ES index.
+
+    This date is the lowest year-month-day date from the interval or (partial)
+    date. The regular publication_date is not in a format ES can use for
+    powerful date queries.
+
+    :param record_dict: dumped Record dict
+    """
+    print("record_dict", record_dict)
```
lnielsen: Leftover?

invenio_rdm_records/marshmallow/fields.py
```diff
@@ -0,0 +1,66 @@
...
+
+"""Custom provided fields."""
+from datetime import date
+
+from edtf.parser.grammar import ParseException, level0Expression
+from flask_babelex import lazy_gettext as _
+from marshmallow import fields, missing
+
+
+class EDTFLvl0DateString(fields.Str):
```
lnielsen: Comment: Personally I would write "Lvl" out in full instead all the places it's used (i.e. not only in this class name). 

``EDTFLvl0DateString`` -> ``EDTFLevel0DateString`` (or just ``EDTFDateString``)

lnielsen: Comment: Likely the marshmallow field could be moved to Invenio-Records-REST later to have a reusable field for other Invenio installations.

invenio_rdm_records/ext.py
```diff
@@ -54,14 +56,36 @@ def init_config(self, app):
...
     :param json: The dumped Record dict which will be indexed.
-    :param record: The correspondng Record object.
+    :param record: The corresponding Record object.
     :param index: The index in which the json will be indexed.
     :param kwargs: Any other parameters.
     """
-    add_es_metadata_extensions(json)  # mutates json
+    # All thee operations mutate the json
+    add_es_metadata_extensions(json)  # TODO: Change for prepare
+    prepare_publication_date(json)
```
lnielsen: Moderate: Perhaps this is easiest to discuss in a video chat. Should the "_publication_date_search" go 1) only to the index or 2) also to the record?

invenio_rdm_records/ext.py
fenekku: Yup. removing!

invenio_rdm_records/marshmallow/fields.py
fenekku: Ah good to know! I was restraining myself here. I'll be a little more liberal with my names.

zzacharo: I agree with being more explicit with names in general. I find it easier after to understand the purpose of the class/variable :)

fenekku: Sure! Made issue for it in invenio-records-rest: https://github.com/inveniosoftware/invenio-records-rest/issues/293

invenio_rdm_records/ext.py
fenekku: Discussed via chat. We will add it to the record as well because:
- it is needed for enough serialization cases (saves us from having to edtf parse each time)
- supports our effort to keep ES record and DB record aligned  

`post_load` approach will be used for now. We will revisit when the deposit flow is hammered out.

fenekku:
Thanks @lnielsen and @zzacharo for the schema comments: it turned out to be *much* more work than I thought, but it caught some hidden bugs! You can re-review now.

PR #82, opened 2020-03-23
datamodel: customize resource type vocab

### LINKED ISSUE #5: customization: resource types

### LINKED ISSUE #76: Validation errors don't get passed for resource_type
Typically when one creates a record with an incorrect field a validation error with helpful message is returned:

e.g. `'titles': [{'title': 'A Romans story', 'type': 'Otherss', 'lang': 'eng'}]` returns an errors field that includes `Invalid title type. Otherss not one of MainTitle, AlternativeTitle, Subtitle, TranslatedTitle, Other`

However when one provides an incorrect resource_type in either the type or subtype field no errors field is returned:

e.g.  `'resource_type': {'type': 'imagess', 'subtype': 'photo'}` only returns `{"status": 400, "message": "Validation error."}`

fenekku:
This PR proposes a general model for controlled vocabulary customization by taking the specific case of resource type customization. 

Essentially, for each controlled vocabulary, a csv file with some conveniences (a way to write hierarchies without repeating oneself which also translates in being able to store general type information once) for the subject experts is used to store the mapping information.

There are a couple of "concept" feedback areas in addition to code feedback:
- The default resource type mapping is placed in this repository, so that its mapping is readily available. A floated alternative was to place it in the cookiecutter-invenio-rdm. This was chosen mostly because it was easier to work in one repository and this way the default was encapsulated (other instances installing this module would get that default). It perhaps makes customization less evident though. Zenodo's `objecttypes.json` was used as the basis for the `resource_types.csv`.
- From discussions, it emerged that requiring a type but not a subtype by default allows us to cover the cases where there is only a type defined. However, **full disclosure**, if subtypes are defined for a type, then a subtype must be passed. This is different than what was discussed, because it is something librarians asked for. The mapping key becomes (`type`, `subtype`) with `subtype` potentially being `''` (the empty string).
- The approach is generalizable to any controlled vocabulary. One just needs to create an appropriate `<Controlled Vocabulary>Vocabulary` class.
- Templates / filters are included.
- Some work has been done to address error messages as per #76, but those should actually be revisited next month when dealing with the deposit page: we will want the error messages to have the right key.
 
closes #5 
closes #76


tests/api/test_schemas_json_load.py
```diff
@@ -366,6 +375,36 @@ def test_resource_type():
         ResourceTypeSchemaV1().load(invalid_subtype)
 
 
+def test_custom_resource_type(config, vocabulary_clear):
```
fenekku: Is there a fixture for modifying the application configuration for the duration of a single test? Forgetting to reset it at the end of the test was very annoying... :sweat:

zzacharo: In invenio-circulation we had created this context manager https://github.com/inveniosoftware/invenio-circulation/blob/master/tests/helpers.py#L26 to address this issue. You can see an example of usage [here](https://github.com/inveniosoftware/invenio-circulation/blob/c3fcc1a4674d5df24221e3c068e8e0c07db68016/tests/test_loan_base_constraints.py#L65). Just note that this is working for "swapping" one config key per time but can easily be extended to an object :)

invenio_rdm_records/vocabularies/resource_types.csv
```diff
@@ -0,0 +1,3 @@
+type,type_name,subtype,subtype_name,datacite_general,datacite_type,csl,openaire_resourceType,openaire_type,schema.org,notes
```
fenekku: Is `datacite_type` necessary? i.e. can it always be filled with the `subtype_name`?

tmorrell: DataCite doesn't have a controlled vocabulary for subtype, so in terms of validating the metadata it would be fine. Datacite does internally use subtypes for some metadata formatting (e.g. using "Dissertation" will result in BibTex with the correct item type). So having the mapping could be useful if you wanted to have different labels in the repository such as "Thesis" and map into "Dissertation" for DataCite.

fenekku: Good to know! Will keep those for now then.

invenio_rdm_records/config.py
```diff
@@ -308,3 +308,29 @@ def _(x):
...
+
+RDM_RECORDS_CUSTOM_VOCABULARIES = {}
+"""Paths to custom controlled vocabularies.
+
+Of the shape:
+
+.. code-block:: python
+
+    {
+        '<vocabulary1>': '<absolute path to CSV file containing it',
```
ppanero: Question: Does it have to be absolute? Could it work with relative? Im just thinking if we will have to edit the Dockerfile, hardcode the path or it would just work specifying a relative path.

invenio_rdm_records/theme/views.py
```diff
@@ -120,3 +121,18 @@ def doi_identifier(pids):
...
+
+@blueprint.app_template_filter('display_resource_type')
+def display_resource_type(resource_type):
+    """Returns formatted resource type string."""
+    vocabulary = Vocabulary.get_vocabulary('resource_types')
+    entry = vocabulary.get_by_dict(resource_type)
+
+    # NOTE: translations could also be done via the CSV file directly
+    result = _(entry.get('type_name'))
+    if entry.get('subtype_name'):
```
ppanero: Doubt: If (from last message IRL) we make mandatory subtype, it would always exist. Therefore, there is no need for an `if`. No?

invenio_rdm_records/vocabularies/__init__.py
```diff
@@ -0,0 +1,134 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+# Copyright (C) 2019 Northwestern University.
```
ppanero: Normal: 2020 😅 happens to me all the time too...

invenio_rdm_records/vocabularies/__init__.py
```diff
@@ -0,0 +1,134 @@
...
+            self.data = {
+                # NOTE: unfilled cells return '' (empty string)
+                (row['type'], row['subtype']): row
+                for row in hierarchized_rows(reader)
+            }
+
+    def get_by_dict(self, type_subtype):
+        """Returns a vocabulary entry as an OrderedDict."""
+        return self.data.get(
+            (type_subtype['type'], type_subtype.get('subtype', ''))
```
ppanero: Minor: links with above question. If there is no subtype (mandatory) we would want it to fail somehow?

invenio_rdm_records/config.py
fenekku: I was wondering about this too. Especially given that the file might be in these different locations (linked and so on)... 

I am doing a full test today and that might be enlightening. If you or others have recommendations for locations / ways to refer to the file, please share.

invenio_rdm_records/theme/views.py
fenekku: My understanding was that: we made the subtype mandatory last week and then yesterday during the call we reversed that decision because Zenodo has types without subtypes (e.g. poster). When @lnielsen takes a look he can clarify.

invenio_rdm_records/vocabularies/__init__.py
fenekku: Haha thanks! I had changed my new file template, but this file wasn't new, it was just the renamed models.py file. :sweat_smile:

invenio_rdm_records/theme/views.py
ppanero: Ah, I understood the decision was reverted once more after your metadata committee hehe sorry for the confusion.

invenio_rdm_records/config.py
lnielsen: We would need an easy way to provide new CSV files in the instance's folder (i.e. it should be possible to provide them and commit them to inveniosoftware/demo-inveniordm

ppanero: What about a "vocabularies" folder? same way we have templates and so on?

lnielsen: Would work yes.

fenekku: See what I did there: https://github.com/inveniosoftware/cookiecutter-invenio-rdm/pull/59 . Is that reasonable? This way we don't have to go back to modifying the cookiecutter and the cli everytime we want to add application data. `vocabularies/` would just be a folder in `app_data/`. 

I couldn't use `data/` because the name is already used for the local files storage folder...

invenio_rdm_records/vocabularies/__init__.py
fenekku: resolved. (we accept empty subtype)

invenio_rdm_records/theme/views.py
```diff
@@ -144,3 +145,18 @@ def doi_identifier(identifiers):
 def doi_locally_managed(pid):
     """Determine if DOI is managed locally."""
     return is_doi_locally_managed(pid)
+
+
+@blueprint.app_template_filter('display_resource_type')
+def display_resource_type(resource_type):
```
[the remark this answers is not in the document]
fenekku: Yes. Will probably address tomorrow.

invenio_rdm_records/theme/views.py
```diff
@@ -144,3 +145,18 @@ def doi_identifier(identifiers):
...
+    """Returns formatted resource type string."""
+    vocabulary = Vocabulary.get_vocabulary('resource_types')
+    entry = vocabulary.get_by_dict(resource_type)
+
+    # NOTE: translations could also be done via the CSV file directly
+    result = _(entry.get('type_name'))
```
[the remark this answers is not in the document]
fenekku: Yes. Will do that change with the one above.

invenio_rdm_records/vocabularies/resource_types.csv
```diff
@@ -0,0 +1,32 @@
+type,type_name,subtype,subtype_name,datacite_general,datacite_type,csl,openaire_resourceType,openaire_type,schema.org
```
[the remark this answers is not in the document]
fenekku: > UTF8 issues

I've asked @saragon02 to try out languages in Excel on that one. I'll see tomorrow how that fares.

> sorting

Excel experts might be able to double column sort... or they can still fill rows or use visual cues to sort manually... Not too worried. 

> translations

My thinking too, but I don't know (haven't investigated at all) how  to connect the translation system with a column selection. (maybe current_language as key...) If someone can show me, we could plug that in.

> IDs

Right, ids (type/subtype here) should never change but names (type_name, subtype_name) should be free to change. I can see how passing a number as id might dissuade people from changing the ID when they change the name...

fenekku:
> possibility to have custom classes

That's something I was wondering about: potentially using the same shape as the internal default vocabulary dictionary in the config dictionary and therefore allowing custom vocabulary classes too. But I think I'd rather be conservative for now. 

> But I think that can come as part of using it for other vocabularies (like the contributor roles).

yes I think that would be the next target to get through this design.

I will push the vocabulary_title changes early tomorrow.

Thanks for looking at this!

invenio_rdm_records/theme/views.py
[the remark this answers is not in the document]
fenekku: Addressed!

invenio_rdm_records/theme/views.py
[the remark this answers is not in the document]
fenekku: Done

ppanero:
Is it worth creating issues for edge cases? or at least pointing them in the RFC?

fenekku:
I think so... What did you have in mind?

PR #90, opened 2020-04-22
ui: dump empty record for formik [more]

fenekku:
This commit is a step in the right direction (should be merged), but we will likely face a problem down the road. Problem is below with potential solutions.

Formik requires at least a full record skeleton in initialValues
(structure required but empty field) for validation (and potentially form filling) to work.
AND
react-invenio-forms' ArrayFields requires an empty array item to
be able to display empty arrays' first item.

Combined those constraints made us generate an empty record.

However, the frontend sends all those empty fields back:
some fields that are empty might cause a validation error on the backend
even though they were untouched by the user
      -> bad user experience

We will need a solution for this either:
  * only send non-null fields from the frontend, OR
  * change ArrayField to not need an empty entry to display AND
    check that Formik can add array elements dynamically

This commit also makes some things uniform:
- refactors allowed controlled vocab Marshmallow test
- uses Nested from invenio-record-rest only
- passes classes to Nested

### DIFF
--- a/invenio_rdm_records/marshmallow/json.py
+++ b/invenio_rdm_records/marshmallow/json.py
@@ -20,6 +20,7 @@ from invenio_records_rest.schemas.fields import DateString, \
 from invenio_rest.serializer import BaseSchema
 from marshmallow import ValidationError, fields, post_load, validate, \
     validates, validates_schema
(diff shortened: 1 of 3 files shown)


fenekku:
PR #91 includes this work, so closing.

PR #91, opened 2020-04-23
Dump vocabularies

fenekku:
- dumps empty record for frontend
- dumps resource type vocabulary for frontend
  - register icons in the resource type file + order resource type according to file 

There will be refactoring when we generalize to all other custom vocabularies.

### DIFF
--- a/invenio_rdm_records/vocabularies/__init__.py
+++ b/invenio_rdm_records/vocabularies/__init__.py
@@ -10,7 +10,7 @@
 
 import csv
 import json
-from collections import defaultdict
+from collections import OrderedDict, defaultdict
 from os.path import dirname, join
 
 from flask import current_app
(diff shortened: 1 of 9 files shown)


invenio_rdm_records/vocabularies/__init__.py
```diff
@@ -106,37 +107,40 @@ def get_invalid(self, type_subtype):
...
+        adapt it to their needs easily).
 
         TODO: Be attentive to generalization for all vocabularies.
         """
-        text_values = {'type': [], 'subtype': []}
+        options = {'type': [], 'subtype': []}
 
         for (_type, subtype), entry in self.data.items():
             type_option = {
+                'icon': entry.get('type_icon'),
```
ppanero, line 123: Curiosity: This is for the UI icon right?

fenekku: Yes it's for the UI. As much as possible I want to concentrate a lot of these customizations there so that it's easy to change by everyone. Labels, icons and order have been asked to be customized frequently by librarians in my group.

PR #93, opened 2020-04-29
release: v0.11.0

fenekku:
How is the registration of the resources going @ppanero?  Let's only bump the version after you have done that or if you deem it too intertwined and would rather wait.

This is blocking merging of invenio-app-rdm since it depends on functionality provided by invenio-rdm-records.

### DIFF
--- a/invenio_rdm_records/version.py
+++ b/invenio_rdm_records/version.py
@@ -12,4 +12,4 @@ This file is imported by ``invenio_rdm_records.__init__``,
 and parsed by ``setup.py``.
 """
 
-__version__ = '0.10.1'
+__version__ = '0.11.0'


ppanero:
@fenekku I discussed a bit with Lars and Zach this morning. And we agreed to just do a separate branch for that work.

I personally think that having records as records-rest and as resources might be confusing and actually lead to errors.

Therefore, the integration work will be done, but in a separate branch. If you agree with this, just go ahead and release :) Thanks for asking tho, appreciated!

fenekku:
Understood. I think that makes sense given the fundamental change you are bringing.

I will merge this and release invenio-rdm-records.

PR #95, opened 2020-05-04
Communities integration

slint:


invenio_rdm_records/marshmallow/json.py
```diff
@@ -433,6 +433,7 @@ class MetadataSchemaV1(BaseSchema):
                                attribute="embargo_date")
     _community = Nested(CommunitySchemaV1, data_key="community",
                         attribute="community")
+    _communities = fields.Raw(data_key="_communities", attribute="_communities")
```
fenekku, line 436: [Discussion] Can this new field replace `_community` or is it complementary? I think it's complementary because it needs additional functionality before it can account for a record that was added and then removed just based on the elasticsearch mapping above. But I would like it to replace, because it would simplifiy communities by making all the community stuff go in one field (which can then be sub-divided).

Here is the structure I see:
 
`_communities` -> `_communities_statuses`

```python
class CommunitiesRequestV1(BaseSchema):
    """Community Request Schema."""

    id = SanitizedUnicode(required=True)  # difference between id and request_id ?
    community_id = SanitizedUnicode(required=True)  # comid
    title = SanitizedUnicode(required=True)
    request_id = SanitizedUnicode(required=True)
    created_by = fields.Integer(required=True)


class CommunityStatusV1(BaseSchema):
    """Status of a community request."""
    pending = fields.List(Nested(CommunitiesRequestV1))
    accepted = fields.List(Nested(CommunitiesRequestV1))
    rejected = fields.List(Nested(CommunitiesRequestV1))


class MetadataSchemaV1(BaseSchema):
    # ....
    _communities_statuses = Nested(CommunityStatusV1)
```

If that is the meaning behind these mapping objects.

invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
```diff
@@ -106,6 +106,71 @@
...
+        "_communities": {
+          "type": "object",
+          "properties": {
+            "accepted": {
+              "type": "object",
+              "properties": {
+                "id": {
+                  "type": "keyword"
+                },
+                "comid": {
```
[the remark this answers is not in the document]
slint: It's the `PersistentIdentifier.pid_value` of the community (similar to the `recid`). This is the "slug" part that e.g. today exists in Zenodo in `/community/<comid>`.

invenio_rdm_records/marshmallow/json.py
slint: I agree that it should be eventually fleshed out into a proper field with schemas, etc. Keep in mind that we also don't dump/store this information in the record's JSON and that's the reason there is no JSONSchema addition in this PR (more on that later down).

Its purpose for now in the integration is to serialize the [ES indexed value](https://github.com/inveniosoftware/invenio-communities/blob/17059f7f10ed5da6b21c0cfa6b2e2c8c83e20656/invenio_communities/records/indexer.py#L15) in the `/api/records` search response so that it can be used in the [curation page](https://zenodo-rdm.web.cern.ch/communities/foo/curate). Actually for completeness, I'll add a `Method` field that dumps this dictionary into the `_communities` field so that the item resource (`/api/records/<recid>`) also has the field serialized.

The basic idea for how we proceed with serialization is that the `RecordCommunitiesMixin` (added to the API Record class) adds a `record.communities` property (similar to `record.files`). You can then as part of `record.dumps()` do a `record.communities.dumps()` (or [`record.communities.as_dict()`](https://github.com/inveniosoftware/invenio-communities/blob/17059f7f10ed5da6b21c0cfa6b2e2c8c83e20656/invenio_communities/records/api.py#L351)) call to get the data that will then become the input to the Marshmallow serializer.

slint:
@fenekku I've added a Marshmallow schema and removed the old `_community` field. I'm not forgetting about the concept of a "primary" + "secondary" communities thought :)

I released invenio-communities v2.0.1 with ES v6 mappings support also so tests should be passing.

invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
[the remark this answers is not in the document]
fenekku: :heavy_check_mark: Ah ok. `comid` and `id` keep same pattern as for Record.

slint:
I see a strange behaviour in the tests... For record indexing we have a receiver which dumps the community information from the DB and assumes the record has a PID (`recid`). In the ES-related tests, I see that [the `$schema` field is added](https://github.com/inveniosoftware/invenio-rdm-records/blob/8564635e1ce96d23ee8515fd142d6e5c47ec8c32/tests/api/test_elasticsearch_mapping.py#L72) (so that the index can be inferred from the `RecordIndexer`), but that introduces the issue that the `recid` field (added by the minter) is not in the JSONSchema, where it goes under [the name `_recid`](https://github.com/inveniosoftware/invenio-rdm-records/blob/8564635e1ce96d23ee8515fd142d6e5c47ec8c32/invenio_rdm_records/jsonschemas/records/record-v1.0.0.json#L110). Should we change [the `PIDSTORE_RECID_FIELD` config](https://github.com/inveniosoftware/invenio-rdm-records/blob/8564635e1ce96d23ee8515fd142d6e5c47ec8c32/invenio_rdm_records/config.py#L213) to `PIDSTORE_RECID_FIELD = "_recid"`?

fenekku:
The fact that the tests are failing given I don't quite understand the comments above is good! Means I am (maybe) consistent in my omission of some crucial part to the data flow. It probably will elucidate some of my misunderstanding about it :)  

I am writing up something that I will post to give you my understanding so far and maybe you will be able to identify the parts we are missing. Thanks!

fenekku:
Ok understanding so far:

The dataflow is an area I have trouble understanding so bear with me ;) 

### This part I understand:

External world json data --> filtered by Marshmallow schema  --> nice record dict used by Record API (from api.py) that inherits from Record in invenio-record-files --> put in DB (jsonschema validates it) + put in ES (optional pre-indexing hook)

search query response <--- pagination/metadata(?) wrapper <--- optional search serialization transformation <-- raw ES index content

### Discussion part:

So there was this discussion about trying to make DB Record and ES record aligned: https://github.com/inveniosoftware/invenio-rdm-records/pull/79#discussion_r395118799
most of the time.

If we need transformations just for the index, we've been placing here:
https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/ext.py#L56

If transformations are needed for ES + DB record (and those are favoured per discussion above), we've been using the `@post_load`:
https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/marshmallow/json.py#L492-496

### In tests, we have 2 fixtures: "input record" and "plain record" 

- input record for external world json record
- plain record for already loaded nice record dict

Refactoring in tests are needed for where things should go:
- test_serializers should be merged with test_json_schema_dump
- test_jsonschema can be dropped since the test_elasticsearch_mappings cover it completely

the right fixture has to be used.

### added comments from new commits:

I thought we wanted a default community, so it must be part of the jsonschema no?

tests/api/test_schemas_json_load.py
```diff
@@ -15,36 +15,14 @@
...
-    data = CommunitySchemaV1().load(valid_primary)
-    assert data == valid_primary
-
-    invalid_no_primary = {
-        "secondary": ["Secondary Community"]
-    }
-    with pytest.raises(ValidationError):
-        CommunitySchemaV1().load(invalid_no_primary)
-
-
```
fenekku, line 47: Still need to be able to account for community choice coming from external world, right?

slint:
> So there was this discussion about trying to make DB Record and ES record aligned: [#79 (comment)](https://github.com/inveniosoftware/invenio-rdm-records/pull/79#discussion_r395118799)
> most of the time.

In general, I disagree with storing computed values in the record's JSON in the DB. Once e.g. the processing logic changes, this implies data migration for all records. E.g. I would consider the `_publication_date_search` a "computed" field, and thus only add it during indexing.

There are two other aspects to computed fields though:

- Dependencies. i.e. the computed field is generated from...
  - ...other record fields (e.g. `publication_date -> _publication_date_search`)
  - ...an entirely other DB model/object (e.g. `CommunityRequest.query(...) -> _communities`)
- Usage. i.e. a computed field may be used for... 
  - ...searching, e.g. `_publication_date_search` is actually only useful for searching dates
  - ...serializing information, e.g. `_communities.pending.{request_id,comid}` is used from both searching for records pending inclusion in a community, and to know how to accept/reject the request)

> If we need transformations just for the index, we've been placing here:
> https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/ext.py#L56

In our case `_communities` is not an ES-only transformation, we include it also in the `GET /api/records/<recid>` response.

Besides that, I fully agree with this, ES-only transformation should take place in indexer-receiver functions. This of course, contradicts the existence of the [communities indexer receiver](https://github.com/inveniosoftware/invenio-communities/blob/c5281bca632d69f1e1c74c9a413efb7cbf768637/invenio_communities/records/indexer.py#L15), which is generating a `_communities` field that is used for both searching and serializing. I believe there's a better solution, described below.

> If transformations are needed for ES + DB record (and those are favoured per discussion above), we've been using the `@post_load`:
> https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/marshmallow/json.py#L492-496

The limitation with `@post_load` is that it happens only during input, i.e. when a record is created and goes through the loader via the REST API or programmatic/shell. If we want to change fields in the future, we have to "re-ingest" all of the records to generate these fields so that the `@post_load` runs or manually run whatever function we call in a "migrator".

### So what do we do?

- For computed fields that are used **only for searching** use the indexer receiver mechanism and skip them during serialization.
- For computed fields that are used for **serialization** include them in the API class [`Record.dumps()` method](https://github.com/inveniosoftware/invenio-indexer/blob/34489ddc5f3942c89bfa66e2a7527e3b72a99f71/invenio_indexer/api.py#L336), which will place them in the indexed JSON. Fun-fact: invenio-indexer v1.2.1 allows you to override the [record class of the RecordIndexer](https://github.com/inveniosoftware/invenio-indexer/blob/34489ddc5f3942c89bfa66e2a7527e3b72a99f71/invenio_indexer/api.py#L53), thus you have now control of what is the starting state of your `json` before it's shipped off to the indexer receiver signal. Then...
  - ...if they're needed for searching define them properly in the record's ES mapping, otherwise mark them with [`enabled: false`](https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html), so you don't waste ES resources. If they need extra transformation for ES indexing (e.g. transform to date), define an indexer receiver for them (and of course include them in the ES mapping accordingly)
  - ...if they need to be serialized, you can safely assume the in the marshmallow serializer that they field will be in the original input, since both the [item](https://github.com/inveniosoftware/invenio-records-rest/blob/c4185c4eac40ccd8c9b532b2ddccce9f91aa63a8/invenio_records_rest/serializers/base.py#L143-L144) and [list/search](https://github.com/inveniosoftware/invenio-indexer/blob/34489ddc5f3942c89bfa66e2a7527e3b72a99f71/invenio_indexer/api.py#L336) input come actually from `Record.dumps()`.

> I thought we wanted a default community, so it must be part of the jsonschema no?

Yes, I should move this into an issue though. The implementation reasoning was that it's easier to implement the `1 record <---> N communities` first and eventually implement the feature to mark one of the `N communities` as being the primary/default.

fenekku:
TLDR; Let's have a chat about this if you are available @slint :)

High-level questions I have are:
- What is the responsibility of Record in api.py (when does it show up other than at record creation?)?
  * What does it mean to bring Record in the Marshmallow class? I can't wrap my head around that.
  * What is the role of dumping in the Marshmallow class, if serializers add a layer and `Record.dumps` adds a layer?
  * Can the concept of "computed fields" be implemented in a centralized fashion? 
- Can `record_communities(record).as_dict()` in the sidebar.html + accompanying react app show the communities a record belongs to just from the CommunityRequest and their status as it is in the PR? or is there additional work needed? (yes for primary/secondary but that can wait for another PR)

Detailed point by by point response: 

> In general, I disagree with storing computed values in the record's JSON in the DB. Once e.g. the processing logic changes, this implies data migration for all records. E.g. I would consider the _publication_date_search a "computed" field, and thus only add it during indexing.

I would tend to agree with the computed field; it's how it originally was. However, if only adding a computed field during indexing, once the processing logic changes, we still need to do a migration of sorts but on the index (re-index) for all records. I gather this kind of migration is less trouble than a DB migration? The second and deeper reason we went the other way is it makes the index content and the db content diverge requiring additional *patchwork* code at the API and the backend representation. To me it is this patchwork aspect that is the most problematic: it is not tightly coupled/centralized. So:
- unrelated changes might break these related pieces of code,
- needed related changes will be omitted,
- duplicate DB and index code and / or attempt at squishing it out will be awkward
- what is the class I should be hanging my thinking on? (this is the exciting knot for me, if it can be cut most problems will go away... I hope :))

> _publication_date_search is actually only useful for searching dates

We also deemed it useful for frontend representation. It's a convenience for the frontend to not have to recompute it.

> serializing information

Don't understand this.

> The limitation with @post_load is that it happens only during input, i.e. when a record is created and goes through the loader via the REST API or programmatic/shell. If we want to change fields in the future, we have to "re-ingest" all of the records to generate these fields so that the @post_load runs or manually run whatever function we call in a "migrator".

This is why we have the post_load functionality extracted out as a function: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/marshmallow/json.py#L388 . This way it can be re-used elsewhere e.g. in a migrator.

> For computed fields that are used only for searching use the indexer receiver mechanism and skip them during serialization.

Why "skip them during serialization" ? Aren't they already serialized in the response from the index and so will appear in the response. Does this mean we have to erase them? 

> For computed fields that are used for serialization include them in the API class Record.dumps() method, which will place them in the indexed JSON

So we don't have that class (Record) in invenio-rdm-records (we could add it). I would like its responsibility to be this 1 class where what we talk about when we say an "invenio-rdm record" is all centralized. That is, it would be a handle on a In-memory-record-dict/RecordMetadata/PersistentIdentifier/Connections-to-Files/Connections-to-Communities/ES-representation. There are probably technical reasons for that not being possible, but I haven't found another way to chunk it. (side-note: Why is it in api.py and not model.py ?)

> In our case _communities is not an ES-only transformation, we include it also in the GET `/api/records/<recid>` response.

This I don't understand (good kind of excited)! Why is there a need for any other operation on the response if communities have been indexed in the record already? They will be in the search API response (which is fueled by the index). And if we want them in the API response for an individual record (which I guess is fueled by the DB)... well saving them in the record would give us that. Plus it would give us that for the jinja rendering. So, I see 2 possible reasons for the aversion here: performance and/or no computed field story. By the latter I mean, there is this feature that we want (computed fields), but don't have, so we are hacking around the simple solution of dumping everything in the DB, while not making it explicit.


Last point: 

As it is in the PR, can `record_communities(record).as_dict()` in the sidebar.html + accompanying react app show the communities a record belongs to just from the CommunityRequest and their status?

lnielsen:
I've just rebased on master and will merge once green again

PR #101, opened 2020-05-19
update dependencies according to invenio 3.4.0a1 bundles

ppanero:
I had to remove the upper limits because they were blocking the integration. When things were more "beta" it was nice to restrict. However, I think now we should do things as they are in Invenio in order to avoid extra work. This means:

- Upper bounds are only set in bundles. In our case, this is set by the `invenio_version` in `invenio-app-rdm` setup.py.
- In modules (i.e. `invenio-rdm-records` and `invenio-records-permissions`) upper bounds are not set, unless they cause troubles (e.g. in Travis). If setting an upper bound we should try to address it because it is an incompatibility with a full Invenio version.

*Why this change?*

I was working on updating InvenioRDM to the v3.4.0a1, Semantic-UI, when the only part of the work in order to integrate it should be done in `cookiecutter-invenio-rdm` and `invenio-app-rdm` I got conflict because other modules. If keeping upper bound would mean that we would have to release every single time we have to update a specific module. When in reality they are compatible.

- Less commits
- Less blockings
- Less time consumed

Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/102

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,20 +16,13 @@ readme = open('README.rst').read()
 history = open('CHANGES.rst').read()
 
 tests_require = [
-    'check-manifest>=0.25',
-    'coverage>=4.0',
-    'isort>=4.3.3',
-    'pydocstyle>=2.0.0',
-    'pytest-cov>=2.5.1',
-    'pytest-pep8>=1.0.6',
-    'pytest-invenio>=1.2.0',
-    'pytest>=4.0.0,<5.3.0',
-    'invenio_app>=1.2.3,<1.3.0'
+    'pytest-invenio>=1.3.2,<2.0.0',
+    'invenio-app>=1.3.0,<2.0.0'
 ]
 
 # Should follow inveniosoftware/invenio versions
-invenio_db_version = '>=1.0.4,<1.1.0'
-invenio_search_version = '>=1.2.0,<1.3.0'
+invenio_db_version = '>=1.0.4,<2.0.0'
+invenio_search_version = '>=1.2.0,<2.0.0'
 
 extras_require = {
     'docs': [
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -16,15 +16,15 @@
 history = open('CHANGES.rst').read()
 
 tests_require = [
-    'check-manifest>=0.25',
-    'coverage>=4.0',
+    'check-manifest>=0.35',
+    'coverage>=4.5.3',
```
lnielsen, line 20: Based on the [new dependency management](https://invenio.readthedocs.io/en/latest/releases/v3.3.0.html#dependency-management) we should be able to remove coverage, isort, check-manifest and pytest from here. 

I would also try with 
- pydocstyle
- pytest-cov

setup.py
```diff
@@ -16,15 +16,15 @@
...
-    'pytest-cov>=2.5.1',
+    'pydocstyle>=3.0.0',
+    'pytest-cov>=2.7.1',
     'pytest-pep8>=1.0.6',
-    'pytest-invenio>=1.2.0',
-    'pytest>=4.0.0,<5.3.0',
-    'invenio_app>=1.2.3,<1.3.0'
+    'pytest-invenio>=1.3.1,<1.4.0',
+    'pytest>=4.6.4,<5.0.0',
+    'invenio_app>=1.3.0'
```
lnielsen, line 27:
```suggestion
    'invenio-app>=1.3.0,<1.4.0'
```

setup.py
```diff
@@ -74,15 +74,15 @@
     'Faker>=2.0.3',
     'Flask-BabelEx>=0.9.4',
     'idutils>=1.1.3',
-    'invenio-formatter[badges]>=1.0.2,<1.1.0',
-    'invenio-jsonschemas>=1.0.1,<1.1.0',
-    'invenio-pidstore>=1.1.0,<1.3.0',
-    'invenio-records-rest>=1.6.4,<1.7.0',
-    'invenio-records>=1.3.0,<1.4.0',
-    'invenio-records-files>=1.2.1,<1.3.0',
+    'invenio-formatter[badges]>=1.0.3',
```
lnielsen, line 77:
```suggestion
    'invenio-formatter[badges]>=1.1.0a1',
```

setup.py
lnielsen: Forget this one except for the underscore to dash

setup.py
ppanero: The tag in invenio still shows 1.0.3 https://github.com/inveniosoftware/invenio/blob/v3.4.0a1/setup.py#L36 Which is where I got the deps versions. Should I check master (wondering about compatibilities)?

setup.py
ppanero: Not really... I just realized `isort`, `coverage` and `check-manifest` are only in `tests_requires`. I am not sure they will get installed here unless they are in `install_requires`. No?

ppanero: Talked with @slint that is missing from pytest-invenio. I'm making a PR now.

setup.py
```diff
@@ -16,15 +16,8 @@
 history = open('CHANGES.rst').read()
 
 tests_require = [
-    'check-manifest>=0.25',
-    'coverage>=4.0',
-    'isort>=4.3.3',
-    'pydocstyle>=2.0.0',
```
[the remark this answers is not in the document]
ppanero: Not any more, migrated to `pytests-invenio` :) Passing now

PR #107, opened 2020-05-20
[follows from#106] store access right choices via a vocabulary

fenekku:
Part of https://github.com/inveniosoftware/react-invenio-deposit/issues/18

Don't know if this is the best way, but it's convenient.


invenio_rdm_records/vocabularies/utils.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+# Copyright (C) 2020 Northwestern University.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Utils for vocabulary code."""
```
ppanero, line 9: moderate: if it is only used for vocabularies, why couldn't it be part of said class? or at least file? totally biased comment since im not a fan of *util* files/modules. **Does not block merge**

invenio_rdm_records/vocabularies/__init__.py
```diff
@@ -8,190 +8,111 @@
...
                 current_app.config
                 .get('RDM_RECORDS_CUSTOM_VOCABULARIES', {})
-                .get(vocabulary_type) or
-                cls.vocabularies
-                .get(vocabulary_type)
-                .get('path')
+                .get(key, {})
+            )
+
+            path = (
```
ppanero, line 72: understanding check: we do this so defaults are the above `cls.vocabularies` but we five preference to the configured ones? If we are gonna get from config anyway, wouldn't it be good to have the defaults defined in config (so they are only in one place)?

invenio_rdm_records/config.py
```diff
@@ -328,9 +331,17 @@ def _(x):
...
+        'resource_type': {
+            'path': os.path.join(
+                os.path.dirname(os.path.abspath(__file__)),
+                'my_resource_types.csv'
+            )
+        },
+        'contributors.role': {
+            'path': os.path.join(
+                os.path.dirname(os.path.abspath(__file__)),
+                'my_resource_types.csv'
```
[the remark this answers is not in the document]
fenekku: Oh that's a completely valid one. Definitely annoying when reading documentation! Thanks for spotting it.

invenio_rdm_records/marshmallow/json.py
```diff
@@ -95,34 +95,15 @@ class CreatorSchemaV1(BaseSchema):
...
+    role = SanitizedUnicode(required=True)
 
-    role = SanitizedUnicode(required=True, validate=validate.OneOf(
-                choices=ROLES,
-                error=_('Invalid role. {input} not one of {choices}.')
-            ))
+    @validates_schema
+    def validate_data(self, data, **kwargs):
+        """Validate role."""
+        vocabulary = Vocabularies.get_vocabulary('contributors.role')
```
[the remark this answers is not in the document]
fenekku: Yes, definitely passed the 3 strike-refactor mark. I made a `validate_entry` function in json.py, so that the Vocabularies class is isn't tied or responsible for marshmallow specific functionality.

invenio_rdm_records/vocabularies/__init__.py
fenekku: The developer experience this allows is more convenient IMHO:

As here:

If the developer wants to override the contributors role, they just have to do:

```python
# invenio.cfg

RDM_RECORDS_CUSTOM_VOCABULARIES = {
    'contributors.role': {
        'path': '/path/to/my/contributor_role.csv'
    }
}
```

Otherwise, if we were to use `RDM_RECORDS_CUSTOM_VOCABULARIES` for all vocabularies:

If the developer wants to override the contributors role, they have to do:

```python
# invenio.cfg
from invenio_rdm_records.config import RDM_RECORDS_CUSTOM_VOCABULARIES

RDM_RECORDS_CUSTOM_VOCABULARIES.update({
    'contributors.role': {
        'path': '/path/to/my/contributor_role.csv'
    }
})
```

It's not a big difference, but it's the little things that matter ;) .

We also store the loaded Vocabulary instances in the Vocabularies' `object` field which should not be exposed to others.

ppanero: Makes sense :)

invenio_rdm_records/vocabularies/utils.py
fenekku: Agreed to move it to Vocabulary.py

Issue #108, opened 2020-05-22
release 0.12.0


ppanero:
Bump dependencies to work with Semantic-UI

PR #104, opened 2020-05-19
Refactor + add contributor role vocabulary

fenekku:
- refactors Vocabulary singleton to Vocabularies and make it support vocabularies for nested fields. 
- closes #98 
- :warning: This introduces a breaking change in the interface to dump vocabularies. (once released we have to bump invenio-app-rdm for sure).

PR #109, opened 2020-05-22
release: v0.12.0

ppanero:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/108


fenekku:
Hey @ppanero , #104 (and following) also introduces a breaking change. Can we get them in before we do that release? Then latest PR for invenio-app-rdm will pass.

PR #111, opened 2020-05-26
more semantic

ppanero:


ppanero:
@fenekku As you pointed out I added the editorconfig plugin. Therefore, some things got formatted to 2 spaces. We should all align so this keeps this way. Do you have it? like that? Also I managed to not commit them, but the current editorconfig changes some indentations:

```
-        {% elif record.access_right == "restricted" %}
-        <span class="label label-warning titlecase">
-        {% else %}
-        <span class="label label-success titlecase">
-        {% endif %}
-        {{ record.access_right }} Access
-        </span>
+          {% elif record.access_right == "restricted" %}
+          <span class="label label-warning titlecase">
+            {% else %}
+            <span class="label label-success titlecase">
+              {% endif %}
+              {{ record.access_right }} Access
+            </span>
```

```
{%- block javascript %}
-  {{ super() }}
-  {{ webpack['invenio-rdm-records-js.js'] }}
-{%- endblock javascript %}
\ No newline at end of file
+{{ super() }}
+{{ webpack['invenio-rdm-records-js.js'] }}
+{%- endblock javascript %}
```

We should get a config that suits all, otherwise its quite a pain :( Would you mind sharing yours if this does not happen in it?

PR #121, opened 2020-06-01
Fix displayed access right

fenekku:
Display access right on landing page using vocab. 

Closes `record landing page: title-ize the access right. Now is "open Access", but should be "Open Access` from https://github.com/inveniosoftware/invenio-app-rdm/issues/154


fenekku:
Using pytest directly fails here because there is a version conflict on `six` : 

```
pkg_resources.ContextualVersionConflict: (six 1.11.0 (/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages), Requirement.parse('six>=1.12.0'), {'jsonresolver', 'invenio-base'})
```

that is solved in `python setup.py test` because it installs: 

```
    'pytest-invenio>=1.3.2,<2.0.0',
    'invenio-app>=1.3.0,<2.0.0'
```

Need to investigate to make sure our dependencies are fine and we don't rely on that to have them work.

PR #126, opened 2020-06-05
datamodel: change creator/contributor identifiers

### LINKED ISSUE #97: Implement identifiers
Identifiers are not quite formalized and there is a number of issues that need to be addressed:

- [x] Add Identifiers front-end field to deposit page for creators/contributors
  * [x] ~~this includes figuring out how to pass potential schemes to the front-end~~ (shelved from planning)
- [x] Should the `{<identifier scheme>: <identifier value>}` pattern be used for other data model fields than Creator/Contributor? If so changes are needed but more reuse can then be achieved. See below for more details.
- [x] Validate identifiers with idutils package and add missing identifiers to idutils ( See #51 and https://github.com/inveniosoftware/idutils/issues/53 )
  * [x] validate orcid with idutils
  * [x] validate ror with idutils https://github.com/inveniosoftware/idutils/pull/59
- [x] Validate that organization ids are used for organizations and person ids are used for people.
- [x] AffiliationSchemaV1 changes to match creator/contributor identifiers
- [x] Link to ROR and ORCID
- [x] Add to frontend deposit form
  * [x] people/org identifiers

Included in this work is making our identifier structure work on the frontend too.

Creator/Contributors have their identifiers as:

```json
{
  "ORCID": "0000-0002-1825-0097"
}
```

but other members of the datamodel don't. For example, the AffiliationFieldSchemaV1 is as follows:

```python
class AffiliationSchemaV1(BaseSchema):
    """Affiliation of a creator/contributor."""

    name = SanitizedUnicode(required=True)
    identifier = SanitizedUnicode(required=True)
    scheme = SanitizedUnicode(required=True)
```

i.e. 

```json
{
  "name": "California Digital Library",
  "identifier": "03yrm5c26",
  "scheme": "ror"
}
```

Should the `identifier`, `scheme` be combined like it is for Creator for other fields exhibiting this pattern? In this case it can reused as looked at by Pablo

> IdentifierScheme could be reused in many other schemas. However, there is no easy and clean way to flatten its attributes. Tested Pluck and Method. The latter worked but it has a more difficult code comprehension, which in my perspective makes is not a good choice.

fenekku:
- [x] use idutils to validate ORCiD identifiers
- [x] check identifier in function of type (Person/Org)
- [x] use idutils to validate ROR identifiers
- [x] link to ORCiD/ROR on landing page

depends on https://github.com/inveniosoftware/idutils/pull/62 + its actual release (fails until then)

closes #97 

Record landing page then looks like (with icons being links):

orcid_ror_record_page


invenio_rdm_records/marshmallow/json.py
```diff
@@ -81,8 +82,19 @@ class AffiliationSchemaV1(BaseSchema):
...
     name = SanitizedUnicode(required=True)
-    identifier = SanitizedUnicode(required=True)
-    scheme = SanitizedUnicode(required=True)
+    identifiers = fields.Dict()
+
+    @validates("identifiers")
+    def validate_identifiers(self, value):
+        """Validate well-formed identifiers are passed."""
+        if len(value) == 0:
+            raise ValidationError(_("Invalid identifier."))
```
zzacharo, line 91: @fenekku raising errors without specifying the field that they refer to, it causes the issue we have in the deposit for not being able to connect an error with the field. Do you think we could pass the field here as per any other default Marshmallow validation error? I am ok if you want to leave it like this and fix all together later but I just see that they are pilling up :) What do you think?

invenio_rdm_records/theme/templates/invenio_rdm_records/details/list_entities.html
```diff
@@ -8,16 +8,19 @@
...
   {% for entity in entities %}
-    {# TODO: Is the first identifier given precedence or is "Orcid" given precedence? #}
-    {# Assume first identifier is given precedence #}
-    {% set main_identifier = entity.identifiers[0] if entity.identifiers and entity.identifiers[0] else {"scheme": ""} %}
-    {%- if main_identifier.scheme == "Orcid" %}
-    <a href="{{ main_identifier.identifier|pid_url('orcid') }}">
-      <img class="inline-orcid" src="{{ url_for('static', filename='images/orcid.svg') }}" />
-    </a>
-    {# TODO: Other potential main identifiers #}
+    {# TODO: Revisit when more than ORCiD accepted #}
```
ppanero, line 11: I think this todo can be removed? We are accepting orcid and ror now?

tests/conftest.py
```diff
@@ -96,19 +96,20 @@ def full_input_record():
                 "given_name": "Julio",
                 "family_name": "Cesar",
                 "identifiers": {
-                    "Orcid": "9999-9999-9999-9999"
+                    "Orcid": "0000-0002-1825-0097"
```
ppanero, line 99: shelved: we might one to normalize when extracting... or do some sort of logic to allow capitalization, upper and lower case? Applies to more than this line, just got the idea here 😅  If agreed I'll create an issue.

zzacharo: Actually, one thought on this: Is/Will be `identifiers` a vocabulary? It can be really messy to have different formats i.e lowercase, uppercase, PascalCase etc... If so, isn't it better to agree on a format and configure via the vocabulary what the users sees?

invenio_rdm_records/marshmallow/json.py
fenekku: > raising errors without specifying the field that they refer to, it causes the issue we have in the deposit for not being able to connect an error with the field. 

Ah thanks for alerting me to that. Since I am looking into errors for flask-resources, I can look into errors for this too after. I made an issue for it: https://github.com/inveniosoftware/invenio-rdm-records/issues/127 since I think it will take more time to wrap my head around a good way to deal with this.

invenio_rdm_records/theme/templates/invenio_rdm_records/details/list_entities.html
fenekku: We might accept other identifiers (Zenodo does like GND...) in the future, but we haven't had a discussion about it too much, so we can leave the TODO there for now.

tests/conftest.py
fenekku: I think all these remarks are covered already in part :grin: . 

- `Orcid` is indeed not visible to the end-user because a small "vocabulary" hides it by displaying a human readable label that corresponds to it. That being said my preference would be for it to be "orcid" instead of "Orcid" for us developers working with it.
- identifier values are normalized according to idutils. (maybe though the remark was just about keys in which case the above point takes care of it)

invenio_rdm_records/marshmallow/json.py
zzacharo: @fenekku inveniosoftware/react-invenio-deposit#22 is not only for styling as the errors are not displaying at all if the error message is not referring to a specific field :)

PR #129, opened 2020-06-26
datamodel: fix affiliation identifiers json+mappings

### LINKED ISSUE #128: fix cli created records should have valid creator/contributors + affiliations identifiers + document
- [x] Need to doublecheck that the cli creates valid record when it comes to creators/contributors fields' identifiers + their affiliations' identifiers. In particular there is:

```json
            "identifiers": {
                "Orcid": "9999-9999-9999-9998",
            },
```

that may be wrong.

- [x] Then these changes need to be reflected in the documentation in docs-invenio-rdm

fenekku:
closes #128

### DIFF
--- a/invenio_rdm_records/cli.py
+++ b/invenio_rdm_records/cli.py
@@ -116,9 +116,6 @@ def create_fake_record():
         "contributors": [{
             "name": fake.name(),
             "type": "Personal",
-            "identifiers": {
-                "Orcid": "9999-9999-9999-9998",
-            },
             "affiliations": [{
                 "name": fake.company(),
                 "identifiers": {
--- a/invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
+++ b/invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
@@ -201,11 +201,8 @@
                 "name": {
                   "type": "text"
                 },
-                "identifier": {
-                  "type": "keyword"
-                },
-                "scheme": {
-                  "type": "keyword"
+                "identifiers": {
+                  "type": "object"
                 }
               }
             }
@@ -273,11 +270,8 @@
                 "name": {
                   "type": "text"
                 },
-                "identifier": {
-                  "type": "keyword"
-                },
-                "scheme": {
-                  "type": "keyword"
+                "identifiers": {
+                  "type": "object"
                 }
               }
             },
--- a/invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
+++ b/invenio_rdm_records/jsonschemas/records/record-v1.0.0.json
@@ -203,11 +203,15 @@
               "items": {
                   "type": "object",
                   "properties": {
-                      "name": {"type": "string"},
-                      "identifier": {"type": "string"},
-                      "scheme": {"type": "string"}
+                    "name": {"type": "string"},
+                    "identifiers": {
+                      "type": "object",
+                      "additionalProperties": {
+                        "type": "string"
+                      }
+                    }
                   },
-                  "required": ["name", "identifier", "scheme"]
+                  "required": ["name", "identifiers"]
               }
             }
         },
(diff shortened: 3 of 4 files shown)


invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
```diff
@@ -273,11 +270,8 @@
                 "name": {
                   "type": "text"
                 },
-                "identifier": {
-                  "type": "keyword"
-                },
-                "scheme": {
-                  "type": "keyword"
+                "identifiers": {
+                  "type": "object"
```
ppanero, line 274: minor: how many key-value pairs do we expect. Since keys will be dynamically created we could have a case of mapping explosion.

invenio_rdm_records/cli.py
```diff
@@ -116,9 +116,6 @@ def create_fake_record():
         "contributors": [{
             "name": fake.name(),
             "type": "Personal",
-            "identifiers": {
```
ppanero, line 119: doubt: what is wrong with this (appart from the value which might not pass the regex?)

invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
fenekku: Very few. In fact right now the marshmallow schema only accepts ROR identfifiers for affiliations so it is 0 or 1.

Is it more problematic that a large number of identifiers are in object mapping rather than array mapping?

invenio_rdm_records/cli.py
fenekku: Indeed that Orcid would not pass the regex. We allow no identifiers for a contributor/creator, so it is good to show this on the generated record. That's mostly it. Do you have a replacement in mind?

invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
ppanero: It depends on the amount of keys... If you have for example:

```
{
 id_type_one: id_value,
 id_type_two: id_value
}
```

Means that **Every** document will have said key, it might be that the value is nulled, but the key will be there. If there are two or few keys is not a problem. However, if it was the case that there were 1000 possible keys every single document would have:

```
{
 id_one: value,
 ...
 id_one_thousand: value
}
```

However, with array the amount of keys would always be two:
```
[ { id: id_one, value: value}, {id: id_one_thousand, value:value} ]
```

I understand from what you said it is not our case. However, if this was to grow a lot in keys it can have impact on search performance and even produce run time errors when indexing since the amount of keys is limited at mapping definition level. More info [here](https://www.elastic.co/guide/en/elasticsearch/reference/master/mapping.html#mapping-limit-settings)

invenio_rdm_records/cli.py
ppanero: Not really, was just nice to display the orcid badge on the contributors, etc. on the demo landing page. But it's okay it will come eventually when we have real world demo records from partners

invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
fenekku: > Means that Every document will have said key, it might be that the value is nulled, but the key will be there.

Oh interesting. So if:

1- index document A with {identifiers: {}}
2- index document B with {identifiers: {ror: blabalba}}

ES reindexes document A with {identifiers: {ror: null}} and so on for each other key in identifiers?

invenio_rdm_records/cli.py
fenekku: Yes and the Orcid badge appears for creators at least so they will be able to see it in action a little.

invenio_rdm_records/mappings/v6/records/record-v1.0.0.json
ppanero: Yes, not fully sure if the default value is "", null or what exactly. Since ES is Java I would say sort of `null`. But yes...

PR #130, opened 2020-06-29
Fix previewer

zzacharo:


zzacharo:
Screenshot 2020-06-29 at 09 33 04

And collapsed version (Will need to revisit the styling):

Screenshot 2020-06-29 at 09 33 25

PR #147, opened 2020-08-04
adapt to indexing split

ppanero:
Requires https://github.com/inveniosoftware/invenio-drafts-resources/pull/24

### DIFF
--- a/invenio_rdm_records/jsonschemas/drafts
+++ b/invenio_rdm_records/jsonschemas/drafts
@@ -1,0 +1,1 @@
+records/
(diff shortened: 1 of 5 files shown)


invenio_rdm_records/jsonschemas/drafts
```diff
@@ -0,0 +1 @@
+records/
```
fenekku, line 1: Should we review this? (this doesn't look like a mapping... :suspect: )

ppanero: Since they are the same... It's a symlink ;) It works like a charm and we do not need to maintain two things. Hopefully at some point converging in one. The main problem is `invenio-search` not being able to create two different indexes out of the same file since it looks for folder name 😅

PR #150, opened 2020-08-07
links: generate links

### LINKED ISSUE #142: naming: resources
As commented here: https://github.com/inveniosoftware/invenio-rdm-records/pull/139 and in many others.

- BlaBlaBlaRecordDraftResourceConfig, etc. are really long names. Shall we make `ResourceConfig` into an acronym? Otherwise shall we just call "Draft" and "Record".
- What is a `deposit`? Shall we use it? if so, does it mean "draft" or "draft+resource"?

### LINKED ISSUE #149: Ensure links are generated
It turns out to be much easier to start from a concrete case (BibliographicRecords) and generate links. Then abstract this away for other cases.  Nevertheless, we need to know what are the links we *do* want.

Here is what I think we need:

**When returning a single published record:**

- [x] self --- api link to record
- [x] self_html --- ui link to record
- [ ] files --- api link to files for the record     (this and the below 2 will depend on permissions) --> https://github.com/inveniosoftware/invenio-rdm-records/issues/151
- [x] edit --- api link to edit the record
- [x] delete --- api link to delete the record

**When returning a single draft record:**

- [x] self --- api link to draft
- [x] self_html --- ui url to deposit page
- [ ] files --- api link to files for the draft      (this and the below will depend on permissions) -> https://github.com/inveniosoftware/invenio-rdm-records/issues/151
- [x] publish --- api linkto publish the draft

**When returning a list of published records**

- [x] self --- api link to results with search parameters
- [x] next --- api link to next set of results with search parameters (next is only present if there is a next page of results)
- [x] prev --- api link to prev set of results with search parameters (prev is only present if there is a prev page of results)

**Are there links missing here?** @lnielsen

fenekku:
- bumps and uses records-resources + drafts-resources to achieve it
- configures service + pid_manager
- renames BibliographicRecordDraftService to
  BibliographicRecordService
- adds create_user fixture
- Per discussion: publication_date is now required and is passed
  in fixture. We will default-fill it on the frontend eventually.
  This is because we are not using load() anymore, so `missing`
  attribute from schema is not relevant anymore.
- Add drafts mapping
  * The addition of "drafts" in setup.py is a pretty blunt solution. 
    Should probably be revisited when $schema is used.
- closes #149 
- closes #142


fenekku:
There is an issue with the way we build indices and elasticsearch 6: probably the blunt approach in `setup.py`. I will come back to it tomorrow...

tests/new_api/conftest.py
```diff
@@ -47,3 +50,38 @@ def app(app):
     app.register_blueprint(draft_bp)
     app.register_blueprint(draft_action_bp)
     return app
+
+
+@pytest.fixture
+def create_user(app, db):
```
lnielsen, line 56: Comment Invenio-Accounts come with a utility function [create_test_user](https://invenio-accounts.readthedocs.io/en/latest/api.html#invenio_accounts.testutils.create_test_user)

fenekku: How did I miss this :smile: ? Using it. Thanks!

PR #156, opened 2020-08-26
config: add community search filters and release v0.18.1

slint:


invenio_rdm_records/config.py
```diff
@@ -71,6 +71,11 @@ def _(x):
             access_right=dict(terms=dict(field='access_right')),
             resource_type=dict(terms=dict(field='resource_type.type'))
         ),
+        # TODO: Move this to `invenio-communities` (or inject via config?)
```
lnielsen, line 74: Records-REST is being disconnected in this release, so not sure this is needed....

slint: Still seems to work though... at least for `/api/records/?community=...`

PR #163, opened 2020-09-04
Add dynamic aggregation [WIP]

Glignos:


Glignos:
Currently, the action request variable is conflicting with the permission action. The first should be renamed to something more specific.

invenio_rdm_records/utils.py
```diff
@@ -8,10 +8,33 @@
...
 def is_doi_locally_managed(doi_value):
     """Determine if a DOI value is locally managed."""
     return any(doi_value.startswith(prefix) for prefix in
                current_app.config['RDM_RECORDS_LOCAL_DOI_PREFIXES'])
+
+
+class DynamicAggregation(dict):
+    """Aggregation that is fully resolved in the request context."""
+
+    def __init__(self, base_agg, function_agg):
```
slint, line 24: You can match the [`dict(mapping, **kwargs)`](https://docs.python.org/3/library/stdtypes.html#dict) constructor, and basically do something like:

```suggestion
    def __init__(self, agg, func=None, **kwargs):
```

slint (approved):

lnielsen:
I'm closing this PR as it is towards Records-REST which is not used any longer. We will have to redo it for records-resources.

Issue #70, opened 2020-02-28
DOI template fails if there is no DOI


dfdan:
Similar to #69 - the DOI template assumes a DOI is present and fails to render if the identifier(s) are of a separate scheme.

**EDIT from discussions below**
A record may not have a DOI. This task is about fixing the template to account for that without crashing.

fenekku:
This: https://github.com/inveniosoftware/invenio-rdm-records/issues/71#issuecomment-592650825 applies here too: all records will have a DOI (@lnielsen can confirm I believe), so the template should be able to rely on it directly.

dfdan:
That's not my understanding - I know that RDM will only be able to *mint* DOI's in its first instance - but I though it would support the use of other PID systems without requiring/privileging DOI. 

It seems like requiring users to mint new DOI's for data with other PIDs that they simply want to import would be a barrier to adoption?

tmorrell:
I agree - thought that the internal record identifier was required and DOI was optional.

fenekku:
My understanding there is that *users* wouldn't mint the DOIs if their record doesn't have one. Rather the system would mint one for them (if none are given), so that every record eventually has one. So it shouldn't be a barrier to adoption, but we would have to be careful with imports. 

If 2 people understood it differently, then I may be wrong on this. Let's wait for @lnielsen to clarify :)

lnielsen:
Records MUST have an internal persistent identifier (record identifier of the style "adf6e-122d2") which will be the primary identifier use by the system. Records MAY have other external system managed persistent identifiers such as DOIs, …

First version of InvenioRDM will only support 1) using the internal identifier or 2) using internal + DOI.

dfdan:
Thanks for clarifying. I think in a lot of cases it would be important to be able to import data with pre-existing identifiers from other schemes - even if you can't mint new ones. Could these at least be stored (/displayed/searchable)?

fenekku:
@dfdan I think we can store those in the `identifiers` field without issue.

@lnielsen Just to be extra clear from your explanation: it may be the case that at rendering/serialization time, a record has no DOI (e.g., record was imported without one), so that kind of identifier may not appear in the `identifiers` list, correct?

PR #189, opened 2020-10-09
templates: fix rendering when no DOI

ppanero:
When re-saving the draft we are adding a DOI... (This should be corrected, maybe in the future when we tackle PIDs/minting)

This issue needs fixing in any case, since a record might not have a minted DOI.
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/70

Below two screenshots: first with DOI second without it.

Screenshot 2020-10-09 at 11 34 54

Screenshot 2020-10-09 at 11 34 39


invenio_rdm_records/theme/templates/invenio_rdm_records/details/doi.html
```diff
@@ -28,4 +29,10 @@ <h4>
       {{ badges_formats_list(url_for('invenio_formatter_badges.badge', title='DOI', value=id_doi, ext='svg', _external=True, _scheme='https'), id_doi|pid_url(scheme='doi')) }}
     </div>
   </div>
+{% else %}
+  <span class="get-badge" data-toggle="tooltip" data-placement="bottom" style="cursor: pointer;"
+    title="Get the DOI badge!">
+    No minted DOI
```
fenekku, line 35: Do we need it to be in the `get=bage` span? If so we could place the if statement inside the span, otherwise we could remove the span in this case.

invenio_rdm_records/theme/templates/invenio_rdm_records/details/doi.html
```diff
@@ -9,6 +9,7 @@
 {%- set id_doi = metadata.identifiers|doi_identifier %}
 <dt>{{_('DOI')}}:</dt>
```
fenekku, line 10: minor: I would remove the "DOI:" label here. If there is a DOI, "DOI" is displayed below. If there is none, the text says "No minted DOI". In both cases, it's clear the entity is a DOI.

fenekku (approved):

fenekku:
I am committing on top to address these and get us a working deposit flow.

invenio_rdm_records/theme/templates/invenio_rdm_records/details/doi.html
fenekku: done

Issue #192, opened 2020-10-12
ESDumper extension for EDTF publication_date


lnielsen:
- Create an [ESDumper extension](https://github.com/inveniosoftware/invenio-records/blob/master/tests/test_api_dumpers.py#L104-L109) which can dump an EDTF date (imprecise) into publication date (precise) which will be used for searching.
- The ESDumper used in the service needs to be configured with the dumper extensions.
- It should be possible to override the dumper extensions from the an instance.

Issue #190, opened 2020-10-09
Split models and api into the corresponding files


ppanero:
We have here https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/models.py both api and models, should be splitted (makes harder to find the classes...)

PR #194, opened 2020-10-13
dumpers: add edtf dumper

ppanero:
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/192
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/190

**Open question**: Should the extension be called `PublicationDateDumperExt` instead?


invenio_rdm_records/dumpers.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
```
lnielsen, line 1: Minor: I would structure the module like Invenio-Records-Resources:

```
invenio_rdm_records/records/api
invenio_rdm_records/records/dumpers.py
invenio_rdm_records/records/models.py
invenio_rdm_records/resources/...
invenio_rdm_records/services/...
```

This way it's clear what belongs to which layer.

lnielsen: Shelved: We should think about where to put these tools. The EDTF dumper is likely to be useful outside of RDM.

tests/test_dumpers.py
```diff
@@ -0,0 +1,61 @@
...
+
+from invenio_rdm_records.api import BibliographicRecord
+from invenio_rdm_records.dumpers import EDTFDumperExt
+
+
+def test_esdumper_with_edtfext(app, db, minimal_input_record):
+    """Test edft extension implementation."""
+    # Create a simple extension that adds a computed field.
+
+    dumper = ElasticsearchDumper(extensions=[EDTFDumperExt()])
```
lnielsen, line 21: Minor: Use fixtures and function to reduce duplication:

```
@pytest.fixture()
def dumper():
     return ElasticsearchDumper(extensions=[EDTFDumperExt()])

def _create_edtf_record(data, datestr):
     data['metadata']['publication_date'] = date_str
     record = BibliographicRecord.create(data)
     db.session.commit()
     return record
```

Also, I think you can skip the dumper, because it's already in the ``BibliographicRecord``

tests/test_dumpers.py
```diff
@@ -0,0 +1,61 @@
...
+
+    minimal_input_record['metadata']['publication_date'] = '2021-01-01'
+
+    # Create the record
+    record = BibliographicRecord.create(minimal_input_record)
+    db.session.commit()
+
+    # Dump it
+    dump = record.dumps(dumper=dumper)
+    assert dump['metadata']['publication_date_start'] == '2021-01-01'
```
lnielsen, line 31: Comment: I would also assert ``publication_date`` to ensure it's not removed by the dumper.

tests/test_dumpers.py
lnielsen: Alternatively, make a loop over different dates and expected result:

```python
[
('2021-01-01',      ('2021-01-01', '2021-01-01'))
('2021-01',         ('2021-01-01', '2021-01-31'))
('2021-01/2021-03', ('2021-01-01', '2021-03-31'))
]
```

Also, add tests for old dates:

1776
576

It's probably not going to work with years BC, but perhaps add a test to see.

ppanero: I commented out the three cases above.

invenio_rdm_records/dumpers.py
ppanero: Refactor done.

Extension could potentially be moved to invenio-records

tests/records/test_dumpers.py
```diff
@@ -0,0 +1,49 @@
...
+
+from invenio_rdm_records.records import BibliographicRecord
+from invenio_rdm_records.records.dumpers import EDTFDumperExt
+
+
+@pytest.mark.parametrize("date, expected_start, expected_end", [
+    ("2021-01-01", "2021-01-01", "2021-01-01"),
+    ("2021-01", "2021-01-01", "2021-01-31"),
+    ("2021", "2021-01-01", "2021-12-31"),
+    # ("1776", "1776-01-01", "1776-12-31"),
```
lnielsen, line 22: Question: is it because it's not working?

tests/test_dumpers.py
lnielsen: Minor: We will need to fix 1776 and 576, but somehow it's strange if EDTF doesn't support it......it's worth spending a bit more time on it, as it's definitely something which will come up (e.g. zenodo already have records from 1776)

ppanero: There is a limitation in python-edtf, it cannot parse anything lower than 1000AD. See: https://github.com/ixc/python-edtf/issues/35

invenio_rdm_records/records/api.py
```diff
@@ -0,0 +1,39 @@
...
+
+class BibliographicRecord(Record):
+    """Bibliographic Record API."""
+
+    model_cls = RecordMetadata
+
+    index = IndexField(
+        'rdmrecords-records-record-v1.0.0', search_alias='rdmrecords-records')
+
+    dumper = ElasticsearchDumper(extensions=[EDTFDumperExt()])
```
lnielsen, line 28: Shelved: We would need to be able to change these from the service config. Can you open an issue with that?

invenio_rdm_records/records/dumpers.py
```diff
@@ -0,0 +1,43 @@
...
+"""DataCite-based data model for Invenio."""
+
+import calendar
+from datetime import date
+
+from edtf import parse_edtf
+from invenio_records.dumpers import ElasticsearchDumperExt
+
+
+class EDTFDumperExt(ElasticsearchDumperExt):
```
lnielsen, line 17: The name I think is fine, but perhaps instead of hardcoding the field name, we could provide it directly on instantiation:

``EDTFDumperExt('metadata.publication_date')``

invenio_rdm_records/records/dumpers.py
```diff
@@ -0,0 +1,43 @@
...
+    which correspond to the lower strict and upper strict bounds. This
+    are required for sorting. In an ascending sort (most recent last),
+    sort by publication_date_start to get a natural sort order. In a
+    descending sort (most recent first), sort by publication_date_end.
+    """
+
+    # NOTE: Hardcoded to publication_date. If the need for other EDTF fields
+    # is found this can be made configurable (e.g. fields list attribute.)
+
+    def dump(self, record, data):
```
lnielsen, line 30: Minor: What if publication date is undefined? and what if there's a parsing error?

invenio_rdm_records/records/api.py
ppanero: https://github.com/inveniosoftware/invenio-records-resources/issues/128

invenio_rdm_records/records/dumpers.py
ppanero: Implemented:

- It requires a dot separated patht to the field: https://github.com/inveniosoftware/invenio-rdm-records/pull/194/files#diff-3062ab2ea652c3ad7e17249676a44495cdc1eaf6f3c7b661c071c94dfccf06aaR34
- Then _dump and _load do the work recursively

invenio_rdm_records/records/dumpers.py
ppanero: Implemented, along with tests:

- The field does not exists (KeyError)
- The field contains an un-parseable string (EDTFParseException)

In both cases, the errors are caught and "ignored".

ppanero:
coveralls is finally happy (+80%) @lnielsen 🚀

invenio_rdm_records/records/dumpers.py
```diff
@@ -0,0 +1,73 @@
...
+        """Constructor.
+
+        :param field: dot separated path to the EDTF field to process.
+        """
+        super(EDTFDumperExt, self).__init__()
+        self.field_path = field.split('.')
+
+    def _dump(self, path, data):
+        """Find recursively and dump EDTF field."""
+        key = path[0]
```
lnielsen, line 38: Better use https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/dictutils.py#L70 (as then we have consistent lookup strategy for field paths.

```python
keys = parse_lookup_key(self.field_path)
data = dict_lookup(data, keys, parent=True)
key = keys[-1]
pd = parse_edtf(data[key])
data[f'{key}_start'] = ...
```

invenio_rdm_records/records/dumpers.py
```diff
@@ -0,0 +1,73 @@
...
+    def dump(self, record, data):
+        """Dump the data."""
+        self._dump(self.field_path, data)
+
+    def _load(self, path, data):
+        """Find recursively and remove the dump fields."""
+        key = path[0]
+
+        if len(path) > 1:
+            data[key] = self._load(path[1:], data[key])
```
lnielsen, line 63:
```python
keys = parse_lookup_key(self.field_path)
key = keys[-1]
data = dict_lookup(data, keys, parent=True)
data.pop(f'{key}_start', None)
...
```

invenio_rdm_records/records/dumpers.py
ppanero: @lnielsen this would work for lookup and parse of the path. however, then we would set the created field in the root of the document. e.g.

```json
{ "metadata: {
   "publication_date": '2021'
  }
}
```

Would become

```json
{ "metadata: {
   "publication_date": "2021"
  },
 "publication_date_start": "2021-01-01",
 "publication_date_end": "2021-12-31",
}
```

Instead of:

```json
{ "metadata: {
   "publication_date": "2021",
   "publication_date_start": "2021-01-01",
   "publication_date_end": "2021-12-31",
  }
}
```

Taking into account that we have at the moment a search problem i.e. the search() method makes ES search in *every* field, including e.g. `access`, and not just inside `metadata`; once we fix that problem, this would not allow search on this type of fields (aggregation, facet and sorting would still be possible).

If search on this fields is not desired this approach would work.

Otherwise:
a- Create a `dict_setup` method (would mean looping twice on the array/field path)
b- Keep this, but inconsistent
c- Assume no search over that content

ppanero: Implemented https://github.com/inveniosoftware/invenio-rdm-records/pull/194/files#diff-3062ab2ea652c3ad7e17249676a44495cdc1eaf6f3c7b661c071c94dfccf06aaR41

invenio_rdm_records/records/dumpers.py
ppanero: Implemented with the restrictions shown above.

invenio_rdm_records/records/dumpers.py
lnielsen: See a solution in https://github.com/inveniosoftware/invenio-rdm-records/pull/201

PR #195, opened 2020-10-13
resource: add ui json serializer

zzacharo:


tests/api-resources/conftest.py
```diff
@@ -35,7 +35,7 @@ def create_app(instance_path):
 def record_resource():
     """Resource."""
     return BibliographicRecordResource(
-        config=RecordResourceConfig(),
+        config=BibliographicRecordResourceConfig(),
```
Glignos, line 38: Changed due to config that needs to be tested.

invenio_rdm_records/resources.py
```diff
@@ -37,6 +39,11 @@ class BibliographicRecordResourceConfig(RecordResourceConfig):
         ValidationError: handle_validation_error,
     }
 
+    response_handlers = {
+        **RecordResourceConfig.response_handlers,
+        "application/vnd.rdmv1+json": RecordResponse(UIJSONSerializer())
```
lnielsen, line 44:
```suggestion
        "application/vnd.inveniordm.v1+json": RecordResponse(UIJSONSerializer())
```

tests/api-resources/test_resources.py
```diff
@@ -228,7 +245,7 @@ def test_create_publish_new_revision(client, minimal_record,
     )
 
     assert response.status_code == 201
-    assert response.json['revision_id'] == 1
+    assert response.json['revision_id'] == 4
```
lnielsen, line 248: Question: why is this changed? Seems strange that this would need to change if it has previously passed.

tests/api-resources/test_resources.py
```diff
@@ -82,6 +90,15 @@ def test_simple_flow(app, client, minimal_record, headers):
     data = res.json['hits']['hits'][0]
     data['metadata']['title'] = 'New title'
 
+    # Search it including UI metadata
```
lnielsen, line 93: Minor: This should IMO be in a different test case. The simple flow tests is about the flow between different requests. The test is about the serialization which is being returned.

Glignos: Indeed I tried to squeeze it in the simple flow to avoid adding a separate case to keep the future testing time shorter but maybe it will be cleaner in a separate one.

lnielsen: In this case I prefer separate tests, to make it more clear what is being tested. This will make maintenance easier in the future.

tests/api-resources/test_resources.py
Glignos: This changed due to the draft being now soft-deleted instead (of hard-deleting). So then when creating a new revision the previously deleted draft is used instead of creating a new one.

lnielsen (approved):

PR #198, opened 2020-10-14
Facets ui config

zzacharo:


invenio_rdm_records/serializers.py
```diff
@@ -23,6 +23,25 @@ def _serialize_access_right(self, obj):
                 category=obj["access"]["access_right"]
             ))
 
+    def _serialize_access_right(self, obj):
```
lnielsen, line 26: Minor: Duplicated (seems like a merge issue)

invenio_rdm_records/serializers.py
```diff
@@ -23,6 +23,25 @@ def _serialize_access_right(self, obj):
...
+        """Dump ui config for `access_right` field."""
+        return dict(
+            access_right=dict(
+                category=obj["access"]["access_right"]
+            ))
+
+    def _serialize_aggregations(self, obj_list):
+        """Dump ui config for aggregations."""
+        aggregations = obj_list["aggregations"]
```
lnielsen, line 35: Minor: the serializer needs to be a little bit more flexible in case it's not there I think. We have use cases where we might not want to add the aggregations again (specifically if you e..g do infinite scroll, you don't need to recompute the aggregation because you're just doing pagination)

invenio_rdm_records/serializers.py
```diff
@@ -23,6 +23,19 @@ def _serialize_access_right(self, obj):
                 category=obj["access"]["access_right"]
             ))
 
+    def _serialize_aggregations(self, obj_list):
+        """Dump ui config for aggregations."""
+        aggregations = obj_list.get("aggregations")
+        if aggregations:
+            aggregations["ui"] = dict(
+                access_right=dict(
+                    open="Open Access",
```
lnielsen, line 32: Text needs to be marked for translation. ``_(....)``

Shelved: we likely need this as a vocabulary.....we need to think a bit more about it, so just merge for now

lnielsen (approved):

PR #200, opened 2020-10-14
WIP records: jsonschema metadata update

lnielsen:
This is a first draft of the update to the internal JSONSchema.

### Quick notes/questions

- Metadata and Extensions have been split apart.
- I've taken the following out of metadata: contact, internal notes, embargo date (their new location i didn't think properly about yet)
- Type definitions i've tried to put in the top, even if there's no controlled vocab (we need to discuss if all of them should be customizable).
- DataCite ``identifiers`` is taken care of by ``pid``, ``conceptpid`` and ``pids`` (i.e. all managed persistent identifiers).
- DataCite alternate identifiers is named identifiers.
- Field names: 
   - I've tried to stay as closed to datacite as possible, however 1) we need shorter names (and is not restricted by xml) 2) we need consistent naming for some properties.
- ``lang`` property - has not been added consistently, and we should disucss if we add them at all.
- Should we rename ``resource_type`` to ``type``? Could resource type be a single identifier?
- There's open questions regarding the structure of the fields that will have linked records (funders, awards, rights, subjects)
- DataCite support only 1 language, while we have requirements for mulitple languages 
- related identifiers have no support for hasMetadata etc, since the usage of it is very complex
- i've left out schemeURIs
- title/description is split in title/additional_titles and description/additional_descriptions in order to make it very clear what is the correct title/description to use for a record (otherwise we end in having different ways in different parts of the code).
- references/related_identifiers I guess we still have to discuss.
- required vs not required has not been properly addressed.

### DIFF
--- a/invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
@@ -7,12 +7,6 @@
 
   "definitions": {
 
-    "lang": {
-      "description": "ISO 639-3 language code.",
-      "type": "string",
-      "maxLength": 3
-    },
-
     "nameType": {
       "description": "Type of name.",
       "type": "string",
@@ -89,10 +83,6 @@
       "additionalProperties": false,
       "required": ["pk", "pid_type", "obj_type", "status"],
       "properties": {
-        "id": {
-          "description": "TODO: The value of the persistent identifier.",
-          "type": "string"
-        },
         "pk": {
           "description": "Primary key of the PID object.",
           "type": "integer"
@@ -125,7 +115,7 @@
       "additionalProperties": false,
       "required": ["id", "provider"],
       "properties": {
-        "id": {
+        "identifier": {
           "description": "The value of the persistent identifier.",
           "type": "string"
         },
@@ -143,7 +133,12 @@
     "identifiers": {
       "description": "Identifiers object (keys being scheme, value being the identifier).",
       "type": "object",
-      "additionalProperties": {"type": "string"}
+      "additionalProperties": {"$ref": "#/definitions/identifier"}
+    },
+
+    "identifier": {
+      "description": "An identifier.",
+      "type": "string",
     },
 
     "resource_type": {
@@ -223,8 +218,9 @@
       }
     },
 
-    "agent-user": {
+    "user": {
       "type": "object",
+      "description": "..",
       "additionalProperties": false,
       "properties": {
         "user": {
@@ -234,9 +230,9 @@
     },
 
     "agent": {
-      "description": "An agent.",
+      "description": "An agent (user, software process, community, ...).",
       "oneOf": [
-        {"$ref": "#/definitions/agent-user"}
+        {"$ref": "#/definitions/user"}
       ]
     }
   },
@@ -260,7 +256,8 @@
     "conceptpid": {"$ref": "#/definitions/internal-pid"},
 
     "pids": {
-      "additionalProperties": {"$ref": "#/definitions/external-pid"}
+      "additionalProperties": {"$ref": "#/definitions/external-pid"},
+      "description": "Managed persistent identifiers for a record including e.g. OAI-PMH identifier, minted DOIs and more. Managed PIDs are registered in the PIDStore"
     },
 
     "metadata": {
@@ -306,7 +303,6 @@
                   "type": "string"
                 },
                 "type": {"$ref": "#/definitions/titleType"},
-                "lang": {"$ref": "#/definitions/lang"}
             },
             "required": ["title"]
           }
@@ -471,7 +467,6 @@
                   "type": "string"
                 },
                 "type": {"$ref": "#/definitions/descriptionType"},
-                "lang": {"$ref": "#/definitions/lang"}
             }
           }
         },
@@ -539,7 +534,10 @@
             "type": "object",
             "additionalProperties": false,
             "properties": {
-              "reference_string": {"type": "string"},
+              "reference": {
+                "type": "string",
+                "description": "A reference string."
+              },
               "identifier": {"type": "string"},
               "scheme": {"type": "string"}
             }
--- a/invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
@@ -1,402 +1,714 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "id": "http://localhost/schemas/records/record-v1.0.0.json",
-  "title": "Invenio Datacite based Record Schema v1.0.0",
+  "id": "https://inveniosoftware.org/schemas/rdm/records/record-v1.0.0.json",
+  "title": "InvenioRDM Record Schema v1.0.0",
   "type": "object",
   "additionalProperties": false,
 
-  "properties": {
-    "$schema": {
-      "description": "This record's jsonschema.",
+  "definitions": {
+
+    "lang": {
+      "description": "ISO 639-3 language code.",
+      "type": "string",
+      "maxLength": 3
+    },
+
+    "nameType": {
+      "description": "Type of name.",
+      "type": "string",
+      "enum": [
+          "organizational",
+          "personal"
+      ]
+    },
+
+    "titleType": {
+      "description": "Type of title.",
+      "type": "string",
+      "enum": [
+          "alternative_title",
+          "subtitle",
+          "translated_title",
+          "other"
+      ]
+    },
+
+    "contributorType": {
       "type": "string"
     },
-    "conceptid": {
-      "description": "Invenio record identifier (alphanumeric).",
+
+    "dateType": {
+      "description": "Type of the date.",
+      "type": "string",
+      "enum": [
+        "accepted",
+        "available",
+        "copyrighted",
+        "collected",
+        "created",
+        "issued",
+        "submitted",
+        "updated",
+        "valid",
+        "withdrawn",
+        "other"
+      ]
+    },
+
+    "relationType": {
       "type": "string"
     },
-    "id": {
-      "description": "Invenio record identifier (alphanumeric).",
-      "type": "string"
+
+    "descriptionType": {
+      "type": "string",
+      "enum": [
+        "abstract",
+        "methods",
+        "seriesinformation",
+        "tableofcontents",
+        "technicalinfo",
+        "other"
+      ]
     },
-    "access": {
-      "access_right": {
-        "default": "open",
-        "description": "Access right for record.",
-        "type": "string"
-      },
-      "created_by": {
-        "description": "ID of user that created the deposit.",
-        "type": "integer"
-      },
-      "metadata_restricted": {
-        "default": false,
-        "description": "Record metadata accesibility. Public by default (False).",
-        "type": "boolean"
-      },
-      "files_restricted": {
-        "default": false,
-        "description": "Record associated files accesibility. Public by default (False).",
-        "type": "boolean"
-      },
-      "owners": {
-        "description": "List of user IDs that are owners of the record.",
-        "type": "array",
-        "minItems": 1,
-        "uniqueItems": true,
-        "items": {
-          "type": "number"
+
+    "longitude": {
+      "type": "number",
+      "minimum": -180,
+      "maximum": 180
+    },
+
+    "latitude": {
+      "type": "number",
+      "minimum": -90,
+      "maximum": 90
+    },
+
+    "internal-pid": {
+      "type": "object",
+      "description": "An internal persistent identifier object.",
+      "additionalProperties": false,
+      "required": ["pk", "pid_type", "obj_type", "status"],
+      "properties": {
+        "id": {
+          "description": "TODO: The value of the persistent identifier.",
+          "type": "string"
+        },
+        "pk": {
+          "description": "Primary key of the PID object.",
+          "type": "integer"
+        },
+        "pid_type": {
+          "description": "The persistent identifier type (e.g. doi or recid).",
+          "type": "string"
+        },
+        "obj_type": {
+          "description": "The type of the assigned object (e.g. rec).",
+          "type": "string"
+        },
+        "status": {
+          "description": "The status of the PID (from Invenio-PIDStore).",
+          "type": "string",
+          "enum": [
+            "N",
+            "K",
+            "R",
+            "M",
+            "D"
+          ]
         }
       }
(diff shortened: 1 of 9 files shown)


lnielsen:
```json
```

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,714 @@
...
+  "id": "https://inveniosoftware.org/schemas/rdm/records/record-v1.0.0.json",
+  "title": "InvenioRDM Record Schema v1.0.0",
   "type": "object",
   "additionalProperties": false,
 
-  "properties": {
-    "$schema": {
-      "description": "This record's jsonschema.",
-      "type": "string"
+  "definitions": {
```
ppanero, line 8: :rocket:

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,714 @@
...
-        "description": "List of user IDs that are owners of the record.",
-        "type": "array",
-        "minItems": 1,
-        "uniqueItems": true,
-        "items": {
-          "type": "number"
+
+    "descriptionType": {
+      "type": "string",
+      "enum": [
```
ppanero, line 64: How could they be changed? e.g. if someone wants to change any vocab that it's on an enum needs to redefine their jsonschema? Otherwise it won't pass validation, no?

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,714 @@
...
+    },
+
+    "internal-pid": {
+      "type": "object",
+      "description": "An internal persistent identifier object.",
+      "additionalProperties": false,
+      "required": ["pk", "pid_type", "obj_type", "status"],
+      "properties": {
+        "id": {
+          "description": "TODO: The value of the persistent identifier.",
```
ppanero, line 93: TODO?
```suggestion
          "description": "The value of the persistent identifier.",
```

ppanero (commented):
> Should we rename resource_type to type? Could resource type be a single identifier?

+1 should be consistent. As is *already* the case for titles, descriptions, etc.

> title/description is split in title/additional_titles and description/additional_descriptions in order to make it very clear what is the correct title/description to use for a record (otherwise we end in having different ways in different parts of the code).

Out of curiosity, how would this look in the deposit form? Separate fields? If not, what if there are more than one `MainTitle`?

> required vs not required has not been properly addressed.

Again on validation, when submitting a draft, if we validate with `required` fields against this schema. It would fail, no? Then we would need two schemas, or not validate the drafts.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,714 @@
...
-            },
-            "type": {
-              "description": "Type of the date interval."
-            }
-          },
-          "required": ["type"]
+      }
+    },
+
+    "agent-user": {
```
ppanero, line 226: question: agent-user? what is the purpose? It is a bit confusing with user-agent.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,714 @@
...
+    },
+    "pid": {"$ref": "#/definitions/internal-pid"},
+
+    "conceptid": {
+      "description": "Persistent concept record identifier (alphanumeric).",
+      "type": "string"
+    },
+    "conceptpid": {"$ref": "#/definitions/internal-pid"},
+
+    "pids": {
```
ppanero, line 262: I assume is not an array because we want the type of pid as key. Would this generate a possible mapping explosion in ES? Or will we have a not 1-to-1 mapping?

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,714 @@
...
             "properties": {
-                "description": {
-                  "description": "Description/abstract for record.",
-                  "type": "string"
-                },
-                "type": {
-                  "description": "Type of description.",
+                "name": {"type": "string"},
+                "type": {"$ref": "#/definitions/nameType"},
+                "given_name": {"type": "string"},
```
ppanero, line 284: this was changed to `name` in the UI, needs consensus. For deduplication and others, having them splitter might be better.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,714 @@
...
+              "reference_string": {"type": "string"},
+              "identifier": {"type": "string"},
+              "scheme": {"type": "string"}
+            }
+          }
         }
+      }
+    },
+
+    "metaext": {
```
ppanero, line 551: Search will be needed on `metadata` and `metaext`, we need to change this on the search obj.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,714 @@
...
+          "items": {"$ref": "#/definitions/agent"}
+        },
+
+        "embargo_date": {
+          "description": "Embargo date of record (ISO8601 formatted date time in UTC). At this time both metadata and files will be made public.",
+          "type": "string",
+          "format": "date-time"
+        },
+
+        "access_condition": {
```
ppanero, line 641: If this is for shared links, isn't it missing a link of some sort?

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,714 @@
...
+          "items": {"$ref": "#/definitions/file"}
+        },
+        "bucket": {
+          "type": "string",
+          "description": "TODO - we need a bucket id, but what about third-party storage? systems"
+        }
+      }
+    },
+
+    "system": {
```
ppanero, line 693: Is this meant for system fields? or for the system that pushes the record?

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,714 @@
...
-                  "required": ["name", "identifiers"]
-                }
-              },
-              "role": {"type": "string"}
-          },
-          "required": ["name", "role"]
+      }
+    },
+
+    "identifiers": {
```
ppanero, line 143: What is the difference between this and pids field?

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: First, see example below.

tests/records/full-record.json
```diff
@@ -0,0 +1,202 @@
...
+        }
+      }]
+    }],
+    "dates": [{
+      "date": "test",
+      "type": "other",
+      "description": "A date"
+    }],
+    "languages": ["da", "en"],
+    "identifiers": [{
```
ppanero, line 95: question above, but this duplicates with `pids`.

tests/records/full-record.json
```diff
@@ -0,0 +1,202 @@
...
+    "status": "R"
+  },
+  "conceptpid": {
+    "pk": 2,
+    "pid_type": "recid",
+    "obj_type": "rec",
+    "status": "R"
+  },
+  "pids": {
+    "doi": {
```
ppanero, line 18: how many keys do we expect? checking on mapping explosion.

tests/records/full-record.json
```diff
@@ -0,0 +1,202 @@
...
+      "provider": "zenodo"
+    }
+  },
+  "metadata": {
+    "resource_type": {
+      "type": "publication",
+      "subtype": "article"
+    },
+    "creators": [{
+      "name": "Nielsen, Lars Holm",
```
ppanero, line 44: will we enforce a validation here? or fill in the other two fields from inference?

tests/records/full-record.json
ppanero: Discussed IRL, we do not expect many pid types (between 1 and 10 aprox)

tests/records/full-record.json
ppanero: Yes, some sort of validation + maybe automatic filling (useful to distinguish personal vs organizational)

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: Maybe need transformation. But not expecting many keys.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: Intended, Personal vs Organization.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: pids is managed pids, identifiers is metadata have no impact on the code

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: To be decided

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: no need, it just specifies details about links but links per see go separated.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: all flattened.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: removing, discussed IRL

ppanero:
Created https://github.com/inveniosoftware/invenio-rdm-records/tree/datamodel branch

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,712 @@
...
+        "owned_by": {
+          "description": "List of user IDs that are owners of the record.",
+          "type": "array",
+          "minItems": 1,
+          "uniqueItems": true,
+          "items": {"$ref": "#/definitions/agent"}
+        },
+
+        "embargo_date": {
+          "description": "Embargo date of record (ISO8601 formatted date time in UTC). At this time both metadata and files will be made public.",
```
slint, line 634: Isn't it that files only are hidden and become public after the embargo date?

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,712 @@
...
+            "condition": {
+              "type": "string",
+              "description": "Textual description under which conditions access is granted."
             },
-            {"type": "boolean"},
-            {"type": "number"},
-            {"type": "string"}
-          ]
+            "default_link_validity": {
+              "type": "integer",
```
slint, line 649: If technically we go with a more fine-grained implementation (e.g. hours), maybe it's better to store this in seconds, to make it easy to uniformly turn into a `timedelta` object

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,712 @@
...
+
+    "access": {
+      "type": "object",
+      "description": "Record access control and ownership.",
+      "additionalProperties": false,
+      "properties": {
+
+        "metadata": {
+          "description": "Metadata visibility (true - public, false - private)",
+          "type": "boolean"
```
slint, line 612: Minor: Making these fields boolean flags might be limiting us in the number of available access "modes" we want to implement. I'm aware that the combination of these booleans translates to different access scenarios (i.e. hidden record, closed files only, etc). Of course, adding a new type of access mode could be achieved via just adding another boolean field, but then the number of total states, along with the boolean logic that has to be performed to translate these states becomes more complex.

A use case could be having "partially" hidden metadata fields. There would a couple of ways to implement:

- With boolean flags only: `metadata: true` and `metadata_hidden_fields: false`
- With enums: `metadata: partial`

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,712 @@
...
+        }
+      }
+    },
+
+    "files": {
+      "type": "object",
+      "description": "Files associated with the record",
+      "additionalProperties": false,
+      "properties": {
+        "disabled": {
```
slint, line 668: I think using the "positive" version, i.e. `enabled: true|false`, might be more readable.

slint (commented):
LGTM, some minor comments here and there.

Something that should be considered to be used more is `uniqueItems`... The question is if it's better to apply on the `marshmallow` level also/instead.

> Type definitions i've tried to put in the top, even if there's no controlled vocab (we need to discuss if all of them should be customizable)

I think a JSONSchema pre/post-processor would be great for that.

> Should we rename resource_type to type? Could resource type be a single identifier?

I prefer `resource_type` since it matches DataCite. Making it a single identifier (e.g. `text-report`), requires defining a convention for displaying hierarchies and translating them into (sub)aggregations.

ppanero:
@slint this PR might be outdated... I am working actively on this, against the `datamodel` branch of the repo. @lnielsen shall we close this PR and open one from that branch against master?

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -1,402 +1,712 @@
...
+
+        "subjects": {
+          "type": "array",
+          "items": {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                  "subject": {"type": "string"},
+                  "identifier": {"type": "string"},
+                  "scheme": {"type": "string"}
```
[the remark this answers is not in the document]
lnielsen: Yes, that was my thinking.....I don't think we need the schemeURI, or if we need, then I'd rather internally have a mapping scheme-> schemeURI that we can use to populate it with.

PR #207, opened 2020-10-16
datamodel: update jsonschema, full flow and validation for minimal record

ppanero:
- Removed `obj_type` and `pid_type` from internal-pid
- `file.file_id` --> `file.id` or even `file.identifier` (we have a definition for it)
- Use `identifier` definition for all identifiers occurences. Did the same for `scheme`.
- Removed `contact` from `provenance` and `access` until send message feature is implemented
- Made `notes` into a simple string on top level

**Open questions**
- In the JSONSchema states languages are ISO 639-**3**, however it is tested with strings of length 2 (it should fail, by definition this ISO has length 3). Should we allow both IS 639-2 and 3 or change tests?
- What about removing `title` from the type:

```diff
    "titleType": {
      "description": "Type of title.",
      "type": "string",
      "enum": [
-         "alternative_title",
+         "alternative",
          "subtitle",
-         "translated_title",
+         "translated",
          "other"
      ]
    },
```
- `resource_type` vs `type`

### DIFF
--- a/invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
@@ -65,6 +65,12 @@
       ]
     },
 
+    "language": {
+      "type": "string",
+      "minLength": 3,
+      "maxLength": 3
+    },
+
     "longitude": {
       "type": "number",
       "minimum": -180,
@@ -77,23 +83,31 @@
       "maximum": 90
     },
 
+    "scheme": {
+      "description": "A scheme.",
+      "type": "string"
+    },
+
+    "identifier": {
+      "description": "An identifier.",
+      "type": "string"
+    },
+
+    "identifiers": {
+      "description": "Identifiers object (keys being scheme, value being the identifier).",
+      "type": "object",
+      "additionalProperties": {"$ref": "#/definitions/identifier"}
+    },
+
     "internal-pid": {
       "type": "object",
       "description": "An internal persistent identifier object.",
       "additionalProperties": false,
-      "required": ["pk", "pid_type", "obj_type", "status"],
+      "required": ["pk", "status"],
       "properties": {
         "pk": {
           "description": "Primary key of the PID object.",
           "type": "integer"
-        },
-        "pid_type": {
-          "description": "The persistent identifier type (e.g. doi or recid).",
-          "type": "string"
-        },
-        "obj_type": {
-          "description": "The type of the assigned object (e.g. rec).",
-          "type": "string"
         },
         "status": {
           "description": "The status of the PID (from Invenio-PIDStore).",
@@ -113,12 +127,9 @@
       "type": "object",
       "description": "An external persistent identifier object.",
       "additionalProperties": false,
-      "required": ["id", "provider"],
+      "required": ["identifier", "provider"],
       "properties": {
-        "identifier": {
-          "description": "The value of the persistent identifier.",
-          "type": "string"
-        },
+        "identifier": {"$ref": "#/definitions/identifier"},
         "provider": {
           "description": "The provider of the persistent identifier.",
           "type": "string"
@@ -130,16 +141,6 @@
       }
     },
 
-    "identifiers": {
-      "description": "Identifiers object (keys being scheme, value being the identifier).",
-      "type": "object",
-      "additionalProperties": {"$ref": "#/definitions/identifier"}
-    },
-
-    "identifier": {
-      "description": "An identifier.",
-      "type": "string",
-    },
 
     "resource_type": {
       "type": "object",
@@ -211,10 +212,7 @@
           "description": "TODO - Bucket id - what about third party storage systems?",
           "type": "string"
         },
-        "file_id": {
-          "description": "TODO - Identifier of the file. - same as for bucket",
-          "type": "string"
-        }
+        "file_id": {"$ref": "#/definitions/identifier"}
       }
     },
 
@@ -303,6 +301,7 @@
                   "type": "string"
                 },
                 "type": {"$ref": "#/definitions/titleType"},
+                "lang": {"ref": "#/definitions/language"}
             },
             "required": ["title"]
           }
@@ -324,8 +323,8 @@
               "additionalProperties": false,
               "properties": {
                   "subject": {"type": "string"},
-                  "identifier": {"type": "string"},
-                  "scheme": {"type": "string"}
+                  "identifier": {"$ref": "#/definitions/identifier"},
+                  "scheme": {"$ref": "#/definitions/scheme"}
               },
               "required": ["subject"]
           }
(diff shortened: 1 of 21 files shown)


invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -77,24 +77,32 @@
...
       "description": "An internal persistent identifier object.",
       "additionalProperties": false,
-      "required": ["pk", "pid_type", "obj_type", "status"],
+      "required": ["pk", "status"],
       "properties": {
         "pk": {
           "description": "Primary key of the PID object.",
           "type": "integer"
         },
-        "pid_type": {
```
lnielsen, line 90: Question: Was this fixed in Invenio-Records-Resources as well?

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -211,10 +206,7 @@
           "description": "TODO - Bucket id - what about third party storage systems?",
           "type": "string"
         },
-        "file_id": {
```
lnielsen, line 214: I would leave this as ``file_id`` becuase it's not the primary identifier for the file. It's the identifier for the FileInstance, which points to the file on disk (this one can be changed, hence not a primary identifier).

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -538,15 +528,15 @@
...
           }
         }
       }
     },
 
-    "metaext": {
+    "ext": {
       "type": "object",
       "description": "Configured additional metadata",
       "propertyNames": {
```
lnielsen, line 542: Question: ``"pattern": "^[A-Za-z][A-Za-z0-9_]*:[A-Za-z0-9_]+$"`` has to be changed as well? Do you keep a list of things to change after merging somewhere?

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -688,25 +672,10 @@
...
-                  "note": {"type": "string"},
-                  "timestamp": {
-                    "description": "ISO8601 formatted date time stamp.",
-                    "type": "string",
-                    "format": "date-time"
-                  }
-              }
-          }
-        }
+    "notes": {
```
lnielsen, line 675: Shelved: We have to think a bit more. I don't like have this as top-level field. Seems wrong to me....

invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -51,22 +47,13 @@ class Meta:
...
+    id = fields.Str()
+    conceptid = fields.Str()
     metadata = NestedAttribute(MetadataSchemaV1)
     access = NestedAttribute(AccessSchemaV1)
-    # files = NestedAttribute(FilesSchemaV1, dump_only=True)
-    # communities = NestedAttribute(CommunitiesSchemaV1)
-    # pids = NestedAttribute(PIDSSchemaV1)
-    # stats = NestedAttribute(StatsSchemaV1, dump_only=True)
-    # relations = NestedAttribute(RelationsSchemaV1, dump_only=True)
+    pids = fields.List(NestedAttribute(PIDSchemaV1))
```
lnielsen, line 54: Comment: Let's drop the ``V1`` prefix on everything and only add it once we need it.

invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -51,22 +47,13 @@ class Meta:
         'files': 'read_files',
     }
 
-    # schema_version = fields.Interger(dump_only=True)
-    # revision = fields.Integer(attribute='revision_id', dump_only=True)
```
lnielsen, line 55: Minor: Revision should also be possible to dump directly.

tests/records/full-record.json
```diff
@@ -59,8 +55,7 @@
     "title": "InvenioRDM",
     "additional_titles": [{
       "title": "a research data management platform",
-      "type": "subtitle",
-      "lang": "en"
```
lnielsen, line 63: Minor: We have to add back the ``lang`` attribute on titles and descriptions. Sorry for the change of mind.

tests/records/full-record.json
```diff
@@ -143,12 +137,12 @@
...
     "references": [{
-      "reference_string": "Nielsen et al,..",
+      "reference": "Nielsen et al,..",
       "identifier": "101.234",
       "scheme": "doi"
     }]
   },
-  "metaext": {
+  "ext": {
     "dwc:collectionCode": "abc",
```
lnielsen, line 146: Minor: Can you change to also:

```
{
  "ext": {
    "dwc": {
      "...": "..."
    }
  }
}
```

lnielsen:
> In the JSONSchema states languages are ISO 639-3, however it is tested with strings of length 2 (it should fail, by definition this ISO has length 3). Should we allow both IS 639-2 and 3 or change tests?

Yes, it should be tested with 3 letter codes.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: Ah, yes, didn't fix the internal part of it. Will do when I arrive to the field itself.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: Nope only here for now

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: Didn't fix the field inside. Waiting to actually fix the metadata extensions since also loading dumping needs fix after all the changes.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: I agree, but `system` wasn't really accurate.

invenio_rdm_records/services/schemas/__init__.py
ppanero: Yes, didnt work on commented fields, there might be many more like this.

tests/records/full-record.json
ppanero: Fixed jsons and removed limitations on jsonschema for now, will fix when working on exts.

tests/records/full-record.json
ppanero: Added back. What is the reasoning behind?

invenio_rdm_records/services/schemas/__init__.py
ppanero: Done to the `schemas` folder, `marshmallow` should be removed. I will work on that now.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: next task is to investigate this

PR #210, opened 2020-10-19
datamodel

ppanero:


ppanero:
Need to take into account and discuss latest comments from @slint here https://github.com/inveniosoftware/invenio-rdm-records/pull/200

PR #211, opened 2020-10-19
serializers: serialize localized pub date

zzacharo:
Depends on https://github.com/inveniosoftware/marshmallow-utils/pull/18

### DIFF
--- a/invenio_rdm_records/resources/serializers.py
+++ b/invenio_rdm_records/resources/serializers.py
@@ -103,9 +103,13 @@ class UIJSONSerializer(JSONSerializer):
             self._serialize_resource_type_agg(
                 aggregations.get("resource_type"))
 
+    def serialize_obj_ui(self, obj):
+        """Dump ui config for object."""
+        obj.setdefault('ui', {}).update(self._serialize_access_right(obj))
+
     def serialize_object(self, obj, response_ctx=None, *args, **kwargs):
         """Dump the object into a json string."""
-        obj['ui'] = self._serialize_access_right(obj)
+        self.serialize_obj_ui(obj)
         return json.dumps(obj)
 
     def serialize_object_list(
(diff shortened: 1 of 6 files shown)


invenio_rdm_records/resources/serializers.py
```diff
@@ -23,6 +23,16 @@ def _serialize_access_right(self, obj):
                 category=obj["access"]["access_right"]
             ))
 
+    def _serialize_localized_publication_date(self, obj):
+        """Dump ui config for `access_right` field."""
+        localized_pub_date = dict(
+            publication_date_l10n=dict(
+                value=obj["metadata"]["publication_date_l10n"]
+            ))
+        del obj["metadata"]["publication_date_l10n"]
```
zzacharo, line 32: @lnielsen this is the point I referred to about not so clean way of handling the serialization. In principle, the new added field is serializing the `publication_date_l10n` already, so this might be redundant here...

invenio_rdm_records/resources/serializers.py
```diff
@@ -23,6 +23,16 @@ def _serialize_access_right(self, obj):
                 category=obj["access"]["access_right"]
             ))
 
+    def _serialize_localized_publication_date(self, obj):
+        """Dump ui config for `access_right` field."""
```
fenekku, line 27: copy-pasted comment!

fenekku (commented):
What about placing that top-level `ui` field in the schema directly like `metadata` is? or have the ui in the `metadata` (granted, the latter probably more complex). Adding back to the serializers goes contrary to the choices we made previously about making the serializers dumb, no?

zzacharo:
Fixed also empty `subtype` in the facets creating empty nested aggregations.

invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -62,6 +63,8 @@ class Meta:
     created = fields.Str(dump_only=True)
     updated = fields.Str(dump_only=True)
     revision = fields.Integer(dump_only=True)
+    ui = NestedAttribute(UISchema, attribute="metadata", dump_only=True)
```
lnielsen, line 66: Comment: I think this should be just the entire object not only the metadata part? You can merge and fix in another PR.

lnielsen (approved):

PR #212, opened 2020-10-19
datamodel

ppanero:


tests/services/test_schemas_json_load.py
```diff
@@ -7,18 +7,16 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 """DEPRECATED. Tests for Invenio RDM Records JSON Schemas."""
-import os
```
fenekku, line 10: This file can be deleted when all the schemas have their own test file like above. A gradual approach might work well:
- remove ResourceTypeSchema from this file because covered above and so on...

fenekku (approved):
Perhaps discuss the enums in the jsonschema (since they are customizable, my vote is to not have them in the jsonschema - and I thought we ran into the same issue in the past :thinking: ) with @lnielsen 

The Orcid -> orcid change has ramifications I think...

other smaller points.

I will let you merge ("approving" in case you want to address those in next PR...)

PR #213, opened 2020-10-20
serializers: inject ui config in aggregation buckets

zzacharo:


invenio_rdm_records/resources/serializers.py
```diff
@@ -9,32 +9,85 @@
...
+        :params labels_map: Dict of transforming bucket values to labels.
+        :params bucket: Current aggregation bucket.
+        """
+        bucket_label = labels_map.get(bucket["key"])
+        if bucket_label:
+            bucket["label"] = bucket_label
+        if bucket.get("subtype"):
+            bucket["subtype"]["buckets"] = list(map(
+                partial(self._set_bucket_label, labels_map),
+                bucket["subtype"]["buckets"]))
```
ppanero, line 63: minor: This logic is repeated a few times. Can it be made a helper function?

ppanero (approved):

invenio_rdm_records/resources/serializers.py
```diff
@@ -9,32 +9,85 @@
...
 
     def _serialize_access_right(self, obj):
-        """Dump ui config for `access_right` field."""
+        """Inject ui config for `access_right` field."""
         return dict(
             access_right=dict(
                 category=obj["access"]["access_right"]
             ))
 
+    def _serialize_ui_options_from_vocabulary(
```
Glignos, line 29: This function works for the currently present Vocabularies but maybe it's worth considering a recursive approach if we expect depth>2 in any future ones.

invenio_rdm_records/resources/serializers.py
```diff
@@ -9,32 +9,85 @@
...
+    def _set_bucket_label(self, labels_map, bucket):
+        """Set bucket label from config.
+
+        :params labels_map: Dict of transforming bucket values to labels.
+        :params bucket: Current aggregation bucket.
+        """
+        bucket_label = labels_map.get(bucket["key"])
+        if bucket_label:
+            bucket["label"] = bucket_label
+        if bucket.get("subtype"):
```
Glignos, line 60: Moderate: `subtype` is a specific key for this vocabulary, I think we need a more general approach. I think we can safely assume in this case that any extra key corresponds to a sub-bucket key, but we could also check the existence of the `buckets` key in the nested object.

invenio_rdm_records/resources/serializers.py
```diff
@@ -9,32 +9,94 @@
...
+
+    def _set_bucket_label(self, labels_map, bucket):
+        """Set bucket label from config.
+
+        :params labels_map: Dict of transforming bucket values to labels.
+        :params bucket: Current aggregation bucket.
+        """
+        bucket_label = labels_map.get(bucket["key"])
+        if bucket_label:
+            bucket["label"] = bucket_label
```
Glignos, line 58: Minor: We should add the `doc_count` to the `label` here to keep the same visuals.

invenio_rdm_records/resources/serializers.py
zzacharo: I checked Zenodo and it seems we use the `subtype` key for child aggregations. Of course, we should force this in our UI configuration. I assumed that we can normalize the interface by having all nested aggregations under `subtype` field.  If not, then it needs some more dynamic checking but not sure if it's worth it... @lnielsen WDYT?

invenio_rdm_records/resources/serializers.py
zzacharo: ping @lnielsen

Glignos (approved):
LGTM 👍

Issue #223, opened 2020-10-21
enable EDTF dumper to support date arrays


ppanero:
## Is your feature request related to a problem? Please describe.
This is needed to support search on the date field

## Describe the solution you'd like
The EDTF dumper should check the type of field that is obtaining. If it is an EDTF date dump it, if it is an array of those it should dump each of them taking into account prefixing/suffixing.

## Additional context
- https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/dumpers.py#L19
- It should also be set in the config: https://github.com/inveniosoftware/invenio-rdm-records/blob/9e8e0158a0507c23c5594d932cac74ee24f1f826/invenio_rdm_records/records/api.py#L28 sort of:

```python
dumper = ElasticsearchDumper(extensions=[
    EDTFDumperExt('metadata.publication_date'),
    EDTFDumperExt('metadata.dates'),
])
```

ppanero:
Closed by #295

PR #225, opened 2020-10-21
more data model

### LINKED ISSUE #222: data model part II
- [x] affiliations validation and tests
- [x] subjects
- [x] dates: refactor schema (use EDTF instead of start-end), change mappings and tests

ppanero:
closes #222 

Enables the start of #223

### DIFF
--- a/invenio_rdm_records/services/schemas/metadata.py
+++ b/invenio_rdm_records/services/schemas/metadata.py
@@ -76,34 +76,6 @@ class DateSchema(Schema):
             )
 
 
-# 'Fake' Identifiers Field
-def _not_blank(error_msg):
-    """Returns a non-blank validation rule with custom error message."""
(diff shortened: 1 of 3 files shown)


ppanero:
Travis is passing, I manually cancelled it so @zzacharo can merge his PR without having to rebase again.

tests/services/test_schemas/test_affiliations.py
```diff
@@ -0,0 +1,76 @@
+# -*- coding: utf-8 -*-
```
lnielsen, line 1: Minor: File naming wise I would tink it should be ``tests/services/schemas/test_affilations.py`` Remember that ``test_affilations.py`` has to be globally unique in all test folders.

ppanero: True, I will rename the folder in a proceeding PR if you agree.

PR #228, opened 2020-10-22
serializers: adds icons to the access_right ui config

Glignos:


invenio_rdm_records/resources/serializers.py
```diff
@@ -19,11 +19,22 @@
 class UIJSONSerializer(JSONSerializer):
     """UI JSON serializer implementation."""
 
+    def icon_options(self, vocabulary_name):
+        """Provides the icon configuration from the vocabulary."""
+        vocabulary_list = Vocabularies.get_vocabulary(
```
zzacharo, line 24: Couldn't this be merged with the `_serialize_ui_options_from_vocabulary` function? It seems that is already there the option but we chose to return only the `text` key...Maybe we can define a pair of keys to return from the vocabulary parsing...WDYT?

ppanero:
cancelled travis cuz it requires a rebase (trying to free travis)

invenio_rdm_records/resources/serializers.py
```diff
@@ -21,87 +20,60 @@ class UIJSONSerializer(JSONSerializer):
...
         """Inject ui config in aggregations."""
         aggregations = obj_list.get("aggregations")
         if aggregations:
-            self._serialize_access_right_agg(aggregations.get("access_right"))
-            self._serialize_resource_type_agg(
-                aggregations.get("resource_type"))
+            for aggregation_key in aggregations:
+                agg_vocabulary = self._serialize_ui_options_from_vocabulary(
+                    aggregation_key)
+                if not agg_vocabulary:
```
zzacharo, line 73: If `agg_vocabulary` value is actually `None`, maybe is better to rewrite this into `if agg_vocabulary is None` :)

invenio_rdm_records/resources/serializers.py
```diff
@@ -21,87 +20,60 @@ class UIJSONSerializer(JSONSerializer):
...
-        self._serialize_agg(resource_type_agg, resource_type_labels["type"])
-
-    def _serialize_access_right_agg(self, access_right_agg):
-        """Inject ui config for `access_right` aggregation."""
-        access_right_labels = self._serialize_ui_options_from_vocabulary(
-            "access_right")
-        self._serialize_agg(access_right_agg, access_right_labels)
+        buckets = agg_obj['buckets']
+        for bucket in buckets:
+            for k in list(bucket):
```
zzacharo, line 60: I am confused here....buckets is an array of objects as far as I remember....why are you iterating through it???

zzacharo (approved):

fenekku:
I rebased and will merge.

Issue #231, opened 2020-10-23
data model: support GeoJSON for geometry locations


ppanero:
This issue extends to `marshmallow-utils`. Approach:

Creating a marshmallow schema for GeoJSON in `marshmallow-utils`. On-going works in [`ppanero/geojson`](https://github.com/ppanero/marshmallow-utils/tree/geojson) branch. This schema is based on the [geojson python library](https://github.com/jazzband/geojson). [Marshmallow-geojson](https://github.com/folt/marshmallow-geojson) was evaluated but discarded due to potential maintainability issues and compatibility issues.

Once implemented it will be used in the `MetadataSchema`. Current works in [`ppanero/geojson`](https://github.com/ppanero/invenio-rdm-records/tree/geojson) branch.

PR #234, opened 2020-10-26
schema: add locations

ppanero:
Requires marshmallow-utils GeoJSONSchema merge and release https://github.com/inveniosoftware/marshmallow-utils/pull/21
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/231

**Important notes**
- The ES mappings `coordinates` field inside the `geometry` is of type`float`, however it correctly stores the locations for both `point` and `multipoint`
```
"coordinates": {
    "type": "float"
}
```

indexed:
```
"locations" : [
              {
                "geometry" : {
                  "type" : "point",
                  "coordinates" : [
                    3.5480075,
                    165.577299
                  ]
                },
                "identifiers" : {
                  "wikidata" : "12345abcde",
                  "geonames" : "12345abcde"
                },
                "place" : "Naga",
                "description" : "Random place on land..."
              },
              {
                "geometry" : {
                  "type" : "multipoint",
                  "coordinates" : [
                    [
                      1.973974,
                      49.308415
                    ],
                    [
                      -9.508502,
                      -21.08845
                    ]
                  ]
                },
                "place" : "Losheim"
              }
            ],
```

### DIFF
--- a/invenio_rdm_records/cli.py
+++ b/invenio_rdm_records/cli.py
@@ -176,6 +176,30 @@ def create_fake_record():
                     "scheme": "openaire"
                 }
             }],
+            "locations": [{
+                'geometry': {
+                    'type': 'Point',
+                    'coordinates': [
+                        float(fake.latitude()), float(fake.longitude())
+                    ]
+                },
+                "place": fake.location_on_land()[2],
+                "description": "Random place on land...",
+                'identifiers': {
+                    'wikidata': '12345abcde',
+                    'geonames': '12345abcde'
+                }
+            }, {
+                'geometry': {
+                    'type': 'MultiPoint',
+                    'coordinates': [
+                        [float(fake.latitude()), float(fake.longitude())],
+                        [float(fake.latitude()), float(fake.longitude())]
+                    ]
+                },
+                "place": fake.location_on_land()[2],
+            }
+            ],
             "references": [{
                 "reference": "Reference to something et al.",
                 "identifier": "9999.99988",
--- a/invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
@@ -471,14 +471,18 @@
             "type": "object",
             "additionalProperties": false,
             "properties": {
-              "point": {
+              "geometry": {
                 "type": "object",
-                "additionalProperties": false,
                 "properties": {
-                  "lat": {"$ref": "#/definitions/latitude"},
-                  "lon": {"$ref": "#/definitions/longitude"}
+                  "type": {
+                    "type": "string"
+                  },
+                  "coordinates": {
+                    "type": "array"
+                  }
                 }
               },
+              "identifiers": {"$ref": "#/definitions/identifiers"},
               "place": {
                 "description": "Place of the location",
                 "type": "string"
--- a/invenio_rdm_records/records/mappings/v6/rdmrecords/drafts/draft-v1.0.0.json
+++ b/invenio_rdm_records/records/mappings/v6/rdmrecords/drafts/draft-v1.0.0.json
@@ -177,21 +177,24 @@
             },
             "locations" : {
               "properties" : {
-                "description" : {
-                  "type" : "text"
+                "geometry": {
+                  "properties" : {
+                    "type" : {
+                      "type" : "keyword"
+                    },
+                    "coordinates" : {
+                      "type" : "float"
+                    }
+                  }
                 },
                 "place" : {
                   "type" : "text"
                 },
-                "point" : {
-                  "properties" : {
-                    "lat" : {
-                      "type" : "double"
-                    },
-                    "lon" : {
-                      "type" : "double"
-                    }
-                  }
+                "identifiers" : {
+                  "type" : "object"
+                },
+                "description" : {
+                  "type" : "text"
                 }
               }
             },
--- a/invenio_rdm_records/records/mappings/v7/rdmrecords/drafts/draft-v1.0.0.json
+++ b/invenio_rdm_records/records/mappings/v7/rdmrecords/drafts/draft-v1.0.0.json
@@ -176,21 +176,24 @@
           },
           "locations" : {
             "properties" : {
-              "description" : {
-                "type" : "text"
+              "geometry": {
+                "properties": {
+                  "type": {
+                    "type": "keyword"
+                  },
+                  "coordinates": {
+                    "type": "float"
+                  }
+                }
               },
               "place" : {
                 "type" : "text"
               },
-              "point" : {
-                "properties" : {
-                  "lat" : {
-                    "type" : "double"
-                  },
-                  "lon" : {
-                    "type" : "double"
-                  }
-                }
+              "identifiers" : {
+                "type" : "object"
+              },
+              "description" : {
+                "type" : "text"
               }
             }
           },
--- a/invenio_rdm_records/services/schemas/metadata.py
+++ b/invenio_rdm_records/services/schemas/metadata.py
@@ -20,6 +20,7 @@ from marshmallow import INCLUDE, Schema, ValidationError, fields, post_load, \
     validate, validates, validates_schema
 from marshmallow_utils.fields import EDTFDateString, GenFunction, \
     ISODateString, ISOLangString, SanitizedUnicode
+from marshmallow_utils.schemas import GeometryObjectSchema
 
 from .utils import validate_entry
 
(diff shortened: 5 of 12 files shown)


invenio_rdm_records/cli.py
```diff
@@ -139,13 +139,30 @@ def create_fake_record():
             }],
             "description": fake.text(max_nb_chars=3000),
             "locations": [{
-                "point": {
-                    "lat": str(fake.latitude()),
-                    "lon": str(fake.longitude())
+                'geometry': {
+                    'type': 'point',
```
lnielsen, line 143:
```suggestion
                    'type': 'Point',
```

invenio_rdm_records/cli.py
```diff
@@ -139,13 +139,30 @@ def create_fake_record():
...
-            }],
+                "description": "Random place on land...",
+                'identifiers': {
+                    'wikidata': '12345abcde',
+                    'geonames': '12345abcde'
+                }
+            },
+            {
+                'geometry': {
+                    'type': 'multipoint',
```
lnielsen, line 157:
```suggestion
                    'type': 'MultiPoint',
```

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -471,14 +471,18 @@
...
                 "type": "object",
-                "additionalProperties": false,
                 "properties": {
-                  "lat": {"$ref": "#/definitions/latitude"},
-                  "lon": {"$ref": "#/definitions/longitude"}
+                  "type": {
+                    "type": "string"
+                  },
+                  "coordinates": {
+                    "type": "object"
```
lnielsen, line 481: Minor: Is this not supposed to be an array?

invenio_rdm_records/records/mappings/v6/rdmrecords/drafts/draft-v1.0.0.json
```diff
@@ -177,21 +177,24 @@
...
                   "properties" : {
-                    "lat" : {
-                      "type" : "double"
+                    "type" : {
+                      "type" : "keyword"
                     },
-                    "lon" : {
-                      "type" : "double"
+                    "coordinates" : {
+                      "type" : "float"
```
lnielsen, line 186: Minor: If we start with an assumption that we can only accept Point objects, then type here should be ``geo_point`` (https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-point.html)

For later:
Later to deal with more than points we would have to extend with ``geo_shape`` (https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-shape.html) but that would require we develop a dumper that's able to split them correctly into separate indexing properties.

invenio_rdm_records/records/mappings/v6/rdmrecords/drafts/draft-v1.0.0.json
```diff
@@ -177,21 +177,24 @@
...
                     }
                   }
+                },
+                "place" : {
+                  "type" : "text"
+                },
+                "identifiers" : {
+                  "type" : "object"
```
lnielsen, line 194: Minor: should this not specify subtypes as well?

invenio_rdm_records/records/mappings/v7/rdmrecords/drafts/draft-v1.0.0.json
```diff
@@ -176,21 +176,24 @@
...
+                  "type": {
+                    "type": "keyword"
+                  },
+                  "coordinates": {
+                    "type": "float"
+                  }
+                }
               },
               "place" : {
                 "type" : "text"
```
lnielsen, line 190: Question: Should ``place`` be text or keyword?

invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -433,8 +434,9 @@ class PointSchema(Schema):
 class LocationSchema(Schema):
     """Location schema."""
 
-    point = fields.Nested(PointSchema)
+    geometry = fields.Nested(GeoJSONSchema, required=True)
```
lnielsen, line 437:
```suggestion
    geometry = fields.Nested(GeometryObjectSchema, required=True)
```

invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -20,6 +20,7 @@
     validate, validates, validates_schema
 from marshmallow_utils.fields import EDTFDateString, GenFunction, \
     ISODateString, ISOLangString, LocalizedEDTFDateString, SanitizedUnicode
+from marshmallow_utils.schemas import GeoJSONSchema
```
lnielsen, line 23:
```suggestion
from marshmallow_utils.schemas import GeometryObjectSchema
```

invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -433,8 +434,9 @@ class PointSchema(Schema):
 class LocationSchema(Schema):
     """Location schema."""
 
-    point = fields.Nested(PointSchema)
+    geometry = fields.Nested(GeoJSONSchema, required=True)
     place = SanitizedUnicode(required=True)
+    identifiers = fields.Dict()
```
lnielsen, line 439: Question: We don't have a schema for identifers that ensures that both keys and values are strings?

tests/records/full-record.json
```diff
@@ -118,12 +118,16 @@
       "lang": "eng"
     }],
     "locations": [{
-      "point": {
-        "lat": 1,
-        "lon": 2
+      "geometry": {
```
lnielsen, line 121: Comment: Here's an example of with Getty TGN and a Geoname (and CERN as example)
```
"locations": [{
      "geometry": {
        "type": "Point",
        "coordinates": [6.05, 46.23333]
      },
      "identifiers": {
        "geonames": "2661235",
        "tgn": "http://vocab.getty.edu/tgn/8703679"
      },
      "place": "CERN",
      "description": "Invenio birth place."
    }],
```

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
ppanero: Yeah, forgot to put it here was in the minutes of long ago. I didn't update the jsonschema/tests until we had an agreement. This was cuz I had to jump to another task.

tests/records/full-record.json
ppanero: Thanks a lot! :100:

invenio_rdm_records/records/mappings/v6/rdmrecords/drafts/draft-v1.0.0.json
ppanero: I would keep it as `float` for now. Mainly because I am worried about:

> Please note that string geo-points are ordered as lat,lon, while array geo-points are ordered as the reverse: lon,lat.
> Originally, lat,lon was used for both array and string, but the array format was changed early on to conform to the format used by GeoJSON.

Also the input will be `[float, float]`, but it will output `{lon: float, lat: float}`. We would need to account for that at (de)serialization. 

I would tackle this when doing the dumper. WDYT?

invenio_rdm_records/records/mappings/v6/rdmrecords/drafts/draft-v1.0.0.json
ppanero: Well, it is of type object which means we can add more key-values (i.e. type and subtype). However, see comment below on identifiers.

invenio_rdm_records/records/mappings/v7/rdmrecords/drafts/draft-v1.0.0.json
ppanero: My thought was that we want some analyzers and queries with certain levenstain distance. But if exact match is enough, I'll change to `keyword`.

invenio_rdm_records/services/schemas/metadata.py
ppanero: This is one of the point raised when developing the data model (cc @zzacharo). We have identifiers such as:

```
{
     identifiers: value,
     scheme: value
}
```

But also as:

```
{
     scheme-value: identifier-value
}
```

The second one was some of the ones than among others could cause the mapping explosion. But it was decided that there are not enough key possibilities for this to happen.

I understood from @zzacharo that when you two discussed about this (related to frontend) was decided that there was not gonna be any change in the data model.

As for a specific schema for this case (the second of the examples above), a Marshmallow schema would be complex because the key is unknown and therefore validation would be complex (and not much more than what we obtain with `Dict`. As for the first example, there is a schema and it is used when appropriate.

lnielsen: Ok. Then what we would need is some sort of identifier validator (on e.g. the Dict field (

```
identifiers = fields.Dict(validate=IdentifiersValidator(...))
```

invenio_rdm_records/records/mappings/v7/rdmrecords/drafts/draft-v1.0.0.json
lnielsen: Let's wait and see how until we know how the search behaves.

invenio_rdm_records/records/mappings/v6/rdmrecords/drafts/draft-v1.0.0.json
lnielsen: I think it's mainly the ES loading that's the issue (but don't we have the original source in ``_source``). Can you log an issue and assign to November board. Thanks.

lnielsen: The order (lon,lat) should be fine, as we have specified that we accept GeoJSON which is in that order.

ppanero: Done https://github.com/inveniosoftware/invenio-rdm-records/issues/265

PR #235, opened 2020-10-27
records-ui: makes version label dependant of the fields existence

Glignos:
* Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/335


invenio_rdm_records/theme/templates/invenio_rdm_records/record_landing_page.html
```diff
@@ -30,7 +30,9 @@
             <div class="left floated left aligned column">
               <span class="ui label teal"
```
lnielsen, line 31: Can you also remove the label around the publication date. I think there's an issue for it reported.

PR #240, opened 2020-11-02
serializers: localizes updated date

Glignos:
* Transfers UISchema to serializer
* Fixes missing access_right title

Closes inveniosoftware/invenio-app-rdm#333


invenio_rdm_records/resources/serializers.py
```diff
@@ -11,13 +11,26 @@
...
 
 from invenio_rdm_records.vocabularies import Vocabularies
 
 
 class UIJSONSerializer(JSONSerializer):
     """UI JSON serializer implementation."""
 
+    def _serialize_obj_dates(self, obj):
+        """Serializes objects dates to localized ones."""
+        localization_serializer = LocalizedEDTFDateString()
```
lnielsen, line 24: Minor: In this case you just have a normal date thus no need to use EDTF. Thus just use the normal Babel ``format_date()`` function. Example:

```python
from flask_babelex import format_date
from datetime import date
format_date(date(2020,3,24), format='long')
```

Glignos: In this case, what we are receiving is a `str(DateTime.now)` string, so we would need to provide a formatting string to convert it into a DateTime, to then format it again. The EDTF can handle non-range strings as well so it kinda "works out of the box" in this case.
I would be hesitant to add formatting strings if we have something more dynamic, unless there is a simpler solution I'm not aware of.

zzacharo: @lnielsen the `publication_date` can be actually an EDTF string no? I mean e.g can be a range date...Also, in the `LocalizedEDTFDateString` I had integrated the case of "Try to check if the date is already an ISO compatible string e.g '2020-10-01', and use flask-babel.format_date()". I guess, is fine to use always the field to produce what we want...

lnielsen: I added the comment in the wrong place..... the ``record.updated`` is an ISO-formatted datetime object which is not in EDTF format, and therefore should not use the EDTF formatting logic. Instead it should be something along the lines of:

```
obj['ui']['updated_date_l10n'] = format_date(arrow.get(obj['updated']), format='long')
```

Issue #244, opened 2020-11-02
landing page: description not being shown 


lnielsen:
Landing page is now showing description, despite it being there:

![Screenshot 2020-11-02 at 16 09 37](https://user-images.githubusercontent.com/1698163/97884061-ea7d2e80-1d25-11eb-9a81-529c7c9c7322.png)

It is being displayed on the search result page: 
![Screenshot 2020-11-02 at 16 09 51](https://user-images.githubusercontent.com/1698163/97884007-d802f500-1d25-11eb-88df-ca6d1ee54573.png)

If indeed there's no description the text should not display at all.

Issue #243, opened 2020-11-02
landing page: contributors roles not capitalised


lnielsen:
The contributor role is not capitalised on the landing page, so either it seems like the vocabulary has some sort of issue.

![Screenshot 2020-11-02 at 16 06 55](https://user-images.githubusercontent.com/1698163/97883750-8e1a0f00-1d25-11eb-990f-2c90d0b14f79.png)

PR #246, opened 2020-11-03
landing page: add more fields and use ui serializer

ppanero:
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/244
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/243
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/305
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/365
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/364

**Changes**

- Enriches the CLI demo records
- Cleans unused config, jinja filters, etc.
- Uses as much as possible the UI serialized object
- Adds bottom *details* section

**Current status/Remaining work** (In a different PR if agreed):

1- The publication date is formated as "long" according to flask-babelext. However, it does not seem to be the same "long" format we used to get from the babel jinja filter.
2- Resource type and access labels seem to not have the exact same size
3- Fix the content of the side bar. **Maybe we should remove place holders (e.g. metrics) since we are gonna have a _release_**
4- Add spacing after Details and before Resource type
Screenshot 2020-11-03 at 17 14 33

5- Add spacing after Details and before Resource type (Same as number 3 above)
6- Style all the content... Note that is just a print of the json objects
7- Add space before upload information
8- None of the 3 values come with the UI serilialized object (it is weird for created and updated to not be there 🤔 ). **How should vibility be calculated? Should it have the same text as the "access.title"**?
Screenshot 2020-11-03 at 17 14 43

### DIFF
--- a/invenio_rdm_records/theme/templates/invenio_rdm_records/record_landing_page.html
+++ b/invenio_rdm_records/theme/templates/invenio_rdm_records/record_landing_page.html
@@ -14,11 +14,11 @@
 {%- endblock css %}
 
 {%- from "invenio_rdm_records/macros/files.html" import file_list_box, preview_file_box %}
+{%- from "invenio_rdm_records/macros/detail.html" import show_detail %}
 
+{%- set record = record|serialize_ui %}
 {%- set metadata = record.metadata %}
-{%- set access = record.access %}
-{# FIXME: Take it from record.ui.access_right.category #}
-{%- set access_right_category = access.access_right %}
+
 {%- block page_body %}
 <div class="ui container">
   <div class="ui padded relaxed grid">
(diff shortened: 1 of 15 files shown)


invenio_rdm_records/theme/views.py
```diff
@@ -160,3 +124,11 @@ def vocabulary_subtitle(dict_key, vocabulary_key):
     vocabulary = Vocabularies.get_vocabulary(vocabulary_key)
```
lnielsen, line 124: I think you can completely remove vocabulary_title and vocabulary_subtitle. Instead the values should be available in the serialised record. @Glignos was working on adding it i think.

invenio_rdm_records/theme/templates/invenio_rdm_records/record_landing_page.html
```diff
@@ -28,23 +28,21 @@
         <div class="ui grid">
           <div class="two column row">
             <div class="left floated left aligned column">
-              <span class="ui"
-                title="Publication date">{{ metadata.publication_date|to_date|format_date(format='long') }}</span>
+              <span class="ui" title="Publication date">{{ record.ui.publication_date_l10n }}</span>
```
lnielsen, line 31: Minor: This should be in long format, so we need both a long and a short formatted publication date.

invenio_rdm_records/theme/views.py
```diff
@@ -160,3 +124,11 @@ def vocabulary_subtitle(dict_key, vocabulary_key):
...
     return vocabulary.get_title_by_dict(dict_key).split('/')[1].strip() \
         if vocabulary else ""
+
+
+@blueprint.app_template_filter('serialize_ui')
+def serialize_ui(record):
+    """Returns the UI serialization of a record."""
+    serializer = UIJSONSerializer()
+    # We need a dict not a string
+    return json.loads(serializer.serialize_object(record))
```
lnielsen, line 134: Minor: Could we instead refactor UIJSONSerializer a bit so that we can get the dict. Something a long the lines:

```
def serialize_object(self, obj, response_ctx=None, *args, **kwargs):
        return json.dumps(self.get_data(obj))

def get_data(self, obj):
      self._serialize_obj_ui(obj)
      self._serialize_obj_dates(obj)
      return obj
```

Then you can this instead:

```return UIJSONSerializer().get_data(record)```

invenio_rdm_records/theme/views.py
fenekku: :+1: on this because `vocabulary.get_title_by_dict(dict_key).split('/')[1].strip()` raises an IndexError when the "Poster" resource type is selected for instance (`[1]` -> `[-1]` other solution, but using the serialized record value is even better if possible)..

ppanero: @fenekku there is a separate issue for that. https://github.com/inveniosoftware/invenio-app-rdm/issues/340 I was planning to address it separately since this PR is growing a bit too much.

invenio_rdm_records/theme/views.py
ppanero: done :)

PR #272, opened 2020-11-11
serializers: makes contributors existance optional

Glignos:
* Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/388

### DIFF
--- a/invenio_rdm_records/resources/serializers/ui/schema.py
+++ b/invenio_rdm_records/resources/serializers/ui/schema.py
@@ -28,7 +28,9 @@ FormatDate = partial(FormatDate_, locale=get_locale)
 def make_affiliation_index(attr, obj, dummy_ctx):
     """Serializes creators/contributors for easier UI consumption."""
     # Copy so we don't modify in place the existing dict.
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -28,7 +28,9 @@
 def make_affiliation_index(attr, obj, dummy_ctx):
     """Serializes creators/contributors for easier UI consumption."""
     # Copy so we don't modify in place the existing dict.
-    creators = deepcopy(obj['metadata'][attr])
+    creators = deepcopy(obj['metadata'].get(attr))
+    if not creators:
+        return {}
```
lnielsen, line 33: Minor:I think this should be returning missing instead?

```suggestion
        return missing
```

lnielsen (commented):
Besides the tiny comment, everything looks good.

invenio_rdm_records/resources/serializers/ui/schema.py
Glignos: Indeed! Fixed.

PR #291, opened 2020-11-20
schema: update creator/contributor

### LINKED ISSUE #224: support name inference on contributors/creators
This feature is needed to properly support naming (and potentially helped deduplication) on the RDM data model.

The creators schema (which is inherited by contributors, therefore it applies to it too) has three attributes:

```python
name = SanitizedUnicode(required=True)
type = SanitizedUnicode(required=True, validate=validate.OneOf(
                choices=NAMES,
                error=_(f'Invalid value. Choose one of {NAMES}.')
            ))
given_name = SanitizedUnicode()
family_name = SanitizedUnicode()
```

`type` can be *organizational* or *personal*. 
- If `type == orgnizational` the required field is `name`.
- if `type == personal` the required fields are `given_name` and `family_name`, and `name` should be filled in with the *family_name, given_name* format (e.g. `Panero, Pablo`).

The successful solution would provide:
- [x] logic to infer names (e.g. `name` for the second case)
- [x] validation for the corresponding cases
- [x] tests

## Additional context

See `invenio_rdm_records/services/schemas/metadata.py` specially `ContributorsSchema` and `CreatorsSchema`.

fenekku:
- closes #224
- adds optional role to Creator (optional is fine, right?)


invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -78,6 +75,21 @@ class CreatorSchema(Schema):
...
+    @post_load
+    def fill_given_family_names(self, data, **kwargs):
+        """Fill the given_name and family_name based on name.
+
+        Allows to override conversion of name to family name / given name
+        if given_name or family_name is explicitly passed.
+        """
+        if data["type"] == "personal":
+            family_name, given_name = (
+                [n.strip() for n in data["name"].partition(",")][0::2]
```
lnielsen, line 87: This will not work in general. For personal names, we need people to explicitly specify the two fields, family name and given name. The only thing we can do is the reverse - i.e. ``name = f"{family_name}, {given_name}"``. 

Even though technically, you can tell people this is how it works, then they won't do it correctly. Zenodo has all the time, had a recommendation to enter the information correctly, but most people just copy/paste whatever name they already have written somewhere.

fenekku: Then what I will do is:

- Frontend: 
 * When "organizational" is chosen, show 1 name input field
 * When "personal" is chosen, show given name and last name input fields (only - don't show name input field)

- Backend:
 * When "personal": construct name from last name and first name
 * When "organizational": empty out first name and last name.

fenekku: Ready for re-review!

fenekku:
Merged in #293

PR #292, opened 2020-11-23
GeoJSON locations

### LINKED ISSUE #265: es: use `geo_point` and related types for geometries
Currently the `coordinates` property of the `locations` field only accepts `floats` (or an array of those).

It would be nice to take advantage of Elasticsearch's [`geo_point`](https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-point.html) and [`geo_shape`](https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-shape.html). Among others to facilitate bounding box searches.

Take into account possible restrictions that might need to be documented. For example:

>Please note that string geo-points are ordered as lat,lon, while array geo-points are ordered as the reverse: lon,lat.
> Originally, lat,lon was used for both array and string, but the array format was changed early on to conform to the format used by GeoJSON.

The idea is to create a `Dumper` that is able to distinguish the shape and dump it in the correct ES field (this would require changes to the mapping). This dumper should take care also of serializing it back (i.e. for searches).

alexsdutton:
`locations` is now a GeoJSON FeatureCollection:

* JSON schema updated to reference the GeoJSON schema, alongside InvenioRDM-specific constraints.
* ES schema now has `centroid` and `geometry` properties within `.locations.features`, as `geo_point` and `geo_shape` respectively
* `centroid` is inferred from `geometry` when indexing, using shapely if present and the geometry isn't a simple point
* properties other than `type`, `geometry`, `place`, `properties`, `description` and `identifiers` are forbidden within Features, to somewhat restrict arbitrary metadata without semantics. This also covers `centroid`, so a user-supplied field isn't silently dropped when re-loading from ES.

Open questions:

* Can we include the GeoJSON schema, so it's not dereferenced from the web at runtime? (Something to do with the geojson RefResolver)
* I found [simplestyle-spec](https://github.com/mapbox/simplestyle-spec/tree/master/1.1.0) which crystalises conventions for properties in GeoJSON features. In particular, it specifies `title` and `description`. Should we move to this for those two fields, instead of foreign members? (I'm aware that [DataCite](https://schema.datacite.org/meta/kernel-4.3/doc/DataCite-MetadataKernel_v4.3.pdf) has `geoLocationPlace` but this seems to be an established convention used by [various tools and services](https://github.com/mapbox/simplestyle-spec/wiki/Implementations))
* Should `shapely` be added as an extra (e.g. `geo`) and then required by `invenio-app-rdm` or the cookiecutter?

Closes #265.


invenio_rdm_records/records/dumpers.py
```diff
@@ -59,3 +65,46 @@ def load(self, data, record_cls):
         # `None` covers the cases where exceptions were raised in _dump
         parent_data.pop(f"{self.key}_start", None)
         parent_data.pop(f"{self.key}_end", None)
+
+
+class LocationsDumper(ElasticsearchDumperExt):
```
ppanero, line 70: Minor: I would create a `dumper` folder and then split this dumpers by "functionality" (i.e. `edtf.py`, `location.py`). It might be clearer.

invenio_rdm_records/records/dumpers.py
```diff
@@ -59,3 +65,46 @@ def load(self, data, record_cls):
...
+    This does a few things:
+
+    * Removes/adds FeatureCollection and Feature types, as those are
+      constant
+    * Calculates a centroid for each location if a geometry is provided
+
+    Centroid calculation requires shapely, except for the simple case where
+    the geometry is a point.
+    """
+
```
ppanero, line 82: moderate: I know we only have one "location" field, but is it worth to have the field `key` configurable as in the EDTF dumper? Not sure if this will ever be the case.

invenio_rdm_records/records/dumpers.py
```diff
@@ -59,3 +65,46 @@ def load(self, data, record_cls):
...
+    def dump(self, record, data):
+        """Dump the data."""
+        if 'locations' not in data:
+            return data
+        data['locations'].pop('type')
```
ppanero, line 87: minor: Is it worth to set it here to `FeatureCollection`? It seems that it is always removed by dumper and set to this value by loader. Not sure if it's value has some implication on ES queries.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -465,36 +465,46 @@
...
-                  "type": {
-                    "type": "string"
-                  },
-                  "coordinates": {
-                    "type": "array"
+          "description": "Geographical locations, as a GeoJSON FeatureCollection, with additional constraints.",
+          "$comment": "Everything here must be a FeatureCollection, but those features can have additional root-level properties (foreign members, in GeoJSON speak; see RFC7946 §6.1.",
+          "allOf": [
+            {
+              "$ref": "https://geojson.org/schema/FeatureCollection.json"
```
ppanero, line 472: moderate: IIRC a conversation with @slint, we avoided having this sort of references due to requests. This will make a request on each validation? It might result in failures when an instance is proxied or cut from the external world with firewalls, etc.

tests/records/test_dumpers.py
```diff
@@ -95,3 +96,181 @@ def test_esdumper_with_edtfext_parse_error(app, db, minimal_record):
...
+    with pytest.warns(UserWarning):
+        dump = record.dumps(dumper=dumper)
+
+    assert 'centroid' not in dump['locations']['features'][0]
+
+
+def test_locationsdumper_with_polygon_and_mock_shapely(
+    app, db, minimal_record
+):
+    with unittest.mock.patch(
```
ppanero, line 210: Nice! much clearer than the `@patch` decorator 🚀

invenio_rdm_records/records/dumpers.py
alexsdutton: Yep. This would certainly be a sensible idea!

invenio_rdm_records/records/dumpers.py
alexsdutton: I should defer to @lnielsen or @slint on this one.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
slint: Exactly, `jsonschema` [will use the `requests` library](https://python-jsonschema.readthedocs.io/en/latest/references/#jsonschema.RefResolver.resolve_remote) to fetch remote schemas, and cache them. We mitigated this in two ways:

- Not using remote schemas :see_no_evil:
- Using [custom ref resolvers](https://github.com/inveniosoftware/invenio-opendefinition/blob/9744f8e9b35bcc059ce6489763e944e5be906a75/invenio_opendefinition/resolvers.py#L23-L36), so that we control how/if any requests get done (e.g. then we could "vendor" the JSON Schema in the package and always use a local copy instead of making a remote request). Keep in mind that this is a bit tricky to implement and test, so I'm not sure if it's the best option at the moment

alexsdutton: I definitely agree that we should avoid external lookups. jsonschema lots you supply a [custom ref resolver](https://python-jsonschema.readthedocs.io/en/stable/references/), which means we can a) supply it locally, and b) forbid external resolution.

I did look at this, and say that we Invenio already uses `jsonresolver`, another Invenio project. invenio-records pays heed to an `invenio_records.jsonresolver` entry point group for hooking stuff in, but there are no instances of it being used. I can play though, and it shouldn't be too difficult to get working.

tests/records/test_dumpers.py
alexsdutton: I tried the decorator first, and pytest complained about the argument not being a fixture :shrug:. But yep, the context manager is good :-)

invenio_rdm_records/records/dumpers.py
alexsdutton: The idea was that because it was always constant, we didn't need to pass it to ES (https://github.com/inveniosoftware/invenio-rdm-records/issues/265#issuecomment-731277669)

lnielsen: Ouch, I think we have misunderstood each other here. What I had in mind here, was that don't store ``"type": "FeatureCollection/Feature"`` in neither the database nor Elasticsearch. Just no ``type`` in the JSONSchema nor the ES mapping. Instead we add the ``type`` in a resource serializer. Right now we only have a [UISerializer](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/ui/__init__.py#L18), but it would be similar......

Perhaps easiest to discuss IRL

invenio_rdm_records/records/dumpers.py
lnielsen: In this case, it would probably just leave it as is.

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
lnielsen: I think here we have to replicate parts of the JSONSchema instead. Would be nicer to just reference GeoJSON schema, but apart from the ref resolvers somehow seeming to work quite magical, then there's other issues:

1) Required properties: When we deposit a draft, we need to store in the same data structure, but we cannot enforce the required properties because then you cannot save a half finished upload.
2) ``properties`` is included, which we don't store at all.
3) There's extra stuff like bbox or geometry collection that we don't know how to deal with, so my preference is for having being more restrictive than what the GeoJSON standard allows.

alexsdutton:
@lnielsen and I had a quick catch-up call about this yesterday, and to summarise:

* We agreed I would change the schema to follow the shape of GeoJSON FeatureCollections, but constrain it to remove the opportunity to provide arbitrary elements. We'll follow datacite, but make it clear how the location data can be transformed into GeoJSON.
* I'll update the code to remove the external dereference of the GeoJSON JSON Schema, by either using a relative $ref, or hooking into jsonresolver as e.g. https://github.com/inveniosoftware/invenio-openaire/blob/master/invenio_openaire/resolvers/funders.py does.

invenio_rdm_records/records/dumpers.py
alexsdutton: :+1:

invenio_rdm_records/records/dumpers.py
alexsdutton: Now done :-)

invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
alexsdutton: We're now using a copy of the Geometry schema which I've embedded in the schema definitions section. The schema for locations is now tied down to not have properties.

invenio_rdm_records/records/dumpers.py
alexsdutton: I'm going to resolve this per the first bullet point in https://github.com/inveniosoftware/invenio-rdm-records/pull/292#issuecomment-734183741.

alexsdutton:
This is ready for re-review. I need to squash the fixup commits and reword the first commit before it's merged.

ppanero (approved):

alexsdutton:
@ppanero All ready to merge! :rocket:

PR #295, opened 2020-11-24
dumpers: add EDTFListDumperExt, change mappings

max-moser:
Addresses issue #223

* add EDTFListDumperExt to handle arrays of dates (e.g. 'metadata.dates') in addition to the existing EDTFDumperExt, which handles scalar values (e.g. 'metadata.publication_date')
* remove the date-type ES mappings KEY_start and KEY_end
* replace them with the date_range-type ES mapping KEY_range
* allows for simple date range queries (e.g. q=date:[1939-01-01 TO 1945-12-31])

### DIFF
--- a/invenio_rdm_records/records/api.py
+++ b/invenio_rdm_records/records/api.py
@@ -20,7 +20,7 @@ from invenio_vocabularies.records.models import VocabularyType
 from werkzeug.local import LocalProxy
 
 from . import models
-from .dumpers import EDTFDumperExt
+from .dumpers import EDTFDumperExt, EDTFListDumperExt
 
 
 class Language(Vocabulary):
@@ -54,11 +54,13 @@ class BibliographicRecord(Record):
     model_cls = models.RecordMetadata
 
     index = IndexField(
-        'rdmrecords-records-record-v1.0.0', search_alias='rdmrecords-records')
+        "rdmrecords-records-record-v1.0.0", search_alias="rdmrecords-records"
+    )
 
     dumper = ElasticsearchDumper(
         extensions=[
             EDTFDumperExt('metadata.publication_date'),
+            EDTFListDumperExt("metadata.dates", "date"),
             RelationDumperExt('relations'),
         ])
 
@@ -91,11 +93,13 @@ class BibliographicDraft(Draft):
     model_cls = models.DraftMetadata
 
     index = IndexField(
-        'rdmrecords-drafts-draft-v1.0.0', search_alias='rdmrecords-drafts')
+        "rdmrecords-drafts-draft-v1.0.0", search_alias="rdmrecords-drafts"
+    )
 
     dumper = ElasticsearchDumper(
         extensions=[
             EDTFDumperExt('metadata.publication_date'),
+            EDTFListDumperExt("metadata.dates", "date"),
             RelationDumperExt('relations'),
         ])
 
--- a/tests/records/test_dumpers.py
+++ b/tests/records/test_dumpers.py
@@ -139,7 +139,7 @@ def test_eslistdumper_with_edtfext_parse_error(app, db, minimal_record):
     """Test edft extension implementation."""
     dumper = ElasticsearchDumper(
         extensions=[
-            EDTFListDumperExt("metadata.creators", "name"),
+            EDTFListDumperExt("metadata.creators", "family_name"),
         ]
     )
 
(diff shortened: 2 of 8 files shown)


invenio_rdm_records/records/api.py
```diff
@@ -23,10 +23,15 @@ class BibliographicRecord(Record):
     model_cls = RecordMetadata
 
     index = IndexField(
-        'rdmrecords-records-record-v1.0.0', search_alias='rdmrecords-records')
+        "rdmrecords-records-record-v1.0.0", search_alias="rdmrecords-records"
```
ppanero, line 26: minor: I am in favour of consistency, but I am not sure if this sort of "styling" changes are good for history tracking :)

invenio_rdm_records/records/dumpers.py
```diff
@@ -17,14 +17,23 @@
...
-    recent last), sort by publication_date_start to get a natural sort order.
-    In a descending sort (most recent first), sort by publication_date_end.
+    It parses the value (i.e. the EDTF level 0 string) of the specified
+    field in the document and adds a dictionary holding the values
+    :code:`{"lte": lower_strict, "gte": upper_strict}`, whose values correspond
+    to the lower strict and upper strict bounds.
+    They are required for sorting.
+    In an ascending sort (most recent last), sort by publication_date_start
+    to get a natural sort order. In a descending sort (most recent first),
+    sort by publication_date_end.
```
ppanero, line 36: moderate: This is not the case anymore, now we need to sort by `publication_date_range.gte` or `lte` right? I think it will still work, just needs updating this docstring.

invenio_rdm_records/records/dumpers.py
```diff
@@ -57,5 +65,59 @@ def load(self, data, record_cls):
...
+
+    It iterates over the items at the specified field in the
+    document and adds dictionaries holding the values
+    :code:`{"lte": lower_strict, "gte": upper_strict}`, as parsed
+    from their EDTF fields (specified by the key).
+    These values correspond to the lower strict and upper strict bounds.
+    They are required for sorting.
+    In an ascending sort (most recent last), sort by publication_date_start
+    to get a natural sort order. In a descending sort (most recent first),
+    sort by publication_date_end.
```
ppanero, line 82: moderate: same as above, update docstring.

invenio_rdm_records/records/dumpers.py
```diff
@@ -57,5 +65,59 @@ def load(self, data, record_cls):
...
+    """
+
+    def __init__(self, list_field, key):
+        """Constructor.
+
+        :param list_field: dot separated path to the array to process.
+        :param key: name of the EDTF field in each array item.
```
ppanero, line 89: minor: should specify it has to be the same name in each item of the list, that `key` does not accept another list.

invenio_rdm_records/records/dumpers.py
```diff
@@ -57,5 +65,59 @@ def load(self, data, record_cls):
...
+        """
+        super(EDTFListDumperExt, self).__init__()
+        self.keys = parse_lookup_key(list_field)
+        self.key = key
+        self.range_key = "{}_range".format(self.key)
+
+    def dump(self, record, data):
+        """Dump the data."""
+        try:
+            date_list = dict_lookup(data, self.keys, parent=False)
```
ppanero, line 99: question: why do we have to use `parent=False` here and in the single date dumper we can use `parent=True`? It has to do with how the `self.key` is assigned I guess. I am just wondering if this can be refactored into a parent class since both the constructor and the code inside the loop is duplicated in both classes.

tests/records/test_dumpers.py
```diff
@@ -12,86 +12,150 @@
...
-    ("2021-01/2021-03", "2021-01-01", "2021-03-31")
-])
-def test_esdumper_with_edtfext(app, db, minimal_record,
-                               date, expected_start, expected_end):
+from invenio_rdm_records.records.dumpers import EDTFDumperExt, \
+    EDTFListDumperExt
+
+
+@pytest.mark.parametrize(
+    "date, expected_start, expected_end",
```
ppanero, line 20: minor: is okay, but for the same reason as above maybe for the future is better not to add change lines for simple styling stuff, mainly to keep history untouched. I am guessing is just a formatter that's configured on-save.

ppanero (approved):
LGTM. To main points:
1- Let's try not to include change lines for styling issues.
2- Some code is duplicated and could be made into auxiliary functions. But how to do it or if it would be clearer is not fully clear to me. So shelved.

Nice job! Thanks for the contribution 🚀

invenio_rdm_records/records/dumpers.py
lnielsen: In ``publication_date``, we need access to add the extra keys at the same level as publication date:

```
{
  "publication_date": "..."
  "publication_date_range": ...
}
```
For dates, we're getting the list, and need to add the objects inside each item, so there's no need to get the parent:

```
[
{
  "date": "..."
  "date_range": ...
}
]
```

tests/records/test_dumpers.py
max-moser: oh yes, sorry, that was some **Black** magic in action

PR #320, opened 2020-12-15
Views registration

ppanero:
Requires https://github.com/inveniosoftware/invenio-drafts-resources/pull/86
Requires https://github.com/inveniosoftware/invenio-records-resources/pull/185

Changes:

- Adapts config loading 
- Adds access component
- Registers views


fenekku (commented):
Trying to be thorough in summary so when coming back from break it's clear.

- Adds access component :+1: 
- Move UI to app-rdm :+1: 
- adapt config loading : I am not sure about...

I will summarize here the main discussion point I discussed with Pablo. Really it applies more to 
- the pattern used in invenio-vocabularies: https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/ext.py#L34 
- the removal of the load from config in : https://github.com/inveniosoftware/invenio-records-resources/pull/185/files#diff-ba4a689688ee572bffc8d2dccf376ed194810cc1ba8e32b6df22832d2acc9ebeL43



What it boils down to is: What is the problem we are trying to solve by removing `load_or_import_from_config` [here](https://github.com/inveniosoftware/invenio-records-resources/pull/185/files#diff-ba4a689688ee572bffc8d2dccf376ed194810cc1ba8e32b6df22832d2acc9ebeL43), and creating Resources + Services in ext.py and registering them in views.py (as opposed to doing both in views.py as before)? I don't grasp the advantages of the refactor and in fact I see one disadvantage. Perhaps I am not seeing something...

I don't see the difference between https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/views.py#L135 (pushing an app context explicitly) and its split in `invenio_rdm_records/ext.py` and `invenio_rdm_records/views.py`. It keeps the status quo, but involves large code move.

The removal of `load_or_import_from_config` is more insidious. It allows a developer to create a given Resource or Service without an app context. In code (not tests), we still only create Resources and Services in app context, so we are good... so far. Nothing prevents other developers to create a given Resource whenever. The moment they do, they have a Resource who's given configuration is NOT overridden by the instance's configuration. This is problematic.

Last thing to note is that Resource creation https://github.com/inveniosoftware/invenio-rdm-records/pull/320/files/bb0cb4660f5c185b02e0d7bc14c27e50e186ae10#diff-2fead0b1b4037156fc74ab4485f4993f64e78c504f3978cea57fa5ec940c8de8R79  makes ext.py need to know the name of the config key when that concern should be encapsulated/defined by the Resource itself.

ppanero:
> Last thing to note is that Resource creation https://github.com/inveniosoftware/invenio-rdm-records/pull/320/files/bb0cb4660f5c185b02e0d7bc14c27e50e186ae10#diff-2fead0b1b4037156fc74ab4485f4993f64e78c504f3978cea57fa5ec940c8de8R79 makes ext.py need to know the name of the config key when that concern should be encapsulated/defined by the Resource itself.

Has been fixed by using the already existing `config_name` attr of the resources/services classes. **e.g**:

```python
        self.resources = Resource(
            service=self.service,
            config=app.config.get(Resource.config_name),
        )
```

As for the rest, in my opinion, the main advantage is being able to declare instances of a service or resource without to **always have to explicitly pass a configuration**. This happens because the `app` context access of the `load_or_import...`. This can also be "fixed" pushing an application context (as done [here](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/theme/views.py#L137)). Therefore the bottom line is:

*Adding always manually the application context (`with...`)* VS *use an already existing `app` in the ext.py*

I wonder which is the cost of pushing many app contexts (7 in this case) instead of using an existing app. Apart from that, is a matter of *choosing a pattern* and sticking to it and which direction this has to go. @lnielsen What is your take on it?

fenekku (approved):
Talked about this with Lars before the break so I get it better now : main reason for moving initialization to `ext.py` is to get localproxy access to the current resource / service. 

For the config_name part, if we keep the config loading from before we won't even need to do:

```python
 self.records_service = BibliographicRecordService(
            config=app.config.get(BibliographicRecordService.config_name),
        )
```
(https://github.com/inveniosoftware/invenio-rdm-records/pull/320/files#diff-2fead0b1b4037156fc74ab4485f4993f64e78c504f3978cea57fa5ec940c8de8R78-R80 and subsequent others)

We will be able to do:

```python
# (push app context if not pushed)
    self.records_service = BibliographicRecordService()
```

Approving. We can keep the config loading now or do it separately.

ppanero:
@fenekku I like the idea. Merging for this release, but let's make it happen. Something like:

```
self.records_service = BibliographicRecordService()
```

Would be ideal, still not happy about pushing a context every time, but we can try to find a way around it.

PR #322, opened 2021-01-08
schema and dumper: account for empty partial metadata in drafts

ppanero:
Why?

Because the partially saved draft with no content sends `metadata: {}` which gets removed by the `clear_none` function in the `services.update_draft` method. This has two consequences:

- Gets indexed without it (therefore the dumper try except)
- Needs to be set back to `{}` in the post_dump serialization. Due to a marshmallow issue the `default` value cannot be used (it does not work when the value is `None` on `Nested` fields).

### DIFF
--- a/invenio_rdm_records/records/dumpers.py
+++ b/invenio_rdm_records/records/dumpers.py
@@ -54,8 +54,13 @@ class EDTFDumperExt(ElasticsearchDumperExt):
 
     def load(self, data, record_cls):
         """Load the data."""
-        parent_data = dict_lookup(data, self.keys, parent=True)
+        try:
+            parent_data = dict_lookup(data, self.keys, parent=True)
 
-        # `None` covers the cases where exceptions were raised in _dump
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/records/dumpers.py
```diff
@@ -54,8 +54,13 @@ def dump(self, record, data):
...
+            parent_data = dict_lookup(data, self.keys, parent=True)
 
-        # `None` covers the cases where exceptions were raised in _dump
-        parent_data.pop(f"{self.key}_start", None)
-        parent_data.pop(f"{self.key}_end", None)
+            # `None` covers the cases where exceptions were raised in _dump
+            parent_data.pop(f"{self.key}_start", None)
+            parent_data.pop(f"{self.key}_end", None)
+        except KeyError:
+            # Drafts partially safe with no data
```
zzacharo, line 64: small typo -> `partially saved`

zzacharo (approved):

PR #332, opened 2021-01-15
tests: split dumpers tests and add edtf query

ppanero:
- Splits dumpers' tests in several files to ease comprehension
- Adds a EDTF test case: search. With year-level range, day-level range, and a year-level range that should not return the result.


max-moser (approved):
LGTM :+1:

Issue #331, opened 2021-01-15
automatic detection of identifiers


ppanero:
## Is your feature request related to a problem? Please describe.

When submitting an identifier (e.g. a creator's ORCID) the backend could automatically detect the scheme.

## Describe the solution you'd like

A generic field e.g. `AutoschemeIdentifier` could be added to `marshmallow-utils`, then used in the service level schema. Then `invenio-rdm-records` will create specific fields with an extra layer of validation, to only allow specific schemas for specific fields (e.g. RoR scheme only allowed for organization, not for person).

## Additional context

Logic can be migrated from [Zenodo's implementation](https://github.com/zenodo/zenodo/blob/8af9ba26bbf6d866b1f3451c6f30d49f098d7799/zenodo/modules/records/serializers/fields/persistentid.py#L36).

It should:

- Detect idenfier's type (scheme)
- If cannot detect, throw a validation error
- If can detect, then:
    - Normalize
    - Save
    - Return it in the response (along with the full record).
- If detected multiple possibilities, then default to the first one returned by `idutils`

ppanero:
The Marshmallow field is created in this PR https://github.com/inveniosoftware/marshmallow-utils/pull/25 while integration in invenio-rdm-records will happen separately (in a PR that will close this issue).

PR #333, opened 2021-01-18
schema: use identifier scheme from marshmallow utils

ppanero:
Depends on https://github.com/inveniosoftware/marshmallow-utils/pull/25
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/331
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/458

Uses Marshmallow-Utils new `IdentifierScheme` which "guesses" the scheme if not given and supports allowing a specific set of schemes.

Migrates identifiers from `dict` to `list` of dicts for contributors, creators and affiliations.

Tests are passing with local installation of marshmallow-utils dependency.

### DIFF
--- a/invenio_rdm_records/services/schemas/metadata.py
+++ b/invenio_rdm_records/services/schemas/metadata.py
@@ -9,45 +9,34 @@
 """RDM record schemas."""
 
 import time
+from functools import partial
 from urllib import parse
 
 import arrow
-import idutils
 from edtf.parser.grammar import level0Expression
 from flask import current_app
 from flask_babelex import lazy_gettext as _
 from marshmallow import EXCLUDE, INCLUDE, Schema, ValidationError, fields, \
-    post_load, validate, validates, validates_schema
+    post_load, pre_load, validate, validates, validates_schema
 from marshmallow_utils.fields import EDTFDateString, GenFunction, \
-    ISODateString, ISOLangString, SanitizedUnicode
-from marshmallow_utils.schemas import GeometryObjectSchema
+    IdentifierSet, ISODateString, ISOLangString, SanitizedUnicode
+from marshmallow_utils.schemas import GeometryObjectSchema, IdentifierSchema
 
 from .utils import validate_entry
 
 
-def _no_duplicates(value_list):
-    str_list = [str(value) for value in value_list]
-    return len(value_list) == len(set(str_list))
+def _not_blank(error_msg):
+    """Returns a non-blank validation rule with custom error message."""
+    return validate.Length(min=1, error=error_msg)
 
 
 class AffiliationSchema(Schema):
     """Affiliation of a creator/contributor."""
 
     name = SanitizedUnicode(required=True)
-    identifiers = fields.Dict()
-
-    @validates("identifiers")
-    def validate_identifiers(self, value):
-        """Validate well-formed identifiers are passed."""
-        if len(value) == 0:
-            raise ValidationError(_("Invalid identifier."))
-
-        for identifier in value.keys():
-            validator = getattr(idutils, 'is_' + identifier, None)
-            # NOTE: identifier key cannot be empty string
-            if not identifier or (validator and
-                                  not validator(value.get(identifier))):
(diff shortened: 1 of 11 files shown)


tests/services/schemas/test_creator_contributor.py
```diff
@@ -205,14 +193,18 @@ def test_creator_invalid_identifiers_scheme():
...
+        "identifiers": [{
+            "scheme": "unapproved scheme",
+            "identifier": "0000-0002-1825-0097",
+        }]
     }
 
     assert_raises_messages(
         lambda: CreatorSchema().load(invalid_scheme),
-        {'identifiers': ["Invalid value. Choose one of ['orcid', 'ror']."]}
+        {'identifiers': {0: {'_schema': [
```
ppanero, line 204: This might affect error handling on the UI, but is simply passing from dict to array (e.g. make use of `0` etc indexes).

tests/services/schemas/test_related_identifier.py
```diff
@@ -159,6 +161,7 @@ def test_invalid_related_identifiers(app, minimal_record):
         data = MetadataSchema().load(metadata)
 
 
+@pytest.mark.skip()
```
lnielsen, line 164: Question: Why the skip?

invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -72,50 +65,26 @@ class CreatibutorSchema(Schema):
...
-        """Validate identifier based on type."""
-        if data['type'] == "personal":
-            person_identifiers = ['orcid']
-            identifiers = data.get('identifiers', {}).keys()
-            if any([i not in person_identifiers for i in identifiers]):
-                messages = [
-                    _(f"Invalid value. Choose one of {person_identifiers}.")
-                ]
-                raise ValidationError({"identifiers": messages})
+    def validate_identifiers_scheme(self, data, **kwargs):
```
lnielsen, line 77: Minor: First of all, it might just be me being confused here 🤔 What's the purpose of this method and why is it not already validated by ``IdentifierSet`` and ``IdentifierSchema``?

ppanero: This is the sort of "extra" case we talked IRL about `ror` only for organizations and `isni/orcid` only for person. The thing is at `IdentifierSchema` level it checks an allowed list of schemes... but knowing the type and this extra check is responsability of the `PersonOrOrgSchema`...

tests/services/schemas/test_related_identifier.py
ppanero: Discussed IRL, deduplication is:
- Incorrect the way it was before. i.e. there can be more than one related DOI work...
- An overkill to do a complex deduplication validation

invenio_rdm_records/services/schemas/metadata.py
ppanero: Discussed IRL, opted to remove this validation. The reasons are:

- Migrating previous data (e.g. Zenodo, and Zenodo-like clones) that for one reason or another did not check this and will have a very difficult time importing data.
- From the UI we can control this by not displaying the PID schemes, and in the backend the correct one will be selected.
- Downside: from the API they can be inserted wrongly... a leap of faith here.

tests/services/schemas/test_location.py
```diff
@@ -66,6 +73,7 @@ def test_invalid_empty(valid_full_location):
         data = LocationSchema().load({})
 
 
+@pytest.mark.skip(reason="idutils cannot validate geonames, wikidata, etc.")
```
ppanero, line 76: Note this happens in several places (subjects, rights, locations..)

setup.py
```diff
@@ -64,12 +64,12 @@
     'arrow>=0.17.0',
     'Faker>=2.0.3',
     'ftfy>=4.4.3,<5.0.0',
-    'idutils>=1.1.7',
```
ppanero, line 67: comes from marshmallow-utils, which now is required here to get the new `IdentifierSet` and `IdentifierScheme`

lnielsen (approved):

PR #334, opened 2021-01-19
refactor contributors

ppanero:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/460

Note that no value (e.g. roles) are made into vocabularies due to cross-works in invenio-vocabularies. In addition, there is a blocker at `Relations` field in `api.py` level. The relation should be set to `metadata.contributors.role` but contributors (and creators) are arrays.

### DIFF
--- a/tests/resources/test_publish_errors.py
+++ b/tests/resources/test_publish_errors.py
@@ -47,14 +47,18 @@ def test_nested_field_error(
         client, minimal_record, location, es_clear, headers):
     minimal_record["metadata"]["creators"] = [
         {
-            "name": "Julio Cesar",
-            "type": "personal",
-            "given_name": "Julio",
-            "family_name": "Cesar",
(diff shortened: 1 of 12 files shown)


tests/resources/test_publish_errors.py
```diff
@@ -68,7 +72,7 @@ def test_nested_field_error(
     assert 400 == response.status_code
     expected = [
         {
-            "field": "metadata.creators.1.name",
+            "field": "metadata.creators.1.person_or_org.name",
```
fenekku, line 75: I've added a note to review frontend Formik fieldpaths in light of these changes in creators task:   https://github.com/inveniosoftware/react-invenio-deposit/issues/154#issuecomment-763700142

fenekku (approved):
Nice!

PR #369, opened 2021-02-09
serializers: call superclass methods in UIJSONSerializer

### LINKED ISSUE #367: Align UIJSONSerializer and JSONSerializer
Recent changes in JSONSerializer are not available in UIJSONSerializer because methods are re-implemented that shouldn't.

e.g. 

In Flask-Resources https://github.com/inveniosoftware/flask-resources/blob/master/flask_resources/serializers/json.py

```python
def serialize_object(self, obj):
        """Dump the object into a json string."""
        encoder = self.encoder
        return json.dumps(obj, cls=encoder, **self.dumps_options)
```

but here:

```python
def serialize_object(self, obj):
        """Dump the object into a JSON string."""
        return json.dumps(self.dump_obj(obj))
```

When it should be:

```python
def serialize_object(self, obj):
        """Dump the object into a JSON string."""
        return super().serialize_object(self.dump_obj(obj))
```

Check for `serialize_list` too.

max-moser:
closes #367 

previously, UIJSONSerializer just called json.dumps() directly, without taking into account the dumps_options set through the constructor (via its superclass, JSONSerializer)

### DIFF
--- a/invenio_rdm_records/resources/serializers/ui/__init__.py
+++ b/invenio_rdm_records/resources/serializers/ui/__init__.py
@@ -8,8 +8,6 @@
 
 """Record response serializers."""
 
-import json
-
 from flask_resources.serializers import JSONSerializer
 
 from .schema import UIListSchema, UIObjectSchema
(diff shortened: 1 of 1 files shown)


fenekku (approved):
:+1: 

(Looking longer at the dump_obj method above, it really shouldn't be modifying the `obj` I think, but that's for another time...)

PR #380, opened 2021-02-12
372 access system field

### LINKED ISSUE #372: access: new system field for managing access
Create a new system field for managing the access data. Main purpose of the system field is to 
- enforce data integrity.
- make working with a record from the service layer easier

### System field

Adding the access field should allow us to use a record similar to:

```
# Add a user into the owners, and ensure that it's a set.
u = User()
record.access.owned_by.add(u)
record.access.owned_by == {User(1)}
# Set grants
grant  = Grant(User(1), 'view')
record.access.grants.add(grant)
# Get allowed needs for a given permission level
record.access.grants.needs('viewmeta')
# Set protection 
record.access.protection.set('restricted')
record.access.protection.set('public', files='restricted')
record.access.protection == {'metadata': 'public', 'files': 'restricted'}
# Set embargo
record.access.embargo.set(datetime(....), reason="....") # active is automatically calculated by timestamp
record.access.embargo.clear()
record.access.embargo.lift()
```

### Grants

The above shows a grant class. This grant class should be responsible for:

- Grants: ensure that there are no duplicates, that only the highest level is retained.
- Grant class: should encode how grant tokens are generated from a single grant.

max-moser:
closes #372 
requires https://github.com/inveniosoftware/invenio-rdm-records/pull/379, because the system field uses the new data structures

### DIFF
--- a/invenio_rdm_records/records/systemfields/access/embargo.py
+++ b/invenio_rdm_records/records/systemfields/access/embargo.py
@@ -15,6 +15,7 @@ class Embargo:
     """Embargo class for the access system field."""
 
     def __init__(self, until, reason=None, active=None):
+        """Create a new Embargo."""
         self.until = until
         if isinstance(self.until, str):
             self.until = arrow.get().datetime
@@ -24,6 +25,7 @@ class Embargo:
 
     @property
     def active(self):
+        """Whether or not the Embargo is (still) active."""
         if self._active is not None:
             return self._active
 
@@ -31,15 +33,18 @@ class Embargo:
 
     @active.setter
     def set_active(self, value):
+        """Set the Embargo's (boolean) active state."""
         self._active = value
 
     def clear(self):
-        # TODO
+        """Completely remove the embargo."""
+        # TODO remove the embargo information entirely
         pass
 
     def lift(self):
-        # TODO
-        pass
+        """Update the embargo active status if it has expired."""
+        if arrow.utcnow().datetime < self.until:
+            self.active = False
 
     def dump(self):
         """Dump the embargo as dictionary."""
@@ -50,11 +55,13 @@ class Embargo:
         }
 
     def __repr__(self):
+        """Return repr(self)."""
         return "<{} (active: {}, until: {}, reason: {})>".format(
             type(self).__name__, self.active, self.until, self.reason
         )
 
     def __bool__(self):
+        """Return bool(self)."""
         return self.active
 
     @classmethod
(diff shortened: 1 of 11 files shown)


invenio_rdm_records/records/mappings/v6/rdmrecords/drafts/draft-v1.0.0.json
```diff
@@ -12,20 +12,47 @@
...
-            "access_right" : {
+            "record" : {
               "type" : "keyword"
             },
-            "created_by" : {
-              "type" : "long"
+            "files" : {
+              "type" : "keyword"
+            },
+            "owned_by" : {
```
lnielsen, line 21: Comment: I think this is correct, but we have to validate that this structure properly supports the queries we have to make.

Say e.g. we need to query for all records owned by a given user:

``acccess.owned_by.user:1``

invenio_rdm_records/records/systemfields/access/embargo.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class Embargo:
+    """Embargo class for the access system field."""
+
+    def __init__(self, until, reason=None, active=None):
+        """Create a new Embargo."""
+        self.until = until
+        if isinstance(self.until, str):
+            self.until = arrow.get().datetime
```
lnielsen, line 21:
```suggestion
            self.until = arrow.get(until).datetime
```
?

invenio_rdm_records/records/systemfields/access/embargo.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def clear(self):
+        """Completely remove the embargo."""
+        # TODO remove the embargo information entirely
+        pass
+
+    def lift(self):
+        """Update the embargo active status if it has expired."""
+        if arrow.utcnow().datetime < self.until:
+            self.active = False
```
lnielsen, line 47: Comment: I would return a boolean that allows the caller to determine if the embargo was lifted or not.

invenio_rdm_records/records/systemfields/access/embargo.py
```diff
@@ -0,0 +1,78 @@
...
+            "reason": self.reason,
+        }
+
+    def __repr__(self):
+        """Return repr(self)."""
+        return "<{} (active: {}, until: {}, reason: {})>".format(
+            type(self).__name__, self.active, self.until, self.reason
+        )
+
+    def __bool__(self):
```
lnielsen, line 63: 👍

lnielsen:
Overall looks good 👍 Just needs some tests to see that it works as expected 😄

PR #382, opened 2021-02-16
generator: IfRestricted

### LINKED ISSUE #376: access: new permission policy
Implement a new permission policy, that uses the access data to restrict access to records. Will need:

- [ ] Need generators for permission levels.
- [ ] Define the permission policy similar to below (perhaps a separate ticket is needed for the handling the files permissions)
- [x] Extensive tests for the permission policy, as well as the usage of the policy inside the service.

We need two need generators:
- ``IfRestricted``
- ``RecordPermissionLevel``

Both need generators have:
- Given a record produce required needs (see (1) below)
- Given an identity produce a query filtering records to only visible records (see (2) below).

### Permission policy

The policy should look something like this:

```python
class RDMRecordPermissionPolicy(...):
    can_search = [AnyUser()]
    can_create = [AuthenticatedUser()]
    can_read = [IfRestricted('metadata', RecordPermissionLevel('viewmeta'), else_=AnyUser())] 
    can_update = [Disable()]
    can_delete = [Disable()]

    can_read_files = [IfRestricted('files', RecordPermissionLevel('viewfull'), else_=AnyUser())]
    can_updates_files = [Disable()]

    can_read_draft = [RecordPermissionLevel('edit)]
    can_update_draft = [RecordPermissionLevel('edit)]
    can_delete_draft = [RecordPermissionLevel('manage')]

    can_read_draft_files = [RecordPermissionLevel('edit')]
    can_read_update_files = [RecordPermissionLevel('edit')]
    
    can_publish = [RecordPermissionLevel('edit')]
    can_manage = [RecordPermissionLevel('manage')]
```

### 1. Record to Needs

Given a record:

```python
{
    "access": {
        "owned_by": [{"user": 1}]
        "record": "restricted",
        "files": "restricted",
        "grants": [
             {"subject": "user", "id": "2", "level": "edit"},
             {"subject": "role", "id": "curator", "level": "edit"},
             {"subject": "user", "id": "3", "level": "viewmeta"},
         ]
    }
}
```

The action ``can_read_files`` defined as ``[IfRestricted('files', RecordPermissionLevel('viewfull'), else_=AnyUser())]`` should require the following needs:

```
# delegate to record.access.needs('viewfull') to make the needs:
UserNeed(1), UserNeed(2), RoleNeed("curator")
```

Note:
- Owners are included because they have implicit "manage" permission.
- User 2 and role curator both have edit level so they have implicit viewfull-level
- User 3 only has viewmeta level so the user is not included.
- All of above should be done and working in the system field (#372)

### 2. Identity to Query

The query would take an identity providing something like (note we may need special handling of any user/authenticated user to handle it efficiently:

```
UserNeed(1), RoleNeed('curator'), SystemRole('authenticated_user'), SystemRole('any_user')
```

and make a query similar to:

```
access.record:public 
OR (
    access.record:restricted AND (
        access.tokens:view-user-1 OR 
        access.tokens:view-role-curator OR          
        access.tokens:view-sysrole-authenticated_user OR
        access.tokens:view-sysrole-any_user
    )
)
```

mb-wali:
partially closes #376

**test failing reason**: requires new version of [invenio-records-permissions](https://github.com/inveniosoftware/invenio-records-permissions) to access ```AuthenticatedUser``` generator.

### DIFF
--- a/invenio_rdm_records/records/systemfields/access/field.py
+++ b/invenio_rdm_records/records/systemfields/access/field.py
@@ -216,6 +216,16 @@ class AccessField(SystemField):
 
         return None
 
+    def set_obj(self, record, obj):
+        """Set the access object."""
+        # We accept both dicts and access class objects.
+        if isinstance(obj, dict):
+            obj = self._access_obj_class.from_dict(obj)
+        assert isinstance(obj, self._access_obj_class)
+        # We do not dump the object until the pre_commit hook
+        # I.e. record.access != record['access']
+        self._set_cache(record, obj)
+
     def __get__(self, record, owner=None):
         """Get the record's access object."""
         if record is None:
@@ -224,6 +234,10 @@ class AccessField(SystemField):
 
         # access by object
         return self.obj(record)
+
+    def __set__(self, record, obj):
+        """Set the records access object."""
+        self.set_obj(record, obj)
 
     def pre_commit(self, record):
         """Dump the configured values before the record is committed."""
--- a/invenio_rdm_records/services/generators.py
+++ b/invenio_rdm_records/services/generators.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 Graz University of Technology.
+# Copyright (C) 2021 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Invenio-RDM-Records Permissions Generators."""
+
+import operator
+from functools import reduce
+from itertools import chain
+
+from elasticsearch_dsl import Q, query
+from invenio_records_permissions.generators import Generator
+
+
+class IfRestricted(Generator):
+    """IfRestricted.
+
+    IfRestricted(
+        'record',
+        then_=[RecordPermissionLevel('view')],
+        else_=[ActionNeed(superuser-access)],
+    )
+
+    A record permission level defines an aggregated set of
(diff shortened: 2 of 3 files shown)


tests/conftest.py
```diff
@@ -300,3 +301,46 @@ def identity_simple():
     i.provides.add(UserNeed(1))
     i.provides.add(Need(method='system_role', value='any_user'))
     return i
+
+
+@pytest.fixture(scope="session")
+def create_record():
```
lnielsen, line 307: Minor: is it possible to somehow use one of the existing minimal records to test with to avoid having too many record test fixtures? Also, to create the record, I think it's better if the service is being used ``service.create(...)``.

tests/test_generators.py
```diff
@@ -0,0 +1,41 @@
...
+                "grants": [
+                    {"subject": "user", "id": 1, "level": "edit"},
+                    ]
+                }
+        }
+    )
+
+    generator = IfRestricted(field=field,
+                             then_=[AuthenticatedUser()], else_=[AnyUser()])
+    assert generator.needs(record=record) == [authenticated_user]
```
lnielsen, line 40: Minor: Can you add a test to see that the ``else_`` part is covered as well. Something along the lines of:

```
generator = IfRestricted(field='files', then_=[AuthenticatedUser()], else_=[AnyUser()])
asser generator.needs(record=record) == [authenticated_user]
generator = IfRestricted(field='record', then_=[AuthenticatedUser()], else_=[AnyUser()])
asser generator.needs(record=record) == [any_user]
```

PR #383, opened 2021-02-16
Access: service component for access control #374 

### LINKED ISSUE #374: access: service component for access control
The access component in ``invenio_rdm_records.services.components.access`` should use the new Access systemfield (#372).

- Set access data on the record using the system
- Errors thrown by the data layer should be handled by the resource level.

max-moser:
closes #374 
requires #372

### DIFF
--- a/invenio_rdm_records/services/components.py
+++ b/invenio_rdm_records/services/components.py
@@ -40,9 +40,11 @@ class AccessComponent(ServiceComponent):
             # populate the record's access field with the data already
             # validated by marshmallow
             record.update({"access": data.get("access")})
+            record.access.refresh_from_dict(record.get("access"))
 
         if record is not None and not record.access.owners and identity.id:
-            # TODO can this even happen? isn't the min length 1 in marshmallow?
+            # NOTE: this should not happen, because the jsonschema demands at
+            #       least one owner to be specified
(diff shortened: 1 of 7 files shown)


invenio_rdm_records/services/components.py
```diff
@@ -10,23 +10,50 @@
...
+                errors.append(e)
+
+        return errors
+
+    def _populate_access_and_validate(self, identity, data, record, **kwargs):
+        """Populate and validate the record's access field."""
+        if "access" in data:
+            # populate the record's access field with the data already
+            # validated by marshmallow
+            record.update({"access": data.get("access")})
```
max-moser, line 42: maybe i should add an explicit sync of the access field (respectively, access object) here?
otherwise, i could miss out on relevant changes made to the `data`?

invenio_rdm_records/services/components.py
```diff
@@ -10,23 +10,50 @@
...
+
+        if record is not None and not record.access.owners and identity.id:
+            # TODO can this even happen? isn't the min length 1 in marshmallow?
+            record.access.owners.add({"user": identity.id})
```
max-moser, line 46: if i'm not mistaken, this should never happen, because the marshmallow (or jsonschema) validation should already complain about a zero-length `owned_by`

PR #387, opened 2021-02-17
service: add mechanism to display uniquely draft and published records

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/563
requires to release https://github.com/inveniosoftware/invenio-drafts-resources/pull/93 first

### DIFF
--- a/tests/services/conftest.py
+++ b/tests/services/conftest.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2020 CERN.
 # Copyright (C) 2020 Northwestern University.
+# Copyright (C) 2021 Graz University of Technology.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
@@ -15,6 +16,10 @@ fixtures are available.
 import pytest
 from flask_principal import Identity
 from invenio_access import any_user
+from invenio_access.models import ActionRoles
+from invenio_access.permissions import any_user, authenticated_user, \
+    superuser_access, system_process
+from invenio_accounts.models import Role
 from invenio_app.factory import create_api
 from invenio_vocabularies.records.models import VocabularyType
 from invenio_vocabularies.services.service import VocabulariesService
@@ -65,3 +70,55 @@ def languages(db):
         languages[lang['props']['id']] = record
 
     return languages
+
+
+@pytest.fixture(scope="function")
+def superuser_role_need(db):
+    """Store 1 role with 'superuser-access' ActionNeed.
+
+    WHY: This is needed because expansion of ActionNeed is
+         done on the basis of a User/Role being associated with that Need.
+         If no User/Role is associated with that Need (in the DB), the
+         permission is expanded to an empty list.
+    """
+    role = Role(name="superuser-access")
+    db.session.add(role)
+
+    action_role = ActionRoles.create(action=superuser_access, role=role)
+    db.session.add(action_role)
+
+    db.session.commit()
+
+    return action_role.need
+
+
+@pytest.fixture(scope="function")
+def anyuser_identity():
+    """System identity."""
+    identity = Identity(1)
+    identity.provides.add(any_user)
+    return identity
+
+
+@pytest.fixture(scope="function")
+def authenticated_identity():
+    """Authenticated identity fixture."""
+    identity = Identity(1)
+    identity.provides.add(authenticated_user)
+    return identity
+
+
+@pytest.fixture(scope="function")
+def superuser_identity():
+    """Superuser identity fixture."""
+    identity = Identity(1)
+    identity.provides.add(superuser_access)
+    return identity
+
+
+@pytest.fixture(scope="function")
+def system_process_identity():
+    """Superuser identity fixture."""
+    identity = Identity(1)
+    identity.provides.add(system_process)
+    return identity
--- a/tests/services/test_permissions_poliy.py
+++ b/tests/services/test_permissions_poliy.py
@@ -1,0 +1,127 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 Graz University of Technology.
+#
+# Invenio-RDM-Records is free software; you can redistribute it
+# and/or modify it under the terms of the MIT License; see LICENSE file for
+# more details.
+
+"""Permissions for Invenio RDM Records."""
+
+from flask_principal import RoleNeed, UserNeed
+from invenio_access.permissions import any_user, authenticated_user, \
+    system_process
+from invenio_drafts_resources.services.records.permissions import \
+    RecordDraftPermissionPolicy
+from invenio_records_permissions.generators import AnyUser, \
+    AuthenticatedUser, Disable, SystemProcess
+
+from invenio_rdm_records.records import RDMRecord
+from invenio_rdm_records.services.generators import IfRestricted, RecordOwners
+
+
+class TestRDMPermissionPolicy(RecordDraftPermissionPolicy):
+    """Define permission policies for RDM Records."""
+    can_search = [AnyUser(), SystemProcess()]
+    can_create = [AuthenticatedUser(), SystemProcess()]
+    can_update = [Disable()]
+    can_delete = [Disable()]
+    can_read = [
+        IfRestricted('record',
+                     then_=[RecordOwners()],
+                     else_=[AnyUser(), SystemProcess()])
+                     ]
+    can_read_files = [
+        IfRestricted('files',
+                     then_=[RecordOwners()],
+                     else_=[AnyUser(), SystemProcess()])
+                     ]
+    can_update_files = [Disable()]
+    can_read_draft = [RecordOwners()]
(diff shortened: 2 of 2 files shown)


invenio_rdm_records/records/api.py
```diff
@@ -80,6 +83,10 @@ class RDMRecord(CommonFieldsMixin, Record):
         delete=False,
     )
 
+    has_draft = HasDraftCheckField(LocalProxy(lambda: RDMDraft))
```
lnielsen, line 86: I think this can be done with just a lambda and alternatively you can probably just define the draft first, so we don't need the proxy all together.

invenio_rdm_records/records/api.py
```diff
@@ -35,7 +36,9 @@ class CommonFieldsMixin:
         extensions=[
             EDTFDumperExt('metadata.publication_date'),
             EDTFListDumperExt("metadata.dates", "date"),
+            RecordStatusDumperExt("metadata.status"),
```
lnielsen, line 39: I think these properties should not be inside the ``metadata`` key but rather as top-level keys.

invenio_rdm_records/records/api.py
```diff
@@ -80,6 +83,10 @@ class RDMRecord(CommonFieldsMixin, Record):
         delete=False,
     )
 
+    has_draft = HasDraftCheckField(LocalProxy(lambda: RDMDraft))
+
+    is_record = True
```
lnielsen, line 88: I understand we need to know if this is a draft or record, but is it not implicit via the class, index or other information we already have available?

invenio_rdm_records/records/api.py
lnielsen: I know this is WIP, so it just a comment to not  forget: We will definitely need some documentation here explaining what this property is used for (like we should have done with ``is_published`` so we don't have to look for how it's used.

invenio_rdm_records/records/api.py
lnielsen: Or, e.g. Draft records doesn't have a ``has_draft`` attribute.

invenio_rdm_records/services/config.py
```diff
@@ -50,13 +50,18 @@ class RDMRecordServiceConfig(RecordDraftServiceConfig):
...
+            'record_status': {
+                'terms': {'field': 'metadata.status'},
+            },
         },
         post_filters={
             'subtype': terms_filter('metadata.resource_type.subtype'),
             'resource_type': terms_filter('metadata.resource_type.type'),
             'access_right': terms_filter('access.access_right'),
             'languages': terms_filter('metadata.languages.id'),
+            'record_status': terms_filter('metadata.status'),
```
lnielsen, line 63:
```suggestion
            'status': terms_filter('status'),
```

invenio_rdm_records/services/config.py
```diff
@@ -50,13 +50,18 @@ class RDMRecordServiceConfig(RecordDraftServiceConfig):
...
+            'has_draft': terms_filter('metadata.has_draft'),
```
lnielsen, line 64: I would not expose the has_draft in the REST API. This filter should be implicitly managed internally.

invenio_rdm_records/records/dumpers/record_status.py
```diff
@@ -0,0 +1,54 @@
...
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""DataCite-based data model for Invenio."""
+
+from invenio_records.dictutils import dict_set
+from invenio_records.dumpers import ElasticsearchDumperExt
+
+
+class RecordStatusDumperExt(ElasticsearchDumperExt):
```
lnielsen, line 14: I think this has to be a system field rather than only something available in the index. E.g. I would assume we would expose the status in the REST API JSON, so that we can e.g. display the right icon.

invenio_rdm_records/records/dumpers/has_draft.py
```diff
@@ -0,0 +1,38 @@
...
+        """Dump the data."""
+        dict_set(
+            data,
+            self.key,
+            getattr(record, 'has_draft', False)
+        )
+
+    def load(self, data, record_cls):
+        """Load the data."""
+        data.pop(self.key, None)
```
lnielsen, line 38: For the caching part, I think here, you'll should just let the value pass through. Hence, the record will be created like:

```
RDMRecord({'has_draft': True, ....})
```

Then, the system field (``HasDraftCheckField``) you can add the ``pre_init`` hook, can consume the value - something like:

```python
def pre_init(self, record, data, model=None, **kwargs):
        """Sets the key in the record during record instantiation."""
        if data is None:
            # A deleted record.
            return
        value = data.pop(self.key, None)
        if value is not None:
              self._set_cache(record, value) 
```

lnielsen (commented):
Overall I think the approach work if I understood correctly.

1) For the ``status`` facet we compute a ``status`` field in the mapping so we can run an aggregation on the field.
2) The ``has_draft`` is an internal field we use to filter out duplicates.

invenio_rdm_records/services/services.py
```diff
@@ -22,12 +22,20 @@ class RDMRecordService(RecordDraftService):
...
 
 
-class RDMUserRecordsService(RecordService):
+class RDMUserRecordsService(RecordDraftService):
     """Bibliographic user records service."""
 
     config_name = "RDM_RECORDS_BIBLIOGRAPHIC_USER_RECORDS_SERVICE_CONFIG"
     default_config = config.RDMUserRecordsServiceConfig
 
+    def user_records_search(self, identity, params=None, links_config=None,
```
zzacharo, line 31: I moved the newly added method here as I thought that this feature is specific to RDM needs....Also all the systemfileds are defined in this module... But, I have some doubts if we want this as part of the invenio-drafts-resources.... In that case, I just need to move the functionality in that module and provide it as core feature...

invenio_rdm_records/services/services.py
```diff
@@ -22,12 +22,20 @@ class RDMRecordService(RecordDraftService):
...
+                            es_preference=None, **kwargs):
+        """Search for records matching the querystring."""
+        index_or_alias = self.record_cls.user_records_index._name
```
zzacharo, line 34: On user records search we use the `user_records_index` provided in the `CommonFieldsMixin` and we point to the common alias of drafts and records.

invenio_rdm_records/records/api.py
```diff
@@ -55,58 +57,61 @@ class CommonFieldsMixin:
 
     access = AccessField()
 
+    status = RecordStatusField('status')
```
lnielsen, line 60: Question: is ``'status'`` needed here?

invenio_rdm_records/records/mappings/v6/rdmrecords/records/record-v1.0.0.json
```diff
@@ -217,6 +220,9 @@
                 }
               }
             },
+            "is_record": {
```
lnielsen, line 223: Minor: I think this one has to be removed?

invenio_rdm_records/records/mappings/v6/rdmrecords/records/record-v1.0.0.json
```diff
@@ -56,6 +56,9 @@
             }
           }
         },
+        "has_draft": {
```
lnielsen, line 59: Minor: Should this one also be added to the drafts mapping to keep the two mappings identical......if we dont' add data it's just ignored, but I'd prefer not having two different mappings for drafts and records.

invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -47,6 +47,8 @@ class Meta:
     # notes = fields.List(fields.Nested(InternalNoteSchema))
     revision = fields.Integer(dump_only=True)
 
+    status = fields.Str(dump_only=True)
```
lnielsen, line 50: Question: In the REST API, what does a record say?

invenio_rdm_records/services/params/has_draft.py
```diff
@@ -0,0 +1,30 @@
...
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Facets parameter interpreter API."""
+
+from functools import partial
+
+from invenio_records_resources.services.records.params import ParamInterpreter
+
+
+class HasDraftParam(ParamInterpreter):
```
lnielsen, line 15: Comment: Perhaps rename to Status since we call the parameter ``status``

lnielsen: Quesiton: is it applying to both ``/api/records``  and ``/api/user/records``

lnielsen: Question: Can you also check that the list of records is filtered to only what the user can actually edit

zzacharo: I renamed it! The param is applied only in the user records resource. Also, I defined the `status` agg and post filter only for the user records resource.

lnielsen (approved):

PR #388, opened 2021-02-17
373 es grant token dumper

### LINKED ISSUE #373: access: create a ES dumper for grant tokens
Depends on #371 and #372

- [ ] Create an ES dumper extension ``AccessDumper`` in ``invenio_rdm_records.dumpers.access`` that dumps the grant tokens on indexing, and removes them again on loading.

The grant tokens themselves should be generated by the system fields Grant class.

max-moser:
closes #373 
**note**: requires the issue with `draft.dump()` in https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/services/records/service.py#L189 to be resolved, in order to work

### DIFF
--- a/invenio_rdm_records/records/dumpers/access.py
+++ b/invenio_rdm_records/records/dumpers/access.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 TU Wien.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""ElasticSearch dumpers for access-control information."""
+
+from invenio_records.dictutils import dict_lookup, parse_lookup_key
+from invenio_records.dumpers import ElasticsearchDumperExt
+
+
+class GrantTokensDumperExt(ElasticsearchDumperExt):
+    """Elasticsearch dumper extension for access grant tokens support.
+
+    On dump, it uses the record's ``Access`` system field to generate tokens
+    from the record's (access) ``Grants`` and dump them in the specified target
+    field in the dictionary (per default: ``access.grant_tokens``).
+    On load, it simply removes the target field from the dictionary again.
+    """
+
+    def __init__(self, target_field):
+        """Constructor.
+
+        :param target_field: dot separated path where to dump the tokens.
+        """
+        super().__init__()
+        self.keys = parse_lookup_key(target_field)
+        self.key = self.keys[-1]
+
+    def dump(self, record, data):
+        """Dump the grant tokens to the data dictionary."""
+        if record.access is not None:
+            tokens = [grant.to_token() for grant in record.access.grants]
+            parent_data = dict_lookup(data, self.keys, parent=True)
+            parent_data[self.key] = tokens
+
+    def load(self, data, record_cls):
+        """Remove the tokens from the data dictionary."""
+        try:
+            parent_data = dict_lookup(data, self.keys, parent=True)
+            if parent_data is not None:
+                parent_data.pop(self.key, None)
+
+        except KeyError:
+            pass
--- a/invenio_rdm_records/records/dumpers/access.py
+++ b/invenio_rdm_records/records/dumpers/access.py
@@ -31,7 +31,7 @@ class GrantTokensDumperExt(ElasticsearchDumperExt):
 
     def dump(self, record, data):
         """Dump the grant tokens to the data dictionary."""
-        if record.access is not None:
+        if record.access:
             tokens = [grant.to_token() for grant in record.access.grants]
             parent_data = dict_lookup(data, self.keys, parent=True)
             parent_data[self.key] = tokens
--- a/invenio_rdm_records/records/systemfields/access/grants.py
+++ b/invenio_rdm_records/records/systemfields/access/grants.py
@@ -118,9 +118,9 @@ class Grant:
         # this ensures that we can always separate the values again
         # (b/c the base64 alphabet doesn't contain the dot)
         return "{}.{}.{}".format(
-            b64encode(str(self.subject_type).encode()).decode(),
-            b64encode(str(self.subject_id).encode()).decode(),
-            b64encode(str(self.permission_level).encode()).decode(),
+            b64encode(str(self.subject_type).encode(), b"-_").decode(),
+            b64encode(str(self.subject_id).encode(), b"-_").decode(),
+            b64encode(str(self.permission_level).encode(), b"-_").decode(),
         )
 
     def to_dict(self):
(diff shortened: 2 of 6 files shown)


invenio_rdm_records/records/dumpers/access.py
```diff
@@ -0,0 +1,47 @@
...
+
+        :param target_field: dot separated path where to dump the tokens.
+        """
+        super().__init__()
+        self.keys = parse_lookup_key(target_field)
+        self.key = self.keys[-1]
+
+    def dump(self, record, data):
+        """Dump the grant tokens to the data dictionary."""
+        if record.access is not None:
```
fenekku, line 34: minor: `if record.access` should be enough

invenio_rdm_records/records/dumpers/access.py
```diff
@@ -0,0 +1,47 @@
...
+            tokens = [grant.to_token() for grant in record.access.grants]
+            parent_data = dict_lookup(data, self.keys, parent=True)
+            parent_data[self.key] = tokens
+
+    def load(self, data, record_cls):
+        """Remove the tokens from the data dictionary."""
+        try:
+            parent_data = dict_lookup(data, self.keys, parent=True)
+            if parent_data is not None:
```
fenekku, line 43: same... but taking a step back: Is it even possible to have a `record.access == None` ? I am not sure that's possible (and to be clear, we both understand that Python behaves differently than Javascript wrt attribute access)

invenio_rdm_records/records/systemfields/access/grants.py
```diff
@@ -118,9 +118,9 @@ def to_token(self):
         # this ensures that we can always separate the values again
         # (b/c the base64 alphabet doesn't contain the dot)
         return "{}.{}.{}".format(
-            b64encode(str(self.subject_type).encode()).decode(),
-            b64encode(str(self.subject_id).encode()).decode(),
-            b64encode(str(self.permission_level).encode()).decode(),
+            b64encode(str(self.subject_type).encode(), b"-_").decode(),
+            b64encode(str(self.subject_id).encode(), b"-_").decode(),
+            b64encode(str(self.permission_level).encode(), b"-_").decode(),
```
fenekku, line 123: :thinking: does that mean a query for permission tokens need to know to b64encode its arguments beforehand to sidestep strange issues? ... I am probably naive, but if we don't want dots, why don't we have the code enforce that and raise an exception if a dot is present in those arguments? Those arguments are all internal code anyway (I don't see anything wrong with that being the way we roll).

invenio_rdm_records/services/schemas/access.py
```diff
@@ -76,6 +76,9 @@ class AccessSchema(Schema):
     # TODO re-enable when the grants are ready
     # grants = fields.List(fields.Nested(Grant))
 
+    # NOTE: this field is only present to avoid errors for unknown fields
+    grant_tokens = fields.List(fields.String, dump_only=True, required=False)
```
fenekku, line 80: `required` is False by default so we can omit it.

fenekku (approved):

invenio_rdm_records/records/dumpers/access.py
max-moser: yes, it's possible for a record's access field to be `None`, if the record's data doesn't have an `access` property defined (which is the case in some unit tests, IIRC): https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/systemfields/access/field.py#L247

if we do initialize a new access field in that case, it will be dumped to the record's data on `.commit()` and thus effectively add a new property (which wasn't the case without the system field in place)

invenio_rdm_records/records/systemfields/access/grants.py
max-moser: we can of course change the encoding to something that's easier to use from a programming perspective, i just selected base64 because i thought it offered some nice benefits:
* with the alternate chars "-" and "_", we always get a string token that's URL-safe
* we're completely independent on the actual values of `subject_type`, `subject_id` and `permission_level` -- they can include any characters
* b64 results never contain dots, so we can be sure that dot-separation will always work

if we know an *entire* token, we can always do something like `Grant.from_dict({"subject": "user", "id": "1", "level": "view"}).to_token()` and not worry about the concrete way that the tokens are generated (in the back-end at least)

however, if we want to generate partial tokens (e.g. from `{"subject": "user"}`), the current approach probably won't work as well...
in that case, we might also want to change the ES mapping however, since i think that the `keyword` type doesn't allow for substring matches:
https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/mappings/v7/rdmrecords/records/record-v1.0.0.json#L54

invenio_rdm_records/records/dumpers/access.py
max-moser: true, `bool(record.access)` should always evaluate to `True` per default, if it's not `None`

invenio_rdm_records/records/dumpers/access.py
fenekku: Hmm I guess since the `access` schema field [is not required right now](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/schemas/__init__.py#L46), it could indeed be the case that we don't have a matching access field. `access` should probably be required :smile: , but @lnielsen should take a look at the fields to lock in before the 1.0 . Thanks for elucidating that!

invenio_rdm_records/records/systemfields/access/grants.py
fenekku: I am thinking about the Generator query filters: those generators would be creating the tokens. Most of those are in invenio-records-permissions, but records-permissions shouldn't have a dependency on invenio-rdm-records (otherwise circular). So  if we are relying on `Grant.from_dict({})`, we should put those generators in invenio-rdm-records... which means all(?) generators should be placed here (not quite what we wanted :smile_cat: ) If we are not relying on `Grant.from_dict({})`, then we need to be careful to copy this encoding in that module and keep them in sync... 

In any case, we will be diving deeper into those questions soon, so we can assess that separately.

max-moser: ahhh, i see!
yes, that's kind of an issue :confused:

lnielsen: Most of the generators will be in Invenio-RDM-Records as they are specific to the RDM data model.

invenio_rdm_records/services/schemas/access.py
```diff
@@ -76,6 +76,9 @@ class AccessSchema(Schema):
     # TODO re-enable when the grants are ready
     # grants = fields.List(fields.Nested(Grant))
 
+    # NOTE: this field is only present to avoid errors for unknown fields
+    grant_tokens = fields.List(fields.String, dump_only=True)
```
lnielsen, line 80: Minor: I'm puzzled why this one is here? The tokens should only exist in the index, and not make their way out in the REST API.

fenekku: Removing that doesn't break the tests now, so it's removed.

PR #395, opened 2021-02-19
fixtures: add UsersFixture for setting up required users

max-moser:
* since the AccessComponent checks if the record owners actually exist
  and each record has to have at least one owner, at least one user is
  required for demo records

### DIFF
--- a/invenio_rdm_records/fixtures/users.py
+++ b/invenio_rdm_records/fixtures/users.py
@@ -11,6 +11,7 @@ import secrets
 import string
 
 import yaml
+from flask import current_app
 from flask_security.utils import hash_password
 from invenio_access.models import ActionUsers
 from invenio_access.proxies import current_access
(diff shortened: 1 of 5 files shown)


invenio_rdm_records/fixtures/users.py
```diff
@@ -0,0 +1,58 @@
...
+            for email, user_data in data.items():
+                self.create_user(email, user_data)
+
+    def create_user(self, email, entry):
+        """Load a single vocabulary."""
+        # for auto-generated passwords use letters, digits,
+        # and some punctuation marks
+        alphabet = string.ascii_letters + string.digits + "+,-_."
+        gen_passwd = "".join(secrets.choice(alphabet) for i in range(20))
+        password = entry["password"] or gen_passwd
```
fenekku, line 42: For these dict access a `get` is a safer bet (no KeyError). 

We can also have the `or` for password so that even an empty string in the yaml file will get a generated password. 

```suggestion
        password = entry.get("password") or gen_passwd
```

That being said, for password, having the default password of a super user be out in the open seems... not so secure ;) . Having it in the config like the password for the db seems better...


```python
entry.get("active", False)
```

```
for role in entry.get("roles", [])
```

```
for action in entry.get("allow", [])
```

fenekku (approved):
The main thing bothering me is how to make the admin password secure by default. Can the `invenio db` command pattern be re-used here?

fenekku (approved):
Nice!

PR #399, opened 2021-02-22
schemas: add field-level permissions for access properties

### LINKED ISSUE #377: access: field-level permissions for managing access
In order to submit records via the REST API it should require the ``manage`` permission-level. This should be defined in the record service schema (#371) via adding ``field_permission_check``. It should take both dumping and loading into account.

max-moser:
closes #377 

* exclude access.grants from being dumped, except for users with "manage" permissions
* require "manage" permissions to make any changes in the access property
* update the initialization of the record's access system field initialization in the AccessComponent
* require the ID of the current_user in the RecordOwners generator, if the supplied record is None (i.e. a new record is being created)

what's **not** in this PR (anymore):
* disallow users from setting access.owned_by themselves; this should only be done by the AccessComponent

tests revealed that this caused issues with publishing drafts (because the `owned_by` property is dumped, and then loaded).

also updates the `edtf_dumpers` test to reflect the split between `service.search` and `service.search_draft` from https://github.com/inveniosoftware/invenio-drafts-resources/commit/1ae974a16784b4fd5745bd9457095d5d348f98fe


fenekku (commented):
Oof that's painful, sorry about that.

Immediate things before a lower-level review:

- ~~There is also login_user_via_view: https://github.com/inveniosoftware/invenio-accounts/blob/0ae2d782f1346b0b8e03810e702cc89875bd762e/invenio_accounts/testutils.py#L68 . There may be some subtle difference, but I've been using it for the files protection and it seems fine so far... (`user.password_plaintext` when logging in...)~~ In the end, it just didn't do it as well, so we can forget about it.
- The `minimal_record` and `full_record` fixtures are for upload. That's their definition. If they are not supposed to have `access.owned_by`, we should remove it there rather than create new fixtures for upload. 
-  `login` or `_login` : settling on one and instead of repeatedly adding it, for most cases it seems like a new fixture called `authenticated_client` or `user1_client` that returns an authenticated client is what we want. I am completely fine to even override the client fixture within a test module if all tests require an authenticated client.

invenio_rdm_records/services/generators.py
```diff
@@ -81,9 +82,17 @@ class RecordOwners(Generator):
 
     def needs(self, record=None, **kwargs):
         """Enabling Needs."""
-        owners = record.access.owners.dump()
-        return [UserNeed(owner.get('user')
-                         ) for owner in owners]
+        if record is None:
+            # 'record is None' means that this must be a 'create'
+            # this should be allowed for any current_user
+            if current_user.is_authenticated:
```
lnielsen, line 88: You can avoid dealing with the current_user object by replacing this with:
```
from invenio_access.permissions import authenticated_user
return authenticated_user
```

lnielsen (approved):
I think it's good to merge, if we just think the tiny comment

PR #402, opened 2021-02-23
permissions: test files API endpoints are protected

### LINKED ISSUE #401: Files protection: record files endpoint
Sub of https://github.com/inveniosoftware/invenio-app-rdm/issues/577

Deal with record files' endpoint: https://inveniordm.docs.cern.ch/reference/rest_api/#record-files
- list
- read file metadata
- download (read file)

fenekku:
- closes #401 
- requires invenio-records-resources PR: https://github.com/inveniosoftware/invenio-records-resources/pull/200

### DIFF
--- a/invenio_rdm_records/services/permissions.py
+++ b/invenio_rdm_records/services/permissions.py
@@ -12,6 +12,8 @@ from invenio_drafts_resources.services.records.permissions import \
     RecordDraftPermissionPolicy
 from invenio_records_permissions.generators import AnyUser, SystemProcess
 
+from invenio_rdm_records.services.generators import IfRestricted, RecordOwners
+
 
 class RDMRecordPermissionPolicy(RecordDraftPermissionPolicy):
     """Access control configuration for records.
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/services/permissions.py
```diff
@@ -27,7 +29,29 @@ class RDMRecordPermissionPolicy(RecordDraftPermissionPolicy):
...
+    # For now, can_read_files means:
+    # - can list files
+    # - can read a file metadata
+    # - can download the file
+    can_read_files = [
+        IfRestricted(
+            'record',
+            then_=[RecordOwners()],
+            else_=[
+                IfRestricted(
```
lnielsen, line 46: Comment: It should be enough to check only ``files``.

fenekku: To clarify...

Do you mean that I should implement dependent validation in the AccessSchema: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/schemas/access.py#L70 , because it's not there right now? By dependent validation, I mean something **like** if `record == "restricted"`, then there needs to be `files == "restricted"`. Otherwise, it's not enough to check `files` only. Actually there is 3 potential implementations I see:

A- have the backend do this dependent check (see note below about how frontend needs to change too then)
B- have the backend *make* the values passed be consistent. For example, if `"record": "restricted"` and `"files": "public"` is passed in, then the backend could change `"files"` to be `"files": "restricted"` so that it is consistent.  
C- have the backend do hierarchical checking (this is what my code is doing). For example, if `"record": "restricted"` and `"files": "public"` is passed in, then the backend doesn't change `"files"` but since it always checks "record" first, it doesn't matter what the other values are.

I can see advantages to each. These scenarios also apply to embargo actually and it should be made to work similarly (embargo has `until` and `reason` that **depend** on `active`). The uncertainty here has led to.... 'complexity' on the AccessComponent side. In the end, my preference is probably for A, but since it wasn't done... I didn't know the intention :sweat_smile:   

Since this dependent validation wasn't implemented, I went with this dependent permission... Also to note, because this kind of validation on the AccessSchema wasn't implemented in the backend, the frontend protection didn't set values that respect it either. We need to revisit frontend to make sure it sends consistent field values.

Which approach should we aim for @lnielsen ?

fenekku:
closed in favour of https://github.com/inveniosoftware/invenio-rdm-records/pull/411

PR #410, opened 2021-02-24
access: introduce null object for Embargo

max-moser:
potentially fixes inveniosoftware/invenio-app-rdm#586

* create a null object for Embargo that is used instead of None
* relax the requirements for creating a 'null embargo' in from_dict to require all properties to be falsy (or not set) instead of being strictly None (or not set)
* change the field type of embargo.until in the AccessSchema, as that didn't allow dumping of null values

### DIFF
--- a/invenio_rdm_records/records/systemfields/access/embargo.py
+++ b/invenio_rdm_records/records/systemfields/access/embargo.py
@@ -14,7 +14,7 @@ import arrow
 class Embargo:
     """Embargo class for the access system field."""
 
-    def __init__(self, until, reason=None, active=None):
+    def __init__(self, until=None, reason=None, active=None):
         """Create a new Embargo."""
         self.until = until
         if isinstance(until, str):
@@ -53,6 +53,12 @@ class Embargo:
 
         return False
 
+    def clear(self):
+        """Clear any information if the embargo was ever active."""
+        self.until = None
+        self.reason = None
+        self._active = None
+
     def dump(self):
         """Dump the embargo as dictionary."""
         until_str = None
@@ -67,11 +73,29 @@ class Embargo:
 
     def __repr__(self):
         """Return repr(self)."""
+        if self == Embargo():
+            return "<No Embargo>"
+
         until_str = self.until or "n/a"
 
         return "<{} (active: {}, until: {}, reason: {})>".format(
             type(self).__name__, self.active, until_str, self.reason
         )
+
+    def __eq__(self, other):
+        """Return self == other."""
+        if type(self) != type(other):
+            return False
+
+        return (
+            self.reason == other.reason
+            and self.until == other.until
+            and self.active == other.active
+        )
+
+    def __ne__(self, other):
+        """Return self != other."""
+        return not (self == other)
 
     def __bool__(self):
         """Return bool(self)."""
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/services/schemas/utils.py
```diff
@@ -54,3 +56,23 @@ def dump_empty(schema_or_field):
...
+
+            return super()._serialize(
+                value, attr, obj, **kwargs
+            )
+        except ParserError:
+            return missing
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        """Deserialize an ISO8601-formatted date."""
+        return super()._deserialize(value, attr, data, **kwargs).isoformat()
```
fenekku, line 78: This is the variation of the marshmallow-utils one... I guess it can't be helped  ╮(￣ω￣;)╭

fenekku (commented):
Hmm took me some time, but it's mostly because of my comment on `__init__` which there isn't much to do about... The rest are all what I think are simplifications. It's hard to make things simpler, but on the long run I think it helps (we'll see if that's even achieved :sweat_smile: ).

invenio_rdm_records/records/systemfields/access/embargo.py
```diff
@@ -67,21 +69,42 @@ def dump(self):
 
     def __repr__(self):
         """Return repr(self)."""
+        if self == Embargo.NONE:
+            return "<No Embargo>"
+
```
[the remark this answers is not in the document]
max-moser: this should never actually be necessary; it's more vanity to print our null embargo as in a visually more distinctive way than `<Embargo (active: False, until: None, reason: None)>`

invenio_rdm_records/records/systemfields/access/field.py
```diff
@@ -107,24 +105,23 @@ def errors(self):
...
-        self._embargo = embargo
+        if embargo is None:
+            embargo = Embargo.NONE
 
-        if embargo is not None:
-            # allow 'clear_embargo()' to be called as 'embargo.clear()'
-            self._embargo.clear = lambda: self.clear_embargo()
+        # allow 'clear_embargo()' to be called as 'embargo.clear()'
+        self._embargo = embargo
+        self._embargo.clear = lambda: self.clear_embargo()
```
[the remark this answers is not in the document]
max-moser: true, that strange construct above was only there because from within the Embargo instance, we couldn't set the reference to itself, that the Access object has, to None...
now that we did away with the None variant, it's not required anymore

invenio_rdm_records/services/schemas/utils.py
max-moser: yes, what we expect (or came to expect) is that exact middle ground between `Date` from marshmallow and `ISODateString` from marshmallow-utils
technically, we could also subclass the latter, and save the `_deserialize`...

fenekku (approved):
Thanks for the changes! Oops I think a merge got in the PR - just rebase on master should do it.

invenio_rdm_records/services/schemas/utils.py
lnielsen: I'm merging and i've opened https://github.com/inveniosoftware/marshmallow-utils/issues/28

PR #418, opened 2021-02-25
permissions: hide a draft's public files

fenekku:


invenio_rdm_records/services/generators.py
```diff
@@ -101,3 +103,42 @@ def query_filter(self, identity=None, **kwargs):
         users = [n.value for n in identity.provides if n.method == "id"]
         if users:
             return Q("terms", **{"access.owned_by.user": users})
+
+
+class IfDraft(Generator):
```
lnielsen, line 108: Is the query filter also needed?

fenekku: is there a way to search for the files without assessing the record/draft first?

PR #424, opened 2021-02-26
access: silently prevent users from changing record owners

max-moser:
* the only exception are system processes, which are allowed to set the
  record owners explicitly

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -331,10 +331,11 @@ def roles(app, db):
     return [role1, role2]
 
 
-@pytest.fixture(scope="module")
-def identity_simple():
+@pytest.fixture(scope="function")
+def identity_simple(users):
     """Simple identity fixture."""
-    i = Identity(1)
-    i.provides.add(UserNeed(1))
+    user = users[0]
+    i = Identity(user.id)
+    i.provides.add(UserNeed(user.id))
     i.provides.add(Need(method='system_role', value='any_user'))
     return i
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -331,10 +331,11 @@ def roles(app, db):
...
+@pytest.fixture(scope="function")
+def identity_simple(users):
     """Simple identity fixture."""
-    i = Identity(1)
-    i.provides.add(UserNeed(1))
+    user = users[0]
+    i = Identity(user.id)
+    i.provides.add(UserNeed(user.id))
     i.provides.add(Need(method='system_role', value='any_user'))
     return i
```
max-moser, line 341: since the created users seem to be assigned incremental IDs over the course of the tests, a hard-coded ID of 1 doesn't always cut it

tests/services/components/test_access_component.py
```diff
@@ -34,11 +35,24 @@ def test_access_component_unknown_owner(
     record = RDMRecord.create(minimal_record)
     component = AccessComponent(RDMRecordService())
 
+    # both should work, since 'access.owned_by' is ignored for anybody
+    # other than system processes
+    component.create(identity_simple, minimal_record, record)
+    component.update_draft(identity_simple, minimal_record, record)
+
+
+def test_access_component_unknown_owner_with_system_process(
```
max-moser, line 44: split up the test in two variants: with a normal user's identity, and with the system identity... since with the new changes, these should behave differently

lnielsen (approved):

PR #433, opened 2021-03-09
dumpers: refactor `has_draft` systemfield to incorporate dump hook

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/655
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/649


invenio_rdm_records/records/systemfields/is_published.py
```diff
@@ -0,0 +1,29 @@
...
+from invenio_records_resources.records.systemfields import PIDStatusCheckField
+from sqlalchemy.orm.exc import NoResultFound
+
+
+class IsPublishedField(PIDStatusCheckField):
+    """Adds a `pre_dump` hook to pid status check field."""
+
+    def pre_dump(self, record, **kwargs):
+        """Called before a record is dumped in a secondary storage system."""
+        dict_set(
```
lnielsen, line 25: I'm changing Invenio-Records so that the signature of pre_dump is:

```python
def pre_dump(record, data, dumper=None):
    pass
```

Also, there's a ``post_dump()`` (same signature).

This avoid modifying the record, because if you commit after having dumped, you'll have the property inside the record. See https://github.com/inveniosoftware/invenio-records/pull/249

invenio_rdm_records/records/systemfields/has_draftcheck.py
```diff
@@ -34,8 +39,25 @@ def __get__(self, record, owner=None):
...
+
         try:
             self.draft_cls.get_record(record.id)
             return True
         except NoResultFound:
             return False
+
+    def pre_dump(self, record, **kwargs):
+        """Called before a record is dumped in a secondary storage system."""
+        dict_set(
```
lnielsen, line 55: See my comment below on method signature and modifying the record.

invenio_rdm_records/records/api.py
```diff
@@ -57,7 +56,9 @@ class CommonFieldsMixin:
 
     access = AccessField()
 
-    status = RecordStatusField()
+    # We redefine the property as we extend the `PIDStatusCheckField` to dump
+    # the property in ES in order to be available for aggregation
+    is_published = IsPublishedField(status=PIDStatus.REGISTERED)
```
lnielsen, line 61: Comment: The name should be aligned with the existing name in the service schema - not sure if it is or not - but from top of my head I think it's named ``published``. Just check and if different we have to update one of the two.

invenio_rdm_records/services/config.py
```diff
@@ -114,16 +114,16 @@ class RDMUserRecordsServiceConfig(RDMRecordServiceConfig):
             # 'languages': {
             #     'terms': {'field': 'metadata.languages.id'},
             # },
-            'status': {
-                'terms': {'field': 'status'},
+            'published': {
```
lnielsen, line 117: Question: We could leave it as ``status`` so that the query is ``?status=published`` and ``?status=new``

invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -49,8 +49,6 @@ class Meta:
     # notes = fields.List(fields.Nested(InternalNoteSchema))
     revision = fields.Integer(dump_only=True)
 
-    status = fields.Str(dump_only=True)
-
     published = fields.Boolean(attribute='is_published', dump_only=True)
```
lnielsen, line 52: Minor: See my comment earlier - about the name ``published`` vs ``is_published``. We need to align these two. Might be we just rename this one to ``is_published``.

invenio_rdm_records/services/config.py
zzacharo: We cannot do that because the actual filter would be `status=true/false` as the `is_published` is a boolean.... The published / new labels will be applied from the UI serializer...

PR #442, opened 2021-03-10
global: add secret links for sharing access permissions

max-moser:
* allow sharing of view permissions on restricted records via links
* add new need type and permission generator
* add separate "links" field in the record for holding the links, as
  they're something different from grants


invenio_rdm_records/ext.py
```diff
@@ -19,6 +22,25 @@
...
+    if token:
+        try:
+            data = SecretLink.load_token(token)
+            if data:
+                session["rdm-records-token"].add(token)
+                link_need = LinkNeed(data["id"])
+                g.identity.provides.add(link_need)
+
+        except SignatureExpired:
+            # TODO flash a message?
```
max-moser, line 40: do we want to notify the user in any way that their token didn't work?

invenio_rdm_records/links/models.py
```diff
@@ -0,0 +1,229 @@
...
+    @classmethod
+    def create(
+        cls,
+        title,
+        owner,
+        permission_level,
+        extra_data,
+        description="",
+        expires_at=None,
+    ):
```
max-moser, line 154: that's gonna be the primary way of creating secret links, i think... along with `record.access.links.create(...)`

invenio_rdm_records/links/models.py
```diff
@@ -0,0 +1,229 @@
...
+    )
+    """Creation timestamp."""
+
+    expires_at = db.Column(db.DateTime, nullable=True)
+    """Expiration date."""
+
+    revoked_at = db.Column(db.DateTime, nullable=True, index=True)
+    """Revocation timestamp."""
+
+    title = db.Column(db.String, nullable=False, default="")
```
lnielsen, line 64: Comment: I would remove title and description - it made sense in the zenodo-accessrequest but not sure it makes sense here.

invenio_rdm_records/links/models.py
```diff
@@ -0,0 +1,229 @@
...
+
+    created = db.Column(
+        db.DateTime, nullable=False, default=datetime.utcnow, index=True
+    )
+    """Creation timestamp."""
+
+    expires_at = db.Column(db.DateTime, nullable=True)
+    """Expiration date."""
+
+    revoked_at = db.Column(db.DateTime, nullable=True, index=True)
```
lnielsen, line 61: Question: Do we need the revoked field? In the previous model, we needed to know it was revoked to invalidate it, but I don't think we have the same issue with this implementation.

invenio_rdm_records/ext.py
```diff
@@ -19,6 +22,25 @@
 from .services.schemas.metadata_extensions import MetadataExtensions
 
 
+def verify_token():
```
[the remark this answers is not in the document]
max-moser: true, the permission generator queries the links in `record.access.links` and only uses those that have the right permission level and haven't expired
so that case should be filtered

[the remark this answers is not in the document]
max-moser: hmm... pretty sure i did that (`app.before_request`), must've been lost on `git add --patch`... i'll just re-add that lost feature.
from the top of my head, i can't really think of a strong reason why we wouldn't want to have the tokens work for the API as well :thinking:

invenio_rdm_records/links/models.py
```diff
@@ -0,0 +1,229 @@
...
+class SecretLink(db.Model):
+    """Secret links for sharing permissions on records."""
+
+    __tablename__ = "rdm_records_secret_links"
+
+    id = db.Column(UUIDType, primary_key=True, default=uuid.uuid4)
+    """Secret link ID."""
+
+    token = db.Column(
+        EncryptedType(type_in=db.Text, key=secret_key), nullable=False
```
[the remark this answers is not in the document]
max-moser: as per our discussion, i've removed the encryption for now, to potentially enable querying for SecretLinks via tokens, and enhance performance
in case the DB is leaked (and we know about it), we can still invalidate all existing shared links
an actual problem could still occur if the database is leaked and we don't know about it, though

invenio_rdm_records/links/models.py
```diff
@@ -0,0 +1,229 @@
...
+    )
+    """Secret token for link (should be stored encrypted)."""
+
+    owner_user_id = db.Column(
```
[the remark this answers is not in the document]
max-moser: as per our discussion, i've removed the ownership information here, as that's implicitly covered by the associated record (unless the links aren't used as intended, and the same link/token are re-used across multiple records)

invenio_rdm_records/links/models.py
max-moser: removed the `revoked_at` field, added a new `revoke_link(cls, link)` static method that removes the specified link from `db.session`

invenio_rdm_records/services/permissions.py
```diff
@@ -39,10 +39,7 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     can_read = [
         IfRestricted('record', then_=[RecordOwners()], else_=[AnyUser()]),
         SystemProcess(),
-    ]
-    can_read_files = [
```
max-moser, line 43: this one was duplicate, and repeated a bit below

invenio_rdm_records/records/systemfields/access/field.py
```diff
@@ -86,6 +95,11 @@ def grants(self):
         """The set of permission grants for the record."""
         return self._grants
 
+    @property
+    def links(self):
+        """The set of secret links for the record."""
+        return self._links
```
fenekku, line 101: question: could we just save `self.links` (not `self._links`) directly and save us this?

invenio_rdm_records/links/models.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+    """Creation timestamp."""
+
+    expires_at = db.Column(db.DateTime, nullable=True)
+    """Expiration date."""
+
+    permission_level = db.Column(db.String, nullable=False, default="")
+    """Permission level of the link."""
+
+    def covers(self, permission):
```
fenekku, line 51: Looking at tests, would go for `permission` -> `action`

(and could also go for `covers` -> `allows`)

but maybe the idea is to change this to an actual permission level ("Metadata curator" / "Owner" ...) down the road?

fenekku (approved):
Most of my comments are minor, but the renaming would really be helpful.

invenio_rdm_records/records/systemfields/access/field.py
```diff
@@ -60,6 +65,10 @@ def __init__(
         if grants is None:
             self._grants = grants_cls()
 
+        self._links = links
+        if links is None:
+            self._links = links_cls()
```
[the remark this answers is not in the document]
max-moser: i changed them to `self._x = x if x is not None else x_cls()`; that still avoids the falsy empty sets, and is more concise

invenio_rdm_records/alembic/0cf260eb8e97_create_table_for_secret_links.py
```diff
@@ -0,0 +1,50 @@
+#
+# This file is part of Invenio.
+# Copyright (C) 2016-2018 CERN.
```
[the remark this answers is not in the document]
max-moser: the woes of ^c ^v

invenio_rdm_records/records/systemfields/access/field.py
max-moser: hmm... i'm not 100% sure about my own reasoning here from back then anymore :thinking: 
the only upside of this `self._x` pattern here that i can think of right now is that it safeguards against unintended direct assignments (e.g. `record.access.links = 5`), which could break things
if that's not an issue though, the code would get much more concise if we did away with that.
your call :smiley_cat:

invenio_rdm_records/records/systemfields/access/links.py
```diff
@@ -0,0 +1,130 @@
...
+            extra_data=extra_data,
+            expires_at=expires_at,
+            commit=commit,
+        )
+        self.add(link)
+        return link
+
+    def needs(self, permission):
+        """Get allowed needs for the given permission level."""
+        needs = {link.need for link in self if link.covers(permission)}
```
lnielsen, line 125: I think you renamed the method to ``allows`` so it would be like below?
```suggestion
        needs = {link.need for link in self if link.allows(permission)}
```

invenio_rdm_records/secret_links/models.py
```diff
@@ -0,0 +1,172 @@
...
+        commit=False,
+    ):
+        """Create a new secret link."""
+        is_date = isinstance(expires_at, date)
+        is_datetime = isinstance(expires_at, datetime)
+        if is_date and not is_datetime:
+            expires_at = datetime.combine(expires_at, datetime.min.time())
+
+        with db.session.begin_nested():
+            obj = cls(
```
lnielsen, line 110: Here I would simply do:
```python
id_ = uuid.uuid4()
obj = cls(
    id=id_
    ....,
    token=serializer.create_token(id_, extra_data)
)
```

There's a super low chance of collission and we also do it other places (i.e. generating the uuid in the app)

invenio_rdm_records/secret_links/models.py
```diff
@@ -0,0 +1,172 @@
...
+            #       referenced in the token
+            if expires_at is None:
+                serializer = SecretLinkSerializer()
+            else:
+                serializer = TimedSecretLinkSerializer(expires_at=expires_at)
+
+            obj.token = serializer.create_token(obj.id, extra_data)
+
+        if commit:
+            db.session.commit()
```
lnielsen, line 131: Question: Is the commit needed here? Usually, I try to avoid having a commit inside the data layer as it can have unwanted side-effects

lnielsen (approved):
Just three minor commets - then it's good to merge

invenio_rdm_records/records/systemfields/access/links.py
max-moser: thanks, missed that one!

max-moser: actually... that should be the same logic as in the `SecretLinks` generator; i'll move the logic from the generator to this place, and call this function from the generator!

invenio_rdm_records/secret_links/models.py
max-moser: no, it was purely for convenience -- now removed, along with the optional commit in `revoke`

invenio_rdm_records/secret_links/models.py
max-moser: done; removed the now no longer required nested sessions

invenio_rdm_records/records/systemfields/access/field.py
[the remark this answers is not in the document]
fenekku: oh, `_links` can hold a set or a "Links class-like" that doesn't behave like a set? or at least doesn't have the `__bool__` of a set?  It's merged now, so we can see later...

invenio_rdm_records/records/systemfields/access/field.py
fenekku: This `e.g. record.access.links = 5` was also the only case I saw but wouldn't really happen (as in: code will break pretty quickly after this anyway and we would never parry all cases due to nature of the language). Was just wondering about it - i.e. if I don't *need* `property` I don't go for it so I was curious.

invenio_rdm_records/records/systemfields/access/field.py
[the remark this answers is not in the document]
max-moser: oh no, i was thinking that as soon as the user explicitly specified anything other than `None` for the `links` argument in the constructor, then that should be used (the default value *is* `None` for all the arguments, causing the fallback to the configured default classes)
in that case, the default semantics of `__bool__` for sets *are* the problem, because empty sets evaluate to `False` -- just like `bool(None)`, causing us to lose the ability to distinguish between the two cases
but what if the user wanted to specify their own thing, and it just happens to be empty?

the primary background of this thought is to give users of this API the possibility of easily overriding the default behavior
a la "not satisfied with the default `Links` list implementation that we supply? just pass something to the constructor that's more to your liking"
though thinking about it, that's probably never gonna be required anyway? :sweat_smile:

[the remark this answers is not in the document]
fenekku: > but what if the user wanted to specify their own thing, and it just happens to be empty?

then wouldn't that be equivalent to the default `links_cls()` anyway ? ... but really the below is the approach for me:

>  though thinking about it, that's probably never gonna be required anyway?

That's my thinking. I am definitely conservative on that front though (and maybe others have differing opinion): you ain't gonna need it yet and all that :smiley: .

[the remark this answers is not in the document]
max-moser: > then wouldn't that be equivalent to the default links_cls() anyway ? ... but really the below is the approach for me:

not if it's an instance of another custom class, different from the supplied `links_cls` :stuck_out_tongue: 
but yeah, let's simplify.

PR #443, opened 2021-03-11
global: initial versioning support

lnielsen:
closes inveniosoftware/invenio-drafts-resources#104


invenio_rdm_records/records/jsonschemas/records/record-v1.0.0.json
```diff
@@ -479,12 +479,6 @@
     },
     "pid": {"$ref": "#/definitions/internal-pid"},
 
-    "conceptid": {
```
lnielsen, line 482: For backward compatibility, we need to create a new v2.0.0 JSONSchema instead.

PR #447, opened 2021-03-15
alembic: add recipe to remove PIDRelations tables if they exist

### LINKED ISSUE #444: installation: remove invenio-pidrelations module
Remove invenio-pidrelations module from InvenioRDM as we're not going to to use it (might be in invenio-drafts-resources).

Add a alembic recipe which will delete the database table.

max-moser:
closes #444 

there's still a number of issues that need to be addressed, e.g. revision "1d4e361b7586" (from invenio-pidrelations) having no revision script but still being referenced from the DB table `alembic_versions` after uninstalling invenio-pidrelations
also, it seems like `alembic stamp` doesn't seem to work properly with the new revision script (it doesn't seem to find a parent?)

### DIFF
--- a/invenio_rdm_records/alembic/a3957490361d_remove_pidrelations_tables.py
+++ b/invenio_rdm_records/alembic/a3957490361d_remove_pidrelations_tables.py
@@ -1,0 +1,33 @@
+#
+# This file is part of Invenio.
+# Copyright (C) 2021 TU Wien.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Remove PIDRelations tables."""
+
+import sqlalchemy as sa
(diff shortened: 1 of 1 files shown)


invenio_rdm_records/alembic/a3957490361d_remove_pidrelations_tables.py
```diff
@@ -0,0 +1,88 @@
...
+branch_labels = ()
+depends_on = None
+
+
+def _get_existing_table_names():
+    """Get the names of existing tables."""
+    conn = op.get_bind()
+    inspector = Inspector.from_engine(conn)
+    tables = inspector.get_table_names()
+    return tables
```
max-moser, line 27: adapted from:
https://stackoverflow.com/questions/31299709/alembic-create-table-check-if-table-exists

PR #449, opened 2021-03-15
search: sort by versions

### LINKED ISSUE #440: versioning: sort by versions
Add a new sort option:

- UI needs to show a new "Versions" sort option
- Backend needs to support the version sort option as well.

fenekku:
- ~~:warning: requires versioning PR with ES indexing.~~
- closes #440 

Note: According to RFC

> `versioning_order` (`parent_order`) - Integer (4 bytes) defining the rank of the record among sibling versions.

is defined for the DB and we expect something similar in the ES index. Furthermore, we expect that every new version is from the latest record. That is:

**Good**
Time 1

```
Concept Record
|
|___ Record V1
```

Time 2

```
Concept Record
|
|___ Record V2
|
|___ Record V1
```

Time 3

```
Concept Record
|
|___ Record V3
|
|___ Record V2
|
|___ Record V1
```

Otherwise, it is extremely hard (I don't know how to) order versions. That is:

**Bad**
Time 4

```
Concept Record
|
|___ Record V3
|
|___ Record new version from V2  <-- how could this be ordered?
|
|___ Record V2
|
|___ Record V1
```
 
ES provides a new version field but only for 7.11+ on: https://www.elastic.co/guide/en/elasticsearch/reference/7.11/version.html .

With these constraints in place, what is the difference between sorting by newest and sorting by versions?

### DIFF
--- a/tests/services/test_sort.py
+++ b/tests/services/test_sort.py
@@ -28,9 +28,12 @@ def test_sort_by_versions(
     record = service.publish(draft.id, superuser_identity)
     # Create version 2
     draft = service.new_version(record.id, superuser_identity)
+    # NOTE: needed because publication_date is stripped from draft
+    service.update_draft(draft.id, superuser_identity, minimal_record)
     record_2 = service.publish(draft.id, superuser_identity)
     # Create version 3
     draft = service.new_version(record_2.id, superuser_identity)
+    service.update_draft(draft.id, superuser_identity, minimal_record)
     record_3 = service.publish(draft.id, superuser_identity)
     # NOTE: we swap version 2 and 3, so that "newest" order is different
     #       than "versions" order
--- a/tests/services/test_sort.py
+++ b/tests/services/test_sort.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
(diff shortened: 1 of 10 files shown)


fenekku:
With these constraints in place, the difference between this and ordering by "newest" is that version ordering can be changed in the future: a software patch to a previous (non-major) version would be created as the most recent version and then its version order/index would be changed to its appropriate position (and it siblings' order would have to be changed to accomodate that too). Then the "newest" sort would yield something different.

tests/services/test_sort.py
```diff
@@ -0,0 +1,54 @@
...
+    # Create version 2
+    draft = service.new_version(record.id, superuser_identity)
+    record_2 = service.publish(draft.id, superuser_identity)
+    # Create version 3
+    draft = service.new_version(record_2.id, superuser_identity)
+    record_3 = service.publish(draft.id, superuser_identity)
+    # NOTE: we swap version 2 and 3, so that "newest" order is different
+    #       than "versions" order
+    # Swap version 2 and 3 - have to reach into the datalayer to do so
+    record_2._record.versions.index = 3
```
fenekku, line 45: Can't assign. Do I reach deeper into the data record or do I modify the systemfield?

PR #450, opened 2021-03-16
services/records: integrate new version support

### LINKED ISSUE #435: versioning: implement version create
**Goal**

Ensure that the "new version" endpoint.

**Tasks**

- [x] Adapt data access and service layers
- [x] Each version should have it's own file buckets, pids, etc.
- [x] Make sure you can completely discard an unpublished draft (and that related objects are removed).
- [x] Ensure ``parent_latest`` is updated correctly.
- [x] On publish, reindex the previous latest version so that search index will show latest version fast

lnielsen:
* Adds support for creating new versions of records.

Requires inveniosoftware/invenio-drafts-resources#110

Closes #435


tests/records/dumpers/test_access_dumpers.py
```diff
@@ -10,25 +10,24 @@
...
     dumper = ElasticsearchDumper(
         extensions=[GrantTokensDumperExt("access.grant_tokens")]
     )
 
+    # Create the record
     minimal_record["access"]["grants"] = [
         {"subject": "user", "id": "1", "level": "view"},
         {"subject": "user", "id": "2", "level": "manage"},
     ]
-
```
lnielsen, line 29: Comment: ``RDMDraft`` is responsible for creating the parent record, so when we use ``RDMRecord`` directly without a draft, we have manually create it. Now, it can make sense to use ``RDMRecord`` because it also doesn't create buckets etc, hence no need for a files location.

PR #452, opened 2021-03-16
434 split access

### LINKED ISSUE #434: versioning: move access owners and grants to concept
**Goal**

Move the owners and grants from the child record to the parent record

**Tasks**

- Split the access system field into two system fields (reuse as much as possible)
   - ParentAccessField: grant, grants, owner, owners
   - ChildAccessField: embargo, protection
- Create a new version of the JSONSchema (``record-v2.0.0.json``) and adapt the parent jsonschema.
- Adapt existing code to work with the new fields.

max-moser:
closes #434

### DIFF
--- a/invenio_rdm_records/records/dumpers/access.py
+++ b/invenio_rdm_records/records/dumpers/access.py
@@ -31,10 +31,14 @@ class GrantTokensDumperExt(ElasticsearchDumperExt):
 
     def dump(self, record, data):
         """Dump the grant tokens to the data dictionary."""
-        if record.access:
-            tokens = [grant.to_token() for grant in record.access.grants]
-            parent_data = dict_lookup(data, self.keys, parent=True)
-            parent_data[self.key] = tokens
+        try:
+            if record.access:
+                tokens = [grant.to_token() for grant in record.access.grants]
+                parent_data = dict_lookup(data, self.keys, parent=True)
+                parent_data[self.key] = tokens
+
+        except KeyError:
+            pass
 
     def load(self, data, record_cls):
         """Remove the tokens from the data dictionary."""
--- a/invenio_rdm_records/services/components.py
+++ b/invenio_rdm_records/services/components.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2020 CERN.
 # Copyright (C) 2020 Northwestern University.
+# Copyright (C) 2021 TU Wien.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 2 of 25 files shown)


tests/services/components/test_access_component.py
```diff
@@ -20,19 +20,23 @@
...
+):
+    record = RDMRecord.create(minimal_record, parent=parent)
     component = AccessComponent(RDMRecordService())
     component.create(identity_simple, minimal_record, record)
 
-    assert len(record.access.owners) > 0
+    assert len(record.parent.access.owners) > 0
 
 
+# TODO this test doesn't currently make sense (the parents will handle owners)
```
max-moser, line 33: these tests test functionality that has been moved to the currently unused `ParentRecordAccessComponent`, so i'm skipping them for now

tests/services/test_generators.py
```diff
@@ -90,7 +99,7 @@ def test_ifrestricted_query():
     }
 
 
-def test_record_owner(mocker):
+def test_record_owner(app, mocker):
```
max-moser, line 102: without the app fixture, i got an error "working outside of app context"

invenio_rdm_records/services/schemas/access.py
```diff
@@ -90,11 +86,12 @@ class AccessSchema(Schema, FieldPermissionsMixin):
...
     files = SanitizedUnicode(required=True)
-    owned_by = fields.List(fields.Nested(Agent))
     embargo = fields.Nested(EmbargoSchema)
-    # TODO re-enable when the grants are ready
+
+    # TODO move them to the parent schema
     # grants = fields.List(fields.Nested(Grant))
-    links = fields.List(fields.Nested(Link))
+    # owned_by = fields.List(fields.Nested(Agent))
+    # links = fields.List(fields.Nested(Link))
```
max-moser, line 94: these should be moved to the parent record marshmallow schema

invenio_rdm_records/services/components.py
```diff
@@ -13,23 +14,24 @@
     ServiceComponent
 from marshmallow import ValidationError
 
-from ..records.systemfields.access import Access
+from ..records.systemfields.access import RecordAccess
 
 
-class AccessComponent(ServiceComponent):
-    """Service component for access integration."""
+class ParentRecordAccessComponent(ServiceComponent):
```
max-moser, line 20: this one's currently unused, but i guess we might need it later?

invenio_rdm_records/records/dumpers/access.py
```diff
@@ -31,10 +31,14 @@ def __init__(self, target_field):
 
     def dump(self, record, data):
         """Dump the grant tokens to the data dictionary."""
-        if record.access:
```
max-moser, line 34: ever since the `record.access` field gets initialized even when there's no `record["access"]` specified, this check was insufficient.
in case that there was a way to have `record["access"]` be different from `data["access"]`, this check would've been wrong even then :thinking:

PR #454, opened 2021-03-17
global: add functionality for reindexing everything

### LINKED ISSUE #441: global: support easy reindexing of entire database
During an upgrade from v1.0 to v2.0 we will upgrade the database. The easiest way for now to handle Elasticsearch indexes is to destroy the index and reindex them.

Today it's non-trivial to reindex records.

We will have to reindex:

- Records
- Drafts
- Vocabularies
- ~~Files / Draft Files~~

The data should be queried and fetched from the database (not the search index because it doesn't exists). This should be implemented via the services. E.g.:

```
service.rebuild_index(identity)
```

Something like this might do the job:
```
# Filter out deleted records/drafts (record.is_deleted)
for record in RDMRecord.query.all():
    service.indexer.index(record)
```

It should be runnable from the CLI for now:
```
invenio index destroy
invenio index init
invenio rdmrecords rebuild-index
```

The goal is that we can run these commands as part of

```
invenio-cli upgrade
```

and that a user can run them:

```
invenio-cli upgrade rebuild-index
```

max-moser:
closes #441 
requires https://github.com/inveniosoftware/invenio-drafts-resources/pull/112 and inveniosoftware/invenio-records-resources/pull/211

* add service methods for reindexing records and drafts
* add new CLI command for reindexing everything, using the new service
  methods of the records service and the vocabularies service

### DIFF
--- a/invenio_rdm_records/cli.py
+++ b/invenio_rdm_records/cli.py
@@ -17,8 +17,11 @@ from faker import Faker
 from flask.cli import with_appcontext
 from flask_principal import Identity
 from invenio_access.permissions import system_identity
+from invenio_vocabularies.proxies import \
+    current_service as current_vocabularies_service
 
 from .fixtures import FixturesEngine
+from .proxies import current_rdm_records
 from .services import RDMDraftFilesService, RDMRecordService
 from .vocabularies import Vocabularies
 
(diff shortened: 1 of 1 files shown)


invenio_rdm_records/cli.py
```diff
@@ -17,8 +17,11 @@
 from flask.cli import with_appcontext
 from flask_principal import Identity
 from invenio_access.permissions import system_identity
+from invenio_vocabularies.proxies import \
+    current_service as current_vocabularies_service
 
 from .fixtures import FixturesEngine
+from .proxies import current_rdm_records
 from .services import RDMDraftFilesService, RDMRecordService
```
max-moser, line 25: should we maybe have the other CLI commands fetch their record/draft services from the `current_rdm_records` too?
i think, that would honor custom service configurations

invenio_rdm_records/services/services.py
```diff
@@ -21,6 +21,36 @@ class RDMRecordService(RecordDraftService):
     config_name = "RDM_RECORDS_BIBLIOGRAPHIC_SERVICE_CONFIG"
     default_config = config.RDMRecordServiceConfig
 
+    def rebuild_index(self, identity):
```
lnielsen, line 24: Will you move this method to drafts-resources similar to your comment in invenio-vocabularies so we only keep the cli command above here?

invenio_rdm_records/cli.py
lnielsen: Yes, that would be great......they should all use the initialised versions instead of creating it on the spot......i fixed another one like this in App-RDM today.

PR #465, opened 2021-03-22
resources: add more links to record serialization

### LINKED ISSUE #438: resource: serialisation should include links
The REST API serialization should include links to:

- The access links endpoint (if user has manage access)
- The latest version api (all) - GET /api/records/:id/latest
- The latest version html (all) - GET /records/:id/latest
- The search all version (all) - GET /records/:id/versions

```python
{
    "id": "...",
    "links": {
        # ...
        "access_links": "https://localhost:5000/api/records/:id/access/links",
        "versions": "https://localhost:5000/api/records/:id/versions",
        "latest": "https://localhost:5000/api/records/:id/versions/latest",
        "latest_html": "https://localhost:5000/records/:id/latest",
    }
}
```

max-moser:
closes #438 
* add links to the secret links endpoint, as well as versions-related endpoints

### DIFF
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -28,6 +28,12 @@ RecordLinks = LinksSchema.create(links={
     "self": ItemLink(template='/api/records/{pid_value}'),
     "self_html": ItemLink(template='/records/{pid_value}'),
     "files": ItemLink(template='/api/records/{pid_value}/files'),
+    "versions": ItemLink(template='/api/records/{pid_value}/versions'),
+    "latest": ItemLink(template='/api/records/{pid_value}/versions/latest'),
+    "latest_html": ItemLink(template='/records/{pid_value}/versions/latest'),
+    "access_links": ItemLink(
+        template='/api/records/{pid_value}/access/links', permission="manage"
+    ),
 })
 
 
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/resources/config.py
```diff
@@ -28,17 +28,29 @@
...
 })
 
 
 DraftLinks = LinksSchema.create(links={
     "self": ItemLink(template="/api/records/{pid_value}/draft"),
     "self_html": ItemLink(template="/uploads/{pid_value}"),
     "files": ItemLink(template="/api/records/{pid_value}/draft/files"),
+    "versions": ItemLink(template='/api/records/{pid_value}/versions'),
+    "latest": ItemLink(template='/api/records/{pid_value}/versions/latest'),
+    "latest_html": ItemLink(template='/records/{pid_value}/versions/latest'),
```
fenekku, line 46: These might point to something not existing right? Is that ok?

invenio_rdm_records/resources/config.py
```diff
@@ -28,17 +28,29 @@
...
+    "access_links": ItemLink(
+        template='/api/records/{pid_value}/access/links', permission="manage"
```
fenekku, line 48: might be nice to test for this.

fenekku (approved):

invenio_rdm_records/resources/config.py
max-moser: yes indeed
right now, those links would simply 404, which isn't *entirely* wrong, but it'd probably be saner to only include them if it wouldn't lead to a 404 (more effort to make this conditional, of course), or simply remove them

fenekku: We have this: https://github.com/inveniosoftware/marshmallow-utils/blob/master/marshmallow_utils/fields/links.py#L52 for conditional rendering.

But it's probably fine for now...

fenekku (approved):

PR #467, opened 2021-03-22
ui: add serialization of version

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/701

### DIFF
--- a/invenio_rdm_records/resources/serializers/ui/schema.py
+++ b/invenio_rdm_records/resources/serializers/ui/schema.py
@@ -86,6 +86,14 @@ def localize_vocabulary_list(field_name, key, attr, obj):
     return localized
 
 
+def record_version(obj):
+    """Return record's version."""
+    field_data = obj.get("metadata", {}).get("version")
+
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -86,6 +86,14 @@ def localize_vocabulary_list(field_name, key, attr, obj):
     return localized
 
 
+def record_version(obj):
+    """Return record's version."""
+    field_data = obj.get("metadata", {}).get("version")
+
+    if not field_data:
+        return obj["versions"]["index"]
```
fenekku, line 94: On the landing page we have `Version {version}` and that looks fine even when this is used. On the frontend search results if this is used we get: `2010-10-02 (1)` which is maybe a little strange ;) . On Zenodo, it shows up as `v<version>` 
image


What about 
```suggestion
        return f"v{obj["versions"]["index"]}"
```
? or something on the frontend...

PR #476, opened 2021-03-24
versions: fix version list when access with secret link

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/719


max-moser (approved):
looking good! :+1:

PR #483, opened 2021-03-24
schema: replace ISODateString with DateTime for secret links

max-moser:


tests/resources/test_resources.py
```diff
@@ -52,7 +52,7 @@ def _validate_access(response, original):
 
         until = arrow.get(embargo["until"]).datetime
         orig_until = arrow.get(orig_embargo["until"]).datetime
-        assert until.strftime("%Y-%m-%d") == orig_until.strftime("%Y-%m-%d")
+        assert until.isoformat() == orig_until.isoformat()
```
max-moser, line 55: this got sucked into the updating spree, but the slightly more precise check shouldn't hurt either

zzacharo (approved):

PR #491, opened 2021-04-01
wip external pids

ntarocco:


invenio_rdm_records/services/components.py
```diff
@@ -139,3 +139,56 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
...
+
+
+class ExternalPIDsComponent(ServiceComponent):
+    """."""
+
+    def _pid_exists(self, external_pid):
+        """Validate external PID with provider."""
+        pid_provider = external_pid["provider"]
+        providers = self.config.external_pids_providers
+        assert pid_provider in providers
```
ppanero, line 151: minor/general practice: It could be that is not there due to config change or migration, right? do we want hard failures (i.e. assert) or better raising a custom exception that could be treated and send back properly via REST to the user?

invenio_rdm_records/records/jsonschemas/records/record-v2.0.0.json
```diff
@@ -123,7 +123,8 @@
       "additionalProperties": false,
       "required": ["identifier", "provider"],
       "properties": {
-        "identifier": {"$ref": "#/definitions/identifier"},
+        "scheme": {"$ref": "#/definitions/scheme"},
```
ppanero, line 126: As per the issue description, the `scheme` would not go here. It would be the key inside the pids object.

```
"pids": {
        "doi": {
            "identifier": "10.5281/zenodo.1234",
            "provider": "datacite",
            "client": "zenodo"
        },
```

we are missing `client` though. What is the benefit of having it as key instead of `scheme`, if its access patterns as dict saves us loops it makes sense. Otherwise, shall we move it inside (as you put it).

invenio_rdm_records/records/jsonschemas/records/record-v2.0.0.json
```diff
@@ -123,7 +123,8 @@
       "additionalProperties": false,
       "required": ["identifier", "provider"],
```
ppanero, line 124: shall we move these restrictions to the marshmallow schema? that way we can properly save drafts that have empty values in some fields but make sure requirements are met upon publishing action.

invenio_rdm_records/external_pids/providers.py
```diff
@@ -0,0 +1,28 @@
+
+
+class Provider:
```
ppanero, line 3: we have a base `provider` and for datacite too in [invenio-pidstore](https://github.com/inveniosoftware/invenio-pidstore/blob/master/invenio_pidstore/providers/datacite.py) this naming is confusing 😅. Shall we move them there/rename? whats the vision on this aspect?

ppanero (commented):
I know is a WIP, just leaving comments not to forget and have discussion points.

invenio_rdm_records/external_pids/providers.py
ppanero: IRL: continue with this, having in mind the base provider of pidstore, to simplify things for now.

PR #499, opened 2021-04-08
global: access status label on search and landing pages

diegodelemos:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/667.

Depends on:
- https://github.com/inveniosoftware/invenio-records/pull/252
- https://github.com/inveniosoftware/invenio-records-resources/pull/233


invenio_rdm_records/cli.py
```diff
@@ -69,14 +72,33 @@ def fake_date(end_date=None):
...
-            "files": "public",
-        },
+            RecordAccessStatusEnum.open: {
+                "record": "public",
+                "files": "public",
+            },
+            RecordAccessStatusEnum.embargoed: {
+                "record": "public",
+                "files": "restricted",
+                "embargo": {
```
diegodelemos, line 87: Conflicts with https://github.com/inveniosoftware/invenio-app-rdm/issues/717. It can be removed, added to make it easier to try out.

diegodelemos:
Status update: The idea was to move `access_status` inside the existing [`AccessSchema`](https://github.com/inveniosoftware/invenio-rdm-records/blob/78ad536dbb95494967bf8de248cf922e5040e844/invenio_rdm_records/services/schemas/access.py#L55), following the next steps:
1. ✅.  Pass a new field from the data access layer to the service layer (`access.status`, which is basically a computed property, https://github.com/inveniosoftware/invenio-rdm-records/pull/499/commits/0f1182cb7ed070f7d009b37a569f35dc0814d338) 
2. ✅.   Use the new `access.status` to create `ui.access_status` which is used in the UI, instead of accessing the data-access layer from the marshmallow schemas (breaking design).
3. 🔴.  At this point there are two scenarios
    a. If you don't remove `access.status` from the response, then the UI sends it back to the server and triggers a validation error when saving drafts even if `dump_only=True` is set
    b. If you remove `access.status` from the response (poorly done in https://github.com/inveniosoftware/invenio-rdm-records/pull/499/commits/14ceaa26de96ca98ed7aa27bd9918205197adb98). In this case, saving a draft shows a sucess message, however, the draft is not listed in uploads. Publishing, on the other hand, works fine.

My questions are:
i. Which path to take `3.a.` or `3.b.`, depends on whether `status` should be present in the UI inside `access` key
ii. If we go for `3.b.` any ideas on how to unblock the described situation in https://github.com/inveniosoftware/invenio-rdm-records/pull/499/commits/14ceaa26de96ca98ed7aa27bd9918205197adb98?

invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -104,6 +105,13 @@ def default_nested(self, data, many, **kwargs):
 
         return data
 
+    @pre_load()
+    def remove_status(self, data, many, **kwargs):
+        """Remove access status if present in incoming serialized object."""
+        if data.get('access', {}).get('status'):
+            data.get('access', {}).pop('status')
+        return data
```
diegodelemos, line 113: @zzacharo I had to add this, otherwise, the UI would send back the `access.status` and a validation error would occur with no message. The service layer schema fails as there is no field `access.status`. Any idea how to better tackle it?

zzacharo: I think the UI should handle the pop of this data....the REST API is fine I think to fail with validation error if the status is passed as it's dump_only field... Now, in the UI we have a serializer class (See https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/DepositRecordSerializer.js#L86). You need to create a class e.g `AccessField` and control the (de)serialization for the [access](https://github.com/inveniosoftware/react-invenio-deposit/blob/master/src/lib/DepositRecordSerializer.js#L59) field!

Issue #501, opened 2021-04-09
global: add `pids` field to record


ppanero:
Currently the `pids` field exists at service.schema level. However, even if it will correctly validate, the data will never reach the DB/ES and in consequence, the user.

This is because it gets stripped before the components run, and no component introduces it back.

Task:

-  [x] Create an `ExternalPIDsComponent`, ideally it would be called `PIDsComponent` but it would become confusing with the existing `PIDComponent` used to manage the internal `pid` field (for *recids*).
- [x] Add a systemfield to both Draft and Record objects (use `CommonFieldsMixin`. This will allow the usage as dict and prop on the `record` object instances.
- [x] Modify the cli.py to add said data.

PR #502, opened 2021-04-09
services: add pids systemfield and component

ppanero:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/501

Preliminary works, this will allow getting in unmanaged pids. Works on provider and more complex operations on the component will come from the work on managed pids.

### DIFF
--- a/invenio_rdm_records/services/components.py
+++ b/invenio_rdm_records/services/components.py
@@ -139,3 +139,27 @@ class MetadataComponent(ServiceComponent):
         # (publication date and version)
         for f in self.new_version_skip_fields:
             draft.metadata.pop(f, None)
+
+
+class ExternalPIDsComponent(ServiceComponent):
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/services/components.py
```diff
@@ -139,3 +139,27 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
         # (publication date and version)
         for f in self.new_version_skip_fields:
             draft.metadata.pop(f, None)
+
+
+class ExternalPIDsComponent(ServiceComponent):
```
ntarocco, line 144: Shall we rename all of them to `PIDsComponent`?

ntarocco (approved):
We should add validation that the `provider` must be `external` in this case, with no `client`, but I guess it will come in the next step!

PR #506, opened 2021-04-11
enhance pids field validation

### LINKED ISSUE #504: service: viability of IdentifierSchema for the `pids` field
The `pids` field at root level of the schema is defined with three fields, but no scheme, since that is the key of the dict. However, marshmallow-utils has implemented a field that could be used here.

How do we deal with validation? The required validation is already implemented in said field, however we cannot easily pass the scheme.

ppanero:
Closes #504 

Issues pending:

marshmallow-utils::IdentifierScheme raises a ValidationError if the scheme cannot be validated (i.e. is not supported by idutils). Currently, this is the [list of supported schemes](https://idutils.readthedocs.io/en/latest/). This makes the following tests fail:

- test_valid_unknown_scheme: because the scheme is unknown (random) and idutils cannot detect it
- test_valid_pid: because OAI is not supported by idutils

Both cases end in the same question:

- Do we want to fail when idutils cannot validate?
- If yes, shall we study what possible identifiers will we have to validate? (e.g. it should not fail on OAI).

### DIFF
--- a/invenio_rdm_records/services/schemas/__init__.py
+++ b/invenio_rdm_records/services/schemas/__init__.py
@@ -14,6 +14,7 @@ from invenio_drafts_resources.services.records.schema import RecordSchema
 from marshmallow import EXCLUDE, ValidationError, fields, post_dump, validates
 from marshmallow_utils.fields import NestedAttribute
 from marshmallow_utils.permissions import FieldPermissionsMixin
+from marshmallow_utils.schemas import IdentifierSchema
 
 from .access import AccessSchema
 from .metadata import MetadataSchema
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -88,14 +86,18 @@ class Meta:
...
-            raise ValidationError(
-                    _("Unkown PID type"),
+        for scheme, pid_attrs in value.items():
+            id_schema = IdentifierSchema(allow_all=True)
+            id_schema.load({
+                "scheme": scheme,
+                "identifier": pid_attrs.get("identifier")
+            })
+
+            is_provider_external = pid_attrs.get("provider") == "external"
```
ntarocco, line 96: for the next PR: I would put this as a "const" `PROVIDER_EXTERNAL = "external" in the service so it can be re-used.

ntarocco (approved):
I think that it is ok NOT to fail, because for sure some organizations will have an ID that is not part of idutils...

invenio_rdm_records/services/schemas/__init__.py
ppanero: Fixed, thanks!

ppanero:
Thanks @ntarocco, I openend an issue (card in our board for now) to support non-idutils schemes, I commented the test so the suite goes correctly. Merging this :)

PR #510, opened 2021-04-12
[WIP] service components: create pids using providers

### LINKED ISSUE #507: service: support managed PIDs
To support managed PIDs the following must be supported:

- [x] If a PID scheme is present, but there is no `identifier` field or value is empty, a new one should be reserved.
- [x] If a PID scheme is present, along with the identifier value, its format validity should be checked, the status of said PID should already be reserved or registered. If not reserved, it should be reserved.
- [x] The provider **must not** be `external`.
- [x] If the PID scheme is not one of the configured providers, fail.
-  [x] Delegate the final (extra) validation/handling to the provider

ppanero:
External PIDs component hooks using providers...

closes #507

### DIFF
--- a/invenio_rdm_records/services/components.py
+++ b/invenio_rdm_records/services/components.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020 CERN.
+# Copyright (C) 2020-2021 CERN.
 # Copyright (C) 2020 Northwestern University.
 # Copyright (C) 2021 TU Wien.
 #
@@ -11,7 +11,9 @@
 
 from copy import copy
 
+from flask_babelex import lazy_gettext as _
 from invenio_access.permissions import system_process
+from invenio_pidstore.models import PIDStatus
 from invenio_records_resources.services.records.components import \
     ServiceComponent
 from marshmallow import ValidationError
@@ -144,22 +146,69 @@ class MetadataComponent(ServiceComponent):
 class ExternalPIDsComponent(ServiceComponent):
     """Service component for pids."""
 
+    def get_provider(self, scheme):
+        """Process a given PID with its provider."""
+        provider = self.service.config.pids_providers[scheme]
+
+        return provider
+
+    def _validate_pids(self, pids):
+        """Validate an iterator of pids."""
+        for scheme, pid_attrs in pids.items():
+            self.get_provider(scheme).validate(pid_attrs=pid_attrs)
+
     def create(self, identity, data=None, record=None, **kwargs):
         """Inject parsed pids to the record."""
-        record.pids = data.get('pids', {})
+        pids = data.get('pids', {})
+        self._validate_pids(pids)
+        record.pids = pids
 
     def update_draft(self, identity, data=None, record=None, **kwargs):
         """Inject parsed pids to the record."""
-        record.pids = data.get('pids', {})
+        pids = data.get('pids', {})
+        self._validate_pids(pids)
+        record.pids = pids
 
     def publish(self, identity, draft=None, record=None, **kwargs):
         """Update draft pids."""
-        record.pids = draft.get('pids', {})
+        pids = draft.get('pids', {})
+
+        for scheme, pid_attrs in pids.items():
+            provider = self.get_provider(scheme)
+            provider.validate(pid_attrs=pid_attrs)
+
+            identifier_value = pid_attrs.get("identifier")
+            if not identifier_value:
+                pid = provider.create()
+                pid_attrs["identifier"] = pid.value
+                pids[scheme] = pid_attrs
+
+            elif not provider.is_external():
+                raise ValidationError(
+                    _(f"Cannot create PID {scheme} with a given value," +
+                      "it must be assigned by the system."),
+                    field_name="pids"
+                )
+            else:
+                pid = provider.get(identifier_value)
+
+            # PIDS-FIXME: Move to provier.validate base class?
+            if pid.status != PIDStatus.RESERVED != PIDStatus.REGISTERED:
+                provider.reserve(pid)
 
     def edit(self, identity, draft=None, record=None, **kwargs):
         """Update draft pids."""
-        record.pids = record.get('pids', {})
+        # NOTE: getting from record instead of draft
+        # Do not allow to edit pids
+        pids = record.get('pids', {})
+        self._validate_pids(pids)
+        record.pids = pids
 
     def new_version(self, identity, draft=None, record=None, **kwargs):
         """Update draft pids."""
-        record.pids = data.get('pids', {})
+        pids = record.get('pids', {})
+
+        # PIDS-FIXME: Remove DOI (maybe all)
+        # new version should be no PIDS just concept?
+        self._validate_pids(pids)
+        record.pids = pids
--- a/invenio_rdm_records/services/config.py
+++ b/invenio_rdm_records/services/config.py
@@ -122,6 +122,10 @@ class RDMRecordServiceConfig(RecordServiceConfig):
     search_drafts = RDMSearchDraftsOptions
     search_versions = SearchVersionsOptions
 
+    # PIDs providers
+    # PIDS-FIXME: setup default providers
+    pids_providers = {}
+
     # Components
     components = [
         MetadataComponent,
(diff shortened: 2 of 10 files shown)


invenio_rdm_records/records/providers/__init__.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
```
ntarocco, line 1: just wondering about the module name, reading the path:
`/records/providers/...` should we have instead `/records/pidproviders/...` ?

invenio_rdm_records/records/providers/base.py
```diff
@@ -0,0 +1,56 @@
...
+class Provider:
+    """Base Provider class."""
+
+    def __init__(self, url, name, client=None, **kwargs):
+        """Constructor."""
+        self.url = url
+        self.name = name
+        self.client = client
+
+    def create(self, status=PIDStatus.RESERVED, **kwargs):
```
ntarocco, line 30: I am not 100% sure about the `create` and `reserve` signatures.
Is the `status` needed for `create`?

invenio_rdm_records/services/components.py
```diff
@@ -144,13 +148,75 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
...
         """Inject parsed pids to the record."""
-        record.pids = data.get('pids', {})
+        pids = data.get('pids', {})
+
+        for scheme, pid_attrs in pids.items():
+            if pid_attrs.get("provider") != PID_PROVIDER_EXTERNAL:
+                provider = self.get_provider(scheme)
+
+                # PIDS-FIXME: Do we assume it fails, or should we re-raise
+                # with ValidationErrors like AccessCompoenent (I think yes)
```
ntarocco, line 172: I would indeed have ValidationErrors as the other cmps given that we want to fail on a specific field.
If it was a generic error, than it probably makes more sense to have a generic one.
Eventually we change all of them at once.

invenio_rdm_records/services/components.py
```diff
@@ -144,13 +148,75 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
...
+                provider.validate(**pid_attrs)
+
+                # NOTE: When reserving a PID the record/draft was created
+                # and therefore any following action is an update
+                if pid_attrs.get("identifier"):
```
ntarocco, line 177: We will have to change the UI for this, I have added a card in the board.

invenio_rdm_records/services/components.py
```diff
@@ -144,13 +148,75 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
...
+        pids = data.get('pids', {})
+
+        for scheme, pid_attrs in pids.items():
+            if pid_attrs.get("provider") != PID_PROVIDER_EXTERNAL:
+                provider = self.get_provider(scheme)
+
+                # NOTE: value format is checked by the marshmallow schema
+                provider.validate(**pid_attrs)
+
+                identifier_value = pid_attrs.get("identifier")
```
ntarocco, line 206: I am not sure we should do this when saving... here my scenario:
1. click on new upload btn
2. input a title
3. hit the Save Draft button
4. suddenly the DOI has been reserved and appears in the deposit form

I think we might want to do that when publishing only?

invenio_rdm_records/services/config.py
```diff
@@ -122,6 +122,10 @@ class RDMRecordServiceConfig(RecordServiceConfig):
     search_drafts = RDMSearchDraftsOptions
     search_versions = SearchVersionsOptions
 
+    # PIDs providers
+    # PIDS-FIXME: setup default providers
+    pids_provides = {}
```
ntarocco, line 127:
```suggestion
    pids_providers = {}
```

invenio_rdm_records/records/providers/__init__.py
ppanero: Do we forsee having minters, resolvers, etc.? If yes, what about `/records/pids/providers/` allowing for `..../minters`, `.../resolvers`, etc.

invenio_rdm_records/records/providers/base.py
ppanero: Will take it into account in a future refactor, try to use the existing provider API fo pidstore.

invenio_rdm_records/services/components.py
ppanero: Adding a card to the board, it's quite self contained might be a good tasks to split

invenio_rdm_records/services/config.py
ppanero: Thanks!

invenio_rdm_records/services/components.py
ppanero: I think that is the difference between "reserve" and "register" and its responsability is even delegated to the Provider (so implementation-dependent). The way I see it:

1 and 2 stay the same
3. Upon save, the provider will be implemented in a way that the reserve only happens locally (i.e. just save in PIDStore).
4. A DOI will appear in the deposit form, but is not available in DataCite

I am not sure if your concern is the fact of reserve goint to DataCite, or also not creating a `DOI` in PIDStore. On the other hand, if we do as you say, there will be less "unused" pids due to random saves of drafts that went nowhere afterwards...

invenio_rdm_records/records/providers/__init__.py
ntarocco: at this time I don't think so, but you could anyway have `pids/providers`! :)

invenio_rdm_records/services/components.py
ntarocco: I think I was not able to explain myself :) Let's have a chat about this :)

invenio_rdm_records/records/pids/providers/datacite.py
```diff
@@ -0,0 +1,111 @@
...
+
+class DataCiteProvider(BaseProvider):
+    """Base Provider class.
+
+    Note that DataCite is only contacted when a DOI is reserved or
+    registered, or any action posterior to it. Its creation happens
+    only at PIDStore level.
+    """
+
+    def __init__(self, name, client=None, pid_type="doi",
```
ntarocco, line 25: I think there might a misunderstanding: for me, the client param is basically one a account of a given provider, e.g. zenodo or CDS on DataCite.
Then, in the `datacite` Python module, client is instead the REST API Python client lib.

What about if make `client` mandatory and not all the rest?
```
def __init__(self, name, client, pid_type="doi", default_status=PIDStatus.NEW, **kwargs):
    self.api_client = DataCiteRESTClient(client.username, client.password, client.prefix, client.test_mode)
```

invenio_rdm_records/services/components.py
```diff
@@ -144,22 +146,73 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
 class ExternalPIDsComponent(ServiceComponent):
     """Service component for pids."""
 
+    def get_provider(self, scheme):
+        """Process a given PID with its provider."""
+        provider = self.service.config.pids_providers.get(scheme)
+
+        if not provider:
+            raise ValidationError(
```
ntarocco, line 154: Here I would raise more a 500, because it is a misconfiguration of the system and not a 400, a wrong user input.
So raise maybe a `KeyError`? Generic `RDM Expection` if any?

invenio_rdm_records/services/components.py
```diff
@@ -144,22 +146,73 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
...
+            provider = self.get_provider(scheme)
+            provider.validate(pid_attrs=pid_attrs)
+
+            identifier_value = pid_attrs.get("identifier")
+            if not identifier_value:
+                pid = provider.create()
+                pid_attrs["identifier"] = pid.value
+                pids[scheme] = pid_attrs
+
+            elif not provider.is_external():
```
ntarocco, line 192: Here it might be that we have the `identifier` because it was reserved and retrieved in the form via btn Get DOI  now, or maybe I am wrong...

invenio_rdm_records/records/pids/providers/datacite.py
ppanero: I see... thanks! changed :)

invenio_rdm_records/services/components.py
ppanero: Im not sure I get this one, then in line 198 we do a get if this is the case so we have the `PersistentIdentifier` instance so we can actually check stuff on it.

invenio_rdm_records/services/components.py
ppanero: there is no generic that I know of, so I'm changing the accessor to [] so it KeyErrors if so.

Issue #518, opened 2021-04-16
invenio-rdm-records: full integration with DataCite


ppanero:
Integrate the full workflow with datacite:

- [x] Use datacite v1.1.0
- [x] Update service level schema and use Marshmallow-Utils v0.4.0
- [x] Add tests for the Base provider
- [x] Configure the service with the default providers
- [x] Use the latest `invenio-app-rdm` to display the DOI badge in the landing page
- [x] Integrate the `reserve` and `unregister` URL endpoints with the service level (e.g. use the DataCite provider). Note that it should be generic enough to be used by any other provider

PR #519, opened 2021-04-16
Enable External PIDs and more

ppanero:
depends on https://github.com/inveniosoftware/invenio-rdm-records/pull/520 and https://github.com/inveniosoftware/datacite/pull/64
partly addresses https://github.com/inveniosoftware/invenio-rdm-records/issues/518

- Re-enables the PIDs functionality. However, DOI is not yet added from the beginning since the serialization is not done yet.
- Adds some tests
- Links resource with service level for reserve and unreserve PIDs

**Important notes**

- Some details like serialization from record to DataCite and some cases of error handling are not yet finalized. Therefer some tests (resources/publish_pid, datacite provider, serializers, etc.) are not present or **removed** from this PR. I have them coded, but will include in a new one to avoid a `mark.skip`.
- Tests will fail unless credentials for datacite are given, we can put them in the CI... how do we do for people testing in local? I will try to mock/patch


invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -24,6 +25,25 @@ def __init__(self, username, password, prefix, test_mode, **kwards):
         self.test_mode = test_mode
 
 
+class DataCiteClientFromConfig(DataCiteClient):
```
ntarocco, line 28: I would simply use this class as `DataCiteClient` since I think 100% of the cases args will come from config.

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -24,6 +25,25 @@ def __init__(self, username, password, prefix, test_mode, **kwards):
...
+    def __init__(self, name, **kwards):
+        """Constructor."""
+        self.name = name
+        name = self.name.upper()
+
+        super(DataCiteClientFromConfig, self).__init__(
+            username=current_app.config[f"{name}_DATACITE_CLIENT_USERNAME"],
+            password=current_app.config[f"{name}_DATACITE_CLIENT_PASSWORD"],
+            prefix=current_app.config[f"{name}_DATACITE_CLIENT_PREFIX"],
+            test_mode=current_app.config[f"{name}_DATACITE_CLIENT_TEST_MODE"],
```
ntarocco, line 43: Ideally `test_mode` does not depend on the specific account on the DataCite, but honestly I kind of like it here... :)

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -32,7 +52,7 @@ class DataCiteProvider(BaseProvider):
     only at PIDStore level.
     """
 
-    def __init__(self, name, client, pid_type="doi",
+    def __init__(self, client, name="datacite", pid_type="doi",
```
ntarocco, line 55: food for thoughts: should name be a property instead that cannot be overridden via constructor?

```
class ..
    name = "datacite"

    def __init__ ....
```

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -41,7 +61,18 @@ def __init__(self, name, client, pid_type="doi",
         )
 
         super(DataCiteProvider, self).__init__(
-            name, client, pid_type, default_status)
+            name, self.client, pid_type, default_status)
+
+    def _generate_id(self, recid, **kwargs):
```
ntarocco, line 66: We need to pass, optionally, the record here. It was a request of a user who would like to customize the the suffix depending on some record fields. Can you add it?

invenio_rdm_records/services/services.py
```diff
@@ -342,19 +342,86 @@ def delete_secret_link(
...
 
-    def get_pid(
-        self, id_, identity,
-    ):
-        """Reserve doi."""
-        #  provider =
-        # find_the_provider_by_the_param_value
-        # (<provider param value>)
-        #  return provider.reserve(...)
+    def get_provider(self, scheme, client=None):
```
ntarocco, line 345: Can we isolate all this about PIDs in a new dedicated service?

invenio_rdm_records/services/services.py
```diff
@@ -342,19 +342,86 @@ def delete_secret_link(
...
+
+        return provider
+
+    def reserve_pid(self, id_, identity, pid_type):
+        """Reserve PID for a given record."""
+        draft = self.draft_cls.pid.resolve(id_, registered_only=False)
 
-        return "1234"
+        # Permissions
+        # PIDS-FIXME: manage is the correct perm type?
```
ntarocco, line 361: I think that it is ok, you can remove the comment.

invenio_rdm_records/services/services.py
```diff
@@ -342,19 +342,86 @@ def delete_secret_link(
...
+        self.require_permission(identity, "manage", record=draft)
 
-    def delete_pid(
-        self, id_, identity, doi,
-    ):
-        """Unreserve doi."""
-        return ""
+        # PIDS-FIXME: Where should client come from? URL?
+        provider = self.get_provider(pid_type, client="datacite")
```
ntarocco, line 365: Yes, it should be an arg and it should be validated that it exists.

invenio_rdm_records/services/services.py
```diff
@@ -342,19 +342,86 @@ def delete_secret_link(
...
+        if not provider:
+            # PIDS-FIXME: raise
+            return
```
ntarocco, line 368: raise ValidationError, `400`

invenio_rdm_records/services/services.py
```diff
@@ -342,19 +342,86 @@ def delete_secret_link(
...
+
+        pid = provider.create(id_)
+
+        # PIDS-FIXME: delegate to a minter?/component
```
ntarocco, line 372: I think that a component is not needed for this service: it should be a rather small simple service that knows what to do and what to answer.
Given that `client` and `provider` are given from the REST client that is making the call and they are validated, do we return only the `identifier` or also `provider` and `client`?
I am ok with both honestly, probably your solution is more flexible.

invenio_rdm_records/services/services.py
```diff
@@ -342,19 +342,86 @@ def delete_secret_link(
...
+        draft.pids[pid_type] = {
+            "identifier": pid.pid_value,
+            "provider": provider.name
+        }
+        if provider.client:
+            draft.pids[pid_type]["client"] = provider._client_credentials.name
+
+        draft.commit()
```
ntarocco, line 380: I am not sure I would this here. It might be that the user request a DOI, but then changes his/her mind and copy/paste another one, external. Wondering if you should simply reserve it in PIDStore, but not touch the draft.

invenio_rdm_records/ext.py
```diff
@@ -141,6 +141,6 @@ def init_resource(self, app):
 
         # PID provider
         self.pid_provider_resource = RDMPIDProviderResource(
-            service=None,   # PIDS-FIXME
+            service=self.records_service,
```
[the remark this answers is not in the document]
ppanero: Hmm Taking into account how it has been growing from parent to child class just adding functionality I thought it was better like this... However checking the `ext.py` registration I see this:
```
self.records_service = RDMRecordService(
            RDMRecordServiceConfig,
            files_service=FileService(RDMFileRecordServiceConfig),
            draft_files_service=FileService(RDMFileDraftServiceConfig),
        )
self.subjects_service = subject_record_type.service_cls(
            config=subject_record_type.service_config_cls,
        )
```

And I start to doubt 😅 What makes a service qualify to be a separate service (like subjects)? or to make it an attribute service of the main RecordsService (like files)? my gut tells me dependency of config/functions.

The potenatially separated `PIDsService` requires knowledge about the `record_cls` in some cases to get the record and add the specific pid.

Linking to your question below about not doing `drafts.commit()`. If we do that (not commit) then we need to **fully rely** on the frontend to have control and remember things (which might be difficult for API requests, i.e. they need to keep the json and know which value was returned). In addition, then when publishing the record the pid needs to come from the data (which in other cases, like access comes from the saved record). Moreover, if we want to merely "reserve" a PID no-matter of the record, then we should not have the `records/pid_value/drafts` in the endpoint...

TL;DR: I could separate, but it needs knowledge of some config and functions, so it will be a separated inheritance. Another option is make it an "attribute" service like the files one.

invenio_rdm_records/resources/config.py
```diff
@@ -188,21 +188,22 @@ class RDMParentRecordLinksResourceConfig(RecordResourceConfig):
 class RDMPIDProviderResourceConfig(RecordResourceConfig):
```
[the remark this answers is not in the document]
ppanero: I agree, I changed all mention to "external" for "unmanaged"... except the component name itself (`ExternalPIDsComponent`), which name does only had to do with the fact that `PIDsComponent` is for internal ones...

invenio_rdm_records/services/services.py
```diff
@@ -342,19 +342,86 @@ def delete_secret_link(
...
-        """Reserve doi."""
-        #  provider =
-        # find_the_provider_by_the_param_value
-        # (<provider param value>)
-        #  return provider.reserve(...)
+    def get_provider(self, scheme, client=None):
+        """Process a given PID with its provider."""
+        provider_class = self.config.pids_providers[scheme]
+        if client:
+            provider = provider_class(
```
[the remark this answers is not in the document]
ppanero: Agree thanks!

invenio_rdm_records/services/pids/providers/datacite.py
ppanero: I think is the easiet way to state it... since we need to put the rest of the variables in config... even if not changed much, it's a nice to have there just for consistency.

invenio_rdm_records/services/pids/providers/datacite.py
ppanero: Wouldnt be enough to pass it as a kwargs? and then inherit and customize the `_generate_id` function?

invenio_rdm_records/services/services.py
ppanero: See first long comment :)

invenio_rdm_records/services/services.py
ppanero: I agree. Now in case of "no client specified by the user" shall we fail? or default? example: Someone requests a DOI, should they have knowledge we accept datacite, cross-ref etc.? Or if nothing specified we simply give one...

ppanero:
Should datacite be an extra? a full `pids` extra or one per pid_type/provider?

invenio_rdm_records/services/pids/providers/datacite.py
ppanero: Agree on this one, I have changed the code to reflect it.

invenio_rdm_records/services/services.py
ppanero: IRL: yes, should end up in something like: ..../pids/doi?client=inveniordm and then the provider should know how to default if no client.

invenio_rdm_records/services/pids/providers/datacite.py
ppanero: IRL, instead of `[]`, do get and default to `True`

invenio_rdm_records/resources/config.py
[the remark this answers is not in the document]
ppanero: Decission pending form Lars, about `provider: "external"` --> `provider: "unamanged"` to avoid confusion.

invenio_rdm_records/ext.py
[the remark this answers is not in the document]
ppanero: IRL, keep as it is :) Let's retake this in the near future, might require refactor.

ppanero:
@lnielsen Tests passing, instance bootstrapped and insalled all latest (there were nothing else merged in invenio-records-resources, invenio-drafts-resources at this time)

Screeshots of it working (plus cli has tests now 🚀 ). Landing page (no doi, since serialization is pending):

Screenshot 2021-04-20 at 15 11 19

working with:

```
invenio==3.4.0
invenio-access==1.4.2
invenio-accounts==1.4.3
invenio-admin==1.3.0
invenio-app==1.3.1
invenio-app-rdm==3.0.0.dev3
invenio-assets==1.2.5
invenio-base==1.2.3
invenio-cache==1.1.0
invenio-celery==1.2.2
invenio-cli==0.23.0
invenio-config==1.0.3
invenio-db==1.0.9
invenio-drafts-resources==0.11.3
invenio-files-rest==1.2.0
invenio-formatter==1.1.0
invenio-i18n==1.3.0
invenio-iiif==1.1.0
invenio-indexer==1.2.1
invenio-jsonschemas==1.1.2
invenio-logging==1.3.0
invenio-mail==1.0.2
invenio-oaiserver==1.2.0
invenio-oauth2server==1.3.2
invenio-oauthclient==1.4.4
invenio-pidstore==1.2.2
invenio-previewer==1.3.2
-e git+https://github.com/ppanero/invenio-rdm-records.git@f1b1bb88ae5ce10edfe551b747763ff427895f90#egg=invenio_rdm_records
invenio-records==1.5.0a5
invenio-records-files==1.2.1
invenio-records-permissions==0.11.0
invenio-records-resources==0.14.8
invenio-records-rest==1.8.0
invenio-records-ui==1.2.0
invenio-rest==1.2.3
invenio-search==1.4.1
invenio-search-ui==2.0.2
invenio-theme==1.3.8
invenio-userprofiles==1.2.1
invenio-vocabularies==0.5.3
```

ntarocco (approved):

PR #521, opened 2021-04-20
schemas: nested to nestedattribute

lnielsen:


fenekku:
something weird with git. trying to close it and opening another.

PR #526, opened 2021-04-20
DataCite serialization

ppanero:
addresses serialization part of https://github.com/inveniosoftware/invenio-rdm-records/issues/518

Based the test in @tmorrell's previous work and [datacite schema examples/definitions](https://github.com/datacite/schema/blob/master/source/json/kernel-4.3/example/datacite-example-full-v4.json#L64). For some reason I cannot request your review @tmorrell... but feel free to do so, I'd be happy to get your feedback. 

Take into account there are corner cases and work to be done still, but this is a first working case.

PR #527, opened 2021-04-20
DOI testing and refactor

ntarocco:
Requires #526 

Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/544

### DIFF
--- a/invenio_rdm_records/services/components.py
+++ b/invenio_rdm_records/services/components.py
@@ -11,9 +11,7 @@
 
 from copy import copy
 
-from flask_babelex import lazy_gettext as _
 from invenio_access.permissions import system_process
-from invenio_pidstore.models import PIDStatus
 from invenio_records_resources.services.records.components import \
     ServiceComponent
 from marshmallow import ValidationError
@@ -149,13 +147,15 @@ class ExternalPIDsComponent(ServiceComponent):
     def _validate_pids(self, pids):
         """Validate an iterator of pids."""
         for scheme, pid_attrs in pids.items():
-            self.service.get_provider(
-                scheme, pid_attrs.get("client")).validate(pid_attrs=pid_attrs)
+            client = pid_attrs.get("client")
+            provider = self.service.get_provider(scheme, client)
+            provider.validate(**pid_attrs)
 
     def create(self, identity, data=None, record=None, **kwargs):
-        """Inject parsed pids to the record."""
+        """Inject parsed pids to the draft record."""
         pids = data.get('pids', {})
         self._validate_pids(pids)
+        # NOTE: record is a draft because we hook to the draft service.
         record.pids = pids
 
     def update_draft(self, identity, data=None, record=None, **kwargs):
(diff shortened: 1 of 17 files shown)


invenio_rdm_records/services/components.py
```diff
@@ -116,7 +115,7 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
 class MetadataComponent(ServiceComponent):
     """Service component for metadata."""
 
-    new_version_skip_fields = ['publication_date', 'version']
+    new_version_skip_fields = ['publication_date', 'version', 'pids']
```
ppanero, line 118: `pids` is an external (root) field, should not need any exception here...

invenio_rdm_records/services/components.py
```diff
@@ -166,47 +166,60 @@ def update_draft(self, identity, data=None, record=None, **kwargs):
...
 
     def new_version(self, identity, draft=None, record=None, **kwargs):
         """Update draft pids."""
-        pids = record.get('pids', {})
-
-        # PIDS-FIXME: Remove DOI (maybe all)
-        # new version should be no PIDS just concept?
-        self._validate_pids(pids)
-        draft.pids = pids
+        draft.pids = {}
```
ppanero, line 225: we should keep concept-doi here no?

invenio_rdm_records/services/pids/providers/__init__.py
```diff
@@ -8,13 +8,13 @@
 """PID Providers module."""
 
 from .base import BaseClient, BasePIDProvider
-from .datacite import DataCiteClient, DataCitePIDProvider
+from .datacite import DOIDataCiteClient, DOIDataCitePIDProvider
```
ppanero, line 11: isn't DOI an overdo in the name? IMO it makes it longer and it does not provide much, since datacite is always doi? or are there more possible pid_types for datacite?

invenio_rdm_records/services/pids/providers/base.py
```diff
@@ -23,19 +24,21 @@ def __init__(self, username, password, url=None, **kwards):
 class BasePIDProvider:
     """Base Provider class."""
 
-    name = "base"
+    name = None
```
ppanero, line 27: None might fail validation, But your idea is that "No PID should be created with the base provider"?

invenio_rdm_records/services/pids/providers/base.py
```diff
@@ -23,19 +24,21 @@ def __init__(self, username, password, url=None, **kwards):
...
         raise NotImplementedError
 
-    def __init__(self, client=None, pid_type=None,
-                 default_status=PIDStatus.NEW, system_managed=True, **kwargs):
+    def __init__(self, api_client=None, pid_type=None,
+                 default_status=PIDStatus.NEW, system_managed=True,
+                 required=False, **kwargs):
         """Constructor."""
-        self.client = client
+        self.api_client = api_client
```
ppanero, line 37: In base... why "api", might not be, right?

invenio_rdm_records/services/components.py
ppanero: removed

invenio_rdm_records/services/components.py
ppanero: no, doing in publish will make it easier

invenio_rdm_records/services/pids/providers/__init__.py
ppanero: used due to `ConceptDOI` provider... feel free to change.

invenio_rdm_records/services/pids/providers/base.py
ppanero: somehow force the user to put the name. Fix the tests.

invenio_rdm_records/services/pids/providers/base.py
ppanero: make more sense to avoid confusion with credentials

PR #528, opened 2021-04-20
Nestedattr

fenekku:
fix the original nested attribute PR

### DIFF
--- a/invenio_rdm_records/services/schemas/__init__.py
+++ b/invenio_rdm_records/services/schemas/__init__.py
@@ -35,13 +35,18 @@ class RDMRecordSchema(RecordSchema, FieldPermissionsMixin):
         'files': 'read_files',
     }
 
+    # ATTENTION: In this schema you should be using the ``NestedAttribute``
+    # instead  of Marshmallow's ``fields.Nested``. Using NestedAttribute
+    # ensures that the nested schema will receive the system field instead of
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -35,13 +35,18 @@ class RDMRecordSchema(RecordSchema, FieldPermissionsMixin):
...
+    # the record dict (i.e. record.myattr instead of record['myattr']).
+
     pids = fields.Dict(keys=fields.String(), values=fields.Nested(PIDSchema))
     metadata = NestedAttribute(MetadataSchema)
     # ext = fields.Method('dump_extensions', 'load_extensions')
     # tombstone
     # provenance
     access = NestedAttribute(AccessSchema)
-    files = fields.Nested(FilesSchema)
+    files = NestedAttribute(FilesSchema)
```
fenekku, line 49: After spending a day to understand the underlying issues... I don't think NestedAttribute gives us much. It just says: for the `files` field, get it via attribute rather than key access when dumping... but we need this kind of functionality for fields that are not `Nested` and `NestedAttribute` doesn't recursively make that the case for the fields of its nested schema (not that this would necessarily make sense anyway).

Implementing `get_attribute()` for this RDMRecordSchema and checking for the relevant attribute works. This is what we are doing in AccessSchema and in FilesSchema....

PR #556, opened 2021-04-26
providers: make datacite credentials optional

ppanero:
Make Datacite credentials optional while keeping the DOI PID configured as required. It will simply not query DataCite if not configured.

You can see the warnings coming form the demo records:
Screenshot 2021-04-26 at 13 03 16

An example record created with no credentials (An no prefix), shows a `None` in the tag.
Screenshot 2021-04-26 at 13 02 42

It will fail (throw an exception) but not interrupt the running course of the instance, because it needs the prefix to create the DOI.
Screenshot 2021-04-26 at 13 02 11

Options:
a) Leave it as is... the error is quite nice to tell the user "configure credentials and prefix or take out the dois"
b) Make the DOI creation aware of no prefix... should it then be "datacite.12345-abdce" without `/` first?


ntarocco (approved):

ntarocco:
LGTM!

PR #557, opened 2021-04-26
resources: add pid resolver endpoint

ppanero:
**Important notes**
- The service part will be refactor at the same time than the `reserve_pid` and `delete_pid` and potentially all PID related operations should be in a sub-service. And be able to do `service.pids.reserve(...)`.
- The resource level is being refactored by @ntarocco, but in principle this will stay as a separate resource since it has nothing to do with `records` (i.e. no `/records/` in the URL).
- There are two endpoint needed, one is specifically meant for DOIs which have a slash (`prefix/client.recid`) in the PID value.
- I **did not** add the `RDM` prefix to all the new classes because I find it unnecessary and making longer names with no extra value. But I can add it if required.

### DIFF
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -122,3 +122,26 @@ class RDMManagedPIDProviderResourceConfig(RecordResourceConfig):
     response_handlers = {
         "application/json": ResponseHandler(JSONSerializer())
     }
+
+
+class PIDResolverResourceConfig(RecordResourceConfig):
+    """PID provider resource configuration."""
+
+    blueprint_name = "record_pids_resolver"
+
+    url_prefix = "/pids/<pid_type>"
+
+    routes = {
+        "item-doi": "/<pid_prefix>/<pid_value>",
+        "item": "/<pid_value>"
+    }
+
+    request_view_args = {
+        "pid_prefix": ma.fields.Str(),
+        "pid_type": ma.fields.Str(),
+        "pid_value": ma.fields.Str(),
+    }
+
+    response_handlers = {
+        "application/json": ResponseHandler(JSONSerializer())
+    }
--- a/invenio_rdm_records/services/services.py
+++ b/invenio_rdm_records/services/services.py
@@ -15,7 +15,7 @@ import arrow
 from flask_babelex import lazy_gettext as _
 from invenio_db import db
 from invenio_drafts_resources.services.records import RecordService
-from invenio_files_rest.models import ObjectVersion
+from invenio_pidstore.models import PersistentIdentifier
 from invenio_records_resources.services.records.schema import \
     ServiceSchemaWrapper
 from marshmallow.exceptions import ValidationError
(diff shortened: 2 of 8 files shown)


invenio_rdm_records/services/services.py
```diff
@@ -405,6 +405,32 @@ def reserve_pid(self, id_, identity, pid_type, pid_client=None):
...
+    def resolve_pid(
+        self, id_, identity, pid_type, pid_prefix=None, pid_client=None
+    ):
+        """Resolve PID to a record."""
+        # PIDS-FIXME: should be migrated to self.record_cls.pids.resolve()
+        provider = self.get_provider(pid_type, client_name=pid_client)
+
+        pid_value = id_
+        if pid_prefix:
+            pid_value = f"{pid_prefix}/{pid_value}"
```
ntarocco, line 417: Suggestion: move this up to the view. Like this, the service can stay generic and not having to know if a prefix exists or not.
You can also then put in common the call to the service in the resource and refactoring to 2 views to all the same function.

invenio_rdm_records/services/services.py
```diff
@@ -405,6 +405,32 @@ def reserve_pid(self, id_, identity, pid_type, pid_client=None):
...
+
+        pid = provider.get(pid_value=pid_value, pid_type=pid_type)
+
+        # get related record/draft
+        record = self.record_cls.get_record(pid.object_uuid)
+
+        # permissions
+        self.require_permission(identity, "manage", record=record)
```
ntarocco, line 425:
```suggestion
        self.require_permission(identity, "read", record=record)
```

invenio_rdm_records/services/services.py
```diff
@@ -405,6 +405,32 @@ def reserve_pid(self, id_, identity, pid_type, pid_client=None):
...
+
+        return self.result_item(
```
ntarocco, line 427: Can we simply return the `link` only?
Ideally, it would be great to simply fetch the PID with `PersistentIdentifer.get` and generate the link, without fetching the record... not sure it is easy.

tests/resources/test_resources.py
```diff
@@ -759,3 +759,25 @@ def test_publish_pid_flow(
     assert response.status_code == 202
     assert response.json["pids"]["doi"]["identifier"]
     assert response.json["pids"]["doi"]["client"] == "datacite"  # default
+
+
+def test_resolve_pid_with_login(
```
ntarocco, line 764: what about adding a test for a 404?

invenio_rdm_records/services/services.py
ppanero: Discussed IRL... Not as it is right now to make it consistent.

tests/resources/test_resources.py
ppanero: Done

ntarocco (approved):

invenio_rdm_records/resources/config.py
```diff
@@ -122,3 +122,26 @@ class RDMManagedPIDProviderResourceConfig(RecordResourceConfig):
...
+class PIDResolverResourceConfig(RecordResourceConfig):
+    """PID provider resource configuration."""
+
+    blueprint_name = "record_pids_resolver"
+
+    url_prefix = "/pids/<pid_type>"
+
+    routes = {
+        "item-doi": "/<pid_prefix>/<pid_value>",
+        "item": "/<pid_value>"
```
lnielsen, line 136: You'll have the same issue in Invenio-App-RDM, so instead of using two endpoints, simply use one:

```
"item": "/<path:pid_value>"
```

invenio_rdm_records/resources/config.py
```diff
@@ -122,3 +122,26 @@ class RDMManagedPIDProviderResourceConfig(RecordResourceConfig):
...
         "application/json": ResponseHandler(JSONSerializer())
     }
+
+
+class PIDResolverResourceConfig(RecordResourceConfig):
+    """PID provider resource configuration."""
+
+    blueprint_name = "record_pids_resolver"
+
+    url_prefix = "/pids/<pid_type>"
```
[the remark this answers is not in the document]
ntarocco: I don't see any comment in #560, did you add any? Anyway, you are right, to be moved.

invenio_rdm_records/resources/config.py
ntarocco: We discussed about this, the issue is the `/` in the doi value, the idea was to avoid to try to split the value by `/` manually.

PR #560, opened 2021-04-26
Refactor PID reserve/discard

ntarocco:


ntarocco:
I still need to add tests!

invenio_rdm_records/services/pids/providers/base.py
```diff
@@ -47,16 +48,40 @@ def is_managed(self):
...
+        query = PersistentIdentifier.query.filter_by(
+            pid_type=pid_type or self.pid_type,
+            object_type=object_type,
+            object_uuid=record_id
+        )
+        if pid_value:
+            query.filter_by(pid_value=pid_value)
+
+        try:
+            return query.one()
```
ppanero, line 82: doubt: This will return the first, right? What happens in the case of "DOI/ConceptDOI"...

The record_id will be the same? (Or not because concept will be linked to the parent), then both will be `pid_type=doi` and `object_type=rec`... IIRC won't happen if `object_uuid=parent_id` for the ConceptDOI... but just to make sure.

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -92,12 +93,24 @@ def _generate_id(self, record, **kwargs):
...
+        except PIDDoesNotExistError:
+            # not existing, create a new one
+            return super().create(
+                pid_value=doi,
+                object_type="rec",
+                object_uuid=record.id,
+                **kwargs
+            )
+
+        # re-activate if previously deleted
```
ppanero, line 108: 🚀

invenio_rdm_records/services/services.py
```diff
@@ -293,31 +291,54 @@ def get_client(self, client_name):
 
         return client_class(name=client_name)
 
-    def get_provider(self, scheme, client_name=None):
+    def _get_managed_provider(self, providers_dict):
```
ppanero, line 294: minor: in both cases we are assuming a preference for the first configure as such (e.g. the first managed)... We chould state this in the docs.

invenio_rdm_records/services/services.py
```diff
@@ -340,18 +370,19 @@ def reserve_pid(self, id_, identity, pid_type, pid_client=None):
         provider.reserve(pid, draft)
         draft.commit()
 
+        db.session.commit()
```
ppanero, line 373: doubt: it's needed because the nested session in the reserve?

invenio_rdm_records/services/services.py
```diff
@@ -376,25 +407,36 @@ def delete_pid(self, id_, identity, pid_type, pid_client=None):
...
+        provider = self.get_provider(pid_type, provider_name=provider_name,
+                                     client_name=pid_client)
         pid_attr = draft.pids.get(pid_type)
 
         if not pid_attr:
             raise ValidationError(
-                message=_("Identifier value or client not given for PID " +
-                          f"type {pid_type}"),
+                message=_(f"No registered PID found for type {pid_type}"),
                 field_name="pids",
```
ppanero, line 424: minor: `pids.pid_type? or not because it might not be and therefore generic `pids` field

ppanero (approved):
LGTM, some comments here and there

invenio_rdm_records/resources/resources.py
```diff
@@ -13,15 +13,63 @@
...
+    def pids_discard(self):
+        """Discard a previously reserved PID."""
+        item = self.service.discard_pid(
+            id_=resource_requestctx.view_args["pid_value"],
+            pid_type=resource_requestctx.view_args["pid_type"],
+            pid_client=resource_requestctx.args.get("client"),
+            identity=g.identity,
+        )
+
+        return item.to_dict(), 204
```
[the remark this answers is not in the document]
ntarocco: Thanks! Already fixed, tests were failing! :)

invenio_rdm_records/services/components.py
```diff
@@ -144,12 +145,40 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
 class ExternalPIDsComponent(ServiceComponent):
     """Service component for pids."""
 
+    def _validate_pid(self, scheme, pid):
+        """Call provider to validate the given PID."""
+        provider_name = pid.get("provider")
+        client = pid.get("client")
+        provider = self.service.get_provider(scheme, provider_name, client)
```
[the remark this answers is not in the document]
ntarocco: Exactly what I thought: this version will loop 2 times, but more readable. The other version is to optimize code execution, but code becomes quite complex... I went for the first.

invenio_rdm_records/services/components.py
```diff
@@ -144,12 +145,40 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
...
+            if not is_enabled:
+                raise Exception(f"The provider {provider_name} is not enabled")
+
+            self._validate_pid(scheme, pid)
+            verified_pids.add(scheme)
+
+        # ensure that there are no extra PIDs for which there is no config
+        all_pids = set(verified_pids)
+        unknown_pis = all_pids - verified_pids
+        if unknown_pis:
```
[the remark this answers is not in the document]
ntarocco: 👍🏼

invenio_rdm_records/services/components.py
```diff
@@ -144,12 +145,40 @@ def new_version(self, identity, draft=None, record=None, **kwargs):
...
+            raise ValidationError(
+                message=_(f"Unknown PIDs {unknown_pis}"),
```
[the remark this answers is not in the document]
ntarocco: 👍🏼

invenio_rdm_records/services/pids/providers/base.py
ntarocco: Yes it should never happen because of the `record_id`. I added this check to make sure that we discard the right pid for the right record, because before it was getting only by pid_type/pid_value and we might have multiple, e.g. concept.

invenio_rdm_records/services/services.py
ntarocco: yes, you are right. Here we should have max 1 managed and max 1 unmanaged.
Maybe the config should enforce this, or we document it when we document when to override.
We cannot have 2 managed for sure...

invenio_rdm_records/services/services.py
ntarocco: without this, there is no `db.session.commit` happening anywhere, since it is a specific new endpoint. Or were you thinking to something else?

invenio_rdm_records/services/services.py
ntarocco: I would keep this check before getting the `pid` to make sure everything is correct.
The pid `.get` will fail anyway if not found.

ntarocco: You are right, fixing it.

invenio_rdm_records/services/components.py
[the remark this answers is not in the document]
ppanero: agree :) shouldn't be maany providers anyway

invenio_rdm_records/services/services.py
ppanero: discussed IRL. `draft.commit` adds to the session but does not commit itself....

ppanero (approved):

PR #567, opened 2021-04-27
Givename

Pineirin:
make only family name mandatory when adding a creator

### DIFF
--- a/tests/resources/serializers/test_ui_serializer.py
+++ b/tests/resources/serializers/test_ui_serializer.py
@@ -67,7 +67,6 @@ def test_ui_serializer(app, full_record_ui):
     with app.app_context():
         serialized_record = UIJSONSerializer().serialize_object_to_dict(
             full_record_ui)
-
     assert serialized_record['ui'] == expected_data
     serialized_records = UIJSONSerializer().serialize_object_list(
         {"hits": {"hits": [full_record_ui]}})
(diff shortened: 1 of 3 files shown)


tests/resources/serializers/test_ui_serializer.py
```diff
@@ -67,7 +67,6 @@ def test_ui_serializer(app, full_record_ui):
     with app.app_context():
         serialized_record = UIJSONSerializer().serialize_object_to_dict(
             full_record_ui)
-
```
ppanero, line 70: minor: not a problem, but try not to change previous things to not change much the git history.

tests/services/schemas/test_creator_contributor.py
```diff
@@ -94,20 +107,6 @@ def test_creator_organization_valid_full():
 
 
 def test_creatibutor_name_edge_cases():
-    # Pass in name and given_name: name is ignored
```
ppanero, line 97: moderate: does the logic of names change when removing the requirement of `given name`? I cannot find in the issue description. If not, this test should not be removed, it should pass :) 

The logic for this is present at https://github.com/inveniosoftware/invenio-rdm-records/pull/567/files#diff-a0d8874bbf9ded2c080996a386a79aa7cc118b7054da68b2281ed5153830b419L99

lnielsen: The test should indeed not be remove, just changed a bit. ``given_name`` should be changed to ``family_name`` in both valid and expected cases, and then it should pass.

Issue #573, opened 2021-04-28
config: support enable/disable of datacite provider


ppanero:
Coming from two cards in the board:

- DATACITE_ENABLE = TRUE flag enabled by default, but disable in Cookiecutter
- add the feature flag for DOI in ext config

PR #574, opened 2021-04-28
pids: add flag to disable DOI

ntarocco:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/573


ntarocco:
@ppanero there is one test failing because of:
`{'status': 400, 'message': 'A validation error occurred.', 'errors': [{'field': 'pids._schema', 'messages': ['Invalid scheme doi.']}]}`

tests/conftest.py
```diff
@@ -48,9 +48,6 @@ def app_config(app_config):
         'PIDSTORE_RECID_FIELD',
         'RECORDS_PERMISSIONS_RECORD_POLICY',
         'RECORDS_REST_ENDPOINTS',
-        'RDM_DATACITE_CLIENT_USERNAME',
```
ppanero, line 51: No need to add the new ones since they are prefixed with `RDM_RECORDS`.

invenio_rdm_records/services/services.py
```diff
@@ -338,8 +338,8 @@ def get_provider(self, scheme, provider_name=None, client_name=None):
                 return provider_class(client)
             elif provider_config["system_managed"]:
                 # use as default the client configured for the provider
-                client_name = provider_class.name
-                client = self.get_client(client_name)
+                provider_name = provider_class.name
```
ppanero, line 341: minor: it was `client_name` because the provider name was the default client name... but I think is more understandable as you  changed it now :)

invenio_rdm_records/config.py
```diff
@@ -174,6 +171,8 @@ def _(x):
 THEME_SITEURL = "http://127.0.0.1:5000"
 
 #: DataCite DOI credentials
-# RDM_DATACITE_CLIENT_USERNAME = ""
-# RDM_DATACITE_CLIENT_PASSWORD = ""
-# RDM_DATACITE_CLIENT_PREFIX = ""
+RDM_RECORDS_DOI_DATACITE_ENABLED = True
+RDM_RECORDS_DOI_DATACITE_USERNAME = ""
+RDM_RECORDS_DOI_DATACITE_PASSWORD = ""
+RDM_RECORDS_DOI_DATACITE_PREFIX = "10.1234"
```
ppanero, line 177: comment: this is the old test prefix, it will not enable datacite credentials since user and psw are still missing, but is more friendly to have this...

ppanero (approved):
Waiting for the green from the tests

PR #578, opened 2021-04-29
pids: disable reservation and saving of invalid values

### LINKED ISSUE #575: pids: saved invalid identifier is not removable
if you enter an invalid identifier you cannot get it out of the system again. Meaning that any consecutive saves will fail.

### LINKED ISSUE #576: pids: avoid storing pids due to soft deletion
When a PID is reserved (e.g. GET DOI button) the pid gets updated to `PIDStatus.RESERVED`. However, this means that *discarding* it will perform a soft deletion. We want to avoid this case by keeping the status to `NEW` and simply jump to `REGISTERED` when publishing.

Make the datacite provider `reserve` method dumb, by returning constant `True`.

ppanero:
closes #576 
closes #575 

This PR tackles two issues:

**Disable local reservation of DOIs**
Do not reserve locally a DOI, to avoid having many soft deleted ones. I.e. a user reserves and discards, if we set status to `RESERVED` then will be soft deleted. With this change it will pass from `NEW` to `REGISTERED`.

**Bug fix: do not store invalid values for pids**
Avoid saving invalid identifiers. If the validation step notes any scheme as having an invalid value, this one will be stripped of the `pids` so they are not stored in db.

This issue has an extra catch.
- There can be several schemes with errors, so at `__init__.py.Schema:validate_pids` level we need to catch and re-raise at the end with *several* messages.
- This makes us not be able to assign a specific field. i.e. it will always be `field_name=pids` and never `field_name=pids.doi`. This is also not possible due to Marshmallo core. The [`error_store`](https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/schema.py#L490) would overwrite the `field_name`.
So even if we re-raise something like:

```
raise ValidationError(
    message=_("Invalid value."),
    field_name=f"pids.{scheme}"
)
```

It will be re-set to `pids` in `error_store.store_error(error.messages, field_name, index=index)`, because the schema field that failed is `pids` (since all the inner ones are dynamic).

Errors are properly returned in the API:

Screenshot 2021-04-29 at 14 20 44

### DIFF
--- a/invenio_rdm_records/services/components.py
+++ b/invenio_rdm_records/services/components.py
@@ -9,11 +9,12 @@
 
 """RDM service components."""
 
+import re
 from copy import copy
 
 from flask_babelex import lazy_gettext as _
 from invenio_access.permissions import system_process
-from invenio_records_resources.services.records.components import \
+from invenio_drafts_resources.services.records.components import \
     ServiceComponent
 from marshmallow import ValidationError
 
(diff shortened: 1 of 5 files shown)


invenio_rdm_records/services/components.py
```diff
@@ -174,16 +175,28 @@ def _validate_pids(self, pids):
...
+        """Remove pids that have validation errors."""
+        errors = errors or []
+        for error in errors:
+            if "pids._schema" == error["field"]:
+                for message in error["messages"]:
+                    # assume format "[some text] scheme {scheme}"
+                    pid_type = message.split("scheme")[1].strip()
+                    pids.pop(pid_type, None)
+
+    def create(self, identity, data=None, record=None,  errors=None):
```
ntarocco, line 188: Did the service interface change or we did a mistake in copying it? I don't recall this `errors` param :(

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -113,14 +113,16 @@ def create(self, record, **kwargs):
...
+        It does not reserve locally, nor externally. This is to avoid storing
+        many PIDs as cause of reserve/discard, which would then be soft
+        deleted. Therefore we want to pass from status.NEW to status.RESERVED.
         :param pid: the PID to reserve.
         :param record: the record.
-        :returns: `True` if is reserved successfully.
+        :returns: `True`
         """
-        return super().reserve(pid, record)
+        return True
```
ntarocco, line 125: I am sure there was a discussion for this, I just find it quite odd :)

Issue #539, opened 2021-04-21
serializers: datacite serialization of publication year


ppanero:
The RDM datamodel contains a `publication_date` which is handled by a `marshmallow_utils.EDTFDateString` field. This can also handle ranges... Datacite serialziation requires a `publicationYear`. 

- [x] How to obtain the year from the given string
- [x] Can the `EDTFDateString` be reused and some how made accessible the `year`?

tmorrell:
`publicationYear` is not a well-loved field in DataCite, since the actual data fields are much more descriptive. It's used for citation purposes, but DataCite will prioritize the date field if they are present. I would keep things simple and just strip out the year, or the first year in a range.

ppanero:
So it will always favor the `Issued` date... Can actually "publication dates" have ranges?

tmorrell:
It's a real uncommon edge case. The only scenario I can imagine is if you're putting a whole bunch of journal articles into a single record. 

I've done some testing, and with the Rest API DataCite actually accepts a range for `publicationYear` but only saves the first year. I think we should follow this implementation and only send the first year. It could be as simple as just returning the first 4 numbers in the EDTFD date.

lnielsen:
Agree with Tom, simply usage first possible year of publication

PR #582, opened 2021-05-03
resources: datacite serialization for publicationYear

ppanero:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/539

@fenekku I recall at some point we had an issue with dates serialization and timezones... i.e. a date was serialized here to 2020 and there to 2019 because of diff... I tried to reproduce but didn't manage... I'm just trying to make sure it does not happen with the edtf parsing since the returned `time.struct` has a `tm_zone`...


tests/services/test_serializers.py
```diff
@@ -1,91 +0,0 @@
-# -*- coding: utf-8 -*-
```
ppanero, line 1: removed because:
1- unused... all skiped
2- serializers are now at resource level, this will be added when extensions are revisited?

fenekku (approved):
I ran the tests for this locally and didn't get any timezone issue (at least at the time of testing :wink:  ), so am going to assume this is all good :+1: .

PR #583, opened 2021-05-03
pids: enhance error logging of datacite provider

### LINKED ISSUE #581: pids: datacite provider errors upon publishing
Package version (if known): v0.29.16

## Describe the bug
Publishing a record with a DataCite DOI logs a warning in the console:

```
WARNING:root:DataCite provider errored when updating DOI for 10.1234/datacite.12345-abcde
```

## Steps to Reproduce
1. Enable DataCite DOI as per [documentation](https://inveniordm.docs.cern.ch/customize/dois/).
2. Create a draft with a DOI and publish it
3. See the terminal/logs

## Expected behavior

Registration to succeed.

## Additional context

There are several errors:

1- The code is sending the full record, while it should be sending only the serialized `metadata`

```
doc = DataCite43JSONSerializer().dump_one(record)
self.api_client.public_doi(metadata=doc, url=url, doi=pid.pid_value)
```

Should send `metdata=doc["metadata"]` or the serializer should return only that part. I would go for the former, since the full record might be useful for other cases (And is consistent with other serializers like the UIJSON one).

2- The resource type is not serialized because is a vocabulary and there are issues there... see https://github.com/inveniosoftware/invenio-rdm-records/issues/529

3- The publication year is not being added to the serialization https://github.com/inveniosoftware/invenio-rdm-records/issues/539

Note the last two are fixed in their specific issues... This issue only needs point 1 to be closed.

ppanero:
closes #581 

Fixes two things:

- Sends only the `metadata` to DataCite
- In case of error, logs the specific error returned by DataCite in order to allow the user to know which field errored e.g.
```
WARNING:root:DataCite provider errored when updating DOI for 10.81088/datacite.bcv6s-a7034
WARNING:root:Error in publication_year: DOI 10.81088/datacite.bcv6s-a7034: [facet 'pattern'] The value '' is not accepted by the pattern '[\d]{4}'. at line 15, column 0
```

### DIFF
--- a/invenio_rdm_records/services/pids/providers/datacite.py
+++ b/invenio_rdm_records/services/pids/providers/datacite.py
@@ -7,6 +7,7 @@
 
 """DataCite DOI Provider."""
 
+import json
 import logging
 
 from datacite import DataCiteRESTClient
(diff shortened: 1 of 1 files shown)


fenekku (approved):
Nice catch on the metadata vs full record passing.

Issue #585, opened 2021-05-03
[doi][april release fix] serialize resource_type from old vocabulary


fenekku:
ResourceType: It is already a vocabulary, but uses the old implementation: obtain the datacite values from csv.

This is just for April release - longer term we want invenio-vocabularies to have a solution for this.

PR #586, opened 2021-05-03
doi: serialize resource type for Datacite

fenekku:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/585

This along with the previous fixes allowed me to register a test DOI successfully!

In Datacite test fabrica (note we use our own resource types):

serialized_doi

### DIFF
--- a/invenio_rdm_records/resources/serializers/datacite/schema.py
+++ b/invenio_rdm_records/resources/serializers/datacite/schema.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2021 CERN.
+# Copyright (C) 2021 Northwestern University.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
@@ -11,6 +12,8 @@ from edtf import parse_edtf
 from edtf.parser.grammar import ParseException
 from marshmallow import Schema, ValidationError, fields, missing, post_dump
 from marshmallow_utils.fields import SanitizedUnicode
+
+from invenio_rdm_records.vocabularies import Vocabularies
 
 
 class PersonOrOrgSchema43(Schema):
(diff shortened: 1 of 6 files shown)


tests/services/schemas/test_resource_type.py
```diff
@@ -8,12 +8,9 @@
 
 """Test metadata resource type schema."""
 
-import os
 from copy import deepcopy
 
-import pytest
 from flask_babelex import lazy_gettext as _
-from marshmallow import ValidationError
```
ppanero, line 16: 🔥

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -177,10 +180,11 @@ class DataCite43Schema(Schema):
     def get_type(self, obj):
         """Get resource type."""
         resource_type = obj["metadata"]["resource_type"]
-
+        vocabulary = Vocabularies.get_vocabulary('resource_type')
+        entry = vocabulary.get_entry_by_dict(resource_type)
         return {
-            'resourceTypeGeneral': "FIXME",
-            'resourceType': "FIXME",
+            'resourceTypeGeneral': entry.get("datacite_general", "Other"),
```
ppanero, line 186: minor: do we want to default to `Other`? Or fail if not found? To avoid pushing "wrong" data or from a not properly configured instance (e.g. *you just migrated and misconfigured, new records will be always other*)

ppanero (approved):
LGTM! Thanks for fixing this!

invenio_rdm_records/resources/serializers/datacite/schema.py
lnielsen: Let's go with ``Other`` to ensure it's registered. Otherwise we have to do a lot of error handling and have ways to fix it easily, and we don't,.

lnielsen (approved):

Issue #580, opened 2021-04-29
pids: publish should not accept duplicated values


ppanero:
The `ExternalPIDsComponent`, when publishing a `pid` with given value, tries to fetch the `pid` with `provider.get` which will find any PID that is registered in the system.
The issue is that it will not get filtering by the current record, so it may fetch the one of another record. It will then add it in the record even if it was already registered to another record.

* Use the `get_by_record` method instead. Ideally, can we change the `provider.get` method to always require the record UUID as param?
* Question: should this dedup also happen at update/create hook? If so, might be easier to move it to the `_validate..` function.

PR #592, opened 2021-05-05
components, providers: implement PID duplicates validation

ppanero:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/580
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/847

### DIFF
--- a/invenio_rdm_records/services/components/pids.py
+++ b/invenio_rdm_records/services/components/pids.py
@@ -20,31 +20,31 @@ from marshmallow import ValidationError
 class ExternalPIDsComponent(ServiceComponent):
     """Service component for pids."""
 
-    def _validate_pid(self, scheme, pid):
+    def _validate_pid(self, scheme, pid, record, provider=None):
         """Call provider to validate the given PID."""
-        provider_name = pid.get("provider")
-        client = pid.get("client")
-        provider = self.service.get_provider(scheme, provider_name, client)
-        if provider:
-            success, errors = provider.validate(**pid)
-            if errors:
-                raise ValidationError(message=errors, field_name="pids")
+        if not provider:  # In case we do not need to calculate it again
+            provider_name = pid.get("provider")
+            client = pid.get("client")
+            provider = self.service.get_provider(scheme, provider_name, client)
 
-    def _validate_pids(self, pids):
+        # NOTE: provider should not be None by now, if not configured should
+        # fail in `_validate_pid_schemes`
+        success, errors = provider.validate(record=record, **pid)
+        if errors:
+            raise ValidationError(message=errors, field_name="pids")
+
+    def _validate_pids(self, pids, record):
         """Validate an iterator of PIDs."""
-        verified_pids = set()
         pids_providers = self.service.config.pids_providers
         for scheme, providers in pids_providers.items():
-            if scheme not in pids:
-                continue
-
             pid = pids.get(scheme, {})
-            self._validate_pid(scheme, pid)
-            verified_pids.add(scheme)
+            self._validate_pid(scheme, pid, record)
 
-        # ensure that there are no extra PIDs for which there is no config
+    def _validate_pid_schemes(self, pids, record):
+        """Validate the pid schemes are supported by the service."""
+        pids_providers = set(self.service.config.pids_providers.keys())
         all_pids = set(pids.keys())
-        unknown_pids = all_pids - verified_pids
+        unknown_pids = all_pids - pids_providers
         if unknown_pids:
             raise ValidationError(
                 message=_(f"No configuration defined for PIDs {unknown_pids}"),
@@ -65,7 +65,8 @@ class ExternalPIDsComponent(ServiceComponent):
         """Inject parsed pids to the draft record."""
         pids = data.get('pids', {})
         self._remove_invalid_pids(pids, errors)
-        self._validate_pids(pids)
+        self._validate_pid_schemes(pids)
+        self._validate_pids(pids, record)
         # NOTE: record is a draft because we hook to the draft service.
         record.pids = pids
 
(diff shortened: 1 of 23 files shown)


tests/services/components/test_external_pids_component.py
```diff
@@ -182,29 +188,6 @@ def test_unmanaged_no_required_no_partial_value(
         component.publish(identity_simple, draft=draft, record=record)
 
 
-def test_unmanaged_no_required_no_value(
```
ppanero, line 185: Removed because is not a valid use case. If it's not required, there is no "default" provider... only required ones are checked to *must have one required provider at least* (i.e. the default)

lnielsen:
Can I ask you to rebase on master. Then afterwards I'd like to take it for a spin in my-site to understand what's going on from a user point of view.

ppanero:
@lnielsen done :)

invenio_rdm_records/services/pids/providers/base.py
```diff
@@ -152,17 +158,30 @@ def get_status(self, identifier, **kwargs):
...
         """
+        errors = []
         if provider and provider != self.name:
-            errors = [
+            errors.append(
                 _(f"Provider name {provider} does not match {self.name}")
-            ]
-            return False, errors
+            )
+        # deduplication check
```
ntarocco, line 175: Does it make sense to continue checking here? I think it should simply return here...

ntarocco:
@ppanero can you please check the last commit in this PR?

We will need to discuss tomorrow a couple of things:
1. the value to put in `pid_provider` column in the DB. It accepts max `8` chars, and the current `unmanaged` has `9` chars. Temporarly changed to `external`
2. I have realized that I have made a mistake and harcoded `unmanaged` in the React app [here](https://github.com/inveniosoftware/react-invenio-deposit/blob/HEAD/src/lib/components/Identifiers/PIDField.js#L17) :'( . After changing that, I can fully test these changes...

ppanero:
LGTM, is just a pitty to have everything named unmanaged and having to put `external` just because of the db constraint hehe

invenio_rdm_records/services/components/pids.py
```diff
@@ -9,7 +9,7 @@
 
 """RDM service component for pids."""
 
-import re
+import logging
```
ppanero, line 12: This is probably my fault 😅 but I think we should use `current_app.logger` instead, that way it can go to e.g. Sentry

invenio_rdm_records/services/components/pids.py
```diff
@@ -20,42 +20,48 @@
...
+            client = pid.get("client")
+            provider = self.service.get_provider(scheme, provider_name, client)
+            if not provider:
+                raise ValidationError(
+                    message=_(f"Provider {provider_name} not found for PID " +
+                              f"type {scheme}"),
+                    field_name="pids",
+                )
+
+        # NOTE: provider should not be None by now, if not configured should
```
ppanero, line 36: mine again 😅 we shouldn't use `NOTE` since this has meaning only for translations 👼

invenio_rdm_records/services/components/pids.py
```diff
@@ -20,42 +20,48 @@
...
-            verified_pids.add(scheme)
+            pid = pids.get(scheme)
+            if pid:
+                self._validate_pid(scheme, pid, record)
 
-        # ensure that there are no extra PIDs for which there is no config
+    def _validate_pid_schemes(self, pids):
+        """Validate the pid schemes are supported by the service."""
+        pids_providers = set(self.service.config.pids_providers.keys())
         all_pids = set(pids.keys())
```
ppanero, line 53: using the `pids.keys()` will give you the type e.g. `doi` that should match the `scheme`.... we are assuming this is a 1-1 match right? because of how the config is done... This might make it fail in many cases... (see tests)

invenio_rdm_records/services/components/pids.py
```diff
@@ -65,19 +71,21 @@ def create(self, identity, data=None, record=None,  errors=None):
         """Inject parsed pids to the draft record."""
         pids = data.get('pids', {})
         self._remove_invalid_pids(pids, errors)
-        self._validate_pids(pids)
+        self._validate_pid_schemes(pids)
+        self._validate_pids(pids, record)
         # NOTE: record is a draft because we hook to the draft service.
```
ppanero, line 76:
```suggestion
        # record is a draft because we hook to the draft service.
```

invenio_rdm_records/services/pids/providers/base.py
```diff
@@ -152,17 +187,36 @@ def get_status(self, identifier, **kwargs):
...
                   array of error messages.
         """
         if provider and provider != self.name:
-            errors = [
-                _(f"Provider name {provider} does not match {self.name}")
-            ]
-            return False, errors
+            logging.error(f"Configuration error: provider name {provider} "
+                          f"does not match {self.name}.")
+            raise   # configuration error
```
ppanero, line 202: is a config error but it will raise at run time when validating a record pids... should we raise with some message?

ppanero (commented):
LGTM, the main change to take into account is enforcing the provider name to be the pid-type on the config dict... (I am doubting if we should *not* do so).
- remove `NOTE`s
- change logger
- Fix tests and can be merged :)

I can help with the first two since they come from my doing 😅

invenio_rdm_records/services/pids/providers/base.py
ntarocco: Here what is happening is that if we raise with a specific message, it might be returned in the API response.
For `400`, it is indeed what we want.

For `500`, what you want to have is to log the error to find it, but to return simply 500 Internal Server error. You don't want, normally, to give specific info on the error, it might a vector of attack e.g. in case of misconfiguration. At least what I personally think :)

invenio_rdm_records/services/components/pids.py
ntarocco: Yes, you are right. Honestly we probably have to rework the config at some point because it can lead to many inconsistencies.
Are you suggesting something in particular?

ppanero: I am out of ideas for this one... But I fully agree it need rewrite at config level to make a structure that actually works...

ppanero (commented):
I cannot approve because I am the author of the PR and GH does not allow me 😅 but it LGTM @ntarocco

PR #593, opened 2021-05-06
relations: language fields tests

lnielsen:
closes inveniosoftware/invenio-app-rdm#833


invenio_rdm_records/records/jsonschemas/records/record-v3.0.0.json
```diff
@@ -13,17 +13,16 @@
...
-      "description": "Persistent concept record identifier (alphanumeric).",
-      "type": "string"
-    },
-    "conceptpid": {"$ref": "local://records/definitions-v1.0.0.json#/internal-pid"},
+    "pid": {"$ref": "local://records/definitions-v1.1.0.json#/internal-pid"},
 
     "pids": {
-      "additionalProperties": {"$ref": "local://records/definitions-v1.0.0.json#/external-pid"},
-      "description": "Managed persistent identifiers for a record including e.g. OAI-PMH identifier, minted DOIs and more. Managed PIDs are registered in the PIDStore"
+      "type": "object",
```
lnielsen, line 19: ``pids`` actually weren't validated properly before ...... 1) tests were run with v1.0.0 schema instead of v2.0.0 schema, and the v2.0.0 schema had a bug

invenio_rdm_records/records/jsonschemas/records/record-v3.0.0.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "id": "local://records/record-v1.0.0.json",
-  "title": "InvenioRDM Record Schema v1.0.0",
+  "id": "local://records/record-v2.0.0.json",
```
lnielsen, line 3: I'm removing v1.0.0 schema since we don't allow upgrading from it anymore.

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -11,23 +11,31 @@
...
+from marshmallow_utils.fields import BabelGettextDictField
 from marshmallow_utils.fields import FormatDate as FormatDate_
 from marshmallow_utils.fields import FormatEDTF as FormatEDTF_
 from marshmallow_utils.fields import StrippedHTML
 
-from invenio_rdm_records.vocabularies import Vocabularies
+from .fields import AccessStatusField, VocabularyTitleField
+
+
+def current_default_locale():
```
lnielsen, line 27: Probably should go somewhere more meaningful (also used in vocabularies).

invenio_rdm_records/services/config.py
```diff
@@ -41,66 +41,52 @@
...
-            'access_status': terms_filter('access.status'),
-            # 'languages': terms_filter('metadata.languages.id'),
-        }
-    )
+    facets = {
+        'resource_type': NestedTermsFacet(
+            field='metadata.resource_type.type',
+            subfield='metadata.resource_type.subtype',
+            label=_("Resource types"),
+            # value_labels=VocabularyLabels('resource_types')
```
lnielsen, line 49: @ppanero This is the class you have to implement.

fenekku: Should I be touching this too when I tackle resource types?

tests/records/test_relations_languages.py
```diff
@@ -0,0 +1,94 @@
...
+
+def test_languages_validation(app, location, minimal_record, lang):
+    """Test languages relationship."""
+    minimal_record["metadata"]["languages"] = [{"id": "eng"}]
+    draft = RDMDraft.create(minimal_record)
+    draft.commit()
+    db.session.commit()
+    assert minimal_record["metadata"]["languages"] == [{"id": "eng"}]
+
+    lang = list(draft.relations.languages())[0]
```
fenekku, line 39: Hmm I thought the descriptors and so on were to make `draft.relations.languages[0]` possible...  And this index is a `Vocabulary` (which should be read as a Vocabulary entry) right?

fenekku (approved):

tests/records/test_relations_languages.py
```diff
@@ -0,0 +1,94 @@
...
+# Tests
+#
+def test_languages_field(app, location, minimal_record):
+    """Languages should be defined as a relation."""
+    assert 'languages' in RDMDraft.relations
+    assert 'languages' in RDMRecord.relations
+    assert RDMDraft.relations.languages
+
+
+def test_languages_validation(app, location, minimal_record, lang):
```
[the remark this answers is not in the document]
ppanero: I think you are right, there is a `db` fixture for that... flushing should not affect the tests, but better safe than sorry.

tests/records/test_relations_languages.py
ppanero: I think is an iterator, thats why the cast to `list`... otherwise needs the `next()`?

Issue #530, opened 2021-04-21
services: factor out secret links as a subservice


lnielsen:
RDMRecordService, has many methods for the secret links, pids, etc. These should be separated out into subservices.

ines-cruz:
i dont know how to link the second PR here  https://github.com/inveniosoftware/invenio-drafts-resources/pull/148

PR #595, opened 2021-05-11
Secret Link as a subservice

ines-cruz:
closes #https://github.com/inveniosoftware/invenio-rdm-records/issues/530

### DIFF
--- a/invenio_rdm_records/ext.py
+++ b/invenio_rdm_records/ext.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2019 CERN.
+# Copyright (C) 2019-2021 CERN.
 # Copyright (C) 2019 Northwestern University.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
@@ -22,7 +22,7 @@ from .resources import RDMDraftFilesResourceConfig, \
     RDMRecordFilesResourceConfig, RDMRecordResource, RDMRecordResourceConfig
 from .secret_links import LinkNeed, SecretLink
 from .services import RDMFileDraftServiceConfig, RDMFileRecordServiceConfig, \
-    RDMRecordService, RDMRecordServiceConfig
+    RDMRecordService, RDMRecordServiceConfig, SecretLinkService
 from .services.schemas.metadata_extensions import MetadataExtensions
 
 
@@ -106,8 +106,8 @@ class InvenioRDMRecords(object):
             self._filter_record_service_config(app, RDMRecordServiceConfig),
             files_service=FileService(RDMFileRecordServiceConfig),
             draft_files_service=FileService(RDMFileDraftServiceConfig),
+            secret_links_service=SecretLinkService(RDMRecordServiceConfig)
         )
-
         self.subjects_service = subject_record_type.service_cls(
             config=subject_record_type.service_config_cls,
         )
--- a/invenio_rdm_records/resources/resources.py
+++ b/invenio_rdm_records/resources/resources.py
@@ -98,7 +98,7 @@ class RDMParentRecordLinksResource(RecordResource):
     @response_handler()
     def create(self):
         """Create a secret link for a record."""
-        item = self.service.create_secret_link(
+        item = self.service.secret_links.create(
             id_=resource_requestctx.view_args["pid_value"],
             identity=g.identity,
             data=resource_requestctx.data,
@@ -110,7 +110,7 @@ class RDMParentRecordLinksResource(RecordResource):
     @response_handler()
     def read(self):
         """Read a secret link for a record."""
-        item = self.service.read_secret_link(
+        item = self.service.secret_links.read(
             id_=resource_requestctx.view_args["pid_value"],
             identity=g.identity,
             link_id=resource_requestctx.view_args["link_id"],
@@ -126,7 +126,7 @@ class RDMParentRecordLinksResource(RecordResource):
     @response_handler()
     def partial_update(self):
         """Patch a secret link for a record."""
-        item = self.service.update_secret_link(
+        item = self.service.secret_links.update(
             id_=resource_requestctx.view_args["pid_value"],
             identity=g.identity,
             link_id=resource_requestctx.view_args["link_id"],
@@ -137,7 +137,7 @@ class RDMParentRecordLinksResource(RecordResource):
     @request_view_args
     def delete(self):
         """Delete a a secret link for a record."""
-        self.service.delete_secret_link(
+        self.service.secret_links.delete(
             id_=resource_requestctx.view_args["pid_value"],
             identity=g.identity,
             link_id=resource_requestctx.view_args["link_id"],
@@ -149,7 +149,7 @@ class RDMParentRecordLinksResource(RecordResource):
     @response_handler(many=True)
     def search(self):
         """List secret links for a record."""
-        items = self.service.read_secret_links(
+        items = self.service.secret_links.read_all(
             id_=resource_requestctx.view_args["pid_value"],
             identity=g.identity,
         )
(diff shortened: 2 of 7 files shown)


invenio_rdm_records/ext.py
```diff
@@ -22,8 +22,9 @@
     RDMRecordFilesResourceConfig, RDMRecordResource, RDMRecordResourceConfig
 from .secret_links import LinkNeed, SecretLink
 from .services import RDMFileDraftServiceConfig, RDMFileRecordServiceConfig, \
-    RDMRecordService, RDMRecordServiceConfig
+    RDMRecordService, RDMRecordServiceConfig, SecretLinkServiceConfig
 from .services.schemas.metadata_extensions import MetadataExtensions
+from .services.secret_links.service import SecretLinkService
```
ppanero, line 27: When refactoring one of the main points is to try not to have to change imports. This is to avoid "breaking compatibilty". In this case you can avoid having this import (and leaving it as it was) by "exporting" the class in the `__init__.py`, note that you can do that successively (until you export it in `.services.__init__.py`).

invenio_rdm_records/ext.py
```diff
@@ -106,8 +107,8 @@ def init_services(self, app):
             self._filter_record_service_config(app, RDMRecordServiceConfig),
             files_service=FileService(RDMFileRecordServiceConfig),
             draft_files_service=FileService(RDMFileDraftServiceConfig),
+            secret_links_service=SecretLinkService(SecretLinkServiceConfig)
```
ppanero, line 110: 💯

invenio_rdm_records/services/__init__.py
```diff
@@ -8,7 +8,7 @@
 """High-level API for wokring with RDM records, files, pids and search."""
 
 from .config import RDMFileDraftServiceConfig, RDMFileRecordServiceConfig, \
-    RDMRecordServiceConfig
+    RDMRecordServiceConfig, SecretLinkServiceConfig
```
ppanero, line 11: For my comment on the `ext.py`... you are actually exporting it, so there is no need to change the import (in `ext.py`)

invenio_rdm_records/services/config.py
```diff
@@ -225,3 +225,50 @@ class RDMFileDraftServiceConfig(FileServiceConfig):
         "content": FileLink("{+api}/records/{id}/draft/files/{key}/content"),
         "commit": FileLink("{+api}/records/{id}/draft/files/{key}/commit"),
     }
+
+
+class SecretLinkServiceConfig(RecordServiceConfig):
```
ppanero, line 230: moderate: The config is 100% the same than the `RecordServiceConfig`? If so there is no need to create it... You can instantiate the subservice by passing the `RecordServiceConfig`.

invenio_rdm_records/services/secret_links/service.py
```diff
@@ -0,0 +1,298 @@
...
+import arrow
+from flask_babelex import lazy_gettext as _
+from invenio_db import db
+from invenio_drafts_resources.services.records import RecordService
+from invenio_records_resources.services.records.schema import \
+    ServiceSchemaWrapper
+from marshmallow.exceptions import ValidationError
+
+
+class SecretLinkService(RecordService):
```
ppanero, line 21: inheritance/design: `SecretLinkService` inherits from `RecordsService` which after a few parental relations end up in [here](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/service.py#L23). 

Before, the functions were named `{create, read, etc.}_secret_link` to avoid overwriting a record one. But now that it is a subservice... if they follow the same interface (they should) they can be renamed to `read`, `create` etc.

invenio_rdm_records/services/secret_links/service.py
```diff
@@ -0,0 +1,298 @@
...
+    """RDM Secret Link service."""
+
+    def __init__(self, config, secret_links_service=None):
+        """Constructor for Secret Link Service."""
+        super().__init__(config)
+        self._secret_links = secret_links_service
```
ppanero, line 27: major: I think there is a conceptual error...  The *secret_links* service should be "independent" like e.g. [`FileService`](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/files/service.py#L22) and inherit from `Service`. Then, it would be `RDMRecordsService` who would have the `secret_links` attribute and property (i.e. `RDMRecordsService` would do something similar to [`RecordsService`](https://github.com/inveniosoftware/invenio-drafts-resources/blob/master/invenio_drafts_resources/services/records/service.py#L31)...

All this said, maybe I'm missing a reason why it has to be like these 😅

ppanero (commented):
Looks geeting in shape :)

TL;DR:
- Try to see if it can be done without changing imports
- Update copyright to YYYY-2021 to the files you touch if they did not have 2021 yet (e.g. 2020-2021)
- major: service inheritance instantiation might not be correct... crosscheck against draft-resources/records-resources links added in comments.

invenio_rdm_records/ext.py
```diff
@@ -24,6 +24,7 @@
 from .services import RDMFileDraftServiceConfig, RDMFileRecordServiceConfig, \
     RDMRecordService, RDMRecordServiceConfig
 from .services.schemas.metadata_extensions import MetadataExtensions
+from .services.secret_links import SecretLinkService
```
ppanero, line 27: moderate: it should be possible to import from "services", by exposing it in the corresponding `__init__.py`.

```suggestion
from .services import RDMFileDraftServiceConfig, RDMFileRecordServiceConfig, \
    RDMRecordService, RDMRecordServiceConfig, SecretLinkService
from .services.schemas.metadata_extensions import MetadataExtensions
```

invenio_rdm_records/resources/resources.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020-2021 CERN.
+# Copyright (C) 2021 CERN.
```
ppanero, line 3: major: should be a range, we should only add... never remove
```suggestion
# Copyright (C) 2020-2021 CERN.
```

invenio_rdm_records/resources/resources.py
```diff
@@ -149,7 +149,7 @@ def delete(self):
     @response_handler(many=True)
     def search(self):
```
ppanero, line 150: minor: I'm doubting if we should actually call this `read_all`... because in the end we are not passing any "search query"

invenio_rdm_records/services/secret_links/service.py
```diff
@@ -0,0 +1,287 @@
...
+        # Fetching
+        links = parent.access.links.resolve_all()
+        return self.link_result_list(
+            service=self,
+            identity=identity,
+            results=links,
+            links_config=links_config,
+        )
+
+    def read_link(
```
ppanero, line 172: minor: just `read`?

invenio_rdm_records/services/services.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2020 CERN.
+# Copyright (C) 2021 CERN.
```
ppanero, line 3: should not remove, just update ranges.

tests/secret_links/test_sharing.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 #
+# Copyright (C) 2021 CERN.
```
ppanero, line 3: info: in this case is correct because is the first time we (CERN) touch it.

invenio_rdm_records/services/__init__.py
ppanero: still present

invenio_rdm_records/services/secret_links/service.py
```diff
@@ -0,0 +1,287 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
```
lnielsen, line 3: You will need to keep the copyright header from the original file here (it's important because it assigns ownership over the source code, and in this specific case it was done by TU Wien :-))

```
# Copyright (C) 2020-2021 CERN.
# Copyright (C) 2020-2021 Northwestern University.
# Copyright (C) 2021 TU Wien.
```

ppanero (approved):

PR #599, opened 2021-05-20
handlers: implement image opener for iiif

ppanero:
requires https://github.com/inveniosoftware/invenio-records-resources/pull/247
requires https://github.com/inveniosoftware/invenio-records-resources/pull/249
requires https://github.com/inveniosoftware/invenio-iiif/pull/39
requires https://github.com/inveniosoftware/invenio-rdm-records/pull/602

closes https://github.com/inveniosoftware/invenio-iiif/issues/38

Status with all other PR requirements - working use case e.g. `https://127.0.0.1:5000/api/iiif/v2/eq18w-nhc19:ouroboros.png/full/250,/0/default.jpg`

### DIFF
--- a/invenio_rdm_records/resources/iiif.py
+++ b/invenio_rdm_records/resources/iiif.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Handler functions for Invenio-RDM-Records."""
+
+
+import tempfile
+
+import pkg_resources
+from flask import g
+
+try:
+    pkg_resources.get_distribution('wand')
+    from wand.image import Image
+    HAS_IMAGEMAGICK = True
+except pkg_resources.DistributionNotFound:
+    # Python module not installed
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/config.py
```diff
@@ -176,3 +176,10 @@ def _(x):
 RDM_RECORDS_DOI_DATACITE_PASSWORD = ""
 RDM_RECORDS_DOI_DATACITE_PREFIX = "10.1234"
 RDM_RECORDS_DOI_DATACITE_TEST_MODE = True
+
+
+# Invenio-IIIF
+# ===================
+
+IIIF_API_DECORATOR_HANDLER = None
```
lnielsen, line 184: These won't be picked up (Invenio-IIIF might initialize before RDM-Records and ext.py is not loading them). Can you remove from this PR and move to Invenio-App-RDM config.py instead?

invenio_rdm_records/handlers.py
```diff
@@ -0,0 +1,64 @@
+# -*- coding: utf-8 -*-
```
lnielsen, line 1: Can you move this file to ``resources/iiif_handler.py``? It belongs to presentation layer (accessing service and g object).

ppanero: discuss IRL, moved to `iiif.py`

Issue #426, opened 2021-02-26
Use `yaml.safe_load` for fixture loading


fenekku:
We should heed the current deprecation notice and security warnings regarding `yaml.load` explained here: https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation . Basically replace `load()` with `safe_load()`.

PR #601, opened 2021-05-20
vocabularies: check app_data first when loading fixtures

fenekku:
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/852
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/426 (will tackle users fixture shortly)

### DIFF
--- a/invenio_rdm_records/fixtures/vocabularies.py
+++ b/invenio_rdm_records/fixtures/vocabularies.py
@@ -35,7 +35,7 @@ class YamlIterator(DataIterator):
     def __iter__(self):
         """Iterate over records."""
         with open(self._data_file) as fp:
-            data = yaml.load(fp)
+            data = yaml.safe_load(fp)
             if data:  # Allow empty files
                 for entry in data:
                     yield entry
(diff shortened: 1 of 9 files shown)


invenio_rdm_records/vocabularies/convert_to_new_vocabulary.py
```diff
@@ -69,31 +68,46 @@ def to_dict(self, csv_row):
...
+                    [self.converter_type.to_dict(row) for row in reader],
+                    yaml_file
+                )
+            click.secho("Conversion done!", fg="green")
+
+
+@click.command()
+@click.argument('csv_filename')
+@click.option('--to', default='.', help='Export destination directory.')
+def convert(csv_filename, to):
```
ppanero, line 96: moderate: should we move this to the `cli.py`?

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -84,29 +84,41 @@ def __iter__(self):
...
+            if not filepath.exists():
+                continue
+
+            with open(filepath) as fp:
+                data = yaml.safe_load(fp)
+                for id_, entry in data.items():
+                    if id_ not in ids:
+                        entry["data-file"] = path / entry["data-file"]
+                        self.load_vocabulary(id_, entry)
+                        ids.add(id_)
```
ppanero, line 113: minor: should we instead fail if there are duplicated? or at least issue a warning so the user might check it?

invenio_rdm_records/vocabularies/convert_to_new_vocabulary.py
fenekku: I was inclined not to since it's like those upgrade scripts: meant to be used for this release and that's it. Also the command would be quite long: `pipenv run invenio rdm-records fixtures convert <filename> <dirname>` ...

invenio_rdm_records/fixtures/vocabularies.py
fenekku: The `ids` here refer to the ids found in the vocabularies.yaml file -not the id of each individual vocabulary record. There are 2 layers of files for vocabularies. 

We allow overridding vocabularies on a per vocabulary basis. That is, an instance can provide a `vocabularies.yaml` that lists only the `languages` id (that entry points to a separate file with all the actual languages entries). The instance will load those languages, but continue looking for any other vocabularies in the next location where it could find a `vocabularies.yaml` file (the one provided with this package). There, it will load the `vocabularies.yaml` but it won't load the `languages` vocabulary this time around, because it already did. This is the intended way to allow load priority.

fenekku:
closed via https://github.com/inveniosoftware/invenio-rdm-records/pull/607

PR #602, opened 2021-05-21
global: add service registration blueprint to ui app

ppanero:
closes https://github.com/inveniosoftware/invenio-records-resources/issues/248

@lnielsen I know this is not optimal AT ALL... but there is no other way (that we know of 😅) at the moment to unblock the fact that we cannot upload a file via UI (`pre_commit` hook fails) 😓


dfdan (approved):
confirm that current_service_registry._services no longer empty with this PR applied.

PR #607, opened 2021-05-24
vocabularies: load custom users fixtures in app_data/

### LINKED ISSUE #425: Revisit UserFixture
As an instance administrator I want to use my chosen email and password for an admin user.

Running `invenio-cli services setup` will load the UserFixture in rdm-records which automatically creates a super-user with email `admin@inveniosoftware.org`. No way out of it. Obviously `admin@inveniosoftware.org` is not an email a typical administrator wants. 
 
- ~~Why was the automatic creation of an admin user needed again? Would like not to need it.~~ Having an admin by default is fine as long as instance administrator can configure it completely from the get-go.
- The mechanism for fixture loading need to be expanded to allow an administrator to use their own yaml file (replacing the file in the virtualenv `site_packages/`is not a desirable method... )
- In that context, revisit `RDM_RECORDS_USER_FIXTURE_PASSWORDS` config and what we want to do with it.

### LINKED ISSUE #426: Use `yaml.safe_load` for fixture loading

### LINKED ISSUE #606: Custom Users fixture loading
Englobes:
- [x] https://github.com/inveniosoftware/invenio-rdm-records/issues/426
- [x] https://github.com/inveniosoftware/invenio-rdm-records/issues/425

fenekku:
- closes #606 
- closes #425 
- closes #426

### DIFF
--- a/invenio_rdm_records/fixtures/users.py
+++ b/invenio_rdm_records/fixtures/users.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 #
+# Copyright (C) 2021 Northwestern University.
 # Copyright (C) 2021 TU Wien.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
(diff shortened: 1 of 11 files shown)


invenio_rdm_records/fixtures/users.py
```diff
@@ -22,20 +23,32 @@
...
+            # Providing a users.yaml file is optional
+            if not filepath.exists():
+                continue
+
+            with open(filepath) as fp:
+                data = yaml.safe_load(fp) or {}
+                for email, user_data in data.items():
+                    self.create_user(email, user_data)
+
+            break
```
ppanero, line 48: we `break` becuase there cannot be more than one *users* file? maybe it can be explained in a comment or in the docstring.

tests/fixtures/conftest.py
```diff
@@ -23,3 +24,12 @@ def vocabularies(system_process_identity):
         [pathlib.Path(__file__).parent / "data"],
         'vocabularies.yaml',
     )
+
+
+@pytest.fixture(scope="function")
+def admin_role(db):
+    """Create an admin role."""
+    role = Role(name="admin")
```
ppanero, line 32: can use this roles? https://github.com/inveniosoftware/invenio-access/blob/87dde7f6b43bda41ec2115ef4a358442d8136703/invenio_access/admin.py#L159

ppanero (approved):

invenio_rdm_records/fixtures/users.py
fenekku: Clarified this in the function description :+1:

tests/fixtures/conftest.py
fenekku: :thinking: I don't think so... we just need a pure role in the database named "admin" like the `invenio-cli services setup` gives us:

```python
            CommandStep(
                cmd=['pipenv', 'run', 'invenio', 'roles', 'create', 'admin'],
                env={'PIPENV_VERBOSITY': "-1"},
                message="Creating admin role..."
            )
```

fenekku:
Merging this in to get things moving.

PR #615, opened 2021-05-27
resources: IIIF manifest support

lnielsen:


egabancho:
> Main issue: Files are not serialized into record projection, so the IIIF serializer has no way of getting the list of files. Also, permissions should be taken into account (e.g. files can be restricted and a user would not be able to see them). Thus the service should be in charge.

I got the files from the corresponding service, either `draft_files` or `files`, which should do all the permission checks and friends for us.
I hope this type of stuff is ok https://github.com/inveniosoftware/invenio-rdm-records/pull/615/files#diff-a78ef7917e10ab1669945ba1a0f2ba7a46ac560e5b1d14e9c3875878b2f5c433R168-R169

> Add links to IIIF manifest in the record JSON (rely on these links in the IIIFPresiSchema). Core problem: The serializer should not know if it deals with a draft (preview) or a record.

Like this https://github.com/inveniosoftware/invenio-rdm-records/pull/615/files#diff-640bfb3d54c64abe82cf753d8ffe63ddc5735cc6cca4388f6c9fc8c339f04102R151-R155?

An issue related to this, which I haven't solved, is the IIIF file links, like the `info.json` link. So far, all these links are generated by Invenio-IIIF (and the underlining Flask-IIIF). More specifically https://github.com/inveniosoftware/flask-iiif/blob/5a250a295e274a23da533dbe923e1072236f1164/flask_iiif/utils.py#L24
Should we add a new RDMFileResource to add this? Should we move part of the logic from Invenio-IIIF to a file service here?

egabancho:
I have been playing around with the file IIIF links, and here are a few thoughts I'd like to share before continuing working on this.

The file links per se are not completed. I guess we can add something like this to the `FileServiceConfig`:

```
"iiif_info": FileLink(
    "{+api}/records/{id}/files/{key}/iiif/{version=v2}/info.json"
    ),
"iiif": FileLink(
    "{+api}/records/{id}/files/{key}/iiif/"
    "{version=v2}/{region=full}/{size=full}/{rotation=0}/{quality=default}.{format=png}"
)
```

(I can change the URL if we don't like it, no biggy)

Now, the more significant issue comes when trying to serve these types of requests. The package responsible for that right now is Flask-IIIF. And the URLs there are [hardcoded](https://github.com/inveniosoftware/flask-iiif/blob/master/flask_iiif/__init__.py#L116-L133). We could bypass the problem by using the same schema (which is the case already), but then the issue comes when trying to find the actual file to open it and manipulate it. 

Currently, it passes what it calls `UUID`, a combination of Bucket id, ObjectVersion id, and file key separated by `:`, but we can change that to something else using the config `IIIF_IMAGE_OPENER_HANDLER`. I guess this is what we tried with https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/iiif.py#L30
This will make us generate pretty ugly URLs 😉 `api/records/1234-abcd/files/image.tiff/iiif/v2/1234-abcd:image.tiff/full/full/0/default.png` (`1234-abcd:image.tiff` is the UUID), I image we can live with that 🤮 
The biggest issue is that the image opener is not aware of the record vs draft situation, and it can't be because the only information it gets from [Flask-IIIF](https://github.com/inveniosoftware/flask-iiif/blob/master/flask_iiif/restful.py#L76) (restful) is the UUID (`<record id>:<file key>`). I supposed we could include that on the URL, but it starts to become too much, IMHO, plus 🤮 

My proposed solution is to create a FileReorce, which extends Invenio-Record-Resources one and add there the handlers for the IIIF image URLs instead of relying on Invenio-IIIF and Flask-IIIF to do that. This should solve both issues I described before.
I am not sure if this resource should leverage the existing Flask-IIIF python API to make the responses, or we should instead move this code over to the service layer, TDB.

ppanero:
Does this PR superseed/discard this one <https://github.com/inveniosoftware/invenio-rdm-records/pull/612/>? seems at least the `manifest` endpoint is in both.

egabancho:
@ppanero yes it does

> I've opened this PR since I cannot push directly to the other PR.

dfdan:
I just spoke with Peter and @egabancho about various things including the URL for the Image API issues.  There are a couple of things to factor into how this (ie the binary blobs of pixels, not the manifest) URLs should work. 

As noted above, the existing flask/invenio-iiif modules make URLs like this -
```https://rsz.biodiversity.hasdai.org/api/iiif/v2/4b15r-86w46:page0001.png/full/full/0/default.jpg```

The advantages of keeping a separate path like /api/iiif (or just /iiif) for Image API operations that I can think of are:

* Compact URLs - the true unique identifier for the file requested is not split up (ie {id}/files/{key}) as the exiting RDM file URL structure leans towards. I'm concerned this may cause some interoperability issues as well as make crazy long URLs as noted above.
* Large scale deployments are likely to want to use separate/optimised processes (as with API/UI split) to do the heavy-lifting of processing and delivering images.
* Its easier to isolate and manage request limits in a separate path - IIIF Image services get lots of requests which will quickly cause 429 errors if not set appropriately.

File: setup.cfg [hunk 1]
```diff
@@ -32,7 +33,9 @@ install_requires =
     datacite>=1.1.1
     dcxml>=0.1.2
     Faker>=2.0.3
+    flask-iiif>=0.6.2
     ftfy>=4.4.3,<5.0.0
+    importlib_metadata>=4.4;python_version<'3.8'
```
lnielsen, line 38: Minor: This one can be removed as it comes in through Invenio-Base.

tests/resources/test_iiif_image_api.py
```diff
@@ -0,0 +1,202 @@
...
+    )
+
+    # Publish the record
+    res = client.post(f"/records/{id_}/draft/actions/publish", headers=headers)
+
+    return id_
+
+
+def test_iiif_base(
+    running_app, es_clear, client_with_login, headers, minimal_record
```
lnielsen, line 71:
```suggestion
    running_app, es_clear, client, uploader, headers, minimal_record
```

tests/resources/test_iiif_image_api.py
```diff
@@ -0,0 +1,202 @@
...
+):
+    client = client_with_login
```
lnielsen, line 73:
```suggestion
    client = uploader.login(client)
```

tests/resources/test_iiif_image_api.py
```diff
@@ -0,0 +1,202 @@
...
+    headers,
+    minimal_record,
+    users,
+):
+    client = client_with_login
+    file_id = "test_image.png"
+    recid = publish_record_with_images(
+        client, file_id, minimal_record, headers, restricted_files=True
+    )
+    logout_user(client)
```
lnielsen, line 127: client = uploader.logout(client)

tests/resources/test_iiif_presentation_api.py
```diff
@@ -0,0 +1,164 @@
...
+
+import flask_security
+from invenio_accounts.testutils import login_user_via_session
+from PIL import Image
+from tripoli import IIIFValidator
+
+from invenio_rdm_records.proxies import current_rdm_records
+
+
+def login_user(client, user):
```
lnielsen, line 20: Ditto here

invenio_rdm_records/resources/resources.py
```diff
@@ -227,3 +241,167 @@ def search(self):
...
+        return self.service.read_record(uuid=uuid, identity=g.identity)
+
+    #
+    # IIIF Manifest - not all clients support content-negotiation so we need a
+    # full endpoint.
+    #
+    # See https://iiif.io/api/presentation/2.1/#responses on
+    # "Access-Control-Allow-Origin: *"
+    #
+    @cross_origin(origin="*", methods=["GET"])
```
lnielsen, line 294: Question: @dfdan What was the reason behind allowing cross origin here? Mirador should be loaded from the same domain so it should be able to access the API?

invenio_rdm_records/resources/resources.py
```diff
@@ -227,3 +241,167 @@ def search(self):
...
+        )
+        # TODO: get from cache on the service image.last_modified
+        last_modified = None
+        send_file_kwargs = {"mimetype": mimetype}
+        # last_modified is not supported before flask 0.12
+        if last_modified:
+            send_file_kwargs.update(last_modified=last_modified)
+
+        if "dl" in request.args:
+            filename = secure_filename(request.args.get("dl", ""))
```
lnielsen, line 390: Minor: The ``dl`` should be parsed and added in the ``resource_requestctx.args``.

invenio_rdm_records/resources/resources.py
```diff
@@ -227,3 +241,167 @@ def search(self):
...
+            if filename.lower() in {"", "1", "true"}:
+                filename = "{0}-{1}-{2}-{3}-{4}.{5}".format(
+                    uuid, region, size, quality, rotation, image_format
+                )
+            send_file_kwargs.update(
+                as_attachment=True,
+                attachment_filename=secure_filename(filename),
+            )
+        if_modified_since_raw = request.headers.get("If-Modified-Since")
```
lnielsen, line 399: Minor: The header should be parsed and added in the ``resource_requestctx.headers``.

invenio_rdm_records/services/config.py
```diff
@@ -209,6 +220,19 @@ class RDMFileRecordServiceConfig(FileServiceConfig, ConfiguratorMixin):
         "RDM_PERMISSION_POLICY", default=RDMRecordPermissionPolicy
     )
 
+    file_links_item = {
+        **FileServiceConfig.file_links_item,
+        # FIXME: should we check if the file is IIIF compatible?
+        # FIXME: filename instead
```
lnielsen, line 226: Question: Do all files get IIIF links and manifest?

invenio_rdm_records/services/services.py
```diff
@@ -4,20 +4,39 @@
...
+from invenio_records_resources.services import LinksTemplate, Service
 from invenio_records_resources.services.uow import RecordCommitOp, unit_of_work
 
 from invenio_rdm_records.services.errors import EmbargoNotLiftedError
 
+try:
+    from importlib import metadata
+except ImportError:
+    # Running on pre-3.8 Python; use importlib-metadata package
+    import importlib_metadata as metadata
```
lnielsen, line 30: Minor: You can just use ``import importlib_metadata as metadata`` as we're bound to it until at least 3.9 is EOL.

invenio_rdm_records/services/services.py
```diff
@@ -125,3 +144,123 @@ def search_community_records(self, identity, community_uuid, params=None,
...
+
+    def read_record(self, identity, uuid):
+        """Read the correct version of the record and its files."""
+        type_, id_ = self._iiif_uuid(uuid)
+        read = (
+            self._records_service.read
+            if type_ == "record"
+            else self._records_service.read_draft
+        )
+        # Kids, don't do this at home
```
lnielsen, line 192: 😂

lnielsen (commented):
Great job!!!

There's some minor comments and improvements. My main concerns is primarily:

1) Cross origin: Why it's needed, and that it's documented it in the code so it's clear when there's exception why it's there.

2) The links in the record REST API response - I'm not sure if the IIIF links are included or not, but I think they should only be included if there's something. Not sure how easy that is.

Last, as I'm unsure about performance/caching, I'm not sure if we should launch it behind a feature flag (especially if also the mirador previewer UX is not in full shape).

invenio_rdm_records/resources/resources.py
egabancho: I believe the reason is to allow external "viewers" to load the content, i.e. mirador running on a different website.

invenio_rdm_records/services/config.py
egabancho: Right now, yes. I'd imagine I can make the links conditional based on file extension and https://github.com/inveniosoftware/flask-iiif/blob/master/flask_iiif/config.py#L86

egabancho: Probably I should add the same logic at the record level. If there are IIIF compatible files, generate the manifest otherwise, don't do it.
But this will be trickier since I don't have access to the files at the record level ...

dfdan:
Related to this - I was going to propose also re-implementing the iiif_simple previewer for the new endpoints - ie use IIIF to load a static image of appropriate size for the preview window. That might at push us to a situation where IIIF was on by default and have a separate one (or just install the m3 module) if people want that. 

@egabancho is correct about the CORS headers. A more concise comment for the code might be 'spec requirement' :).

PR #619, opened 2021-05-28
fixtures: load resource types from files

ntarocco:


invenio_rdm_records/fixtures/demo.py
```diff
@@ -34,12 +36,14 @@ class CachedVocabularies:
...
+
+            dir_ = Path(__file__).parent
+
+            cls._resource_type_ids = list(VocabulariesFixture(
+                system_identity,
+                [Path("./app_data"), dir_ / "data"],
+                "vocabularies.yaml",
+            ).get_records_by_vocabulary("resource_types"))
 
         random_id = random.choice(cls._resource_type_ids)
```
ppanero, line 48: This will return a full record but here we only need `random_id["id"]`?

ntarocco: Thanks! Fixed! :)

ntarocco: I will remove the `print` in a minute

ppanero (approved):

PR #623, opened 2021-06-01
mappings: migrate resource_type to vocbulary style

### LINKED ISSUE #622: facets not present on search
The rest API response has an empty aggregation bucket:
```
"aggregations": {
    "resource_type": {
        "buckets": [], 
        "label": "Resource types"
    },
    "languages": {
        "buckets": [],
        "label": "Languages"
    }
}
```

## Possible issues:
**Resource type**
- Value labels was not migrated [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/config.py#L44)
- A simple search on ES returns empty buckets... `id` is created a `text` not keyword.
```
curl -X GET "localhost:9200/_search?pretty" -H "Content-Type: application/json" -d '
{   "query": {
        "match_all": {}
    },
    "aggs": {"resource_type": {"terms": {"field": "metadata.resource_type.id"}}}
}'
```

The mapping reports the keyword being inside, the term filter needs to go to it... or check the mapping for possible errors
```
"resource_type": {
              "properties": {
                "id": {
                  "type": "text",
                  "fields": {
                    "keyword": {
                      "type": "keyword",
                      "ignore_above": 256
                    }
                  }
                },
            [...]
```

**Languages**
- They are not indexed in ES...
- The mapping does not contain a `keyword` attribute, `id` is already of said type (no `text` as in resource_type).
```
"languages":{
    "properties":{
        "id":{"type":"keyword"}
```

ppanero:
closes #622 

Facets are back

Screenshot 2021-06-01 at 16 11 34

### DIFF
--- a/invenio_rdm_records/fixtures/demo.py
+++ b/invenio_rdm_records/fixtures/demo.py
@@ -52,6 +52,12 @@ class CachedVocabularies:
         random_id = random.choice(cls._resource_type_ids)
         return {"id": random_id}
 
+    @classmethod
+    def fake_language(cls):
+        """Generate a random resource_type."""
+        random_id = random.choice(["eng", "aah", "aag"])
+        return {"id": random_id}
+
 
 def fake_edtf_level_0():
     """Generates a fake publication_date string."""
@@ -159,8 +165,7 @@ def create_fake_record():
             #     "description": "Random test date",
             #     "type": "other"
             # }],
-            # TODO: Add when we have PIDs for languages vocabulary
-            # "languages": [{"id": "eng"}],
+            "languages": [CachedVocabularies.fake_language()],
             # "related_identifiers": [{
             #     "identifier": "10.9999/rdm.9999988",
             #     "scheme": "doi",
--- a/invenio_rdm_records/resources/serializers/datacite/schema.py
+++ b/invenio_rdm_records/resources/serializers/datacite/schema.py
@@ -206,11 +206,9 @@ class DataCite43Schema(Schema):
             type_ = add_title.get("type")
             if type_:
                 title["titleType"] = type_.capitalize()
-            lang = add_title.get("lang")
-            if lang:
-                id = lang.get("id")
-                if id:
-                    title["lang"] = id
+            lang_id = add_title.get("lang", {}).get("id")
+            if lang_id:
+                title["lang"] = lang_id
 
             titles.append(title)
 
--- a/tests/resources/conftest.py
+++ b/tests/resources/conftest.py
@@ -13,10 +13,14 @@ See https://pytest-invenio.readthedocs.io/ for documentation on which test
 fixtures are available.
 """
 
+from collections import namedtuple
 from copy import deepcopy
 
 import pytest
+from invenio_access.permissions import system_identity
 from invenio_app.factory import create_api
+from invenio_vocabularies.proxies import current_service as vocabulary_service
+from invenio_vocabularies.records.api import Vocabulary
 
 from invenio_rdm_records import config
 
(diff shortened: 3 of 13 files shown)


invenio_rdm_records/fixtures/demo.py
```diff
@@ -31,27 +31,38 @@ class CachedVocabularies:
...
             for res in res_types:
                 cls._resource_type_ids.append(res["id"])
 
         random_id = random.choice(cls._resource_type_ids)
         return {"id": random_id}
 
+    @classmethod
+    def fake_language(cls):
+        """Generate a random resource_type."""
+        random_id = random.choice(["eng", "aah", "aag"])
```
fenekku, line 63: much simpler ;)

invenio_rdm_records/records/mappings/v6/rdmrecords/drafts/draft-v2.0.0.json
```diff
@@ -363,12 +368,17 @@
...
+                  "type": "keyword"
                 },
-                "type" : {
-                  "type" : "keyword"
+                "id": {
+                  "type": "keyword"
+                },
+                "title": {
+                  "type": "object",
+                  "dynamic": true
```
fenekku, line 381: Since this is the old mapping, we should leave this file alone and not provide these changes. Previous resource types should operate the same as before.

invenio_rdm_records/records/mappings/v6/rdmrecords/records/record-v2.0.0.json
```diff
@@ -363,12 +368,17 @@
...
+                  "type": "keyword"
                 },
-                "type" : {
-                  "type" : "keyword"
+                "id": {
+                  "type": "keyword"
+                },
+                "title": {
+                  "type": "object",
+                  "dynamic": true
```
fenekku, line 381: Same as above. We should leave this file alone and only change v3.0.0.

invenio_rdm_records/records/mappings/v7/rdmrecords/drafts/draft-v2.0.0.json
```diff
@@ -362,12 +367,17 @@
...
+                "type": "keyword"
               },
-              "type" : {
-                "type" : "keyword"
+              "id": {
+                "type": "keyword"
+              },
+              "title": {
+                "type": "object",
+                "dynamic": true
```
fenekku, line 380: Same

invenio_rdm_records/records/mappings/v7/rdmrecords/records/record-v2.0.0.json
```diff
@@ -362,12 +367,17 @@
...
+                "type": "keyword"
               },
-              "type" : {
-                "type" : "keyword"
+              "id": {
+                "type": "keyword"
+              },
+              "title": {
+                "type": "object",
+                "dynamic": true
```
fenekku, line 380: same

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -206,11 +206,9 @@ def get_titles(self, obj):
             type_ = add_title.get("type")
             if type_:
                 title["titleType"] = type_.capitalize()
-            lang = add_title.get("lang")
-            if lang:
-                id = lang.get("id")
-                if id:
-                    title["lang"] = id
+            lang_id = add_title.get("lang", {}).get("id")
```
fenekku, line 209: :+1:

invenio_rdm_records/fixtures/demo.py
```diff
@@ -31,27 +31,38 @@ class CachedVocabularies:
...
+    @staticmethod
+    def _get_vocab(vocabulary):
+        """Get all records of a vocabulary."""
+        dir_ = Path(__file__).parent
+
+        return VocabulariesFixture(
+            system_identity,
+            [Path("./app_data"), dir_ / "data"],
+            "vocabularies.yaml",
+        ).get_records_by_vocabulary(vocabulary)
```
fenekku, line 45: minor: Now that we are considering multiple vocabularies and we have dealt with importing them, it might be that we want to rely on the vocabularies services instead: `vocabularies_service.read_many` might give us a pool of id to randomize.

fenekku (approved):
Thanks, I knew there were going to be resource_types remnants. 

Leaving previous versions alone is important, approving though so you are unblocked when the tests are fixed.

invenio_rdm_records/fixtures/demo.py
ppanero: `read_many` requires the ids to be queried, we could use `read_all`... but we would be in the same situation than querying the db... When creating demo records or anything in bulk, it could be that the vocab records are not yet there (commited or refreshed) and therefore it will fail.

invenio_rdm_records/records/mappings/v6/rdmrecords/drafts/draft-v2.0.0.json
ppanero: true, nice catch sorry about that

tests/resources/conftest.py
```diff
@@ -61,3 +65,54 @@ def headers():
...
+
+
+@pytest.fixture(scope="function")
+def resource_type_type(app):
+    """Resource type vocabulary type."""
+    return vocabulary_service.create_type(
+        system_identity, "resource_types", "rsrct")
+
+
+@pytest.fixture(scope="function")
```
ppanero, line 77: Had to make it function scoped and therefore remove it from `runnint_app`. The issue comes when used along `es_clear`. Since the vocabulary will be indexed and the labels come from ES, any previous test with `es_clear` will remove it and it will not be present when needed (e.g. labels). Note: this only happens at `tests/resources/` level since es_clear is not used otherwise.

invenio_rdm_records/fixtures/demo.py
fenekku: Ah right, now that vocabulaires are actually loaded when the worker starts i.e. after services setup. oh well.

PR #645, opened 2021-06-08
datacite: serialize languages following ISO 639-1

### LINKED ISSUE #629: Languages in Datacite are defined by RFC3066 but the internal representation is exported
Package version (if known): 4.0

## Describe the bug
The datacite serializer serializes the languages as they are. While in datacite schema a language is defined by RFC3066, basically a 2-letter code plus optionally - and the country code. Might there be a problem with minting a DOI?!

## Steps to Reproduce
1. Not really reproducible?!

## Expected behavior
Languages should be exported according to RFC3066.

ppanero:
closes #629 

~~How do we deal with those that have no alpha_2? There are many...~~ --> `missing` since it's not a mandatory field in datacite

### DIFF
--- a/invenio_rdm_records/resources/serializers/datacite/schema.py
+++ b/invenio_rdm_records/resources/serializers/datacite/schema.py
@@ -184,6 +184,17 @@ class DataCite43Schema(Schema):
             system_identity
         )._record
 
+    def _read_language(self, id_):
+        """Retrieve resource type record using service."""
+        lang = vocabulary_service.read(
+            ('languages', id_),
+            system_identity
+        )._record
+
+        alpha_2 = lang["props"].get("alpha_2", None)
+
+        return alpha_2 or missing
+
     def get_type(self, obj):
         """Get resource type."""
         resource_type = obj["metadata"]["resource_type"]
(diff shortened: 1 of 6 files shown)


invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -184,6 +184,17 @@ def _read_resource_type(self, id_):
...
+    def _read_language(self, id_):
+        """Retrieve resource type record using service."""
+        lang = vocabulary_service.read(
+            ('languages', id_),
+            system_identity
+        )._record
+
+        alpha_2 = lang.get("props", {}).get("alpha_2", '')
+
+        return alpha_2 if alpha_2 else missing  # account for ''
```
ntarocco, line 196: cosmetic: another option (`props` should be always there?):
```suggestion
        alpha_2 = lang["props"].get("alpha_2", None)
        return alpha_2 or missing
```

ntarocco (approved):

invenio_rdm_records/resources/serializers/datacite/schema.py
ppanero: > cosmetic: another option (props should be always there?):

Should be if people use our `language.py` script... We sort of need it for functionality, so I would say *yes*... but can make it optional too. WDYT?

PR #647, opened 2021-06-08
schemas: remove 'required' properties from jsonschema

max-moser:
* previously, missing required properties would lead to generic 500
  errors during the upload of metadata for drafts without useful
  information
* also, the draft would not be saved
* this was because during draft creation, marshmallow validation
  errors wouldn't lead to an abort, but jsonschema validation errors
  would
* as per discussion, the role of jsonschema is primarily for basic
  structural validation nowadays

### DIFF
--- a/invenio_rdm_records/records/jsonschemas/records/record-v3.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/record-v3.0.0.json
@@ -39,7 +39,6 @@
           "items": {
             "type": "object",
             "additionalProperties": false,
-            "required": ["person_or_org"],
             "properties": {
                 "person_or_org": {"$ref": "local://records/definitions-v1.1.0.json#/person_or_org"},
                 "role": {"type": "string"},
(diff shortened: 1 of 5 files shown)


max-moser:
removed many of the `required` fields where there exist such rules in marshmallow
exceptions: the internal PIDs (field `pid`) because there's no marshmallow equivalent for that, and many of the GEO JSON things, as they're still WIP

invenio_rdm_records/records/jsonschemas/records/record-v3.0.0.json
```diff
@@ -226,7 +223,6 @@
           "$comment": "This part of the schema is modelled on GeoJSON, but without the scope to embed arbitrary metadata.\n\nWe nest `features` within `locations` to give scope later to say something about the locations as a whole.",
           "type": "object",
           "additionalProperties": false,
-          "required": ["features"],
```
ppanero, line 229: might this be the cause of the issue? I do not see features anywhere in marshmallow...

ppanero (approved):
LGTM, @slint could you confirm this is the correct approach?

ntarocco (approved):

PR #648, opened 2021-06-09
CSL string serializer

Pineirin:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/646
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/651
* Add citation string serializer.

### DIFF
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -14,7 +14,8 @@ from flask_resources import HTTPJSONException, JSONSerializer, \
 from invenio_drafts_resources.resources import RecordResourceConfig
 from invenio_records_resources.resources.files import FileResourceConfig
 
-from .serializers import CSLJSONSerializer, UIJSONSerializer
+from .serializers import CSLJSONSerializer, StringCitationSerializer, \
+    UIJSONSerializer
 
 #
 # Response handlers
@@ -22,8 +23,12 @@ from .serializers import CSLJSONSerializer, UIJSONSerializer
 record_serializers = {
     "application/json": ResponseHandler(JSONSerializer()),
     "application/vnd.inveniordm.v1+json": ResponseHandler(UIJSONSerializer()),
-    "application/vnd.citationstyles.csl+json":
-        ResponseHandler(CSLJSONSerializer()),
+    "application/vnd.citationstyles.csl+json": ResponseHandler(
+        CSLJSONSerializer()
+    ),
+    "text/x-bibliography": ResponseHandler(
+        StringCitationSerializer(), headers={"content-type": "text/plain"}
+    ),
 }
 
 
@@ -76,14 +81,16 @@ class RDMDraftFilesResourceConfig(FileResourceConfig):
 record_links_error_handlers = RecordResourceConfig.error_handlers.copy()
 
 
-record_links_error_handlers.update({
-    LookupError: create_error_handler(
-        HTTPJSONException(
-            code=404,
-            description="No secret link found with the given ID.",
-        )
-    ),
-})
+record_links_error_handlers.update(
+    {
+        LookupError: create_error_handler(
+            HTTPJSONException(
+                code=404,
+                description="No secret link found with the given ID.",
+            )
+        ),
+    }
+)
 
 
 class RDMParentRecordLinksResourceConfig(RecordResourceConfig):
@@ -102,11 +109,9 @@ class RDMParentRecordLinksResourceConfig(RecordResourceConfig):
 
     request_view_args = {
         "pid_value": ma.fields.Str(),
-        "link_id": ma.fields.Str()
+        "link_id": ma.fields.Str(),
     }
 
-    response_handlers = {
-        "application/json": ResponseHandler(JSONSerializer())
-    }
+    response_handlers = {"application/json": ResponseHandler(JSONSerializer())}
 
     error_handlers = record_links_error_handlers
--- a/invenio_rdm_records/resources/serializers/__init__.py
+++ b/invenio_rdm_records/resources/serializers/__init__.py
@@ -8,13 +8,13 @@
 
 """Record response serializers."""
 
-from .csl import CSLJSONSerializer
+from .csl import CSLJSONSerializer, StringCitationSerializer
 from .datacite import DataCite43JSONSerializer
 from .ui import UIJSONSerializer
 
 __all__ = (
-    "UIJSONSerializer",
     "CSLJSONSerializer",
     "DataCite43JSONSerializer",
-
+    "StringCitationSerializer",
+    "UIJSONSerializer",
 )
--- a/invenio_rdm_records/resources/serializers/csl/__init__.py
+++ b/invenio_rdm_records/resources/serializers/csl/__init__.py
@@ -5,9 +5,19 @@
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
 
-"""CSL JSON Serializers for Invenio RDM Records."""
+"""CSL JSON and  citation string serializers for Invenio RDM Records."""
 
+import re
+
+from citeproc import Citation, CitationItem, CitationStylesBibliography, \
+    CitationStylesStyle, formatter
+from citeproc.source.json import CiteProcJSON
+from citeproc_styles import get_style_filepath
+from citeproc_styles.errors import StyleNotFoundError
+from flask import abort, request
 from flask_resources.serializers import MarshmallowJSONSerializer
+from webargs import fields
+from webargs.flaskparser import FlaskParser
 
 from .schema import CSLJSONSchema
 
@@ -18,3 +28,96 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
     def __init__(self, **options):
         """Constructor."""
         super().__init__(schema_cls=CSLJSONSchema, **options)
+
+
+def get_citation_string(json, id, style, locale):
+    """Get the citation string from Citeproc."""
+
+    def _clean_result(text):
+        """Remove double spaces, punctuation and escapes apostrophes."""
+        text = re.sub(r"\s\s+", " ", text)
+        text = re.sub(r"\.\.+", ".", text)
+        text = text.replace("'", "\\'")
+        return text
+
+    source = CiteProcJSON([json])
+    citation_style = CitationStylesStyle(
+        validate=False, style=style, locale=locale
+    )
+    bib = CitationStylesBibliography(citation_style, source, formatter.plain)
+    citation = Citation([CitationItem(id)])
+    bib.register(citation)
+
+    return _clean_result(str(bib.bibliography()[0]))
+
+
+class StringCitationSerializer(MarshmallowJSONSerializer):
+    """CSL Citation Formatter serializer for records.
+
+    In order to produce a formatted citation of a record through citeproc-py,
+    we need a CSL-JSON serialized version of it.
+    """
+
+    _default_style = "harvard1"
+    """The `citeproc-py` library supports by default the 'harvard1' style."""
+
+    _default_locale = "en-US"
+    """The `citeproc-py` library supports by default the 'harvard1' style."""
+
+    _user_args = {
+        "style": fields.Str(missing=_default_style),
+        "locale": fields.Str(missing=_default_locale),
+    }
+    """Arguments for the webargs parser."""
+
+    _valid_formats = ("csl", "bibtex")
+    """Supported formats by citeproc-py."""
+
+    def __init__(self, **options):
+        """Constructor."""
+        super().__init__(schema_cls=CSLJSONSchema, **options)
+
+    @classmethod
+    def _get_args(cls, **kwargs):
+        """Parse style and locale.
+
+        Argument location precedence: kwargs > view_args > query
+        """
+        csl_args = {"style": cls._default_style, "locale": cls._default_locale}
+        parser = FlaskParser(locations=("view_args", "query"))
+        csl_args.update(parser.parse(cls._user_args, request))
+
+        csl_args.update(
+            {k: kwargs[k] for k in ("style", "locale") if k in kwargs}
+        )
+
+        try:
+            csl_args["style"] = get_style_filepath(csl_args["style"].lower())
+        except StyleNotFoundError:
+            abort(
+                400,
+                description=f"Style {csl_args['style']} could not be found.",
+            )
+
+        return csl_args
+
+    def serialize_object(self, record, **kwargs):
+        """Serialize a single record.
+
+        :param record: Record instance.
+        """
+        return get_citation_string(
+            self.dump_one(record), record["id"], **self._get_args(**kwargs)
+        )
+
+    def serialize_object_list(self, records, **kwargs):
+        """Serialize a list of records.
+
+        :param records: List of records instance.
+        """
+        return "\n".join(
+            [
+                self.serialize_object(rec, **kwargs)
+                for rec in records["hits"]["hits"]
+            ]
+        )
--- a/setup.py
+++ b/setup.py
@@ -64,6 +64,8 @@ setup_requires = [
 
 install_requires = [
     'arrow>=0.17.0',
+    'citeproc-py>=0.6.0',
+    'citeproc-py-styles>=0.1.2',
     'datacite>=1.1.1',
     'Faker>=2.0.3',
     'flask>=1.1,<2',
--- a/tests/resources/serializers/test_csl_serializer.py
+++ b/tests/resources/serializers/test_csl_serializer.py
@@ -1,4 +1,3 @@
-
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2021 CERN.
@@ -8,14 +7,20 @@
 
 """Resources serializers tests."""
 
-import pytest
+from copy import deepcopy
 
-from invenio_rdm_records.resources.serializers import CSLJSONSerializer
+from citeproc_styles import get_style_filepath
+
+from invenio_rdm_records.proxies import current_rdm_records
+from invenio_rdm_records.records.api import RDMRecord
+from invenio_rdm_records.resources.serializers import CSLJSONSerializer, \
+    StringCitationSerializer
+from invenio_rdm_records.resources.serializers.csl import get_citation_string
+from invenio_rdm_records.resources.serializers.csl.schema import CSLJSONSchema
 
 
-def test_csl_serializer(
-        running_app, full_record, vocabulary_clear):
-    """Test serializer to DayaCide 4.3 JSON"""
+def test_csl_json_serializer(running_app, full_record, vocabulary_clear):
+    """Test JSON CLS Serializer."""
     # if the record is created this field will be present
     full_record["id"] = "12345-abcde"
 
@@ -26,9 +31,11 @@ def test_csl_serializer(
         "title": "InvenioRDM",
         "issued": {"date-parts": [["2018"], ["2020", "09"]]},
         "abstract": "Test",
-        "author": [{
-            "family": "Nielsen, Lars Holm",
-        }],
+        "author": [
+            {
+                "family": "Nielsen, Lars Holm",
+            }
+        ],
         "note": "Funding by European Commission ROR 1234.",
         "version": "v1.0",
         "type": "graphic",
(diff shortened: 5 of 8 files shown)


invenio_rdm_records/resources/config.py
```diff
@@ -24,6 +25,8 @@
     "application/vnd.inveniordm.v1+json": ResponseHandler(UIJSONSerializer()),
     "application/vnd.citationstyles.csl+json":
         ResponseHandler(CSLJSONSerializer()),
+    "text/x-bibliography":
+        ResponseHandler(StringCitationSerializer()),
```
ppanero, line 29: Fits in one line:
```suggestion
    "text/x-bibliography": ResponseHandler(StringCitationSerializer()),
```

invenio_rdm_records/resources/serializers/__init__.py
```diff
@@ -8,12 +8,13 @@
...
-from .csl import CSLJSONSerializer
+from .csl import CSLJSONSerializer, StringCitationSerializer
 from .datacite import DataCite43JSONSerializer
 from .ui import UIJSONSerializer
 
 __all__ = (
     "UIJSONSerializer",
+    "StringCitationSerializer",
     "CSLJSONSerializer",
     "DataCite43JSONSerializer",
```
ppanero, line 19: I know the UI was already there... But could you changed them and sort them alphabetically?
```suggestion
    "CSLJSONSerializer",
    "DataCite43JSONSerializer",
    "StringCitationSerializer",
    "UIJSONSerializer",
```

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -5,9 +5,27 @@
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
 
-"""CSL JSON Serializers for Invenio RDM Records."""
+"""CSL JSON and string Serializers for Invenio RDM Records."""
```
ppanero, line 8:
```suggestion
"""CSL JSON and  citation string serializers for Invenio RDM Records."""
```

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +36,89 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+    _user_args = {
+        'style': fields.Str(missing=_default_style),
+        'locale': fields.Str(missing=_default_locale)
+    }
+    """Arguments for the webargs parser."""
+
+    _valid_formats = ('csl', 'bibtex')
+    """Supported formats by citeproc-py."""
+
+    def __init__(self, record_format='csl'):
```
ppanero, line 63: This makes is fully schema dependent. When we talked about it, it seemed like a good option... Now I am wondering if we should inherit from `Marshmallow....` and receive the `schema` parameter (defaulting to `CSLJSONSchema`). WDYT?

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +36,89 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+        """Parse style and locale.
+
+        Argument location precedence: kwargs > view_args > query
+        """
+        csl_args = {
+            'style': cls._default_style,
+            'locale': cls._default_locale
+        }
+
+        if has_request_context():
```
ppanero, line 82: I am wondering if we can avoid using `request` and many other `flask` objects... and use the `resource_requestctx`... this would require adding some parameters in the config...  e.g. in the new `resources` paradigm we already implemented an arg parser and here we are using `FlaskParser`, is a bit contradictory.



Maybe it can be improved in a different PR. @ntarocco this is maybe a discussion point for later on.

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +36,89 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+
+        try:
+            csl_args['style'] = get_style_filepath(csl_args['style'].lower())
+        except StyleNotFoundError:
+            if has_request_context():
+                raise RuntimeError(csl_args['style'])
+            raise
+        return csl_args
+
+    def _get_source(self, data):
```
ppanero, line 97: Where is this function used? I cannot find any reference in the code...

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +36,89 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+        text = re.sub(r'\.\.+', '.', text)
+        text = text.replace("'", "\\'")
+        return text
+
+    def serialize_object(self, record, **kwargs):
+        """Serialize a single record.
+
+        :param record: Record instance.
+        """
+        serialized_json = super().serialize_object(record)
```
ppanero, line 116: you can use `dump_one` (see [here](https://github.com/inveniosoftware/flask-resources/blob/master/flask_resources/serializers/json.py#L92)), and then you will not need to do `json.loads` below since you will already have a dict.

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +36,89 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+        source = CiteProcJSON([json.loads(serialized_json)])
+        style = CitationStylesStyle(validate=False, **self._get_args(**kwargs))
+        bib = CitationStylesBibliography(style, source, formatter.plain)
+
+        citation = Citation([CitationItem(record["id"])])
```
ppanero, line 121: Can we access the attribute?
```suggestion
        citation = Citation([CitationItem(record.id)])
```

tests/resources/serializers/test_csl_serializer.py
```diff
@@ -39,3 +40,23 @@ def test_csl_serializer(
     serialized_record = serializer.dump_one(full_record)
 
     assert serialized_record == expected_data
+
+
+def test_csl_serializer(
+        running_app, full_record, vocabulary_clear):
+    """Test serializer to DayaCide 4.3 JSON"""
```
ppanero, line 47:
```suggestion
    """Test Citation String Serializer."""
```

tests/resources/serializers/test_csl_serializer.py
```diff
@@ -39,3 +40,23 @@ def test_csl_serializer(
     serialized_record = serializer.dump_one(full_record)
 
     assert serialized_record == expected_data
+
+
+def test_csl_serializer(
+        running_app, full_record, vocabulary_clear):
```
ppanero, line 46: Fits in one line:
```suggestion
def test_csl_serializer( running_app, full_record, vocabulary_clear):
```

tests/resources/serializers/test_csl_serializer.py
```diff
@@ -39,3 +40,23 @@ def test_csl_serializer(
...
+                     'issued': {'date-parts': [['2018'], ['2020', '09']]},
+                     'type': 'graphic',
+                     'note': 'Funding by European Commission ROR 1234.',
+                     'version': 'v1.0', 'title': 'InvenioRDM',
+                     'author': [{'family': 'Nielsen, Lars Holm'}],
+                     'id': '12345-abcde', 'publisher': 'InvenioRDM',
+                     'DOI': '10.5281/inveniordm.1234'}
+
+    serializer = StringCitationSerializer()
+    serialized_record = serializer.dump_one(full_record)
```
ppanero, line 61: This test is nice, I wonder if we should add a random one for a specific format... Although we are relying on an 3rd party dependency so we should not have to test it... WDYT?

invenio_rdm_records/resources/serializers/csl/__init__.py
ntarocco: I am not sure if it can be really improved with more params in the config?
At some point, the `view_args` needs to be parsed...

I would remove `if has_request_context` because here we are at the resource layer, I think that we are sure that we are in a request context. WDYT?
Question: why using `FlaskParser` and not simply `request.args.get("style")`? Maybe it was in Zenodo code?

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +36,89 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+            parser = FlaskParser(locations=('view_args', 'query'))
+            csl_args.update(parser.parse(cls._user_args, request))
+
+        csl_args.update({k: kwargs[k]
+                         for k in ('style', 'locale') if k in kwargs})
+
+        try:
+            csl_args['style'] = get_style_filepath(csl_args['style'].lower())
+        except StyleNotFoundError:
+            if has_request_context():
```
ntarocco, line 92:
```suggestion

```

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +36,89 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+                raise RuntimeError(csl_args['style'])
```
ntarocco, line 93: I think that this should throw a `400` instead. Will `RuntimeError` throw a `500`?

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +36,89 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+        """
+        serialized_json = super().serialize_object(record)
+        source = CiteProcJSON([json.loads(serialized_json)])
+        style = CitationStylesStyle(validate=False, **self._get_args(**kwargs))
+        bib = CitationStylesBibliography(style, source, formatter.plain)
+
+        citation = Citation([CitationItem(record["id"])])
+        bib.register(citation)
+
+        return self._clean_result(''.join(bib.bibliography()[0]))
```
ntarocco, line 124: Missing the implementation of the method `serialize_object_list` to serialize back a list of results.

tests/resources/serializers/test_csl_serializer.py
ntarocco: Yes and please add tests for the view_args param. Test that your code will not fail when:
* passing bad values for args `style` and `locale` (it should throw a 400 error)
* not passing at all `style` and `locale` (it should use default)

invenio_rdm_records/resources/serializers/csl/__init__.py
ppanero: Indeed, correct. Should be a 400.

invenio_rdm_records/resources/serializers/csl/__init__.py
Pineirin: I don't think so, but I will test it again.

invenio_rdm_records/resources/serializers/csl/__init__.py
ppanero: > Not need to check context...

I fully agree, in this layer we can assume we are in a request context.

> About parsing view_args

Maybe the parsing itself is needed and cannot be done in the same fashion than other args, I agree. However, we should tyr to be consistent and access it through the [`resource_requestctx.view_args`](https://github.com/inveniosoftware/flask-resources/blob/5ca81aa51715f2bdd12be25c2e71d6c6fb1d7671/flask_resources/context.py#L61)

invenio_rdm_records/resources/serializers/csl/__init__.py
ppanero: IIRC, at this point `record` should be a `ResultItem` which has an `id` attribute, see [here](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/results.py#L33)

setup.py
```diff
@@ -73,6 +73,8 @@
     'invenio-communities>=2.3.0',
     'pytz>=2020.4',
     'pyyaml>=5.4.0',
+    'citeproc-py>=0.6.0',
+    'citeproc-py-styles>=0.1.2'
```
ppanero, line 77: Please add them in alphabetical order not at the end

invenio_rdm_records/resources/serializers/csl/__init__.py
Pineirin: Ok, I will leave this for now.

invenio_rdm_records/resources/serializers/csl/__init__.py
Pineirin: It depends, is there any option that we want the String Citation Serializer to be able to use different schemas?

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -5,9 +5,26 @@
...
+    CitationStylesStyle, formatter
+from citeproc.source.json import CiteProcJSON
+from flask import has_request_context, request
 from flask_resources.serializers import MarshmallowJSONSerializer
+from webargs import fields
+from webargs.flaskparser import FlaskParser
+from ...errors import StyleNotFoundRDMError
+
+try:
+    from citeproc_styles import get_style_filepath
```
ntarocco, line 22: Why this try/except? I think it is in the `requirements` and it is not an extra.

invenio_rdm_records/resources/serializers/csl/__init__.py
ntarocco: If `CSLJSONSerializer` is not needed as parent class, I would remove it. The code is using `self.dump_one`, should it use it from `CSLJSONSerializer` or from `Marshmallow...`?

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +35,90 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+        parser = FlaskParser(locations=('view_args', 'query'))
+        csl_args.update(parser.parse(cls._user_args, request))
+
+        csl_args.update({k: kwargs[k]
+                         for k in ('style', 'locale') if k in kwargs})
+
+        try:
+            csl_args['style'] = get_style_filepath(csl_args['style'].lower())
+        except StyleNotFoundError:
+            if has_request_context():
```
ntarocco, line 89:
```suggestion

```

tests/resources/serializers/test_csl_serializer.py
```diff
@@ -8,14 +8,34 @@
...
+from collections import namedtuple
+
 import pytest
 
-from invenio_rdm_records.resources.serializers import CSLJSONSerializer
+from invenio_rdm_records.proxies import current_rdm_records
+from invenio_rdm_records.resources.serializers import CSLJSONSerializer, \
+    StringCitationSerializer
 
+RunningApp = namedtuple("RunningApp", [
```
ntarocco, line 19: Is all this needed?

tests/resources/serializers/test_csl_serializer.py
```diff
@@ -39,3 +59,100 @@ def test_csl_serializer(
...
+    service = current_rdm_records.records_service
+    superuser_identity = running_app.superuser_identity
+    # create the draft
+    draft = service.create(superuser_identity, minimal_record)
+    # publish the record
+    record = service.publish(draft.id, superuser_identity)
+    doi = record.to_dict()["pids"]["doi"]["identifier"]
+
+    # test resolution
+    resolved_record = service.resolve_pid(
```
ntarocco, line 97: looks unused?

tests/resources/serializers/test_csl_serializer.py
```diff
@@ -39,3 +59,100 @@ def test_csl_serializer(
...
+    expected_data = {'abstract': 'Test', 'language': 'dan',
+                     'issued': {'date-parts': [['2018'], ['2020', '09']]},
+                     'type': 'graphic',
+                     'note': 'Funding by European Commission ROR 1234.',
+                     'version': 'v1.0', 'title': 'InvenioRDM',
+                     'author': [{'family': 'Nielsen, Lars Holm'}],
+                     'id': '12345-abcde', 'publisher': 'InvenioRDM',
+                     'DOI': '10.5281/inveniordm.1234'}
+
+    assert response == expected_data
```
ntarocco, line 158: I would test also from no locale. Maybe you can put all these similar tests into one function:
```
def test_..._params(...):

   - test default params
   - test different locale param
   - test different style param
   - test wrong local param
   - test wrong style param
```

tests/resources/serializers/test_csl_serializer.py
```diff
@@ -39,3 +43,56 @@ def test_csl_serializer(
...
+    record = service.publish(draft.id, superuser_identity)
+    id = record.id
+    headers = {"Accept": "text/x-bibliography"}
+
+    default_style = StringCitationSerializer._default_style
+    default_locale = StringCitationSerializer._default_locale
+
+    _url = f"/records/{id}"
+
+    test_cases = [
```
ppanero, line 76: This is nice! But I see an issue on debugging/understanding what could fail. There would be no proper message. Taking into account that most (if not all) fixtures are module-scoped. There is no penalty in using `pytest.mark.parametrize`
e.g. https://github.com/inveniosoftware/invenio-rdm-records/blob/c321c06893b92eb087cfcb3ba0f32b187aadf6fa/tests/services/schemas/test_location.py#L43 (there are more in that tests folder).

tests/resources/serializers/test_csl_serializer.py
```diff
@@ -39,3 +43,56 @@ def test_csl_serializer(
     serialized_record = serializer.dump_one(full_record)
 
     assert serialized_record == expected_data
+
```
ppanero, line 46: Would it be worth adding a tests for `serialize_object_list`? I see no other serializer test it. 🤔

ppanero (approved):
Cosmetic details. I am approving so it can be merged in case it is blocking. However, I think is nice exercise to learn to use `pytest.mark.parametrize`. @Pineirin

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +28,92 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+                  description=f"Style {csl_args['style']} could not be found.")
+
+        return csl_args
+
+    def serialize_object(self, record, **kwargs):
+        """Serialize a single record.
+
+        :param record: Record instance.
+        """
+        return get_citation_string(self.dump_one(record), record["id"],
```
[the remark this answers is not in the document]
Pineirin: As far as I tested record.id doesn't work in this case.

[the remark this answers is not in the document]
ppanero: You are right... I missed the `item.to_dict()` happening at resource level. 
-

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -18,3 +28,92 @@ class CSLJSONSerializer(MarshmallowJSONSerializer):
...
+        return text
+
+    source = CiteProcJSON([json])
+    citation_style = CitationStylesStyle(validate=False, style=style,
+                                         locale=locale)
+    bib = CitationStylesBibliography(citation_style, source, formatter.plain)
+    citation = Citation([CitationItem(id)])
+    bib.register(citation)
+
+    return _clean_result(''.join(bib.bibliography()[0]))
```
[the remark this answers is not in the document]
ntarocco: Yes, it should always return a list. Every time you call `register`, it will add the each `citation.cite` to the internal list of items, see [here](https://github.com/brechtm/citeproc-py/blob/9079882f5a5ca43be616dbe88c8e897c7374ed5b/citeproc/frontend.py#L85).

I think that since we only have 1 record, we always only one `cite`. No need to use `' '`, it will be always 1 element max.

tests/resources/serializers/test_csl_serializer.py
ntarocco: Good idea!

tests/resources/serializers/test_csl_serializer.py
ntarocco: It was me. We can use `parametrize`, but it will clean/re-create fixtures all the times. In ILS we basically removed all of them and replaced with this pattern to increase tests speed.
The drawback is that it is not easy to debug... you will have to put print/breakpoints.

ppanero: Yeah... Most of the fixtures seem to be `module` scope tho... but it's true that many others have `pytest.fixture` which will default to `function`.... This package takes ages to test already, so I agree with you lets sacrifice a bit of debugging for some speed :)

tests/resources/serializers/test_csl_serializer.py
```diff
@@ -39,3 +43,75 @@ def test_csl_serializer(
...
+    """Test Citation String Serializer given a list of records."""
+    # if the record is created this field will be present
+    service = current_rdm_records.records_service
+    records = []
+    expected_data = []
+    for i in range(3):
+        draft = service.create(superuser_identity, minimal_record)
+        record = service.publish(draft.id, superuser_identity)
+        records.append(record)
+        expected_data.append(f"Brown, Troy and Troy Inc. (2020) A Romans story. doi: 10.1234/datacite.{draft.id}.")  # noqa
```
ntarocco, line 73:
```suggestion
       expected_data.append(get_citation_string(
                CSLJSONSchema().dump(record), id, locale=default_locale,
                style=get_style_filepath(default_style)))
```

tests/resources/serializers/test_csl_serializer.py
```diff
@@ -39,3 +43,75 @@ def test_csl_serializer(
...
+
+    serializer = StringCitationSerializer()
+    serialized_records = serializer.serialize_object_list(records)
+    assert serialized_records == expected_data
```
ntarocco, line 77:
```
response = client.get("/records", headers=headers)
assert expected_data == response.get_data(as_text=True)
```

Issue #663, opened 2021-06-15
serializer: add Dublin Core serializer


rekt-hard:
Add schema for Dublin Core
Add Dublin Core serializer

lnielsen:
https://openaire-guidelines-for-literature-repository-managers.readthedocs.io/en/v4.0.0/application_profile.html

ppanero:
@rekt-hard this is the PR to review right <https://github.com/inveniosoftware/invenio-rdm-records/pull/665> ?
If so could you link it? (add closes ... statement in the description) :) thanks!

PR #665, opened 2021-06-16
schemas: add dublin core schema and json serializer

rekt-hard:
add dublin core schema
add dublin core json serializer


invenio_rdm_records/resources/serializers/dublincore/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class DublinCoreJSONSerializer(MarshmallowJSONSerializer):
+    """Marshmallow based Dublin Core serializer for records."""
+
+    def __init__(self, **options):
+        """Constructor."""
+        super().__init__(schema_cls=DublinCore, **options)
+
+    def dump_one(self, obj):
```
ppanero, line 24: I think it does not defer on implementation from the [parent class](https://github.com/inveniosoftware/flask-resources/blob/master/flask_resources/serializers/json.py#L92), I think is not needed :)

rekt-hard:
closes #663

invenio_rdm_records/resources/serializers/dublincore/__init__.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
```
ppanero, line 3: Hmm not sure if CERN is the copyright for your institution?

invenio_rdm_records/resources/serializers/dublincore/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+
+    Note: This serializer is not suitable for serializing large number of
+    records.
+    """
+
+    def __init__(self, **options):
+        """Constructor."""
+        self._schema_cls = DublinCore
+
+    def serialize_object(self, obj) -> str:
```
ppanero, line 41: It is nice to use the type hints, but we are not really using it accross Invenio... so just to make it consistent :)
```suggestion
    def serialize_object(self, obj):
```

invenio_rdm_records/resources/serializers/dublincore/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+        """Constructor."""
+        super().__init__(schema_cls=DublinCore, **options)
+
+    def dump_one(self, obj):
+        """Dump the object with extra information."""
+        json = self._schema_cls().dump(obj)
+        return json
+
+
+class DublinCoreXMLSerializer(SerializerMixin):
```
ppanero, line 30: you might need to implement `serialize_object_list` too (see mixin)... otherwise it will fail for a list of records.

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
```
ppanero, line 3: same as above, check copyright on the files you created/touched.

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Dublin Core based Schema for Invenio RDM Records."""
+
+from __future__ import absolute_import, print_function
```
ppanero, line 10: Pythong 2.x is not supported anymore 🚀  so this is not really needed

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+
+from invenio_access.permissions import system_identity
+from invenio_vocabularies.proxies import current_service as vocabulary_service
+from marshmallow import Schema, fields, missing
+
+
+class DublinCore(Schema):
```
ppanero, line 17: In general we suffix the name with `Schema` to be more clear in other places when importing the class :)
```suggestion
class DublinCoreSchema(Schema):
```

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+        """Get relations."""
+        rels = []
+
+        # Fundings
+        fundings = obj['metadata'].get('funding', [])
+        for f in fundings:
+            award = f.get('award')
+            if award:
+                rels.append(
+                    u'{0}:{1}'.format(award['scheme'], award['identifier'])
```
ppanero, line 67: nothing agains this, feel free not to change it :) I just find f-strings more friendly 👼 . Since you also used f-strings below (locations) just to be consistent use always one or the other :)
```suggestion
                    f"{award['scheme']}:{award['identifier']}"
```

I hope I'm missing a reason to use u-strings 😅

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+            i['identifier'] for i in obj['metadata'].get('identifiers', [])
+        )
+        items.extend(p['identifier'] for p in obj.get('pids', {}).values())
+
+        return items
+
+    def get_creators(self, obj):
+        """Get creators."""
+        return [
+            c['person_or_org']['name']
```
ppanero, line 53: `name` is not a required value so it might be empty, see [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/schemas/metadata.py#L63). You need to use `.get` instead

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+        for a in obj['metadata'].get('related_identifiers', []):
+            rels.append(u'{0}:{1}'.format(a['scheme'], a['identifier']))
+
+        return rels
+
+    def get_rights(self, obj):
+        """Get rights."""
+        rights = []
+        for right in obj['metadata'].get('rights', []):
+            rights.append(right['title'])
```
ppanero, line 90: `title` is not required so it might not be present, see [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/schemas/metadata.py#L204) you might need to use `.get`

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+        for add_desc in additional_descriptions:
+            descriptions.append(add_desc['description'])
+
+        return descriptions
+
+    def get_subjects(self, obj):
+        """Get subjects."""
+        metadata = obj['metadata']
+        subjects = []
+        subjects.extend(metadata.get('keywords', []))
```
ppanero, line 126: There is no attribute `keywords` in the [metadata schema](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/schemas/metadata.py#L406)

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+                s['subject']
+                for s in metadata.get('subjects', [])
+                if s.get('subject')
+            )
+        )
+        return subjects
+
+    def get_publishers(self, obj):
+        """Get publishers."""
+        return [obj['metadata']['publisher']]
```
ppanero, line 138: `publisher` is not required so it might not be present, see here you might need to use .get

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def get_contributors(self, obj):
+        """Get contributors."""
+        return [
+            c['person_or_org']['name']
```
ppanero, line 143: `name` is not required so it might not be present, see here you might need to use .get

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+            for c in obj['metadata'].get('contributors', [])
+        ]
+
+    def get_types(self, obj):
+        """Get resource type."""
+        type = obj['metadata']['resource_type']
+        return [type['id']]
```
ppanero, line 150: Not sure which is the convention, but with `id` you will be sending sort of an internal value that is `type-subtype` (e.g. [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/data/vocabularies/resource_types.yaml#L16))... You might want to send a labelled one? I'm not sure about the spec... see for example the [datacite](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/datacite/schema.py#L197) serializer

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def get_sources(self, obj):
+        """Get sources."""
+        items = []
```
ppanero, line 154: unused?
```suggestion
```

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+        return missing
+
+    def get_languages(self, obj):
+        """Get language."""
+        metadata = obj["metadata"]
+        languages = metadata.get("languages")
+
+        if languages:
+            return [language['id'] for language in languages]
```
ppanero, line 163: This is the same case as with the resource_type, but IIRC from yesterdya's standup you want to send the 3-letter encoding (which is the ID), so this is fine *I think*.

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def get_sources(self, obj):
+        """Get sources."""
+        items = []
+        return missing
+
+    def get_languages(self, obj):
+        """Get language."""
+        metadata = obj["metadata"]
+        languages = metadata.get("languages")
```
ppanero, line 160: Simpler, and consistent with the rest of the code :) 👼 
```suggestion
        languages = obj["metadata"].get("languages")
```

tests/resources/serializers/test_dublincore_serializer.py
```diff
@@ -0,0 +1,84 @@
...
+# Copyright (C) 2021 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Resources serializers tests."""
+
+from invenio_rdm_records.resources.serializers.dublincore import \
+    DublinCoreJSONSerializer, DublinCoreXMLSerializer
+
```
ppanero, line 12: It might be good to add a test for:
- list serializer? If needs implementing - see one of the above comments about it
- In the schema many of the fields would return an empty list `[]` instead of `missing`, for example subjects or descriptions. What is the expected output for the `minimal_record`. I.e. should the serialized field be an empty array or not be present?. Adding a test for it might be a good idea.

ppanero (commented):
Very nice work! A few minor comments and tiny polishing details :)

invenio_rdm_records/resources/serializers/dublincore/__init__.py
lnielsen: The correct header to add is ``# Copyright (C) 2021 Graz University of Technology.`` (this is the on user other places).

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+        return [
+            c['person_or_org']['name']
+            for c in obj['metadata'].get('creators', [])
+        ]
+
+    def get_relations(self, obj):
+        """Get relations."""
+        rels = []
+
+        # Fundings
```
lnielsen, line 61: There's a special syntax to use for grants which is not trivial, so I would leave out these funding links for now. Also, it's not possible via UI yet to add it (only supported in the REST API

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+        for a in obj['metadata'].get('alternate_identifiers', []):
+            rels.append(u'{0}:{1}'.format(a['scheme'], a['identifier']))
+
+        # Related identifiers
+        for a in obj['metadata'].get('related_identifiers', []):
+            rels.append(u'{0}:{1}'.format(a['scheme'], a['identifier']))
+
+        return rels
+
+    def get_rights(self, obj):
```
lnielsen, line 86: This one should also add the access level as defined in - https://guidelines.openaire.eu/en/latest/literature/field_accesslevel.html 

You should be able to get the access level from ``record.access.status`` and should be one-of ``open``, ``embargoed``, ``restricted`` and ``metadata-only`` (i.e. closed)

lnielsen: If the access is embargoed, we should also add:

```xml
<dc:date>info:eu-repo/date/embargoEnd/2015-12-31</dc:date>
```

See https://guidelines.openaire.eu/en/latest/literature/field_embargoenddate.html

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+            if license_url:
+                rights.append(license_url)
+
+        return rights
+
+    def get_dates(self, obj):
+        """Get dates."""
+        dates = [obj["metadata"]["publication_date"]]
+
+        if obj['access']['status'] == 'embargoed':
```
lnielsen, line 102: There's a special encoding for dates embargo dates: https://guidelines.openaire.eu/en/latest/literature/field_embargoenddate.html

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -0,0 +1,189 @@
...
+            dates.append(obj['access']['embargo']['until'])
+
+        return dates or missing
+
+    def get_descriptions(self, obj):
+        """Get descriptions."""
+        metadata = obj['metadata']
+        descriptions = []
+
+        description = metadata.get("description")
```
lnielsen, line 112: Descriptions may contain HTML we need to strip them from HTML.

Something like this should do the job:

```python
import bleach
bleach.clean(..., strip=True, strip_comments=True, tags=[], attributes=[], styles=[])

Comment (lnielsen) [reply invenio_rdm_records/resources/serializers/dublincore/schema.py:112]: THere's also this field: https://github.com/inveniosoftware/marshmallow-utils/blob/master/marshmallow_utils/fields/strippedhtml.py#L16

File: invenio_rdm_records/resources/serializers/dublincore/schema.py (hunk 14 above)
Comment (lnielsen) [reply invenio_rdm_records/resources/serializers/dublincore/schema.py:150]: The resource type vocabulary should provide a mapping to the eu-repo vocabulary, but I don't think this is yet the case, so perhaps this is best we can do for now:

https://guidelines.openaire.eu/en/latest/literature/field_publicationtype.html

File: invenio_rdm_records/resources/serializers/dublincore/schema.py [hunk 23]
```diff
@@ -0,0 +1,189 @@
...
+        """Get language."""
+        metadata = obj["metadata"]
+        languages = metadata.get("languages")
+
+        if languages:
+            return [language['id'] for language in languages]
+
+        return missing
+
+    def get_locations(self, obj):
```
lnielsen, line 167: I tihnk for now we should leave out locations. It's not yet there properly, and no support in deposit form for it.

invenio_rdm_records/resources/serializers/dublincore/schema.py
rekt-hard: for organizational creators it is a required field, see [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/4fa591355cf8c6a94cec082e72ab8848b86e3ba2/invenio_rdm_records/services/schemas/metadata.py#L86)
for personal creators, the name field will be updated, see [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/4fa591355cf8c6a94cec082e72ab8848b86e3ba2/invenio_rdm_records/services/schemas/metadata.py#L98)
i think there will always be a value for `name`

invenio_rdm_records/resources/serializers/dublincore/schema.py
rekt-hard: same as [here](https://github.com/inveniosoftware/invenio-rdm-records/pull/665#discussion_r656150220)

invenio_rdm_records/resources/serializers/dublincore/schema.py
rekt-hard: should we use a labelled type (openaire maybe?) until the eu-repo vocabulary is added?

invenio_rdm_records/resources/serializers/dublincore/__init__.py
ntarocco: Let's use type hints!!!! Revolution! :D

ppanero: I'm up for it!

invenio_rdm_records/resources/serializers/dublincore/__init__.py
rekt-hard: implemented now. there is also a test case for it

tests/resources/serializers/test_dublincore_serializer.py
rekt-hard: added test cases for minimal records and list serializer

invenio_rdm_records/resources/serializers/dublincore/schema.py
rekt-hard: removed fundings for now. added FIXME note

invenio_rdm_records/resources/serializers/dublincore/schema.py
rekt-hard: commented it out for now and added FIXME note

PR #667, opened 2021-06-16
schemas: adds resource_type props to the record index

Glignos:


Glignos:
Depends on https://github.com/inveniosoftware/invenio-records/pull/260

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -92,7 +92,7 @@ class ResourceTypeL10NSchema(Schema):
 
     id = fields.String()
     title = L10NString(data_key='title_l10n')
-
+    props = fields.Dict(dump_only=True)
```
lnielsen, line 95: Do we need them here?

invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -140,12 +140,12 @@ class VocabularySchema(Schema):
 
     id = SanitizedUnicode(required=True)
     title = fields.Dict(dump_only=True)
+    props = fields.Dict(dump_only=True)
```
lnielsen, line 143: Is the reason they are needed here, because otherwise you can create the facet over them?

Glignos: No, actually that was a leftover, since we expect these fields only to be indexed and we don't need to persist them in the DB I will remove them from the schema.

invenio_rdm_records/resources/serializers/ui/schema.py
Glignos: ditto

PR #675, opened 2021-06-18
subjects: load from 1-app_data 2-extenions 3-this

fenekku:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/677

### DIFF
--- a/invenio_rdm_records/fixtures/demo.py
+++ b/invenio_rdm_records/fixtures/demo.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2019 CERN.
-# Copyright (C) 2019 Northwestern University.
+# Copyright (C) 2019-2021 CERN.
+# Copyright (C) 2019-2021 Northwestern University.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
@@ -35,12 +35,25 @@ class CachedVocabularies:
 
     @classmethod
     def _read_vocabulary(cls, vocabulary):
-        dir_ = Path(__file__).parent
+        """Iterate over records of a vocabulary."""
+        # Simplification: we don't consider extensions when generating demo
+        #                 data for now
+        filepath = Path("./app_data") / "vocabularies.yaml"
+        if filepath.exists():
+            app_fixture = VocabulariesFixture(
+                system_identity,
+                filepath
+            )
+            app_vocabulary_iter = (
+                app_fixture.get_records_by_vocabulary(vocabulary)
+            )
+            if any(True for _ in app_vocabulary_iter):
+                return app_fixture.get_records_by_vocabulary(vocabulary)
 
         return VocabulariesFixture(
             system_identity,
-            [Path("./app_data"), dir_ / "data"],
-            "vocabularies.yaml",
+            # We know this file exists
+            Path(__file__).parent / "data" / "vocabularies.yaml",
         ).get_records_by_vocabulary(vocabulary)
 
     @classmethod
@@ -58,22 +71,6 @@ class CachedVocabularies:
 
         random_id = random.choice(cls._resource_type_ids)
         return {"id": random_id}
-
-    @classmethod
-    def fake_subjects(cls):
-        """Generate random subjects."""
-        if not cls._subject_ids:
-            subjects = cls._read_vocabulary("subjects")
-
-            for subj in subjects:
-                cls._subject_ids.append(subj["id"])
-
-        if not cls._subject_ids:
-            return []
-
-        n = random.choice([0, 1, 2])
-        random_ids = random.sample(cls._subject_ids, n)
-        return [{"id": i} for i in random_ids]
 
     @classmethod
     def fake_language(cls):
@@ -158,7 +155,7 @@ def create_fake_record():
             }],
             "publisher": "InvenioRDM",
             "publication_date": fake_edtf_level_0(),
-            "subjects": CachedVocabularies.fake_subjects(),
+            "subjects": [],  # Too complex to generate for now
             "contributors": [{
                 "person_or_org": {
                     "family_name": fake.last_name(),
--- a/invenio_rdm_records/fixtures/vocabularies.py
+++ b/invenio_rdm_records/fixtures/vocabularies.py
@@ -10,8 +10,11 @@
 
 import csv
 import json
+from collections import defaultdict
 from os.path import splitext
+from pathlib import Path
 
+import pkg_resources
 import yaml
 from flask import current_app
 from invenio_vocabularies.proxies import current_service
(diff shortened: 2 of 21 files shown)


fenekku:
Test error seems to have to do with citation string serializer and facet aggregations :shrug:

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -105,20 +108,157 @@ def __iter__(self):
             yield from create_iterator(data_file)
 
 
+#
+# Exceptions
+#
+class ConflictingFixturesError(Exception):
```
ppanero, line 114: minor: usually we create an `errors.py` for this type of classes

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -105,20 +108,157 @@ def __iter__(self):
...
+        """Constructor."""
+        message = "\n".join(errors)
+        super().__init__(message)
+        self.errors = errors
+
+
 #
 # Fixture
 #
+class VocabulariesFixtures:
```
ppanero, line 127: naming: to be honest it is quite confusing to have `VocabulariesFixtures` and `VocabulariesFixture`. It makes it quite hard to discern what's the different - I'm actually not sure I understand 😅 

- The first one allows to load (potentially) from external packages. being the priority `app_data`, `packages` then `rdm-records` (e.g. `data`)
- The second one loads as before: first `app_data` then `data`

In addition, `VocabulariesFixtures` uses `VocabularyFixture` to load from `data` at the end. But the loading from `app_data` seems a bit duplicated since we need to put the packages/extensions in the middle of the loading right?

I'm not sure if this would be sensible, and I *do not want* to make you refactor a lot... But what about just 1 class that takes a "loading order" and has components to do so... Somehow similar to what we do in services with the `components`... So a `PackageFixture`, `AppDataFixture`, etc. and then there is just an order and they have a common interface. In some sort of "common" structure we can do the duplication check by passing it to the current loading component.

That said, maybe I totally misunderstood the code... hope not 😅

ppanero (commented):
the `_A`, `_B` etc is a MeSH thing right? This is a personal feeling, but I dislike the capitals in the file names and so on... Could the examples not have them? - Feel more than free to ignore this 😅

invenio_rdm_records/fixtures/vocabularies.py
fenekku: The idea was to re-use the existing code as much as possible because we are only adding another level on top :sweat_smile: . tldr; I think changing `VocabulariesFixture._search_paths/_filename` to `filepath` would be a sensible change that diffuses some of the complexity since `VocabulariesFixture` really only represents a single `vocabularies.yaml` file and is the same functionality as each of the `PackageFixture`, `AppDataFixture` mentioned above.

Here is what we are trying to do:

Load fixtures in general (Vocabularies fixtures and others like users fixtures). Especially for vocabularies it means: loading vocabularies fixtures in the priority order app_data, extensions, data/ (package) BUT allowing for this priority to be at the subject type level rather than the vocabulary level for subjects (or any vocabulary with subtypes). It's that wrinkle that makes things a little strange I think. We need to revisit for the use case of updating an instance with a new voabulary anyway though.

Vocabularies fixtures are very layered:
 - there is the vocabuaries.yaml file with vocabularies in it
 - each vocabulary entry in that file has 1 or more data-files
 - each datafile has 0 or more records
so having a dedicated class for them is sensible.

invenio_rdm_records/fixtures/vocabularies.py
fenekku: Perhaps let's see if another exception is needed and create the file then?

fenekku:
The `A / B` is not MeSH. It was but just a way to denote variations. You would rather have: `mock_module_a`, `mock_module_b` and use `a` / `b` in the vocabulary names? The `A-D...` is for identifier namespacing. I don't mind changing it, but I wonder if there is a deeper cause of confusion here that I could address?

ppanero:
> but I wonder if there is a deeper cause of confusion here that I could address?

Not really, was just trying to understand things 😅. I saw in the example something like A-... and then `Abdomen`... Thanks for the answer :)

invenio_rdm_records/fixtures/vocabularies.py
ppanero: I see... It is indeed complex. I will try to understand the code more in depth... I find it a bit complex, but it is a complex task 🤷‍♂️ . I dont want to block you, so maybe just a more descriptive docstring to differentiate between `VocabulariesFixture` and `VocabulariesFixtures` 👼

fenekku:
I've added more explanation in the classes' docstrings + made `VocabulariesFixture` just take a path to its vocabulary file (versus different directories to look into). Tests are passing locally. If they pass on CI by end of day I will merge.

I will revisit to change for array of subtypes in another PR.

fenekku:
Hmmm locally tests pass and this case [3.6, pypi, postgresql9, elasticsearch6] was the one failing only on CI before (CSL serializer and nesting). I think @Glignos is supposed to merge/release something that would help?

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -106,84 +109,249 @@ def __iter__(self):
...
+
+    This class' responsibility is to load different vocabularies in
+    this priority order: app_data, then extensions and then this package.
+
+    Earlier vocabularies in this hierarchy are chosen first. But in the case
+    of subvocabularies (e.g. subject types), having loaded a subvocabulary
+    before, shouldn't prevent loading another one. Only if the same
+    subvocabulary is encountered again should it be ignored.
+
+    Concretely, having loaded MeSH subjects shouldn't prevent loading FAST
```
ppanero, line 138: sanity check: imagine an extension is loading some subjects (e.g. mesh) and the user doesn't want it. the way to exclude it is by not installing `invenio-mesh`, and since there is no default (at package level) the user can fully customize his/her choice by modifying the `vocabularies.yaml` at instance (cookiecutter) level and/or adding new extensions.

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -106,84 +109,249 @@ def __iter__(self):
...
+        """
         self._identity = identity
+        # Path("./app_data") assumes app_data is in current working directory
+        self._app_data_folder = app_data_folder or Path("./app_data")
+        self._pkg_data_folder = (
+            pkg_data_folder or Path(__file__).parent / "data"
+        )
+        self._filename = filename
+        self._delay = delay
+        # self._ids = set()
```
ppanero, line 162: remove?
```suggestion
```

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -106,84 +109,249 @@ def __iter__(self):
...
+
+        identity: Identity to use when loading
+        app_data_folder: Path object to instance data folder
+        pkg_data_folder: Path object to this package's data folder.
+                         Defaults to `./data` and really only changeable for
+                         tests.
+        filename: vocabularies filename to check at each location
+        """
         self._identity = identity
+        # Path("./app_data") assumes app_data is in current working directory
```
ppanero, line 155: remove?
```suggestion
```

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -106,84 +109,249 @@ def __iter__(self):
...
-            with open(filepath) as fp:
-                data = yaml.safe_load(fp) or {}
-                for id_, entry in data.items():
-                    if isinstance(entry["data-file"], dict):
-                        entry["data-file"] = {
-                            k: path / v for k, v in entry["data-file"].items()
-                        }
-                    else:
-                        entry["data-file"] = path / entry["data-file"]
+        but Invenio is much invested in pkg_resources (``entry_points``
```
ppanero, line 173: maybe is worth bringing this up so it can be evaluated by architects and potentially worked on later in the future?

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -106,84 +109,249 @@ def __iter__(self):
...
+        if errors:
+            raise ConflictingFixturesError(errors)
+
+        # Then load
+        for module in extensions:
+            directory = Path(module.__file__).parent
+            filepath = directory / self._filename
+            self.load_vocabularies(filepath)
+
+    def peek_vocabularies(self, filepath):
```
ppanero, line 235: what about something like `read_vocabularies_locations` or even `read_locations`? I also feel that every function in this class has `vocabularies` on its name and it's sort of understood from being instide a class that has the name `Vocabularies` on it... e.g. `File` has a `.read`, `.write` method not `.read_file`, `.write_file`.

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -106,84 +109,249 @@ def __iter__(self):
...
+    def __init__(self, errors):
+        """Constructor."""
+        message = "\n".join(errors)
+        super().__init__(message)
+        self.errors = errors
+
+
+#
+# Fixture
+#
```
ppanero, line 126: What about `MultiLocationVocabulariesFixture` and below `FileVocabulariesFixture`... then the interface inside can be sort of similar, being `load` the most important method to the users of this classes.

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -106,84 +109,249 @@ def __iter__(self):
...
         """Load the fixture.
 
-        Fixtures found sooner in self._search_paths have priority.
+        ignore: iterable of ids to ignore
+
+        For subjects (or any vocabulary with a dict of data-file), the id
+        that counts is ``<vocabulary id>.<dict key>``.
         """
-        ids = set()
+        ids = set().union(ignore) if ignore else set()
```
ppanero, line 308: `ignore` is always called with `_loaded_vocabularies` which is already a set, right? (this could be so easy with typing 😅 😂 )
```suggestion
        ids = ignore if ignore else set()
```

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -106,84 +109,249 @@ def __iter__(self):
...
 
-        for id_, entry in self._open_vocabularies():
+        for id_, entry in self.read():
+            # TODO: We need to carefully think about what the workflow for
```
ppanero, line 311: What would be an action for this TODO? e.g. is not clear to me what's expected of someone finding this... just wondering if it can be removed and is a bigger taks i.e. issue.

I think tend to forget TODO/FIXME in general...

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -106,84 +109,249 @@ def __iter__(self):
...
+            #       adding new subjects later is and then modify the code to
+            #       allow for it.
             if id_ not in ids:
                 try:
-                    self.load_vocabulary(id_, entry)
+                    self.create_vocabulary_type(id_, entry)
                 except IntegrityError:
+                    # Can only get here when running the code multiple times
```
ppanero, line 318: Agree, nice

ppanero (approved):
A few minors and sanity checks... mostly on naming, I am not a fan of having two classes with just one `s` at the end as difference. So I tried to convey their difference: *one loads form file the other from multiple places*.

- `MultiLocationVocabulariesFixture`
- `FileVocabulariesFixture`

WDYT?

invenio_rdm_records/fixtures/demo.py
```diff
@@ -35,12 +35,25 @@ class CachedVocabularies:
...
+        filepath = Path("./app_data") / "vocabularies.yaml"
+        if filepath.exists():
+            app_fixture = VocabulariesFixture(
+                system_identity,
+                filepath
+            )
+            app_vocabulary_iter = (
+                app_fixture.get_records_by_vocabulary(vocabulary)
+            )
+            if any(True for _ in app_vocabulary_iter):
```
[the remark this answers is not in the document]
fenekku: :dart:  right on the money

invenio_rdm_records/fixtures/demo.py
```diff
@@ -158,7 +155,7 @@ def create_fake_record():
             }],
             "publisher": "InvenioRDM",
             "publication_date": fake_edtf_level_0(),
-            "subjects": CachedVocabularies.fake_subjects(),
+            "subjects": [],  # Too complex to generate for now
```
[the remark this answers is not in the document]
fenekku: Yes, I will mention it at standup again. The demo instance could also have an `app_data/` with some small number of subjects so that when people trying out the online demo can add a subject to a record.

invenio_rdm_records/fixtures/vocabularies.py
fenekku: > the way to exclude it is by [...]

Yes there are a couple way for the administrator to be in full control in this situation. They can either:

1- not install invenio-subjects-mesh in the first place (there is no default MeSH subjects so they won't get any)
2- define in `app_data/vocabularies.yaml` a `subjects` `mesh` datafile that points to an empty yaml file. The loading priority will make sure that the later defined one is ignored. Obviously 1- is preferable. 

They can install other extensions to get other subjects or again just stick a specific subjects yaml file in their `app_data/`(e.g. if they don't want the overhead of creating an extension - not something shareable with other instances).

invenio_rdm_records/fixtures/vocabularies.py
fenekku: It applies to the line below it, so would keep. Basically, `Path('app_data/')` relies on wider knowledge of the structure of our application implicitly: it knows that when `invenio-cli rdm-records fixtures` is called, it is called within the main directory of the instance, where there is an `app_data/` folder. For something so tucked away in the code base it's good to flag it.

invenio_rdm_records/fixtures/vocabularies.py
fenekku: How about:
 
`VocabulariesFixtures` -> `PrioritizedVocabulariesFixtures`

and keeping `VocabulariesFixture` the same?

invenio_rdm_records/fixtures/vocabularies.py
fenekku: > this could be so easy with typing

Yes *laughs because otherwise would cry...*

Maybe `set().union()` was too cute here... It allows for passing any iterable to `load` not just sets (i.e. a list of ids to ignore could be passed). Can be changed later if too much.

invenio_rdm_records/fixtures/vocabularies.py
fenekku: Yeah... I am keeping track of those in my personal notes but I should really create issues for them. I created the following:

- Discussion about other use cases around subject loading: https://github.com/inveniosoftware/invenio-rdm-records/issues/687 . I will bring this up in the chat
- A concrete issue about using the content of database to enforce priority (needs other PR in invenio-vocabularies first): https://github.com/inveniosoftware/invenio-rdm-records/issues/686

invenio_rdm_records/fixtures/vocabularies.py
ppanero: What about `PrioritizedVocabulariesFixture`, vocabularies is already plural 😊

ppanero: Discussed "IRL", stays as `PrioritizedVocabulariesFixtures`

invenio_rdm_records/fixtures/vocabularies.py
ppanero: > Yes, haha laughs because otherwise would cry...

😂 

I see, is that expected? or can we assume is receiving a set always? fine either way was just a minor :)

ppanero:
Thanks for all the answers 🚀 merge when you consider 😃

fenekku:
finally it passed :tada: . merging!

PR #672, opened 2021-06-18
resources

Pineirin:
refactor URL parameters parsing for style and locale

PR #682, opened 2021-06-23
CSL

Pineirin:
escape properly the apostrophes when cleaning the citation string


ntarocco:
The comment in the previous PR https://github.com/inveniosoftware/invenio-rdm-records/pull/672#discussion_r656866255 should be solved here

lnielsen (approved):

Issue #680, opened 2021-06-23
vocabularies: integrate affiliations with rdm data model


lnielsen:
**Data layer**

- [x] Define relation in CommonFIeldsMixin.
    - [x] Determine which fields to dump in ES index.
- [x] JSONShema: Do not modify an already released schema, create new version instead.
- [x] JSONSchema: Define or review existing property
    - List or single?
    - Mixed linked with non-linked?
- [x] Mappings adapt to JSONSchema and dumped fields.

**Service layer**

- [x] Schema: Fix metadata schema.
- [x] Facets: Do we need facets?

**Presentation layer**

- [x] Fix REST API serializations
- [x] Fix other serialization formats.

**Tests**

- [x] Add tests that record linking is wokring as expected (e.g. from service layer).
    - Bad and good data.
- [x] Add REST API tests for serialization, facets
- [x] Ensure that I18N is tested properly if required.

ppanero:
Discusses IRL
> Ensure that I18N is tested properly if required.

No i18n, UI will display `name`.

PR #684, opened 2021-06-23
vocabularies: integrate affiliations with rdm data model

ppanero:
requires https://github.com/inveniosoftware/invenio-records-resources/pull/271
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/680

### DIFF
--- a/invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
@@ -1,0 +1,431 @@
+{
+  "affiliation": {
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
+      "id": {
+        "$ref": "local://definitions-v1.0.0.json#/identifier"
+      },
+      "name": {
+        "type": "string"
+      }
+    }
+  },
+  "affiliations": {
+    "type": "array",
+    "items": {"$ref": "#/affiliation"},
+    "uniqueItems": true
+  },
+  "agent": {
+    "description": "An agent (user, software process, community, ...).",
+    "oneOf": [
+      {
+        "$ref": "#/user"
+      }
+    ]
+  },
+  "dateType": {
+    "description": "Type of the date.",
+    "type": "string"
+  },
+  "description_type": {
+    "type": "object",
+    "additionalProperties": false,
+    "description": "A description type.",
+    "properties": {
+      "id": {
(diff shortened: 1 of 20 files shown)


tests/conftest.py
```diff
@@ -533,22 +517,62 @@ def subject_v(app, subject_type):
...
+def affiliations_v(app):
+    """Affiliations vocabulary record."""
+    affiliations_service = (
+        current_service_registry.get("rdm-affiliations")
+    )
+    aff = affiliations_service.create(system_identity, {
+        "id": "cern",
+        "name": "CERN",
+        "acronym": "CERN",
+        "identifiers": [{
```
jrcastro2, line 530: I think `identifiers` and `acronym` should not be there ?

jrcastro2 (commented):
LGTM!
There are some minor fixes and I am not sure I understand the jsonschemas changes with the new definitions-v2.0.0, it might be I am missing context so feel free to ping and chat IRL if it's faster

invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
```diff
@@ -0,0 +1,424 @@
+{
```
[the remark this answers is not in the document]
ppanero: discussed IRL, the usage of the definitions coming from records-resources is considered a breaking change. This is the same reason applied to the other definitions comments :)

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -60,13 +61,23 @@ def get_name_identifiers(self, obj):
 
         return serialized_identifiers
 
+    def _read_affiliation(self, id_):
+        """Retrieve resource type record using service."""
+        affiliations_service = (
+            current_service_registry.get("rdm-affiliations")
+        )
+        return affiliations_service.read(id_, system_identity)._record
```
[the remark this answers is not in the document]
ppanero: Nice catch 👀

tests/conftest.py
ppanero: Yes because affiliations are full records, with a structure a bit different from a generic vocabulary, and in this case we are creating it, then in the `full_records` we only set the `id`.

jrcastro2: I see, makes sense, thanks!

jrcastro2 (approved):

PR #690, opened 2021-06-24
datacite 

Pineirin:
Allow partial customization in DOI generation

### DIFF
--- a/invenio_rdm_records/services/pids/providers/datacite.py
+++ b/invenio_rdm_records/services/pids/providers/datacite.py
@@ -12,6 +12,7 @@ import json
 from datacite import DataCiteRESTClient
 from datacite.errors import DataCiteError
 from flask import current_app
+from invenio_pidstore.errors import PIDAlreadyExists, PIDDoesNotExistError
 from invenio_pidstore.models import PIDStatus
 
 from invenio_rdm_records.resources.serializers import DataCite43JSONSerializer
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -87,19 +87,22 @@ def _log_errors(errors):
...
+    def _generate_doi(self, record, **kwargs):
         """Generate a unique DOI."""
         prefix = self.client.prefix
-        suffix = self.generate_suffix(record, self.client, **kwargs)
-        return f"{prefix}/{suffix}"
+        id = self._generate_id(record, **kwargs)
+        if current_app:
+            format = current_app.config['RDM_RECORDS_DOI_DATACITE_FORMAT']
+            return format.format(prefix, id)
+        return "{prefix}/datacite.{id}"
```
ntarocco, line 105: I think we should be more flexible: `RDM_RECORDS_DOI_DATACITE_FORMAT` should have a better name and should accept a `string` or a func (a callable):
1. if string, it mean it is a format. You have to document very well what pattern can be inserted in the string, {prefix} {id}. As a developer, I cannot insert {ciao}
2. if callable, it should be used instead of the `_generate_doi` function, which has been designed exactly for this use case. Example:

```
def myfunc(record, **kwargs):
    ... my implementation ...
    ... for example: ....
    return "{prefix}/{record['resource_type']}.{record.pid.pid_value}"

RDM_RECORDS_DOI_DATACITE_FORMAT = myfunc
```

PR #704, opened 2021-07-02
serializers: DataCite serializer fixes

lnielsen:
Missing:

- [ ] Creator/contributors affiliations should be optimized to use ``read_many`` (reads from ES instead of DB, and sends a single query vs one per affiliation - think of CERN paper with 5000 people with affiliations).
- [x] Subjects should be migrated to efficient querying.

### DIFF
--- a/invenio_rdm_records/resources/serializers/datacite/schema.py
+++ b/invenio_rdm_records/resources/serializers/datacite/schema.py
@@ -211,71 +211,59 @@ class DataCite43Schema(Schema):
         props = {h["id"]: h["props"] for h in res.hits}[id_]
         return props
 
-    def _read_resource_type(self, id_):
-        """Retrieve resource type record using service."""
-        return vocabulary_service.read(
-            ('resource_types', id_),
-            system_identity
-        )
-
-    def _read_title_type(self, id_):
-        """Retrieve title type record using service."""
-        return vocabulary_service.read(
-            ('title_types', id_),
-            system_identity
-        )
-
-    def _read_description_type(self, id_):
-        """Retrieve description type record using service."""
-        return vocabulary_service.read(
-            ('descriptiontypes', id_),
-            system_identity
-        )
-
-    def _read_language(self, id_):
-        """Retrieve resource type record using service."""
-        lang = vocabulary_service.read(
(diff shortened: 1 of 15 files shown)


lnielsen:
Subjects are missing to migrated

jrcastro2:
Missing creator/contributors affiliations

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -201,71 +197,68 @@ class DataCite43Schema(Schema):
...
-            ('resource_types', id_),
-            system_identity
-        )
-
-    def _read_title_type(self, id_):
-        """Retrieve title type record using service."""
-        return vocabulary_service.read(
-            ('title_types', id_),
-            system_identity
+    def _map_type(self, vocabulary, fields, id_):
```
ppanero, line 200: This pattern with the props is used also before with the `Contributor.role`, `locations` below... maybe we can extract it a bit more? MIght not be worth if the code becomes too messy since there is only a few appearances...

ppanero (approved):
few minors, can be merged once its rebase if blocking

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -110,29 +110,25 @@ class CreatorSchema43(PersonOrOrgSchema43):
...
+        if not role:
+            return missing
 
-        return data
+        role_vocabulary = vocabulary_service.read_all(
+            system_identity,
+            ['id', 'props'],
+            "contributorsroles"
+        )
+        roles_props = {h["id"]: h["props"] for h in role_vocabulary.hits}
```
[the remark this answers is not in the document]
jrcastro2: Yes, it's cached, but I guess we can add it his anyway, thanks!

invenio_rdm_records/resources/serializers/datacite/schema.py
jrcastro2: For now whe have the map_type that is used widely and also the one used for `contributors.role`, I would keep it as it for now, and refactor it, if needed when we finish subjects/affiliations

Issue #703, opened 2021-07-02
serialisers: datacite JSON and XML needs a full review


lnielsen:
Here's examples of possible current issues:

```
-  <identifier identifierType="DOI">/datacite.158m7-jk782</identifier> # <---- missing prefix
-  HTML tags in description field.
- Publisher set to default to InvenioRDM. Seems wrong and normally there's a default for the institution i think
```

ppanero:
> -  <identifier identifierType="DOI">/datacite.158m7-jk782</identifier> # <---- missing prefix

Most likely due to lack of prefix configured in the invenio.cfg file? Tests are passing including the prefix.

> Publisher set to default to InvenioRDM. Seems wrong and normally there's a default for the institution i think

There is no default in the `services.schemas.metadata`, nor in the datacite serialization.... It is set to `InvenioRDM` demo data and the confests record fixture. We support this value from the deposit form.

> -  HTML tags in description field.

Fixed in the `_get_text` function. Using the `StrippedHTML` field from marshmallow utils.

PR #714, opened 2021-07-07
serializers: minor fixes to datacite json and add tests for xml

ppanero:
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/703
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/727

Differences I can observe in comparison with the full example of [DataCite 4.3 XML](https://github.com/inveniosoftware/datacite/blob/master/tests/data/datacite-v4.3-full-example.xml), which are not yet implemented are:
- All languages should use the 2 letter encoding, instead of 3. This should be fixed for the JSON schema too.
- We do not have alternate identifiers, we put them in `identifiers`. This would differ from the JSON schema. We should:
    - pids --> identifiers
    - identifiers --> alternate identifiers

### DIFF
--- a/tests/resources/serializers/test_datacite_serializer.py
+++ b/tests/resources/serializers/test_datacite_serializer.py
@@ -16,7 +16,8 @@ from invenio_records_resources.proxies import current_service_registry
 from invenio_vocabularies.proxies import current_service as vocabulary_service
 from invenio_vocabularies.records.api import Vocabulary
 
-from invenio_rdm_records.resources.serializers import DataCite43JSONSerializer
+from invenio_rdm_records.resources.serializers import \
+    DataCite43JSONSerializer, DataCite43XMLSerializer
 
 
 @pytest.fixture(scope="function")
@@ -97,7 +98,7 @@ def title_type_v(app, title_type):
         "id": "subtitle",
         "props": {
             "datacite": "Subtitle"
-         },
+        },
         "title": {
             "en": "Subtitle"
         },
@@ -125,6 +126,9 @@ def running_app(
 
 def test_datacite43_serializer(running_app, full_record):
     """Test serializer to DataCite 4.3 JSON"""
+    # for HTML stripping test purposes
+    full_record["metadata"]["description"] = "<h1><p>Test</p></h1>"
+
     expected_data = {
         "types": {
             "resourceTypeGeneral": "Image",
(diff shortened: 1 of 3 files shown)


ppanero:
> All languages should use the 2 letter encoding, instead of 3. This should be fixed for the JSON schema too.

BCP 47 includes ISO 639-3 so we are good.

ppanero:
> We do not have alternate identifiers, we put them in identifiers. This would differ from the JSON schema. We should:
pids --> identifiers
identifiers --> alternate identifiers

Not needed, is taken care by datacite, [ref](https://github.com/inveniosoftware/datacite/blob/master/datacite/schema43.py#L58)

lnielsen:
Other issues I've seen:

1) Dates should also serialize the embargo date.
Screenshot 2021-07-07 at 15 02 19

ppanero:
Rebased on latest master (There were conflicts) and added the content of https://github.com/inveniosoftware/invenio-rdm-records/pull/731 to close https://github.com/inveniosoftware/invenio-rdm-records/issues/727

tests/resources/serializers/test_datacite_serializer.py
```diff
@@ -257,3 +261,88 @@ def test_datacite43_serializer(running_app, full_record):
...
+        "  </relatedIdentifiers>",
+        "  <sizes>",
+        "    <size>11 pages</size>",
+        "  </sizes>",
+        "  <formats>",
+        "    <format>application/pdf</format>",
+        "  </formats>",
+        "  <version>v1.0</version>",
+        "  <rightsList>",
+        "    <rights rightsURI=\"https://creativecommons.org/licenses/by/4.0/\" rightsIdentifierScheme=\"spdx\" rightsIdentifier=\"cc-by-4.0\">Creative Commons Attribution 4.0 International</rights>",  # noqa
```
lnielsen, line 317: The tests are failing here however. I saw another PR that made a change the record ``rights`` property, thus that's likely the cause. However, the string above is correct, and we should serialize as above, so we'll need to check how to fix it.

lnielsen (approved):
Besides the "rigths" field issue, all is good.

tests/resources/serializers/test_datacite_serializer.py
ppanero: It was modified to make tests passed. @jrcastro2 in your PR they should pass with 
```
 "  <rightsList>",
"    <rights rightsURI=\"https://creativecommons.org/licenses/by/4.0/\" rightsIdentifierScheme=\"spdx\" rightsIdentifier=\"cc-by-4.0\">Creative Commons Attribution 4.0 International</rights>",  # noqa
"  </rightsList>",
```
You might need to add something to the conftests values maybe? Worth to add a custom one too?

PR #730, opened 2021-07-12
I18n: removing f-strings in order to extract messages

mb-wali:
Todo:
* create a new resource for this module in transifex
* push the `.pot` file to created resource on transifex

### DIFF
--- a/invenio_rdm_records/services/pids/providers/base.py
+++ b/invenio_rdm_records/services/pids/providers/base.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2021 CERN.
+# Copyright (C) 2021 Graz University of Technology.
 #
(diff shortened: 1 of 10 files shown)


invenio_rdm_records/services/pids/providers/base.py
```diff
@@ -211,7 +212,7 @@ def validate(
                 )
                 return False, [
                     _(
-                        f"{self.pid_type}:{identifier} is already registered" +
+                        "{self.pid_type}:{identifier} is already registered" +
```
zzacharo, line 215: I think this won't work...I think the goal is to replace f strings with the `'{myvar}'.format(myvar='foo')` format....

mb-wali: yes, i just got it.

zzacharo (approved):

PR #733, opened 2021-07-12
vocabularies: fix subjects and affiliations registration

ppanero:
closes https://github.com/inveniosoftware/invenio-vocabularies/issues/66

Both affiliations and subjects were being installed without potential changes in the config since it was being taken directly from the `record_type` and not from the corresponding class.

### DIFF
--- a/invenio_rdm_records/ext.py
+++ b/invenio_rdm_records/ext.py
@@ -13,10 +13,14 @@ from flask_babelex import _
 from flask_principal import identity_loaded
 from invenio_records_resources.resources.files import FileResource
 from invenio_records_resources.services import FileService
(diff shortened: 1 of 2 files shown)


ppanero:
@fenekku Im fine with closing this but as long as we include the full fix... otherwise:
- There is no point on exposing the services/resources as such
- We might end up in not realizing this is happining in the future

Just for consistency :)

fenekku:
I snuck the fix in https://github.com/inveniosoftware/invenio-rdm-records/pull/734 so closing this.

ppanero:
The test was not added as far as I can see... Maybe is not so important to test it... just wanted to make sure the way the links are sent back (urlencoded) was resolving properly on the ID (i.e. we encode-decode properly)

Issue #687, opened 2021-06-23
[question][meta] Define the installation of subjects workflow


fenekku:
[Now](https://github.com/inveniosoftware/invenio-rdm-records/pull/675) that extensions can be installed to provide subjects, we need to think about what the user experience for administrators installing a subject is like and what the use cases we need to make sure we cover. This in turn will lead to futher work on refining how subjects are installed.

As it stands, a subject extension should be added to the instance via 
- pipfile and will be installed when `invenio-cli install` is run
- pipenv install 
- invenio-cli packages install

The subjects will be loaded when 

1- `invenio-cli services setup` is run or 
2- when `invenio rdm-records fixtures` is run

But the current code  assumes it is run in  1- so if subjects already exist in the database, there will be some issues when running 2- at a later point in the lifecycle of the instance. So some tweaks need to be done: see this issue https://github.com/inveniosoftware/invenio-rdm-records/issues/686 . 

What about updating a subject? (e.g. MeSH terms are updated every year at least) What is the story there?  What other lifecyle cases do wee need to take into account?

PR #734, opened 2021-07-12
subjects: account for pre-existing vocabularies when loading [+]

### LINKED ISSUE #686: subjects: load new structure
**EDIT 2021-07-09**

Load subjects into separate table/subject.

prerequisite: https://github.com/inveniosoftware/invenio-vocabularies/pull/55

- [x] Change dict of data-file to an array of dicts so that each subject type can have metadata in the yaml file that will be loaded in db
- [x] When loading subjects, priority is enforced by relying on content of yaml files only and not on database content. Database content should be used, so that running `invenio rdm-records fixtures` again can execute correctly (ignore previously existing database subjects).

fenekku:
- depends on https://github.com/inveniosoftware/invenio-vocabularies/pull/67
- integrate new separate subjects vocabulary (invenio-vocabularies 0.7.8)
- rely on database content before loading vocabulary
- closes #686 (and #687 for now)
- manage to close https://github.com/inveniosoftware/invenio-vocabularies/issues/66 in here too


invenio_rdm_records/fixtures/tasks.py
```diff
@@ -7,20 +7,21 @@
...
 
 @shared_task
-def create_vocabulary_record(data):
+def create_vocabulary_record(service_str, data):
     """Create a vocabulary record."""
-    vocabulary_service.create(system_identity, data)
+    if service_str == "vocabulary_service":
+        service = vocabulary_service
+    else:
+        service = getattr(current_rdm_records, service_str)
```
ppanero, line 23: I do agree with favouring the import of a specific service when available instead of using the registry... But here, why not use the registry? It exists and is a bit nicer than `getattr` IMO

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -188,7 +178,17 @@ def load(self):
 
         Fixtures found later are ignored.
         """
-        self._loaded_vocabularies = set()
+        # Prime with existing (sub)vocabularies
```
ppanero, line 181: minor: I am not 100% I understand the comment 😅

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -293,64 +286,135 @@ def get_records_by_vocabulary(self, vocabulary_id):
...
-                ids.add(id_)
+            ids.update(
+                entry.load(self._identity, ignore=ids, delay=self._delay)
+            )
 
         return ids
 
-    def create_vocabulary_type(self, id_, entry):
+
+class GenericVocabularyEntry:
```
ppanero, line 312: This and the class below seem to share a common structure, is it worth it to create an interface and maybe move some functionality there? Might be useful when doing this for affiliations.

ppanero (approved):
I am approving this (assuming tests passing before merging) not to block and so we can continue but I have the feeling that vocabularies loading is getting a bit out of hand and not in the correct place.

- Do we have somewhere a list of use cases of what we want to support and what not to? I feel we are removing and re-adding code all over the place. i.e. all loading PRs are always huge.
- Should the "mechanics" to load vocabularies be moved to `invenio-vocabularies`, generalized, and be used here? i.e. also tested there. I feel there are many files and tests (which are needed) but they do not have much to do with rdm-records but 100% with vocabularies.

With this, I am not trying to point fingers of any sort. I am just trying to say that I think we should take a step back after the release and re-think the loading to make it easier to work on, document use cases and maybe split responsibilities in a clearer manner.

invenio_rdm_records/fixtures/tasks.py
fenekku: We discussed this during the standup (at least this is what I took out of it): we don't need to use the registry and we let the registry be its thing in its own corner.

The registry is a "workaround" for access to services by id in a module that cannot simply import the service because of circular dependency. The concrete case why it was created is this file: https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/tasks.py (read the commit comments there too). Basically, a generic task was placed in records-resources that would use a service in packages depending on records-resources (it can't just import them). 

Other packages (invenio-rdm-records) that have direct import access don't need to use the registry.

invenio_rdm_records/fixtures/vocabularies.py
fenekku: Ah, is it "prime" (any of the verb definition here: https://www.wordnik.com/words/prime)? Basically, we don't start with an empty set of loaded vocabularies. Instead we start with the vocabularies already present in the database, so that we don't run into integrity errors / we skip previously loaded vocabularies. Would no comment be better?

invenio_rdm_records/fixtures/vocabularies.py
fenekku: 3-strike refactor. The loading of affiliations will tell us once we get to it.

fenekku:
I would actually agree that most of the vocabulary loading would make sense in invenio-vocabularies. It wasn't moved there because:

a- Initial loading was in here, so just continued from it
b- Dealing with with default vocabularies if the loading logic is placed in invenio-vocabularies is a complex lift in itself best done separately. (more below)

Placing the loading responsibility in invenio-vocabularies makes sense, but placing the default vocabularies values in invenio-vocabularies doesn't (at least for me). It's the schema defining package (invenio-rdm-records) that sets out those default values because it knows the kind of records it's dealing with (e.g. InvenioRDM has these default resource types, but other users of invenio-vocabularies may not even have resource types). So then we would need another mechanism for invenio-vocabularies to use things defined in rdm-records without importing it (circular dependency). Entrypoints or a registry-like pattern like seen above could be it. But those are both complex and I already spent my complexity points in the loading code for this PR.

invenio_rdm_records/fixtures/tasks.py
ppanero: I do recall that, that's why I agree on favouring importing specific services... However, since the workaround already exists I  think it's better to use it than `getattr`... But I guess that's a design choice.

invenio_rdm_records/fixtures/vocabularies.py
ppanero: Nope, leave the comment, is good. Just wanted to make sure that I understood it. Thanks 👼

ppanero:
> Placing the loading responsibility in invenio-vocabularies makes sense, but placing the default vocabularies values in invenio-
vocabularies doesn't (at least for me)

Agree completely

>  So then we would need another mechanism for invenio-vocabularies to use things defined in rdm-records without importing it (circular dependency)

I see... Need to check but if the `vocabularies.yaml` is made "generic" (sort of as it is now already) it can be used as a point of entry for vocabularies. Then rdm-records would define the defaults in its module. And the mechanics to get from entry points is already coded.

Issue #732, opened 2021-07-12
fixtures: Do we wanna use prioritized loading for affiliations


ppanero:
Guillaume: sort of looking at this in https://github.com/inveniosoftware/invenio-rdm-records/issues/686

PR #738, opened 2021-07-14
prioritized affiliations loading

fenekku:
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/732
- the refactor breakdown of VocabularyEntry is almost a troll/pointless :shrug:


invenio_rdm_records/fixtures/data/vocabularies.yaml
```diff
@@ -1,3 +1,10 @@
+affiliations:
+  pid-type: aff
+  schemes:
+    - id: ROR
```
ppanero, line 4: 🚀  nice

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -309,112 +316,114 @@ def load(self, ignore=None):
...
 
-class SubjectsVocabularyEntry:
-    """Vocabulary fixture for subjects vocabulary."""
 
-    def __init__(self, directory, id_, entry):
-        """Constructor."""
-        self._dir = directory
-        self._parent_id = id_
-        self._entry = entry
+class SpecificVocabularyEntryWithSchemes(VocabularyEntry):
```
ppanero, line 378: What about `VocabularyEntryWithSchemes` to make it a bit shorter... I think is more or less clear that when using schemes they cannot go to the generic vocab db/index?

invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -309,112 +316,114 @@ def load(self, ignore=None):
...
+    """Vocabulary fixture for specific vocabulary with schemes."""
 
-        ids = [s["id"] for s in self.schemes()]
-        # Raise if duplicate (conflicting)
-        if len(ids) != len(set(ids)):
-            raise ConflictingFixturesError(
-                f"Duplicate subtypes found for {self._parent_id}"
-            )
+    # service_str = "subjects_service"
```
ppanero, line 381:
```suggestion
```

run-tests.sh
```diff
@@ -51,7 +51,7 @@ if [[ ${keep_services} -eq 0 ]]; then
 	trap cleanup EXIT
 fi
 
-python -m check_manifest --ignore ".*-requirements.txt"
+python -m check_manifest --ignore ".*-requirements.txt,pyproject.toml"
```
ppanero, line 54: 😅

tests/fixtures/test_fixtures.py
```diff
@@ -172,20 +177,28 @@ def test_load_users(app, db, admin_role):
...
+    affiliations.load(system_identity, delay=False)
 
-    # app_data/users.yaml doesn't create an admin@inveniosoftware.org user
-    service = current_service_registry.get("rdm-affiliations")
-    cern = service.read(identity=system_identity, id_="01ggx4157")
+    cern = affiliations_service.read(identity=system_identity, id_="01ggx4157")
     assert cern["acronym"] == "CERN"
-    pytest.raises(PIDDoesNotExistError, service.read, "cern", system_identity)
+
+    with pytest.raises(PIDDoesNotExistError):
```
ppanero, line 203: nothing against this change... just wondering if it is a style preference or there is something to be gained from using the context descriptor.

ppanero (approved):
Just one minor on naming, merge when address/answered

invenio_rdm_records/fixtures/vocabularies.py
fenekku: :+1: I was ahead of myself

run-tests.sh
fenekku: I can't win: every time I do a `git add -u` it sneaks in :sob:

tests/fixtures/test_fixtures.py
fenekku: I find it easier to read, but in this scenario I was just copying the test structure in the tests above to make sure we were testing the same things and made the raise tests the same. What I also like about this style is that it's easier to extend. It can easily become:

```python
    with pytest.raises(PIDDoesNotExistError) as e:
        affiliations_service.read("cern", system_identity)

    # check something about the raised error
    print(e)
```

fenekku:
Comments addressed. Merging.

PR #744, opened 2021-07-15
fixtures: fix missing resource type mappings

lnielsen:


ppanero (commented):
needed for dublin core 🚀

Issue #741, opened 2021-07-15
Fix DataCite relation types


lnielsen:
Tom opened a PR to fix the relation type mapping to DataCite (see #697) for the old-style vocabulary. In the meantime, the vocabulary was moved to a generic vocabulary but the fixture is not correct wrt to the values used for the relation types.

Thus, 
- 1) we need to update the [fixture](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/data/vocabularies/relation_types.yaml) to have a correct value as in #697
- 2) we need to test if the relation types are serializer correctly in the DataCite XML.

PR #747, opened 2021-07-15
serialization: fix relation type datacite mapping

ppanero:
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/741

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -149,7 +149,7 @@ def full_record(users):
             "related_identifiers": [{
                 "identifier": "10.1234/foo.bar",
                 "scheme": "doi",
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -566,12 +566,12 @@ def relation_type(app):
 def relation_type_v(app, relation_type):
     """Relation type vocabulary record."""
     vocab = vocabulary_service.create(system_identity, {
-        "id": "cites",
+        "id": "iscitedby",
```
ppanero, line 569: Changing this to be able to test properly the mapping since `Cites` is only capitalized the first work... Wanted to check the proper camlecased.

PR #752, opened 2021-07-16
fixture: add identifiers to affiliations

ppanero:
this solution is not ideal... Maybe some logistics to include the `id` with scheme = to the file id (e.g. ROR.lower())... But then we also need the possibility of extra identifiers... For now I think is good like this even if it duplicates a tiny bit.

- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/967

Screenshot 2021-07-16 at 17 29 11
Screenshot 2021-07-16 at 17 29 04

### DIFF
--- a/invenio_rdm_records/fixtures/data/vocabularies/affiliations_ror.yaml
+++ b/invenio_rdm_records/fixtures/data/vocabularies/affiliations_ror.yaml
@@ -5,26 +5,41 @@
     de: "\_Europ\xE4ische Organisation f\xFCr Kernforschung"
     en: European Organization for Nuclear Research
     fr: "Organisation europ\xE9enne pour la recherche nucl\xE9aire"
+  identifiers:
+    - scheme: ror
+      identifier: 01ggx4157
 - acronym: BNL
   id: 02ex6cf31
   name: Brookhaven National Laboratory
   title:
     en: Brookhaven National Laboratory
+  identifiers:
+    - scheme: ror
+      identifier: 02ex6cf31
 - id: 047h63042
   name: Eko-Konnect
   title:
     en: Eko-Konnect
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/fixtures/data/vocabularies/affiliations_ror.yaml
```diff
@@ -5,78 +5,123 @@
     de: "\_Europ\xE4ische Organisation f\xFCr Kernforschung"
     en: European Organization for Nuclear Research
     fr: "Organisation europ\xE9enne pour la recherche nucl\xE9aire"
+  identifiers:
+    - scheme: ror
+      identifier: 01ggx4157
```
lnielsen, line 10: I think we need the full ROR ``https://ror.org/01ggx4157``

https://ror.org/display-guidelines/

ppanero: We could, I was just trying not to duplicate/mix content of ID and URI... If you see in the screenshots above, it is displayed correctly. Therefore, the guidelines are met.

If we want a more in depth fix, we should do some refactor as mentioned in this PR description:
- Add `scheme` to IdentifierScheme in Marshmallow-Utils
- Modify loading to add the `id + scheme` from the current `subtype` as each affiliation identifier. This would mean removing the current yaml changes.

lnielsen: So ROR website is not really clear about what is considered the ID. Initially, it seemed to me that they considered the entire URL as the ID, hence we should store the entire ID. I've tried to re-read everything and there's not a clear definition, so let's go with the short form.

Issue #750, opened 2021-07-16
datacite xml: incorrect schemes being used


lnielsen:
The relation scheme type is using wrong capitalization.

![Screenshot 2021-07-16 at 15 14 17](https://user-images.githubusercontent.com/1698163/125953728-69bee060-b947-4fd0-b6d3-8803f31ff37e.png)

Values should be one defined in: 
- http://schema.datacite.org/meta/kernel-4.4/include/datacite-relatedIdentifierType-v4.xsd
- https://github.com/inveniosoftware/datacite/blob/master/datacite/schemas/datacite-v4.3.json#L126-L147

Subtask:

- Add a test that validates a generated DataCite JSON or XML to ensure that it's correct.

Issue #735, opened 2021-07-13
revist name identifier and other hardcoded URIs in rdm records datacite seriliazation


ppanero:
i.e. https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/datacite/schema.py#L40

ppanero:
Could use the [LANDING_URLS](https://github.com/inveniosoftware/idutils/blob/master/idutils/__init__.py#L878) from idutils... however not all schemes are there (e.g. missing grid)

PR #753, opened 2021-07-19
schemas: use new allowed identifiers structure

ppanero:
- requires https://github.com/inveniosoftware/invenio-vocabularies/pull/76
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/962
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/735
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/750 (incorporated changes from #756 to make things easier)

Screenshot 2021-07-19 at 13 30 22
Screenshot 2021-07-19 at 13 30 06

### DIFF
--- a/invenio_rdm_records/resources/serializers/datacite/schema.py
+++ b/invenio_rdm_records/resources/serializers/datacite/schema.py
@@ -26,16 +26,13 @@ from invenio_rdm_records.resources.serializers.ui.schema import \
 from ..utils import map_type
 
 
+def get_scheme_label(scheme, labels):
+    """Returns a scheme label."""
+    return current_app.config[labels][scheme].get("label", scheme)
+
+
 class PersonOrOrgSchema43(Schema):
     """Creator/contributor common schema for v4."""
-
-    # PIDS-FIXME: need a more escalable solution for URIs
-    URIS = {
-        "orcid": "http://orcid.org/",
-        "gnd": "http://d-nb.info/",  # PIDS-FIXME: is this correct?
-        "ror": "https://ror.org/",
-        "isni": "https://isni.org",
-    }
 
     name = fields.Str(attribute="person_or_org.name")
     nameType = fields.Str(attribute="person_or_org.type")
@@ -50,19 +47,13 @@ class PersonOrOrgSchema43(Schema):
         identifiers = obj["person_or_org"].get("identifiers", [])
 
         for identifier in identifiers:
-            scheme = identifier["scheme"]
-            value = identifier["identifier"]
-            uri = self.URIS.get(scheme)
-
             name_id = {
-                "nameIdentifier": value,
-                "nameIdentifierScheme": scheme.upper(),
+                "nameIdentifier": identifier["identifier"],
+                "nameIdentifierScheme": get_scheme_label(
+                    identifier["scheme"],
+                    "RDM_RECORDS_PERSONORG_SCHEMES"
+                ),
             }
-
-            if uri:
-                name_id["nameIdentifier"] = uri + value
-                name_id["schemeURI"] = uri
-
             serialized_identifiers.append(name_id)
 
         return serialized_identifiers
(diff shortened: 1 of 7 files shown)


invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -52,16 +45,17 @@ def get_name_identifiers(self, obj):
...
                 "nameIdentifier": value,
                 "nameIdentifierScheme": scheme.upper(),
             }
 
+            uri = to_url(value, scheme)
             if uri:
-                name_id["nameIdentifier"] = uri + value
-                name_id["schemeURI"] = uri
+                name_id["nameIdentifier"] = uri
+                name_id["schemeURI"] = LANDING_URLS.get(scheme).format(
```
lnielsen, line 57: This shoudl be using [``idutils.to_url()``](https://github.com/inveniosoftware/idutils/blob/master/idutils/__init__.py#L908) instead

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -52,16 +45,17 @@ def get_name_identifiers(self, obj):
         for identifier in identifiers:
             scheme = identifier["scheme"]
             value = identifier["identifier"]
-            uri = self.URIS.get(scheme)
 
             name_id = {
                 "nameIdentifier": value,
                 "nameIdentifierScheme": scheme.upper(),
```
lnielsen, line 51: This has to be a mapping as well that defines the correct name identifier scheme value for DataCite.

invenio_rdm_records/resources/serializers/datacite/schema.py
lnielsen, line 57: I don't think we can do it like this as they are for the landing pages and might not be the same as the Scheme's URI or the scheme might not have a URI. As long as we include the ``nameIdentifierScheme`` we can remove the ``schemeURI``.

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -52,16 +45,17 @@ def get_name_identifiers(self, obj):
...
             }
 
+            uri = to_url(value, scheme)
             if uri:
-                name_id["nameIdentifier"] = uri + value
-                name_id["schemeURI"] = uri
+                name_id["nameIdentifier"] = uri
```
lnielsen, line 56: Problem here is the same as below with the ``affiliationIdentifier``.

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -104,9 +98,9 @@ def get_affiliation(self, obj):
                     id_value = identifier["identifier"]
                     aff["affiliationIdentifier"] = id_value
                     aff["affiliationIdentifierScheme"] = scheme.upper()
-                    uri = self.URIS.get(scheme)
+                    uri = to_url(id_value, scheme)
                     if uri:
-                        aff["affiliationIdentifier"] = uri + id_value
+                        aff["affiliationIdentifier"] = uri
```
lnielsen, line 103: This logic is not correct as it will always use the landing page URL instead of the normalised identifier.

- affiliationIdentifier: should be the normalised identifier value
- affilationIdentifierScheme: use the correct scheme acronym e.g. ``arXiv``

lnielsen: I see the nameIdentifierScheme/affiliationIdentifierScheme is fixed in #756

ppanero: > affiliationIdentifier: should be the normalised identifier value

[Here](https://github.com/inveniosoftware/idutils/blob/master/idutils/__init__.py#L919) you can see the identifier value is normalized. Or do you mean that is hould **not have a URI** in it? This would result e.g. 

```xml
<affiliation affiliationIdentifier="01ggx4157" affiliationIdentifierScheme="ROR">CERN</affiliation>",
```

The rest of the comment I understand is about labelling/types that got fixed in a subsequent PR.

invenio_rdm_records/resources/serializers/datacite/schema.py
ppanero: :+1: Removing

invenio_rdm_records/resources/serializers/datacite/schema.py
ppanero: Discussed IRL:
- Identifiers should be assumed to be normalized (Marshmallow-Utils: IdentifierScheme does this already)
- No URI should be calculated
- IDUtils normalization functions need to be reviewed (e.g. ROR should contain the URL)

tests/resources/serializers/test_datacite_serializer.py
```diff
@@ -139,15 +139,14 @@ def test_datacite43_serializer(running_app, full_record):
                 "givenName": "Lars Holm",
                 "familyName": "Nielsen",
                 "nameIdentifiers": [{
-                    "nameIdentifier": "http://orcid.org/0000-0001-8135-3489",
```
ppanero, line 142: assuming normalization means removing urls...

lnielsen (approved):

PR #756, opened 2021-07-19
Datacite

ppanero:
- requires #753 
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/750

**Open questions**
- CSL fixes the labelling by using the marshmallow [data_key](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/csl/schema.py#L47) which I think it is fine since it is only specific identifiers that are allowed.
- UI, is not being touched at the moment although it will be necessary for https://github.com/inveniosoftware/invenio-app-rdm/issues/980
- Dublin Core uses the `scheme` directly e.g. [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/dublincore/schema.py#L68). I am quite sure that:
    - It should use the label
    - In cases like `arXiv` where idutils allows (optionally) the prefix, it might actually get double prefixed

```
>>> idutils.is_arxiv("1904.00007")
<re.Match object; span=(0, 10), match='1904.00007'>
>>> idutils.is_arxiv("arxiv:1904.00007")
<re.Match object; span=(0, 16), match='arxiv:1904.00007'>
```

Screenshot 2021-07-19 at 15 03 13
Screenshot 2021-07-19 at 15 03 08

### DIFF
--- a/invenio_rdm_records/resources/serializers/datacite/schema.py
+++ b/invenio_rdm_records/resources/serializers/datacite/schema.py
@@ -27,6 +27,11 @@ from invenio_rdm_records.resources.serializers.ui.schema import \
 from ..utils import map_type
 
 
+def get_scheme_label(scheme, labels):
+    """Returns a scheme label."""
+    return current_app.config[labels][scheme]["label"]
+
+
 class PersonOrOrgSchema43(Schema):
     """Creator/contributor common schema for v4."""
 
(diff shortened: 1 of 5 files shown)


tests/resources/serializers/test_datacite_serializer.py
```diff
@@ -207,7 +207,7 @@ def test_datacite43_serializer(running_app, full_record):
             },
             {
                 "identifier": "1924MNRAS..84..308E",
-                "identifierType": "bibcode"
+                "identifierType": "Bibcode"
```
ppanero, line 210: There are many cases where this labelling is being tested (e.g. `doi` to `DOI`) but this is a clear one.

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -48,13 +53,17 @@ def get_name_identifiers(self, obj):
...
+                    scheme,
+                    "RDM_RECORDS_PERSONORG_SCHEMES"
+                ),
             }
 
             uri = to_url(value, scheme)
             if uri:
                 name_id["nameIdentifier"] = uri
-                name_id["schemeURI"] = LANDING_URLS.get(scheme)
+                name_id["schemeURI"] = LANDING_URLS.get(scheme).format(
```
lnielsen, line 65: See comment in #753

ppanero:
superseeded by https://github.com/inveniosoftware/invenio-rdm-records/pull/753

PR #757, opened 2021-07-19
subjects: fetch and cache long scheme names

zzacharo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/977


lnielsen (commented):
I have a couple of worries:

1) I'm slightly worried that we're adding quite a lot of data into the UI serializer and it's used for search results

2) We're doing direct access to the data layer from the presentation layer.

lnielsen:
I'm closing this PR as it needs another method to be used. Keeping issue open and it's still linked with this PR

PR #759, opened 2021-07-19
services: move embargo lifting to data layer

lnielsen:


invenio_rdm_records/records/systemfields/access/embargo.py
```diff
@@ -39,8 +39,8 @@ def active(self, value):
         """Set the Embargo's (boolean) active state."""
         self._active = bool(value)
 
-    def lift(self):
-        """Update the embargo active status if it has expired.
+    def _lift(self):
```
ppanero, line 42: question: this function is called with the field but from another file/module (i.e. line 97 of record.py). Doesn't that break a bit the idea of "private" (underscored function name)?

invenio_rdm_records/records/systemfields/access/field/record.py
```diff
@@ -10,6 +10,7 @@
 
 from enum import Enum
 
+import arrow
```
ppanero, line 13: Is this needed?

invenio_rdm_records/services/services.py
```diff
@@ -198,44 +199,41 @@ def discard_pid(self, id_, identity, pid_type, pid_client=None):
...
-                lifted_embargo_from_draft = True
-
-        # Lifts embargo from record
-        self._lift_embargo_from(record)
-        # Commit and index
+            if record.access == draft.access:
+                draft.access.lift_embargo()
+                draft.commit()
+
+        if not record.access.lift_embargo():
```
ppanero, line 222: The embargo from the draft is being lifted before, and it's result is ignored (and commited). Since the access conditions are checked right before. Would it be sensible to raise also from the embargo lift of the draft (line 219 above)? The previous implementation was raising on both cases.

It might save a few DB queries and the result would be the same? Alternatively, lift first the record and then the draft.

ppanero (approved):
Just some doubts and minors

invenio_rdm_records/records/systemfields/access/embargo.py
lnielsen: I make it private to make it clear that you should not be calling this function as part of the API: - e.g. ``record.access.embargo._lift()``. The call from the ``RecordAccess`` and ``Embargo`` I consider as intimately related objects (RecordAccess is in charge of creating and working with the Embargo object), and hence it's more about making the API clear for users of the Record class here.

invenio_rdm_records/services/services.py
lnielsen: Yep, i'll move the record lift up first.

invenio_rdm_records/records/systemfields/access/embargo.py
lnielsen: Didn't work in the end, because the record.access != draft.access

Issue #729, opened 2021-07-12
preview: citation box breaks in preview


ppanero:
```
Traceback (most recent call last):
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/flask/app.py", line 2464, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/werkzeug/middleware/proxy_fix.py", line 169, in __call__
    return self.app(environ, start_response)
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/werkzeug/middleware/dispatcher.py", line 66, in __call__
    return app(environ, start_response)
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/flask/app.py", line 2464, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/werkzeug/middleware/proxy_fix.py", line 169, in __call__
    return self.app(environ, start_response)
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/flask/app.py", line 2450, in wsgi_app
    response = self.handle_exception(e)
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/flask_restful/__init__.py", line 271, in error_router
    return original_handler(e)
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/flask/app.py", line 1867, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/ppanero/.virtualenvs/june-late/lib/python3.7/site-packages/flask_restful/__init__.py", line 271, in error_router
    return original_handler(e)
[...]

PR #780, opened 2021-07-23
serializers: csl should validate existance of resource_type

ppanero:
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/729

This PR makes the REST API return a validation error for the citation, as it should since the resource type is mandatory for CiteProc to work.

It is up to app-rdm to display one message or another.

### DIFF
--- a/invenio_rdm_records/resources/serializers/csl/schema.py
+++ b/invenio_rdm_records/resources/serializers/csl/schema.py
@@ -67,7 +67,11 @@ class CSLJSONSchema(Schema):
 
     def get_type(self, obj):
         """Get resource type."""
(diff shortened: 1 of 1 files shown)


lnielsen:
I think it's better we use a default resource type like ``article`` than sending an error. Otherwise users will see an error message instead of an approximate citation string?

ppanero:
Examples of output with minimal record and empty record

**Publication-Article**

Screenshot 2021-07-23 at 14 26 18
Screenshot 2021-07-23 at 14 25 02


**Publication-Report**
Screenshot 2021-07-23 at 14 26 34
Screenshot 2021-07-23 at 14 24 43

PR #790, opened 2021-08-02
facets: configure subjects facets

fenekku:
- facets: add subjects faceting
- dependencies: bump invenio-vocabularies
- release: v0.32.2
- pre-requisite: https://github.com/inveniosoftware/invenio-vocabularies/pull/80
- closes: https://github.com/inveniosoftware/invenio-vocabularies/issues/68


invenio_rdm_records/services/config.py
```diff
@@ -54,10 +55,11 @@
     value_labels=VocabularyLabels('languages')
 )
 
-subject_facet = TermsFacet(
-    field='metadata.subjects.id',
+subject_facet = NestedTermsFacet(
```
lnielsen, line 58: See comment in inveniosoftware/invenio-app-rdm#1036. The rest of the PR makes full sense independent of if we choose to go with TermsFacet or NestedTermsFacet.

fenekku: Is it ok if I make a ticket for that one and we can revisit? Issues I see:

- We have a lot of subjects at NU (at least 2 other categories than MeSH will be coming) and displaying them all as 1 is awkward (huge list (or with caching not big enough!) and searchers can't drill down/ focus on category of interest) 
- Resource types would have subfaceting but not this one; it's a little strange 
- Heavier: have to rely on `read_all` + more development time

PR #794, opened 2021-08-04
bugfix: current_default_locale was not called

mb-wali:


invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -73,8 +74,8 @@ def apply_idx(affiliation):
         if 'role' in creator:
             creator['role']['title'] = gettext_from_dict(
                 creator['role']['title'],
-                get_locale(),
-                current_default_locale
+                locale=get_locale(),
```
lnielsen, line 77: Function signature is without keyword arguments: ``def gettext_from_dict(catalog, locale, default_locale):``
```suggestion
                get_locale(),
```

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -73,8 +74,8 @@ def apply_idx(affiliation):
...
+                default_locale=current_default_locale()
```
lnielsen, line 78:
```suggestion
                current_default_locale()
```

lnielsen (approved):
Good catch. Just a minor thing about keyword to positional args.

PR #795, opened 2021-08-04
global: facets, sort and permission customizations

lnielsen:


invenio_rdm_records/config.py
```diff
@@ -338,4 +304,118 @@ def always_valid(identifier):
         "validator": always_valid
     }
 }
-RDM_RECORDS_DOI_DATACITE_FORMAT = "{prefix}/{id}"
+
+
+#
+# Record permission policy
+#
+RDM_PERMISSION_POLICY = None
```
lnielsen, line 312: Intended to be set from ``invenio.cfg`` if you want to override it.

invenio_rdm_records/config.py
```diff
@@ -338,4 +304,118 @@ def always_valid(identifier):
...
+"""Override the default record permission policy."""
+
+
+#
+# Search configuration
+#
+RDM_RECORD_FACETS = {
```
lnielsen, line 319: Intended to be defined here, and only overwritten by advanced users.

invenio_rdm_records/config.py
```diff
@@ -338,4 +304,118 @@ def always_valid(identifier):
...
+    },
+    'is_published': {
+        'facet': facets.is_published,
+        'ui': {
+            'field': 'is_published',
+        }
+    }
+}
+
+RDM_RECORD_SORT_OPTIONS = {
```
lnielsen, line 349: Ditto.

invenio_rdm_records/config.py
```diff
@@ -338,4 +304,118 @@ def always_valid(identifier):
...
+.. code-block:
+
+    "<option name>": dict(
+        title=_('<title>'),
+        fields=['-updated'],
+    ),
+
+"""
+
+RDM_RECORD_SEARCH = {
```
lnielsen, line 386: Intended to be set from ``invenio.cfg`` if you want to override it.

invenio_rdm_records/config.py
```diff
@@ -338,4 +304,118 @@ def always_valid(identifier):
...
+  a single sort option name which must have been defined in
+  ``RDM_RECORD_SORT_OPTIONS``. If not provided, will use the first element of
+  the ``sort`` list.
+- ``sort_default_no_query`` - The default sort option when no query is
+  provided. Must be a single sort option name which must have been defined in
+  ``RDM_RECORD_SORT_OPTIONS``. If not provided, will use the second element of
+  the ``sort`` list.
+"""
+
+RDM_RECORD_SEARCH_DRAFTS = {
```
lnielsen, line 408: Intended to be set from ``invenio.cfg`` if you want to override it.

invenio_rdm_records/searchconfig.py
```diff
@@ -0,0 +1,90 @@
+# -*- coding: utf-8 -*-
```
lnielsen, line 1: This is reused in App-RDM for the RSK app config generation to ensure there's a coherent way of selecting among available options.

invenio_rdm_records/services/config.py
```diff
@@ -21,92 +21,55 @@
...
-    value_labels={"true": _("Published"), "false": _("Unpublished")}
-)
-
 
 #
-# Search options
+# Default search configuration
 #
-class RDMSearchOptions(SearchOptions):
+class RDMSearchOptions(SearchOptions, SearchOptionsMixin):
```
lnielsen, line 43: I keep the customize methods apart (in``customizations.py``), because I want a clean overview of the config.

invenio_rdm_records/config.py
lnielsen: I purposedly name the variables ``RDM_*`` instead of ``RDM_RECORDS_*``. These variables are intended to be overwritten in ``invenio.cfg`` by end-users, and long configuration variable names are not very user friendly.

We should align on existing ones as well.

invenio_rdm_records/config.py
```diff
@@ -108,43 +110,6 @@ def _(x):
...
-        'contributors.role': {
-            'path': os.path.join(
-                os.path.dirname(os.path.abspath(__file__)),
-                'my_contributor_roles.csv'
-            )
-        }
-    }
-"""
-
 RDM_RECORDS_USER_FIXTURE_PASSWORDS = {
```
ppanero, line 113: shelvable: This was there before but... Why do we *only* allow the password through env var? e.g. in ES we allow it fully or partially. I'd say the consistent thing would be:
- Allow full configuration (not partial)
- If one wants to do it through yaml can (maybe their deployment can have the passwords secure somehow)
- If one wants to do all via env var, can do also

invenio_rdm_records/ext.py
```diff
@@ -92,30 +93,74 @@ def init_config(self, app):
...
+            sort=sort_opts,
+            facets=facet_opts,
+        )
+        search_versions_opts = SearchConfig(
+            app.config.get('RDM_RECORD_SEARCH_VERSIONING'),
+            sort=sort_opts,
+            facets=facet_opts,
+        )
+
+        class ServiceConfigs:
```
ppanero, line 128: could this become a de-facto `RDMServiceConfig`? Do we think this might be useful in another context?

ppanero (approved):
Most of my comments are on design and probably shelvable for a next iteration. My take on this:
- ➕ it introduces the longed for "configuration validation" that we have mentioned at some point in the past
- ➕ allows users to configure services in a quite easy manner
- ➖ It is not a clear cut when a variable shold be `RDM_`, when `RDM_RECORDS` and it gets confusing when something is `RDM_` but records related cuz it's `RDM_RECORD`. i.e. in the next iteration we should consolidate configuration and set a clear cut

invenio_rdm_records/config.py
lnielsen: Yeah, I guess it should just be part of the fixtures and not here.

invenio_rdm_records/ext.py
lnielsen: Hmm 🤔 Could be..... I must admit I mainly did it so I could have simple syntax in ``init_services()`` that didn't involve dicts..... went pretty fast.

lnielsen:
> It is not a clear cut when a variable shold be RDM_, when RDM_RECORDS and it gets confusing when something is RDM_ but records related cuz it's RDM_RECORD. i.e. in the next iteration we should consolidate configuration and set a clear cut

Yes I agree. I really want to go through all our config variables and define clear and consistent naming for all. There's too much ad-hoc naming when needed, so we definitely need a full review.

invenio_rdm_records/services/customizations.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # Create the config class
+        new_cls = _make_cls(cls, attrs)
+
+        # TODO: Change how to customize PID providers. Was moved here from
+        # ext.py
+        if not doi_registration:
+            new_cls.pids_providers.pop("doi", None)
+
+        return new_cls
```
fenekku, line 79: Ouf... This is to achieve: `RDM<Service>Config` -> set some attributes -> use this configuration with the set attributes but leave the global `RDM<Service>Config` untouched so not mixing with others, right?

What about just instantiating the config class? 

```python
class CustomizableConfigMixin:
    """Mixin to allow local config modifications."""

    @classmethod
    def customize(cls, customizations**):
        """Class method to customize the config for an instance."""
        config = cls()
        for attribute, value in customizations.items():
            if attribute in config:  # if we only want to allow customizations of existing attributes
                setattr(config, attribute, value)
        return config
```  

The instance of the config class has all the same attributes, but now their modification is local to that instance. `customize` can also be edited to restrict the fields that can be customized if so desired, although leaving it open would make any future additions easier.

When we drop support for Python 3.6, we can even use [dataclasses](https://docs.python.org/3.7/library/dataclasses.html).

fenekku (commented):
Ok I think I got it.

Main points:
- `RDM_` naming conventions to be seen
- minor-ish interface tidying up
- simplification of configuration customization.

invenio_rdm_records/config.py
lnielsen: For maintainers, this naming convention works for navigating the code, but I think it fails for users who just install the system and need to configure it. I think for these users, feature-wise grouping in the naming works well. E.g. the feature of search is split over both App-RDM, RDM-Records and more, but that doesn't matter for the user - for them it's just search. 

One possible solution is naming everything in ``App-RDM`` and ``RDM-Records`` with the ``RDM_`` prefix.

invenio_rdm_records/config.py
lnielsen: I was tempted at simply calling it ``RDM_FACETS``. But then there might be ``RDM_AFFILIATIONS_FACETS``?

My preference would be for ``RDM_FACETS`` (similar below).

invenio_rdm_records/services/customizations.py
lnielsen: > Ouf... This is to achieve: RDM<Service>Config -> set some attributes -> use this configuration with the set attributes but leave the global RDM<Service>Config untouched so not mixing with others, right?

Yes, the intention is to leave it untouched.

1. I agree that instantiation of the config class ``RDMRecordsService(permission_policy, ...)`` is a better way to customize the class. The problem is I find it very dirty if we start sometimes passing classes and other times we pass instances, so I'd prefer to keep it like this and add a comment in the code that ideally this should refactor just be a class instantiation.

2. It's intentional that you can only customize some specific attributes (to keep a clean API of which customizations are supported). I.e. I want to make it a conscious decision that something is customizable.

PR #797, opened 2021-08-05
search: fix internal _sort/_facet [+]

fenekku:
If 'sort' or 'facets' is omitted/forgotten in RDM_SEARCH the code breaks
because self._sort/self._facets are expected to be iterable. By setting
them to be [] by default, that interface is respected.


lnielsen:
I've changed the branch to ``maint-0.32``. I'll cherry-pick and put on master as well

fenekku:
Got it, I will aim for the `maint-` branches (across modules) as we fix bugs and then chery pick the merged commits into respective masters. 

We can discuss workflow at standup if we want something else.

PR #812, opened 2021-09-10
serializer: fix datacite alt identifier doi handling

### LINKED ISSUE #811: duplicate DOI makes DataCite export fail
Package version (if known): v0.33.15

## Describe the bug
If the record is created with a DOI, and the same DOI gets added as alternate identifier. The DataCite export will add the same `{identifier, type}` pair to the list of identifiers and it will fail.

```
{
  "pids": {
    "doi": {
      "identifier": "10.81088/12345-abcde",
      "provider": "datacite",
      "client": "datacite"
    }
  },
  ...
  "metadata": {
    "identifiers": [
      {
        "identifier": "10.81088/12345-abcde",
        "scheme": "doi"
      }
    ],
  }
}
```

## Expected behavior

The serializer should be able to deal with duplicates.

## Additional context

Sentry error link: https://inveniordm-sentry.web.cern.ch/share/issue/6b70f6d5cef74c3cbb3139775e680340/

tmorrell:
Resolves #811 by removing alternative identifiers that have type DOI from the DataCite serialization, since DataCite doesn't accept DOIs that are not the registered DOI.

### DIFF
--- a/invenio_rdm_records/resources/serializers/datacite/schema.py
+++ b/invenio_rdm_records/resources/serializers/datacite/schema.py
@@ -302,8 +302,7 @@ class DataCite43Schema(Schema):
         """Get (main and alternate) identifiers list."""
         serialized_identifiers = []
 
-        # pids go first so the DOI from the record is the main doi
-        # otherwise a doi from identifiers can be used
+        # pids go first so the DOI from the record is included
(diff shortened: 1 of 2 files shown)


tmorrell:
Superseded by #829

Issue #763, opened 2021-07-20
api: cannot provide existing DOI


lnielsen:
Package version (if known): <version>

## Describe the bug

![Screenshot 2021-07-20 at 14 15 00](https://user-images.githubusercontent.com/1698163/126322027-695e5dc0-907d-4295-b5b5-36fa77325e20.png)

## Steps to Reproduce

Make sure you have enabled DOI minting according to the docs.

1. Start a new upload
2. Select "Yes" to "Do you already have a DOI"
3. Enter a valid DOI - e.g. ``10.5281/zenodo.1234``
4. Save the draft
5. Reload and see that it's not saved.

## Additional context

```
127.0.0.1 - - [20/Jul/2021 14:11:34] "POST /api/records HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/flask/app.py", line 2464, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/werkzeug/middleware/proxy_fix.py", line 169, in __call__
    return self.app(environ, start_response)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/werkzeug/middleware/dispatcher.py", line 66, in __call__
    return app(environ, start_response)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/flask/app.py", line 2464, in __call__
    return self.wsgi_app(environ, start_response)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/werkzeug/middleware/proxy_fix.py", line 169, in __call__
    return self.app(environ, start_response)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/flask/app.py", line 2450, in wsgi_app
    response = self.handle_exception(e)
  File "/Users/lnielsen/.virtualenvs/my-site-ZNC6mTtx/lib/python3.8/site-packages/flask_restful/__init__.py", line 271, in error_router
    return original_handler(e)
[...]

Issue #761, opened 2021-07-20
pids: valid system-managed values accepted


slint:
It is currently possible to create (and possibly update) a draft with a "system-managed" provider/client configuration, e.g. a "datacite" DOI with a prefix different from the one configured. The following service (and REST API) input is possible:

```python
def test_pid_creation_invalid_format_value_managed(
    running_app, es_clear, minimal_record
):
    superuser_identity = running_app.superuser_identity
    service = current_rdm_records.records_service

    doi = {
        # valid PID, but not with the configured managed prefix (i.e. 10.1234)
        "identifier": "10.5678/valid.doi",
        "provider": "datacite",
        "client": "datacite"
    }
    minimal_record["pids"] = {"doi": doi}

    # Create should fail with a validation error (see below)
    draft = service.create(superuser_identity, minimal_record)
    
    # In case we don't fail "hard" this should be an empty dict (see below)
    assert draft.to_dict()["pids"] == {
        "identifier": "10.5678/valid.doi",
        "provider": "datacite",
        "client": "datacite"
    }

    # Expected result
    assert draft.to_dict()["pids"] == {}
    assert draft.errors == [
        {'field': 'pids.doi', 'messages': [
            # Or a more user friendly error message?
            _('The "datacite" provider for "doi" is system-managed.')
        ]}
    ]
```

Thankfully, even though creation is possible, on publish there will be a failure, but for an unrelated reason (it will try to lookup the PID with value `10.5678/valid.doi`, and fail with `PIDNotExists`).

This issue also affects the OAI PID registration (since it makes it possible to submit OAI values, which of course don't make sense, since OAI PIDs are generated only locally by the instance).

### Proposed solution

- PIDs configurations with `system_managed=True`, should not accept user input at all (i.e. it shouldn't be possible to submit `provider=datacite` and also an `identiier=...` value)
- It should only be possible to reserve `system_managed=True` PIDs, in case one wants to know their value before publishing

ppanero:
The issue refers to being able to create a pid (managed) with a different prefix than the configured one.
[...]

Issue #590, opened 2021-05-05
service: extract pids and links to subservices


ppanero:
Extract pids and links to subservices, in a similar fashion than the files. However, this requires design on how to create subservices in general.

Issue #584, opened 2021-05-03
DOI registration in async task


ppanero:

ppanero:
closing in favour of https://github.com/inveniosoftware/invenio-rdm-records/issues/563

PR #814, opened 2021-09-14
pids: rewrite

ppanero:
The detailed design can be found [here](https://codimd.web.cern.ch/PWRdWnQEQd6XLZnrrj2HLQ).

**Breaking compatibility changes**
- PIDs become a subservice
- `reserve` changes name to `create` since the pid will be in status `NEW`.
- Some responses change (e.g. receiving a `{pids: doi: {}}` payload would create a doi before, but now it throws a validation error.
- `BaseClient` does not have a `api_client` attribute but a `client` one. The `api_client` seems specific to the DataCite case, while a client might be more common (e.g. some clients might interact through programmatic api).
- Providers do not support multiclient. The way to support multiple clients is to create different providers with different clients.
- DataCite Provider requires credentials. Having it instantiated without an actual client was a test requirement that can be solved by mocking it and not adding if and complexity to the provider itself.
- Service level config for pids changed. Now we use `pids_providers` (different structure from before) and `pids_required` which before was set per provider in the `pids_providers` config. This was forcing us to loop through it every time we needed to know the list.
- As per design `reserve` is not call upon creation, but *on publish*. This means that the behaviour of the service functions is the same, but the `reserve` function of the datacite provider now reserves the pid in the system (before it was left as `NEW`).
- `post_publish` hook is used to trigger async registration task for pids.
- probably something more I am forgetting....

**Other issues tackled**
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/580
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/590
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/763
- closes https://github.com/inveniosoftware/react-invenio-deposit/issues/293
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1050
- enables (needs work at UI level) https://github.com/inveniosoftware/react-invenio-deposit/issues/293 
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/584
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/563
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/761
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/562

**Requires**
- https://github.com/inveniosoftware/invenio-drafts-resources/pull/164 and release
- https://github.com/inveniosoftware/invenio-drafts-resources/pull/163
- https://github.com/inveniosoftware/invenio-drafts-resources/pull/170


invenio_rdm_records/services/pids/tasks.py
```diff
@@ -9,8 +9,16 @@
...
+from invenio_rdm_records.proxies import current_rdm_records
+
 
 @shared_task(ignore_result=True)
-def register_pid(recid, pid_type):
-    """Register a PID."""
-    pass
+def register_pid(pid, scheme, record_id, provider_name=None):
+    """Registers a PID."""
+    record = current_rdm_records.records_service.record_cls.get_record(
```
ppanero, line 18: Can we do `service.read`? not to go that low to data access layer (i.e. `record_cls`)?

invenio_rdm_records/services/pids/tasks.py
```diff
@@ -9,8 +9,16 @@
...
 from celery import shared_task
 
+from invenio_rdm_records.proxies import current_rdm_records
+
 
 @shared_task(ignore_result=True)
-def register_pid(recid, pid_type):
-    """Register a PID."""
-    pass
+def register_pid(pid, scheme, record_id, provider_name=None):
```
ppanero, line 16: ocd: would style call it pid_type instead of scheme since a record level is what we use... scheme is more at config level because is more understandable for the users.

Also maybe change the order of record_id and scheme to have the related ones closer?

tests/services/pids/test_pids_tasks.py
```diff
@@ -0,0 +1,67 @@
...
+        assert pid.status == PIDStatus.REGISTERED
+
+
+def test_publish_record(running_app, es_clear, minimal_record):
+    """No pid provided, creating one by default."""
+    service = current_rdm_records.records_service
+    superuser_identity = running_app.superuser_identity
+    draft = service.create(superuser_identity, minimal_record)
+    record = service.publish(draft.id, superuser_identity)
+    dict_record = record.to_dict()
```
ppanero, line 53: is this needed? cant you do `record["pids"]` directly?  e.g. https://github.com/inveniosoftware/invenio-rdm-records/pull/814/files#diff-4d69e1b21e25061eb03ca154d0f090414925e0529fede5b88a2b6044d8d9e913R41

ppanero (commented):
@jrcastro2 small suggestions! nice work! Im wondering about the eager execution of the tasks... Do the tests always pass? Wondering what would happen if the publish takes a bit more to actually send the task etc. WDYT?

invenio_rdm_records/services/components/pids.py
```diff
@@ -246,10 +243,26 @@ def edit(self, identity, draft=None, record=None):
...
             raise ValidationError(message=errors)
         draft.pids = record_pids
+
+    def post_publish(self, identity, draft=None, record=None):
+        """Post publish handler."""
+        record_pids = record.get('pids', {})
+        for scheme, pid_dict in record_pids.items():
+            provider_name = pid_dict.get("provider")
+            identifier_value = pid_dict.get("identifier")
+            provider = current_rdm_records.records_service.pids.get_provider(
```
ppanero, line 260: minor: just for code readability maybe we can assignthe service before the for loop.

```python
service = current_rdm_records.records_service
for ...
    service.pids....
```

invenio_rdm_records/services/components/pids.py
```diff
@@ -246,10 +243,26 @@ def edit(self, identity, draft=None, record=None):
         # FIXME: this should be moved to marshmallow level
         # validation happens before to be able to fail before reserving pids
         # it means a double looping but validation should eventually be moved
-        # to marshmellow see invenio-rdm-records/issues/796
-        errors = {}
+        # to marshmallow see invenio-rdm-records/issues/796
+        errors = []
```
ppanero, line 247: 🚀 nice catch!

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -130,7 +129,7 @@ def register(self, pid, record, **kwargs):
             # PIDS-FIXME: move to async task, exception handling included
             try:
                 doc = DataCite43JSONSerializer().dump_one(record)
-                url = kwargs["url"]
+                url = record.to_dict()["links"]["self_html"]
```
ppanero, line 132: sanity check: can we do `record.links` or `record["links"]. At this stage is a `RDMRecord` no? and thats why we need `to_dict()`?

tests/services/pids/providers/test_datacite_pid_provider.py
```diff
@@ -72,11 +87,13 @@ def test_datacite_provider_reserve(record, datacite_provider):
     assert db_pid.status == PIDStatus.RESERVED
 
 
-def test_datacite_provider_register(record, datacite_provider, mocker):
+def test_datacite_provider_register(record_w_links, datacite_provider, mocker):
     mocker.patch("invenio_rdm_records.services.pids.providers.datacite." +
                  "DataCite43JSONSerializer")
-    created_pid = datacite_provider.create(record)
-    assert datacite_provider.register(pid=created_pid, record=record, url=None)
+    created_pid = datacite_provider.create(record_w_links._record)
```
ppanero, line 93: sanity check: in the code the flow goes as follows
1. the component launches the `register_pid` task passing the recid.
2. the task gets the record using `service.read`, this should return a `RecordItem` object.
3. this `RecordItem` is sent to the provider.

Shouldn't we need to pass just `record_w_links` here? because the `service.publish` of the fixture should return also ` RecordItem`. We should not need the `service.read` no? (line 48 above)

Could it be that the provider expects different object in the `create` and `publish` method? maybe we need to harmonize this.

invenio_rdm_records/services/pids/tasks.py
```diff
@@ -8,17 +8,22 @@
...
 from celery import shared_task
+from invenio_db import db
+from invenio_access.permissions import system_identity
 
 from invenio_rdm_records.proxies import current_rdm_records
 
 
 @shared_task(ignore_result=True)
-def register_pid(pid, scheme, record_id, provider_name=None):
+def register_pid(pid_type, pid_value, record_id, provider_name=None):
```
ppanero, line 18: naming: wondering if we should call it `recid` instead of `record_id`, what do you think?

ppanero (commented):
just a few minor comments. we should mostly check to see if the provider api is consistent

invenio_rdm_records/services/pids/providers/datacite.py
jrcastro2: After discussion, going back to `kwargs["url"]`.

invenio_rdm_records/services/components/pids.py
```diff
@@ -244,10 +243,27 @@ def edit(self, identity, draft=None, record=None):
...
         self._validate_pids_schemes(record_pids)
         self._validate_pids(record_pids, record, errors)
         if errors:
             raise ValidationError(message=errors)
         draft.pids = record_pids
+
+    def post_publish(self, identity, record=None):
+        """Post publish handler."""
+        service = current_rdm_records.records_service
```
ppanero, line 256: my bad, I overlooked this. The `components` have the `service` in them (i.e. `self.service`) so there is no need to access it through the proxy. See [here](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/components.py#L22)

invenio_rdm_records/services/components/pids.py
```diff
@@ -244,10 +243,27 @@ def edit(self, identity, draft=None, record=None):
...
+        record_pids = record.get('pids', {})
+        for scheme, pid_dict in record_pids.items():
+            provider_name = pid_dict.get("provider")
+            identifier_value = pid_dict.get("identifier")
```
ppanero, line 260: At this point, *after published*, pids **must** have the `provider` and `identifier` attributes. So we can access with `[]` so it fails earlier (not below with a `PIDDoesNotExists` later on)

invenio_rdm_records/services/components/pids.py
```diff
@@ -244,10 +243,27 @@ def edit(self, identity, draft=None, record=None):
...
+            provider = service.pids.get_provider(
+                scheme, provider_name
+            )
```
ppanero, line 263: fits in one line?

tests/services/pids/providers/test_datacite_pid_provider.py
```diff
@@ -37,6 +39,19 @@ def record(location):
...
 
+@pytest.fixture(scope="function")
+def record_w_links(running_app, minimal_record):
+    """Creates an empty record."""
+    service = current_rdm_records.records_service
+    draft = service.create(system_identity, minimal_record)
+    record = service.publish(draft.id, system_identity)
+    record = current_rdm_records.records_service.read(
+        record["id"], system_identity
+    )
```
ppanero, line 50: why is the `read` needed? `publish` should already be returning the record?

jrcastro2: You are right! Thanks!

invenio_rdm_records/services/permissions.py
```diff
@@ -70,8 +71,15 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
...
     #
-    can_pid_reserve = can_curate
-    can_pid_delete = can_curate
+    can_pid_create = can_curate
+    can_pid_discard = can_curate  # discard an unregistered pid
+    can_pid_delete = [
+        Admin(),
+        SuperUser(),
+        SystemProcess(),
+        RecordOwnersIfExternalPID(),
```
lnielsen, line 80: I would follow a but more closely the what we have for public vs restricted records:

``IfProvider(scheme='doi', provider='external', then_=can_curate)``

This will more easily allow others to reuse it I tihnk.

Similarly below:

``IfProvider(scheme='doi', provider='datacite', then_=IfPIDstatus(status=pidstatus.NEW, then_=can_curate))``

lnielsen (commented):
First review iteration to discuss IRL

ppanero:
@lnielsen 
Refactor done and tests passing at service and resource level (fails on some serialization tests due to fixtures scopes).


**What I like**
- Flow is way easier to follow from the component level 
- We have complete service methods, i.e. someone can re-use them.

**What I do not like**
- There is no consistent way for validation, we need more context than what we have at Marshmallow level. Soft vs hard validation this becomes complex/ad-hoc (see `_create_required`). I am thinking of adding a `hard/soft` argument to the validate function (not sure if/how it would look like).
- Calling `_` methods from the component. We need to do that to avoid having service methods that do not check permissions (1) and do not commit/answer with a result item (2). This would cause problems in performance (permissions) and in SQL versioning (out of order commits without a nested session).
- For the `update` we cannot make just one perm check since it is PID dependent. However, we can make the assumption that the user will have them if s/he has them for 1 PID (i.e. assume is record dependent).

**Extras**
- Generator names I'll change once we have decided on the rest 👼

invenio_rdm_records/services/permissions.py
ppanero: In the end the generators are not needed.

invenio_rdm_records/services/components/pids.py
```diff
@@ -9,177 +9,257 @@
...
+        # validate new pids
+        self._validate_pids_schemes(data_pids)
+        self._validate_pids(data_pids, record, errors)
+
+        # update content
+        for scheme, pid in existing_pids.items():
+            updated_pid = data_pids.get(scheme)
+            if updated_pid != pid:
+                try:  # record and scheme reach the generator as "over"
+                    self.service.require_permission(
```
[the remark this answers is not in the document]
ppanero: I haven't checked. But...
- Since the generators are not needed in the end, and it is checked at *provider* level there are not many queries done (just one per PID, to get it and check status).
- Per service method, only one perm check is done (as in any other service method).

PR #818, opened 2021-09-30
gh actions: disable runs on draft prs

ppanero:
You can see below how the tests won't run on this PR since it is a draft. Shall we also disable CodeQL?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,6 +25,7 @@ on:
 
 jobs:
   Tests:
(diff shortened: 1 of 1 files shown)


lnielsen:
Yes, let's disable CodeQL (right now it doesn't do anything). Regarding not running tests on drafts, I think the problem is only the hanging builds which I hope we've sorted now. Also, Max's work on support Flask 2 should help alleviate the problem. 

If there's a setting for it we could perhaps try to lower the max time a job can run to avoid having them idle for ever.

PR #828, opened 2021-11-10
records: initial integration of communities

lnielsen:
Still missing updates of the index mappings.

Depends on https://github.com/inveniosoftware/invenio-communities/pull/361

### DIFF
--- a/invenio_rdm_records/services/schemas/parent/access.py
+++ b/invenio_rdm_records/services/schemas/parent/access.py
@@ -9,6 +9,8 @@
 
 """Access schema for RDM parent record."""
 
+# TODO: Replace with invenio_records_resources.services.base.schema import *
+
 
 from datetime import timezone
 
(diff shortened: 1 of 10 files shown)


invenio_rdm_records/services/schemas/parent/access.py
```diff
@@ -9,6 +9,8 @@
 
 """Access schema for RDM parent record."""
 
+# TODO: Replace with invenio_records_resources.services.base.schema import *
```
max-moser, line 12: i.e. move the entire parent schema to `invenio-records-resources.base.schema`, also the schema for the secret links?
should the secret link code also be moved?

max-moser (approved):

invenio_rdm_records/services/schemas/parent/access.py
lnielsen: Initially I was thinking just the parts that make sense - specifically first AgentSchema, but we likely need more as well that could be reused.

Issue #831, opened 2021-11-08
vocabularies: integrate names


slint:
- [x] Ext.py: register service and resource.
- [x] setup.py: entry points

PR #832, opened 2021-11-22
vocabularies: add names

ppanero:
- requires https://github.com/inveniosoftware/invenio-vocabularies/issues/85
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/831


invenio_rdm_records/ext.py
```diff
@@ -91,12 +93,13 @@ def init_config(self, app):
...
         overriding_configurations = [
             'PREVIEWER_RECORD_FILE_FACTORY',
+            'VOCABULARIES_DATASTREAM_TRANSFORMERS',
         ]
 
         for k in dir(config):
             if k in supported_configurations or k.startswith('RDM_'):
                 app.config.setdefault(k, getattr(config, k))
-            if k in overriding_configurations and not app.config.get(k):
+            if k in overriding_configurations and app.config.get(k):
```
ppanero, line 102: @slint special attention to this. IMO the previous logic was faulty:
- If the variable is in `overriding_configurations` it should be _always_ added, existing previously or not.
- The previous behaviour was "set if it does not exists", which is basically the `setdefault`

invenio_rdm_records/fixtures/data/vocabularies.yaml
```diff
@@ -26,6 +26,9 @@ descriptiontypes:
 datetypes:
   pid-type: dat
   data-file: vocabularies/date_types.yaml
+names:
+  pid-type: nameid
+  data-file: vocabularies/names.yaml
```
ppanero, line 31: empty file to provide compatibility with previous `FixturesEngine`

tests/resources/vocabularies/test_names_vocabulary.py
```diff
@@ -0,0 +1,71 @@
...
+                "identifier": "0000-0001-8135-3489",
+                "scheme": "orcid"
+            }, {
+                "identifier": "gnd:4079154-3",
+                "scheme": "gnd"
+            }
+        ],
+        "affiliations": [
+            {
+                "name": "CustomORG"
```
ppanero, line 40: Does not tests the inner relation link with affiliations, but:
- We are testing it in invenio-vocabularies
- The transformer does not support searching and adding `id` anyway, so all will be added as `name`.

invenio_rdm_records/ext.py
ppanero: On a second round to this... I see two options:
- As before, overwrite explicitly in the `ext.py`, which is a functionality that seems to be deprecated.
- Do it in `invenio-app-rdm` (where the config entrypoint is registered), which is the "glue" module. However, this means that `invenio-rdm-records` will be somehow "incomplete" by itself (for tests it is fixable by overwriting the `app_config` fixture).
    - If we do it via `invenio-app-rdm`, would this be overwritable via `invenio.cfg` file? If I understand the [docs](https://invenio-config.readthedocs.io/en/latest/usage.html#factory-pattern) correctly, it would not since the module (`invenio-vocabularies) would be loaded before the cfg file, no?

ppanero: @lnielsen 

What would be the desired way of overwriting/adding config from other modules in here? Asking this because is a recurrent question and will hit us again in funders/grants.

- As before, overwrite explicitly in the `ext.py`, which is a functionality that seems to be deprecated.
- Do it in `invenio-app-rdm` (where the config entrypoint is registered), which is the "glue" module. However, this means that `invenio-rdm-records` will be somehow "incomplete" by itself (for tests it is fixable by overwriting the `app_config` fixture).
    - If we do it via `invenio-app-rdm`, would this be overwritable via `invenio.cfg` file? If I understand the [docs](https://invenio-config.readthedocs.io/en/latest/usage.html#factory-pattern) correctly, it would not since the module (`invenio-vocabularies) would be loaded before the cfg file, no?

lnielsen: My five cents:

- Extensions defines **default configuration** - i.e. they only make sure a given configuration variable has been set.
- Setting a configuration variable should be done in ``ENV``, ``invenio.cfg`` or ``invenio_config.module`` (i.e. App-RDM).

Above means, Invenio-RDM-Records should not try to set ``VOCABULARIES_DATASTREAM_TRANSFORMERS`` in ``ext.py``. The options then is:

1. Set in ``invenio_app_rdm/config.py`` AND in ``app_config`` test fixture in RDM-Records.
2. or, change Invenio-Vocabularies to use entry points.

It looks to me what you're trying to do is more likely (2). I.e. RDM-Records provides something to Invenio-Vocabularies. This is similar to RDM-Records providing JSONSchemas, Mappings, DB models etc for which we use entry points.

ppanero: Thanks, it makes sense. I will opt for implementing 1 due to time constrains, and the fact that this is how other contribs have been implemented. Nonetheless, I think 2 is the correct way beacuse otherwise `invenio-rdm-records` is left incomplete and cannot be use _as is_ without adding config (i.e. in `invenio-app-rdm`).

I will open a detailed issue in `invenio-vocabularies` to work on this.

ppanero:
This PR is ready @lnielsen @slint if I get approval I can merge/release. Tested:

- Bootstrap a new RDM instance with `-c master`  and app rdm v8
- User registration, search, record publishing, and new version publishing work
- The `/api/names` endpoint is accesible and vocabularies can be imported correctly. Tested with the zenodo subset:
```
invenio vocabularies import -v names -f zenodo-rdm/app_data/vocabularies-future.yaml
```

slint (approved):

PR #834, opened 2021-11-26
services: create, read, delete and submit review

lnielsen:


invenio_rdm_records/services/schemas/parent/communities.py
```diff
@@ -7,39 +7,7 @@
 
 """Community schema."""
 
-from datetime import timezone
```
lnielsen, line 10: Copy/paste error

PR #864, opened 2021-12-20
refactor: increase vocabulary prop lookup performance

rekt-hard:
adding filter to only match the provided id so it returns one or no item. 
iterating over all the items and comparing them with the provided id takes more time and the further back in the results the wanted item is, the longer it takes.

performance wise, this is at least 2 times quicker - even going up to 50 times quicker (as of right now) depending on the returned list. usually it is around 10 times quicker
Screenshot from 2021-12-20 13-55-05


fenekku (approved):
Awesome!

PR #865, opened 2022-01-06
services: refactor configuration customization

fenekku:
At its core, service config.py (and others really) is a place to
- define the config interface
- define how those config attributes are populated with special attention to
  the fact that InvenioRDM is inherently configurable per instance.

So the idea is to
- make that core functionality clear and predominant
- group related things together e.g. configuration customization is extracted
  out of ext.py completely and placed in customization.py
- favor independent composition of reusable configuration attribute definition
  (using data accessors, an already common pattern in InvenioRDM)

### DIFF
--- a/invenio_rdm_records/ext.py
+++ b/invenio_rdm_records/ext.py
@@ -26,7 +26,6 @@ from . import config
 from .resources import RDMDraftFilesResourceConfig, \
     RDMParentRecordLinksResource, RDMParentRecordLinksResourceConfig, \
     RDMRecordFilesResourceConfig, RDMRecordResource, RDMRecordResourceConfig
-from .searchconfig import SearchConfig
 from .secret_links import LinkNeed, SecretLink
 from .services import RDMFileDraftServiceConfig, RDMFileRecordServiceConfig, \
     RDMRecordService, RDMRecordServiceConfig, SecretLinkService
@@ -129,49 +128,11 @@ class InvenioRDMRecords(object):
 
     def service_configs(self, app):
         """Customized service configs."""
-        # Overall record permission policy
-        permission_policy = app.config.get('RDM_PERMISSION_POLICY')
-        pid_providers = app.config['RDM_PERSISTENT_IDENTIFIER_PROVIDERS']
-        pids = app.config['RDM_PERSISTENT_IDENTIFIERS']
-        doi_enabled = app.config['DATACITE_ENABLED']
-
-        # Available facets and sort options
-        sort_opts = app.config.get('RDM_SORT_OPTIONS')
-        facet_opts = app.config.get('RDM_FACETS')
-
-        # Specific configuration for each search endpoint
-        search_opts = SearchConfig(
-            app.config.get('RDM_SEARCH'),
-            sort=sort_opts,
-            facets=facet_opts,
-        )
-        search_drafts_opts = SearchConfig(
-            app.config.get('RDM_SEARCH_DRAFTS'),
-            sort=sort_opts,
-            facets=facet_opts,
-        )
-        search_versions_opts = SearchConfig(
-            app.config.get('RDM_SEARCH_VERSIONING'),
-            sort=sort_opts,
-            facets=facet_opts,
-        )
 
         class ServiceConfigs:
-            record = RDMRecordServiceConfig.customize(
-                permission_policy=permission_policy,
-                pid_providers=pid_providers,
-                pids=pids,
-                doi_enabled=doi_enabled,
-                search=search_opts,
-                search_drafts=search_drafts_opts,
-                search_versions=search_versions_opts,
-            )
-            file = RDMFileRecordServiceConfig.customize(
-                permission_policy=permission_policy,
-            )
-            file_draft = RDMFileDraftServiceConfig.customize(
-                permission_policy=permission_policy,
-            )
+            record = RDMRecordServiceConfig.build()
+            file = RDMFileRecordServiceConfig.build()
+            file_draft = RDMFileDraftServiceConfig.build()
             affiliations = AffiliationsServiceConfig
             subjects = SubjectsServiceConfig
 
--- a/invenio_rdm_records/services/config.py
+++ b/invenio_rdm_records/services/config.py
@@ -28,7 +28,8 @@ from ..records import RDMDraft, RDMRecord
 from . import facets
 from .components import AccessComponent, MetadataComponent, PIDsComponent, \
     RelationsComponent, ReviewComponent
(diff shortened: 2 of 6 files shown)


invenio_rdm_records/services/customizations.py
```diff
@@ -41,57 +41,133 @@ def customize(cls, opts):
...
+    def __set__(self, obj, value):
+        """Set value on grafted_field of obj (descriptor protocol)."""
+        # If we want to allow setting it we can implement this.
+        pass
+
+
+class FromConfigPIDsProviders:
+    """Data descriptor for pid providers configuration."""
+
+    def __get__(self, obj, objtype=None):
```
max-moser, line 92: one of the effects of using descriptors here would be that the service config properties would be loaded lazily (and in fact, the `__get__(...)` would be re-executed on each access) instead of eagerly as it was until now

this means that we need to be careful regarding how heavy the operation of fetching and building the config properties is (not an issue here, i suppose) and maybe do some caching
but this kind of behavior would also allow for `app.config` to change during run-time, and have the changes take effect without restarting the application :thinking: 

just a rough thought:
if it would become a thing to populate parts of `app.config` via Vault (a secrets management engine) during runtime, this kind of approach could come in handy with [Vault's dynamic secrets](https://learn.hashicorp.com/tutorials/vault/getting-started-dynamic-secrets)

invenio_rdm_records/services/customizations.py
```diff
@@ -41,57 +41,133 @@ def customize(cls, opts):
...
+        return self.search_option_cls.customize(search_config)
+
+
+class ConfiguratorMixin:
+    """Shared customization for requests service config."""
+
+    @classmethod
+    def build(cls):
+        """Build the config object."""
+        return type(f"Custom{cls.__name__}", (cls,), {})
```
max-moser, line 173: question: what does this do?
aren't the config properties now calculated lazily when they're accessed via the descriptors, removing the necessity for an eager build?

max-moser (approved):

invenio_rdm_records/services/pids/service.py
```diff
@@ -19,7 +19,19 @@ class PIDsService(RecordService):
...
+    def _manager(self):
+        """Transitive manager property.
+
+        This is done to:
+        - only access self.config attributes when in an application context
+        - limit code change
+        - limit side-effects (in case using pre-existing `pid_manager` would
+                              cause some.)
+        """
+        return self.manager_cls(self.config.pids_providers)
```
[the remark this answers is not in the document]
fenekku: That's right. The way it worked before, the constructor was setting the manager (making that `manager_cls(self.config.pids_providers)`) with a passed `config` that had just been customized in `ext.py`. Since we moved that customization out and it does the work of connecting to the `app.config`, we need to delay the creation of the manager (the `_manager` property) until the service is used i.e. until the app context is pushed which is always the case when a service is used (but not when it's created).

invenio_rdm_records/services/customizations.py
fenekku: Hmmm if I recall, I was just keeping the additional layer that was already there here. Since we were creating this class before via `_make_cls` I kept it around. I'll make a ticket to investigate if this can be removed or if we somehow needed this additional layer for some reason: https://github.com/inveniosoftware/invenio-rdm-records/issues/872

invenio_rdm_records/services/customizations.py
```diff
@@ -41,57 +41,133 @@ def customize(cls, opts):
...
+        self.import_string = import_string
+
+    def __get__(self, obj, objtype=None):
+        """Return value that was grafted on obj (descriptor protocol)."""
+        if self.import_string:
+            return load_or_import_from_config(
+                key=self.config_key, default=self.default
+            )
+        else:
+            return current_app.config.get(self.config_key, self.default)
```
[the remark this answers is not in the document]
fenekku: Yes thanks for catching that! :+1:

invenio_rdm_records/services/customizations.py
fenekku: > one of the effects of using descriptors here would be that the service config properties would be loaded lazily (and in fact, the __get__(...) would be re-executed on each access) instead of eagerly as it was until now

Yes, I was thinking about that too. All configs are memory based operations (no disk or network access), so I am not worried about performance at all. The dynamic nature of it is a nice added benefit for the future. There would be some work to be done in [invenio-config](https://github.com/inveniosoftware/invenio-config) to really get there though.

Vaults sounds cool! Hashicorp stuff is really on point.

invenio_rdm_records/ext.py
```diff
@@ -129,49 +128,11 @@ def init_config(self, app):
...
-                search_drafts=search_drafts_opts,
-                search_versions=search_versions_opts,
-            )
-            file = RDMFileRecordServiceConfig.customize(
-                permission_policy=permission_policy,
-            )
-            file_draft = RDMFileDraftServiceConfig.customize(
-                permission_policy=permission_policy,
-            )
+            record = RDMRecordServiceConfig.build()
```
lnielsen, line 133: Given the issue with the app context, I think here the ``app`` has to be passed in:

```
RDMRecordServiceConfig.build(app)
```

invenio_rdm_records/services/config.py
```diff
@@ -95,20 +96,29 @@ class RDMRecordServiceConfig(RecordServiceConfig, RecordConfigMixin):
     schema_secret_link = SecretLink
 
     # Permission policy
-    permission_policy_cls = RDMRecordPermissionPolicy
+    permission_policy_cls = FromConfig(
+        "RDM_PERMISSION_POLICY", default=RDMRecordPermissionPolicy,
```
lnielsen, line 100: Comment: Should the default only be defined one place?

Right now it's in ``config.py`` and here?

invenio_rdm_records/services/customizations.py
```diff
@@ -41,57 +41,133 @@ def customize(cls, opts):
         return _make_cls(cls, attrs) if attrs else cls
 
 
-class FileConfigMixin:
-    """Shared customization for file service configs."""
+class FromConfig:
```
lnielsen, line 44: ``FromConfig*``s  ought to live in Invenio-Records-Resources so that it's reusable.

lnielsen (commented):
I have some further comments. So overall, I'm in favour of a more declarative API so that it's clear from the config class what is being loaded.

That said, I think the approach of evaluating the property upon access should be changed to building the config upon instantiation of the service as previously.

To discuss IRL probably 😄

PR #867, opened 2022-01-11
tests: remove network calls to datacite in tests

### LINKED ISSUE #866: Tests make requests to DataCite servers
Package version (if known): v0.33.0+

## Describe the bug
The test suite is very slow (`476 passed, 559 skipped, 666 warnings in 615.66s (0:10min15sec)` on my local machine) and upon inspection calls to datacite service are being made during tests. 

For example: `invenio-rdm-records/tests/resources/test_serialized_links.py` makes a call to register a public DOI for each test where the record is published. 

This pattern occurs more widely for every test that publishes, deletes, updates (published) records.

## Expected behavior
Not make external network calls during tests to speed up their execution and not mistakenly make remote changes.

## Sugested solution

Make a fake DOI provider and use it in all tests, so that it mimics the DataCite API while not making external calls.

fenekku:
closes #866 

This drops the test suite speed by ~45s from ~10min15sec to ~9min30sec .  This is good in a sense, but also underwhelming
given that the full test suite still takes that long :sob: . The app loading seems the biggest culprit.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2019-2021 CERN.
-# Copyright (C) 2019-2021 Northwestern University.
+# Copyright (C) 2019-2022 Northwestern University.
 # Copyright (C) 2021 TU Wien.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
@@ -42,6 +42,9 @@ from invenio_vocabularies.records.api import Vocabulary
 from invenio_rdm_records import config
 from invenio_rdm_records.proxies import current_rdm_records
 from invenio_rdm_records.records.api import RDMDraft, RDMParent, RDMRecord
+from invenio_rdm_records.services.pids import providers
+
+from .fake_datacite_client import FakeDataCiteClient
 
 
 #
@@ -184,6 +187,11 @@ def celery_config():
     return {}
 
 
+def _(x):
+    """Identity function for string extraction."""
+    return x
+
+
 @pytest.fixture(scope='module')
 def app_config(app_config):
     """Override pytest-invenio app_config fixture.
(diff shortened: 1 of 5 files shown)


tests/fake_datacite_client.py
```diff
@@ -0,0 +1,125 @@
...
+        else:
+            self.api_url = url or "https://api.datacite.org/"
+
+        if not self.api_url.endswith('/'):
+            self.api_url += '/'
+
+        self.timeout = timeout
+
+    def public_doi(self, metadata, url, doi=None):
+        """Create a public doi ... not.
```
max-moser, line 48: image

tests/conftest.py
```diff
@@ -184,6 +187,11 @@ def celery_config():
     return {}
 
 
+def _(x):
+    """Identity function for string extraction."""
+    return x
```
max-moser, line 192: do we need this?
AFAIK, this is usually just a placeholder for `from flask_babelex import lazy_gettext as _` when translations are planned, but not yet done... but i don't think translations will be relevant here in the future?

max-moser (approved):

tests/conftest.py
```diff
@@ -221,11 +229,32 @@ def app_config(app_config):
     # Enable communities while in preview
     app_config['COMMUNITIES_ENABLED'] = True
 
-    # Enable DOI miting
+    # Enable DOI minting...
     app_config['DATACITE_ENABLED'] = True
     app_config['DATACITE_USERNAME'] = 'INVALID'
     app_config['DATACITE_PASSWORD'] = 'INVALID'
     app_config['DATACITE_PREFIX'] = '10.1234'
```
max-moser, line 236: just out of curiosity: this configuration wouldn't have worked ever anyway, right?
so previously, the tests were sending requests to datacite's test fabrica that would get rejected because of invalid credentials?
but this wouldn't have an impact on the local DOI minting workflow?

tests/fake_datacite_client.py
fenekku: :smile: I couldn't put the exclamation point because of pycodestyle but I am glad someone got it.

tests/conftest.py
fenekku: > tests were sending requests to datacite's test fabrica that would get rejected because of invalid credentials?

Exactly. And since we don't seem to care about the feedback from DataCite, this is a silent error.
It wouldn't impact "local" minting because we mint before we send the request to DataCite.

tests/conftest.py
fenekku: This was copied from what is used in config.py (https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/config.py#L19) because the `_()` calls are used in the config we copied e.g., https://github.com/inveniosoftware/invenio-rdm-records/pull/867/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R256  . Just keeping that as is.

PR #876, opened 2022-01-26
Fixes error message.

chriz-uniba:


ppanero (approved):
wondering about the whole phrasing... shouldn't it be "errored"? but for the time being, fixing the ~updating~ is necessary :)

PR #877, opened 2022-01-26
Demo requests: create command to add random types of community submission requests

Pineirin:
Create command to add community membership requests.
Create command to add community invitation requests.
Allow using the user for request creation commands.
Create tests for the above mentioned commands.

Commands implemented:
- `invenio rdm-records demo`. Creates a batch of demo data, including records and requests of different types (mostly, community submission requests with different status).
- `invenio rdm-records generate records`. Creates a batch of records given a optional user and number of records.
- `invenio rdm-records generate submission_requests`. Creates a bunch of community submission requests. Each request is created for the review of a draft in a community (both the draft and the community are generated by this command). The command takes as optional parameters a user used to create, the number the communities to create and the number of drafts for each community.
- `invenio rdm-records generate membership_requests`. Creates a bunch of community membership requests. Each request is created for a community (generated by the command). The command takes as optional parameters a user used to create, the number the communities to create and the number of requests for each community.
- `invenio rdm-records generate invitation_requests`. Creates a bunch of community invitation requests. Each request is created for a community (generated by the command). The command takes as optional parameters a user used to create, the number the communities to create and the number of requests for each community.

closes: https://github.com/inveniosoftware/invenio-rdm-records/issues/878


invenio_rdm_records/fixtures/tasks.py
```diff
@@ -30,3 +32,46 @@ def create_demo_record(data):
...
+
+    # Create requests for the draft
+    for _ in range(n_request):
+        _create_requests.delay(records_service, draft)
+
+
+@shared_task
+def _create_requests(records_service, draft):
+    # Create a demo request
+    records_service.review.submit(system_identity, draft.id)
```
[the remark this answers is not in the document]
kprzerwa: additionally it would be good to create not only the review request but also community invitation, since we have both in the mockups

kprzerwa, line 77: I still see the system identity here? I thought it was not correct, did something change?

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -30,3 +32,46 @@ def create_demo_record(data):
...
+@shared_task
+def create_demo_request(community_data, drafts_data, n_drafts=1, n_requests=1):
+    """Create n_requests requests for n_drafts drafts in a new community."""
+    # Create a demo community
+    community = current_communities.service.create(
+        system_identity,
+        community_data
+    )
+    Community.index.refresh()
+    records_service = current_rdm_records.records_service
```
kprzerwa, line 46:
```
from invenio_rdm_records.proxies import current_rdm_records_service
```
you can import directly `current_rdm_records_service` from proxies

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -27,6 +34,187 @@ def create_vocabulary_record(service_str, data):
...
+    if (random_choice > 0):
+        options[random_choice](draft, community, records_service)
+
+
+def _open_request(draft, community, records_service):
+    # Create a demo community submission request with open status.
+    records_service.review.submit(system_identity, draft.id)
+
+
+def _accept_request(draft, community, records_service):
```
kprzerwa, line 107: could you also add a comment event type?

invenio_rdm_records/fixtures/tasks.py
kprzerwa: I can still see here system_identity
as we discussed yesterday it would be good to have community manager, regular user to see those requests from different points of views

invenio_rdm_records/requests/community_invitation.py
```diff
@@ -0,0 +1,21 @@
...
+"""Community submission request."""
+
+from flask_babelex import lazy_gettext as _
+from invenio_requests.customizations import DefaultRequestType
+
+
+#
+# Request
+#
+class CommunityInvitation(DefaultRequestType):
```
kprzerwa, line 17: was this discussed ? maybe this class exists somewhere already?

invenio_rdm_records/requests/community_submission.py
```diff
@@ -18,6 +18,21 @@
 #
 # Actions
 #
+class CommentAction(RequestAction):
```
kprzerwa, line 21: you don't need to create the class for Comment action, it is already implemented but not explicitly as a separate class. Check the requests service, it basically needs only using the create method with a specific type in the payload
```
"type": RequestEventType.COMMENT.value
```

invenio_rdm_records/cli.py
```diff
@@ -39,6 +42,86 @@ def demo():
...
+            fake_community,
+            fake_requests,
+            n_requests=n_requests,
+        )
+
+    click.secho('Created requests!', fg='green')
+
+
+@rdm_records.command('demo_invitation_requests')
+@click.option("-u", "--user", type=click.STRING)
```
kprzerwa, line 99: do we pass user email or id ?

invenio_rdm_records/cli.py
```diff
@@ -39,6 +42,86 @@ def demo():
     click.secho('Created records!', fg='green')
 
 
+@rdm_records.command('demo_submission_requests')
+@click.option("-u", "--user", type=click.STRING)
+@with_appcontext
+def demo_submission_requests(user=system_identity.id):
```
ntarocco, line 48: Open question: should this be moved above in the `demo` func? Otherwise, we will have to document that this extra step should also be done.

invenio_rdm_records/cli.py
```diff
@@ -39,6 +42,86 @@ def demo():
...
+    # Create 25 drafts in each community.
+    n_drafts = 25
+    # Create 4 communties.
+    for _ in range(4):
+        fake_community = create_fake_community()
+        fake_drafts = []
+        for _ in range(n_drafts):
+            fake_drafts.append(create_fake_draft())
+        create_demo_community_submission_request.delay(
+            user,
```
ntarocco, line 61: IMO, the user should be hardcoded here, not passed as a param. As a dev/user, I have no idea what user to pass.

invenio_rdm_records/cli.py
```diff
@@ -39,6 +42,86 @@ def demo():
...
+            n_requests=1
+        )
+
+    click.secho('Created requests!', fg='green')
+
+
+@rdm_records.command('demo_membership_requests')
+@click.option("-u", "--user", type=click.STRING)
+@with_appcontext
+def demo_membership_requests(user=system_identity.id):
```
ntarocco, line 74: Same comment on this func here.

invenio_rdm_records/cli.py
```diff
@@ -39,6 +42,86 @@ def demo():
...
+            n_requests=n_requests,
+        )
+
+    click.secho('Created requests!', fg='green')
+
+
+@rdm_records.command('demo_invitation_requests')
+@click.option("-u", "--user", type=click.STRING)
+@with_appcontext
+def demo_invitation_requests(user=system_identity.id):
```
ntarocco, line 101: Same comment on this func here.

I would rather split all this in inner private func, but only have one `demo`. As a dev/user, I have no idea on which order I should run these various commands.

invenio_rdm_records/fixtures/demo.py
```diff
@@ -312,3 +312,61 @@ def create_fake_record():
...
+    fake = Faker()
+    return {
+        "id": fake.unique.domain_word(),
+        "access": {
+            "visibility": "public",
+        },
+        "metadata": {
+            "title": fake.sentence(nb_words=5, variable_nb_words=True),
+            "type": random.choice([
+                "organization", "event", "topic", "project"
```
ntarocco, line 328: Is this en enum already defined somewhere else?

invenio_rdm_records/fixtures/demo.py
```diff
@@ -312,3 +312,61 @@ def create_fake_record():
...
+            ]),
+        }
+    }
+
+
+def create_fake_draft():
```
ntarocco, line 334: don't we already have a `fake_data` record used in the `demo` func?

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -27,6 +36,241 @@ def create_vocabulary_record(service_str, data):
...
+        user_id,
+        community_data,
+        drafts_data,
+        n_drafts=1,
+        n_requests=1
+):
+    """Create n_requests community_submission requests for n_drafts drafts."""
+    """Drafts are created in a new community."""
+    # Create a demo community
+    user = current_datastore.get_user(user_id)
```
ntarocco, line 55: try/catch + nice error if user does not exist? Only if the user was passed as a param: if hardcoded, no need because it should always exist.

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -27,6 +36,241 @@ def create_vocabulary_record(service_str, data):
...
+            'type': 'community-submission',
+            'receiver': {'community': community.data['uuid']}
+        }
+    }
+
+    draft = records_service.create(
+        data=draft_data, identity=user
+    )
+
+    # Create requests for the draft
```
ntarocco, line 95: Please be more explicit:
```suggestion
    # Create a new request to incude a draft to a community
```

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -27,6 +36,241 @@ def create_vocabulary_record(service_str, data):
...
+        _create_community_submission_request.delay(
+            user_id,
+            records_service,
+            draft,
+            community
+        )
+
+
+@shared_task
+def _create_community_submission_request(
```
ntarocco, line 106: Suggestion:
```suggestion
def _create_draft_inclusion_request_for_community(
```

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -27,6 +36,241 @@ def create_vocabulary_record(service_str, data):
...
+        _accept_request,
+        _draft_request,
+        _comment_request
+    ]
+    if (random_choice > 0):
+        options[random_choice](draft, community, records_service, user)
+
+
+def _open_request(draft, community, records_service, user):
+    # Create a demo community submission request with open status.
```
ntarocco, line 126: Prefer doc strings instead: """Create a ...."""

invenio_rdm_records/requests/community_invitation.py
ntarocco, line 17: Shouldn't this be in `invenio-communities`?

invenio_rdm_records/requests/community_membership.py
```diff
@@ -0,0 +1,21 @@
...
+"""Community submission request."""
+
+from flask_babelex import lazy_gettext as _
+from invenio_requests.customizations import DefaultRequestType
+
+
+#
+# Request
+#
+class CommunityMembership(DefaultRequestType):
```
ntarocco, line 17: Shouldn't this be in `invenio-communities`?

invenio_rdm_records/cli.py
Pineirin: Both are accepted.

invenio_rdm_records/requests/community_invitation.py
Pineirin: I think it shouldn't.
This is a not part of the logic of a community, is just a request type that just happens to be related to a community.
Maybe it could go into a invenio-requests folder, but, as far as I understand, the non default request types are defined by the package needs the new type of request, and existence of this requests folder in invenio_rdm_records seems to point in that direction.

invenio_rdm_records/requests/community_membership.py
Pineirin: Same as above

invenio_rdm_records/fixtures/demo.py
Pineirin: There is a TODO for injecting this from the vocabularies, but no enum can be found of the configuration right now.

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -25,8 +39,234 @@ def create_vocabulary_record(service_str, data):
 
 
 @shared_task
-def create_demo_record(data):
+def create_demo_record(data, user_id):
```
kpsherva, line 42: how do you make sure the user exists for given `user_id`?

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -25,8 +39,234 @@ def create_vocabulary_record(service_str, data):
...
+
+@shared_task
+def create_demo_community_submission_request(
+        user_id,
+        community_data,
+        drafts_data
+):
+    """Create a community_submission requests for the param drafts."""
+    """Drafts are created in a new community."""
+    # Create a demo community
```
kpsherva, line 59: two separate docstrings

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -25,8 +39,234 @@ def create_vocabulary_record(service_str, data):
...
+        )
+
+
+@shared_task
+def _create_draft(
+        user_id,
+        community_uuid,
+        draft_data
+):
+    """Create a community_submission request."""
```
kpsherva, line 81: two separate docstrings

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -25,8 +39,234 @@ def create_vocabulary_record(service_str, data):
...
+    """The request is created for a draft."""
+    # Create a demo draft
+    identity = _identity(user_id)
+    identity.provides.add(CommunityNeed(community_uuid))
```
kpsherva, line 85: I am not sure which use case this covers, I would guess the user does not need a community membership to create a draft, correct me if I am wrong, but that's why we have the requests, so that anyone can create a draft and then request to add it to the community?

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -25,8 +39,234 @@ def create_vocabulary_record(service_str, data):
...
+    current_rdm_records_service.review.update(
+        identity,
+        draft_id,
+        data,
+        revision_id=revision_id
+    )
+
+
+def _comment_request(draft_id, revision_id, community_uuid, identity):
+    """Set the status of the community submission request to accept status."""
```
kpsherva, line 151: docstring is mismatching the functions purpose

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -25,8 +39,234 @@ def create_vocabulary_record(service_str, data):
...
+        )
+
+
+@shared_task
+def _create_community_invitation_request(
+        user_id,
+        community_uuid,
+        data,
+        index,
+):
```
kpsherva, line 244: docstring missing

invenio_rdm_records/requests/community_invitation.py
kpsherva: I also think it should be in the communities. Like you said, it is related to community logic, and it does not refer to `rdm_records`, so in my opinion, it should belong to invenio-communities.

invenio_rdm_records/fixtures/demo.py
kpsherva: @Pineirin this comment was not addressed

kpsherva (approved):
LGTM! Do we add it to the setup, or for now it's optional?

jrcastro2 (commented):
LGTM!

Issue #880, opened 2022-01-27
DataCite metadata formats on OAI-PMH


slint:
See [comment](https://github.com/inveniosoftware/product-rdm/issues/27#issuecomment-1009793303)

mitsosf:
Two PRs

invenio-oaiserver: [213](https://github.com/inveniosoftware/invenio-oaiserver/pull/213)
invenio-app-rdm: [1205](https://github.com/inveniosoftware/invenio-app-rdm/pull/1205)

The config in invenio-oaiserver is overwritten by the config in invenio-app-rdm.

PR #888, opened 2022-02-04
oai: add datacite and oai_datacite support

mitsosf:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/880


invenio_rdm_records/oai.py
```diff
@@ -27,6 +30,46 @@ def dublincore_etree(pid, record):
...
+    nsmap = {
+        None: "http://schema.datacite.org/oai/oai-1.1/",
+        "xsi": "http://www.w3.org/2001/XMLSchema-instance",
+    }
+
+    attribs = {
+        f"{{{nsmap['xsi']}}}schemaLocation": "http://schema.datacite.org/"
+                                             "oai/oai-1.1/ "
+                                             "http://schema.datacite.org/"
+                                             "oai/oai-1.1/oai.xsd",
```
slint, line 53:
```suggestion
        f"{{{nsmap['xsi']}}}schemaLocation": (
            "http://schema.datacite.org/oai/oai-1.1/ "
            "http://schema.datacite.org/oai/oai-1.1/oai.xsd"
        ),
```

I tried with `black` as well, but didn't get anything more intuitive. So in the end "hand-splitted" looks the most logical I think.

invenio_rdm_records/oai.py
```diff
@@ -27,6 +30,46 @@ def dublincore_etree(pid, record):
...
 
+def datacite_etree(pid, record):
+    """Dump the record (search result) into a DataCite XML format for OAI-PMH."""
+    data_dict = DataCite43XMLSerializer().dump_one(record["_source"])
+    return schema43.dump_etree(data_dict)
+
+
+def oai_datacite_etree(pid, record):
+    """Dump the record (search result) into a OAI DataCite XML format for OAI-PMH."""
+    datacentre_symbol_text = None
```
slint, line 41: - This is one of the things that should be configurable. For Zenodo e.g. that is [`CERN.ZENODO`](https://zenodo.org/oai2d?verb=ListRecords&metadataPrefix=oai_datacite), which is our "datacenter symbol" on DataCite (see from [their REST API](https://api.datacite.org/clients/cern.zenodo)).
- The variable itself also doesn't need to be initialized up here, but just directly get set below at `datacentre_symbol.text = current_app.config.get('DATACITE_DATACENTER_SYMBOL', None)`.
- For the config name, there's already a [`DATACITE_` set of variables used](https://github.com/inveniosoftware/invenio-rdm-records/blob/436bd2f4f6a75713db87ebf19d9cb962176d3899/invenio_rdm_records/ext.py#L105-L110), which we might actually piggy-back on.

invenio_rdm_records/oai.py
```diff
@@ -27,6 +30,46 @@ def dublincore_etree(pid, record):
     return DublinCoreXMLSerializer().serialize_object_xml(record["_source"])
 
 
+def datacite_etree(pid, record):
```
slint, line 33: I think it's time we also add some test coverage for this file in general... I didn't find any tests for this part, so if that's really the case, you can kick-off a `tests/tests_oai.py` :)

slint (commented):
LGTM, one point on the configuration, and about adding tests for the entire file.

invenio_rdm_records/oai.py
ppanero: fixed

invenio_rdm_records/oai.py
ppanero: done

invenio_rdm_records/config.py
```diff
@@ -523,3 +523,6 @@ def make_doi(prefix, record):
 
     DATACITE_FORMAT = make_doi
 """
+
+DATACITE_DATACENTER_SYMBOL = ""
```
ppanero, line 527: @slint default to empty?

tests/test_oai.py
```diff
@@ -0,0 +1,233 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Test OAI-PMH support."""
+
+from lxml import etree
+
+from invenio_rdm_records.oai import datacite_etree, dublincore_etree, \
+    oai_datacite_etree
+
+# This tests would ideally be E2E. However, due to the lack of assets building
```
ppanero, line 16: @slint note this

tests/test_oai.py
```diff
@@ -0,0 +1,233 @@
...
+        '        <fundingReference>\n'
+        '          <funderName>European Commission</funderName>\n'
+        '          <funderIdentifier funderIdentifierType="ROR">1234</funderIdentifier>\n'  # noqa
+        '          <awardNumber>246686</awardNumber>\n'
+        '          <awardTitle>OpenAIRE</awardTitle>\n'
+        '        </fundingReference>\n'
+        '      </fundingReferences>\n'
+        '    </resource>\n'
+        '  </payload>\n'
+        '</oai_datacite>\n'
```
ppanero, line 226: @slint is this wrapping correct?

invenio_rdm_records/config.py
slint: Yes, It seems that from the [XML Schema](http://schema.datacite.org/oai/oai-1.0/oai.xsd), the element is required, but there's nothing about the text...

tests/test_oai.py
slint: Yes, see also an [example from Zenodo](https://zenodo.org/oai2d?verb=GetRecord&metadataPrefix=oai_datacite&identifier=oai:zenodo.org:1228465)

PR #897, opened 2022-02-17
permissions: define ones over records for community members

### LINKED ISSUE #890: Add membership permissions to Record permission policy

fenekku:
- depends on https://github.com/inveniosoftware/invenio-requests/pull/136
- depends on https://github.com/inveniosoftware/invenio-communities/pull/410
- closes #890


fenekku:
I am on it.

lnielsen (changes requested):
I'm blocking just make sure we don't merge the PR now. After discussion with @ntarocco we think the best approach at this point is we revert communites/requests so that we have as few possible issues for the release. The PR will be taken up again after the release.

PR #901, opened 2022-02-25
babelex: use lazy_gettext instead of _= lambda x:x

mb-wali:
closes inveniosoftware/invenio-app-rdm#1303


kpsherva (commented):
thats great!
are there any not desired consequences of this change?

kpsherva (approved):

PR #907, opened 2022-03-02
fixtures: add custom names

ppanero:
closes https://github.com/inveniosoftware/cookiecutter-invenio-rdm/issues/204

Support names with the _old_ fixtures mechanisms, due to lack of migration to new _datastreams_. Changes are:

- Base methods are moved to `VocabularyEntry`. Previously this class was merely an interface, which lack of implementation of certain methods.
- Change the `iterate` method to avoid adding `type` to non-generic vocabularies
- Names file with 6585 entries (all checked, non failing entries).

After running `services setup` and `run`, there is no error in the console. Names can be found via REST API and search works as expected (therefore other vocabularies were also loaded correctly, since demo data uses those).

Screenshot 2022-03-02 at 13 46 54

Screenshot 2022-03-02 at 13 47 16

For example, subjects and resource types are correctly loaded:

Screenshot 2022-03-02 at 13 52 14

Screenshot 2022-03-02 at 13 52 03


invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -344,9 +348,9 @@ def create_record(self, data, delay=False):
 class GenericVocabularyEntry(VocabularyEntry):
     """Vocabulary fixture with single data-file."""
 
-    def __init__(self, directory, id_, entry):
+    def __init__(self, directory, id_, entry, service_str="vocabulary_service"):
```
ppanero, line 351: This is not the most beautiful change ever, but:
- is back compat
- allows usage of custom service for vocabularies without schemes

Alternative: move all templated methods to `VocabularyEntry`. This would be better, but since we are moving to DataStreams soon, maybe is better not to make big changes here.

PR #935, opened 2022-04-07
tasks: add vocabulary updates propagation

ppanero:
- closes


ppanero:
This seems to be a stale and incomplete PR, the work for change notifications was already merged

PR #937, opened 2022-04-12
Communities: create search community records endpoint

Pineirin:
Define tests for the created endpoint.
closes: inveniosoftware/invenio-communities#408

### DIFF
--- a/invenio_rdm_records/records/mappings/v7/rdmrecords/records/record-v4.0.0.json
+++ b/invenio_rdm_records/records/mappings/v7/rdmrecords/records/record-v4.0.0.json
@@ -37,6 +37,16 @@
           "id" : {
             "type" : "keyword"
           },
+          "communities": {
+            "properties": {
+              "default": {
+                "type": "keyword"
+              },
+              "ids": {
+                "type": "keyword"
+              }
+            }
+          },
           "access": {
             "properties" : {
               "owned_by" : {
--- a/invenio_rdm_records/resources/resources.py
+++ b/invenio_rdm_records/resources/resources.py
@@ -15,6 +15,7 @@ from flask_resources import resource_requestctx, response_handler, route
 from invenio_drafts_resources.resources import RecordResource
 from invenio_records_resources.resources.records.resource import \
     request_data, request_headers, request_search_args, request_view_args
+from invenio_records_resources.resources.records.utils import es_preference
 
 
 class RDMRecordResource(RecordResource):
@@ -35,7 +36,12 @@ class RDMRecordResource(RecordResource):
             route("GET", p(routes["item-review"]), self.review_read),
             route("PUT", p(routes["item-review"]), self.review_update),
             route("DELETE", p(routes["item-review"]), self.review_delete),
-            route("POST", p(routes["item-actions-review"]), self.review_submit)
+            route("POST", p(routes[
+                "item-actions-review"
+                ]), self.review_submit),
+            route("GET", routes[
+                "community-records"
+                ], self.search_community_records)
         ]
 
         return url_rules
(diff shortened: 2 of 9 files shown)


invenio_rdm_records/records/mappings/v7/rdmrecords/records/record-v4.0.0.json
```diff
@@ -37,6 +37,13 @@
           "id" : {
             "type" : "keyword"
           },
+          "communities": {
+            "properties": {
+              "default": {
+                "type": "keyword"
```
ntarocco, line 43: can you please add the `ids` field too? 
Did you check if there is any other field missing?

invenio_rdm_records/resources/resources.py
```diff
@@ -43,6 +44,22 @@ def p(route):
     #
     # Review request
     #
+    @request_search_args
+    @request_view_args
+    @response_handler(many=True)
+    def search_community_records(self):
```
ntarocco, line 50: Can you please move this down at the very bottom with a similar comment on top?
the comment here says `# Review request` and it is not releated to that

tests/services/test_rdm_service.py
```diff
@@ -132,3 +180,87 @@ def test_embargo_lift_with_error(running_app, es_clear, minimal_record):
...
+    )
+    draft = service.create(superuser_identity, minimal_record)
+    record = service.publish(id_=draft.id, identity=superuser_identity)
+
+    # Record should not be lifted since it didn't expire (until 3220)
+    with pytest.raises(EmbargoNotLiftedError):
+        service.lift_embargo(_id=record['id'], identity=superuser_identity)
+
+
+def prueba(app, location, db, es_clear, vocabularies, users):
```
ntarocco, line 219: ?

tests/services/test_rdm_service.py
```diff
@@ -132,3 +180,87 @@ def test_embargo_lift_with_error(running_app, es_clear, minimal_record):
     # Record should not be lifted since it didn't expire (until 3220)
     with pytest.raises(EmbargoNotLiftedError):
         service.lift_embargo(_id=record['id'], identity=superuser_identity)
+
+
+def test_embargo_lift_with_error(running_app, es_clear, minimal_record):
```
ntarocco, line 185: did you add these tests?

invenio_rdm_records/resources/resources.py
ntarocco, line 50: It does not look like that you have created a route for this endpoint, and there is no test for this resource. The endpoint cannot be used.
The URL, I think, should be `/communities/<pid_value>/records`

invenio_rdm_records/services/services.py
```diff
@@ -86,3 +88,38 @@ def scan_expired_embargos(self, identity):
             f"AND access.embargo.until:[* TO {today}]"
 
         return self.scan(identity=identity, q=embargoed_q)
+
+    def search_community_records(self, identity, community, params=None,
```
ntarocco, line 92: This signature does not match with how you are using it is the resource.

invenio_rdm_records/services/services.py
```diff
@@ -86,3 +88,38 @@ def scan_expired_embargos(self, identity):
...
+                                 es_preference=None, **kwargs):
+        """Search for drafts records matching the querystring."""
+        community_uuid = community.to_dict()['uuid']
```
ntarocco, line 95: You are already getting the `id` from the resource, not the community.

invenio_rdm_records/services/services.py
```diff
@@ -86,3 +88,38 @@ def scan_expired_embargos(self, identity):
...
+
+        search_result = self._search(
+            "search",
+            identity,
+            params,
+            es_preference,
+            record_cls=self.record_cls,
+            search_opts=self.config.search,
+            extra_filter=Q(
+                "term", **{"parent.communities.default": str(community_uuid)}
```
ntarocco, line 109: I am not sure this is enough. A record can be in the default community, or in one of the `ids` list.
Please add tests for both cases.

ntarocco (commented):
Much better, good job! Added some more comments because I think that there is still something to add.

Pineirin (commented):
Why was this appearing as an error to you but not to me?

zzacharo (approved):

PR #949, opened 2022-04-27
Communities relations

alejandromumo:
closes https://github.com/inveniosoftware/invenio-communities/issues/500


tests/fixtures/data/vocabularies.yaml
```diff
@@ -19,3 +19,6 @@ descriptiontypes:
 relationtypes:
   pid-type: rlt
   data-file: vocabularies/relation_types.yaml
+communitytypes:
```
ppanero, line 22: is this needed? IIRC we overwrite in tests when we wanted to load just a subset for tests (e.g. the main languages.yaml has 10k entries but we want just a few for tests). However, in the case of community types is still just 4. Doesn't it pick up defaults if not overwritten here?

tests/fixtures/test_fixtures.py
```diff
@@ -72,6 +72,28 @@ def test_load_resource_types(app, db, es_clear):
...
 
 
+def test_load_community_types(app, db, es_clear):
+    id_ = 'communitytypes'
+    resource_types = GenericVocabularyEntry(
+        Path(__file__).parent / "data",
+        id_,
+        {
+            "pid-type": "comtyp",
+            "data-file": "vocabularies/community_types.yaml"
```
ppanero, line 82: might be we just need to specify the package to use defaults? (might need fixtures/data in the path)
```suggestion
            "data-file": "invenio_rdm_records/vocabularies/community_types.yaml"
```

ppanero:
tests failing due to known `KeyError: ''`

```python
            if key_prefix:
                full_key = key_prefix + self._splitchar + full_key
            bucket_out = {
                "key": key,
                "doc_count": self.get_metric(bucket),
>               "label": label_map[key],
                "is_selected": self.is_filtered(full_key, filter_values)
            }
E           KeyError: ''
```

tests/fixtures/data/vocabularies.yaml
alejandromumo: tried this change locally but tests fail when importing the fixtures.

tests/fixtures/test_fixtures.py
alejandromumo: tried this change locally but tests fail when importing the fixtures.

tests/fixtures/data/vocabularies.yaml
ppanero: it's because the passed `dir` is the parent (`/tests/fixtures`) and therefore it won't look outside of it. But its not a problem, let's keep it as is.

ppanero (approved):

lnielsen (approved):

PR #959, opened 2022-05-06
serializers: use g.identity instead of system

ppanero:
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/958

Since all the occurrences happen at _resources_ level it is accepted to access the `g` object. The `current_user` does not provide an identity (it provides `id` and `roles`) and we would need to build the identity from it, which is already stored in `g.identity`.


ntarocco (approved):

ppanero:
It doesnt let me revert the PR, maybe someone with higher powers?

Issue #944, opened 2022-04-22
IIIF caching mechanishim


egabancho:
Port IIIF caching from [Flask-IIIF](https://github.com/inveniosoftware/flask-iiif/tree/master/flask_iiif/cache)

PR #989, opened 2022-06-26
IIIF enhacements

egabancho:

### Description

Please describe briefly your pull request.

### Checklist

**Third-party code**

**Reminder**


egabancho:
I still need to find a way to filter files when generating the canvas to only IIIF-compatible files.

egabancho:
I'm not very proud of the filtering part on the manifest or the links. If you can think of a better way to do that, I'll be more than happy to change it.

invenio_rdm_records/resources/config.py
```diff
@@ -218,6 +218,14 @@ class IIIFResourceConfig(ResourceConfig):
...
+        "dl": ma.fields.Str(),
+    }
+
+    request_headers = {
+        "If-Modified-Since": ma.fields.DateTime(),
+    }
+
     response_handler = {"application/json": ResponseHandler(JSONSerializer())}
 
     supported_formats = {
```
ntarocco, line 231: I am a bit out of context, sorry... do you know we have `IIIF_FORMATS` config but also this `supported_formats`?
Should they match?

invenio_rdm_records/services/config.py
```diff
@@ -71,6 +74,12 @@ def has_doi(record, ctx):
     return "doi" in pids
 
 
+def is_iiif_conpatible(file_, ctx):
```
ntarocco, line 77:
```suggestion
def is_iiif_compatible(file_, ctx):
```

invenio_rdm_records/services/config.py
```diff
@@ -71,6 +74,12 @@ def has_doi(record, ctx):
...
+    """Determine if a file is IIIF compatible."""
+    file_ext = splitext(file_.key)[1].replace(".", "").lower()
```
ntarocco, line 79: Will the filename always have the extension? Basically, does the `[1]` always work?

ntarocco:
> I'm not very proud of the filtering part on the manifest or the links. If you can think of a better way to do that, I'll be more than happy to change it.

I think it is ok, it is just a function of the filename. Not sure it can be done any better?
The only thing I can think of is to add a cache at request level to avoid the check `is extension compatible) each time, but I don't think it is worth it, given that everything happens in memory.

invenio_rdm_records/services/config.py
egabancho: I think it should be alright https://docs.python.org/3/library/os.path.html#os.path.splitext

invenio_rdm_records/resources/config.py
egabancho: You are right, and I missed the `supported_formats`. 
Now I have a question for you 😄 
How can I access the resource config from the service? (I don't think I should 😂 )

Also, and answering your second question, `IIIF_FORMATS` it's used by Flask-IIIF to do its thing. A good alternative might be to remove the `supported_formats` and rely solely on `IIIF_FORMATS` everywhere. 
I can't think of any reason I added it in the first place ...

egabancho:
I made an issue for the caching some time ago, https://github.com/inveniosoftware/invenio-rdm-records/issues/944. Probably there is worth exploring the possibility of caching not only the generated images but also the manifest files.

egabancho:
Quick question, should I also add the feature flag? Something similar to what you did for the Communities at the beginning, Lars and I talked about it, but then v9 got released, and I don't know if it makes sense anymore ...

kpsherva (approved):

ntarocco:
After discussion, no need for the feature flag. Enabled by default.

PR #990, opened 2022-06-29
piece together notification and broadcast

rekt-hard:
* closes https://github.com/inveniosoftware/invenio-communities/issues/502
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1179

depends on:
- https://github.com/inveniosoftware/invenio-users-resources/pull/78
- https://github.com/inveniosoftware/invenio-communities/pull/954
- https://github.com/inveniosoftware/invenio-notifications/pull/11

### Description

Send notification on community submission created

### Checklist

- [ ] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [ ] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).

**Third-party code**

**Reminder**


rekt-hard:
**Current state**

When submitting a record for review, it will create a notification and send it to every community member via mail.

There are a few TODO and NOTE markers. These are placed there for minor adaptations before merging and to encourage discussion on the current implementation (e.g. [passing a record item vs id vs other object] (https://github.com/inveniosoftware/invenio-rdm-records/commit/4a16f4faa658bc0dbec04ed7527828ac1ca09a6c#diff-df26bf0f6fb4e0e211832e1872e2369d580a83e67659d8b1b2b4026dd341a37eR97-R99)).

**needed packages/commits for local testing:**

- [invenio-notifications](https://github.com/inveniosoftware/invenio-notifications/pull/8/commits/dc110a91aa31b7932f2e4a27ccbdddd3c9f45cba)
- [invenio-records-resources](https://github.com/inveniosoftware/invenio-records-resources/tree/feature/notifications)
- [invenio-users-resources](https://github.com/inveniosoftware/invenio-users-resources/pull/65/commits/c3543c938d089ed7d047095d7298e1d1a5e27e78)

**Video showing**
- community members (2)
- non community member submitting a record
- the notification being sent to community members (printed via console as sending mails is not configured in the local instance)
[Screencast from 2023-03-21 11:12:41.webm](https://user-images.githubusercontent.com/72449192/226581749-dae0288b-2666-42a6-b20f-77e6ffca0474.webm)

invenio_rdm_records/notifications/utils.py
ppanero, line 1: can this file be split in two? for examples `payload_generators.py` and `builders.py`? Also some names are quite lengthy e.g. `CommunitySubmissionSubmittedNotificationBuilder` can they be reduced? maybe with less repetition on the name?

invenio_rdm_records/services/review/service.py
```diff
@@ -12,17 +12,23 @@
...
-from invenio_requests import current_request_type_registry, current_requests_service
+from invenio_requests import (
+    current_request_type_registry,
+    current_requests_service,
+)
 from invenio_requests.resolvers.registry import ResolverRegistry
+from invenio_users_resources.records.api import UserAggregate
 from marshmallow import ValidationError
 
 from invenio_rdm_records.proxies import current_rdm_records
```
ppanero, line 28: why do we have relative and absolute imports? I think this was already like this but I think these should be made relative.

invenio_rdm_records/services/review/service.py
```diff
@@ -86,6 +96,16 @@ def create(self, identity, data, record, uow=None):
         # Set the request on the record and commit the record
         record.parent.review = request_item._request
         uow.register(RecordCommitOp(record.parent))
+
+        # TODO: remove before merge. dev only
```
ppanero, line 100: remove?

invenio_rdm_records/notifications/events.py
```diff
@@ -0,0 +1,97 @@
...
+@dataclass
+class Event:
+    """Base event."""
+
+    type: str
+    handling_key: str
+
+
+@dataclass
+class CommunitySubmissionEvent(Event):
```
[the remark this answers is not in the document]
rekt-hard: Goal would be to have `Event` in `invenio-records-resources` and community related events in `invenio-communities`. Currently here to not have another dependency for the implementation :smile:

invenio_rdm_records/notifications/utils.py
```diff
@@ -0,0 +1,148 @@
...
+
+# NOTE: Move to invenio-records-resources?
+class BaseRecordPayloadGenerator(PayloadGenerator):
+    """Record payload for a notification."""
+
+    # NOTE: Since we want links and everything resolved, we should use a result item from the service level.
+    def __init__(self, key, record_item, **kwargs):
+        """Ctor."""
+        self._record_item = record_item
+        self._key = key
```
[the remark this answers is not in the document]
rekt-hard: refactored into resolvers

invenio_rdm_records/services/review/service.py
rekt-hard: It is only here for development purposes, to more easily trigger the whole workflow of the submission. Will remove before merge :)

invenio_rdm_records/services/review/service.py
```diff
@@ -190,5 +199,17 @@ def submit(self, identity, id_, data=None, require_review=False, uow=None):
                 identity, community, request, uow
             )
 
+            uow.register(
+                NotificationOp(
+                    CommunityInclusionNotificationBuilder.build(request_item)
+                )
+            )
```
slint, line 206:
```suggestion
            uow.register(
                NotificationOp('community-submission', request=request_item._request)
            )
```

minor/shelve: maybe we could simplify this call to something like this? let's shelve and revisit maybe after we have the full example integrated.

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
```diff
@@ -0,0 +1,24 @@
+{# notifications/community-submission.submitted.jinja #}
+
+{%- block subject -%}
+New record submission for your community {{ notification.context.get("request").get("receiver").get("metadata").get("title") }} submitted by {{ notification.context.get("request").get("created_by").get("username") }}
+{%- endblock subject -%}
```
slint, line 5:
```suggestion
{# notifications/community-submission.submitted.jinja #}

{% set community = notification.request.receiver %}
{% set creator = notification.request.created_by %}

{%- block subject -%}
New record submission for your community {{ community.metadata.title }} submitted by {{ creator.username }}
{%- endblock subject -%}
```

minor: would something like this work? Since we're in Jinja, we should take advantage such tricks as much as we can, otherwise templating will become very tedious/verbose :)

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
```diff
@@ -0,0 +1,24 @@
...
+
+{%- block html_body -%}
```
slint, line 7: shelve: not for now, but we should explore how to address email template headers/footers and overall styling e.g. via extending some base templates.

slint (commented):
Some naming, clean-up, and usability comments, otherwise looks very good.

I'll give a go at shelving some of the comments into separate issues to be addressed later on after we've merged a solid base.

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
rekt-hard: We could use the `dict_lookup` method (if provided for jinja templates) or at least remove the `.get(...)'  calls and just access them via ["..."].

Also taking into account setting variables as suggested by you.

```suggestion
{# notifications/community-submission.submitted.jinja #}

{% set community = notification["context"]["request"]["receiver"] %}
{% set creator = dict_lookup(notification, "request.created_by") %}

{%- block subject -%}
New record submission for your community {{ community["metadata"]["title"] }} submitted by {{ dict_lookup(creator, "username") }}
{%- endblock subject -%}
```

Not sure which one is better though :sweat_smile:

jrcastro2 (commented):
LGTM! Just left a minor comment about when the notification is send.

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
ntarocco: This is what I was suggesting when we had the demo.

invenio_rdm_records/services/review/service.py
```diff
@@ -190,5 +192,13 @@ def submit(self, identity, id_, data=None, require_review=False, uow=None):
                 identity, community, request, uow
             )
 
+            uow.register(
+                NotificationOp(
+                    CommunityInclusionSubmittedNotificationBuilder.build(
+                        request_item._request
+                    )
+                )
+            )
```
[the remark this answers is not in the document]
rekt-hard: Thank you for pointing it out. When I was testing it via the UI, the `require_review` parameter was always `False`, even though the record had to be reviewed - hence why I put it here. Just realized that the `include()` method will not throw, but silently just read the request again and not change the request status (kind of different to all other methods). That is probably also why all the test cases did not have to updated to explicitly set the `require_review` parameter.

Anyway, the question now is in which case this specific notification should be sent:
- on every new submission
- only if a review is required

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
rekt-hard: Update: since we render blocks individually, they do not have access to variables set in another scope (i.e. outside of the block). With the change from notifications [here](https://github.com/inveniosoftware/invenio-notifications/pull/11), an initial block can take care of setting everything the template needs without touching `notifications` or `recipients`

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
```diff
@@ -0,0 +1,35 @@
...
+{% endblock ctx_setup %}
+
+{%- block subject -%}
+    New record submission for your community {{ ctx["community"]["metadata"]["title"] }} submitted by {{ ctx["creator"]["username"] }}
+{%- endblock subject -%}
+
+{%- block html_body -%}
+<p>The record "{{ ctx["record"]["metadata"]["title"] }}" was submitted to your community {{ ctx["community"]["metadata"]["title"] }} by {{ ctx["creator"]["username"] }}.</p>
+
+<a href="{{ ctx["submission_request"]["links"]["self_html"] }}" class="button">Review the submission_request</a>
```
rekt-hard, line 21: This link is not correct. It will point to something like `<base_url>/requests/<request_uuid>`
A request, for wich user has access to, should be accessible via `<base_url>/me/requests/<request_uuid>`. But I am not sure about other implications this might have. For a community manager, it would make sense to direct them to `<base_url>/communities/<community_slug>/requests/<request_uuid>` , as this is the context where the request should be reviewed.

We could of course construct the request url per template like
```
{{ ctx["community"]["links"]["self_html] + "/requests/` + ctx["submission_request"]["id"] }}`
```
Any suggestions on this? @slint @ntarocco

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
```diff
@@ -0,0 +1,35 @@
...
+{% block ctx_setup %}
+    {% set submission_request = notification.context["request"] %}
+    {%- set dummy = ctx.update({
+        "submission_request": submission_request,
+        "community": submission_request["receiver"],
+        "creator": submission_request["created_by"],
+        "record": submission_request["topic"],
+        })
+    %}
+{% endblock ctx_setup %}
```
slint, line 12: minor: that's a bit unfortunate, I thought the context would be shared inside the template blocks... In that case, we have to make sure that this block runs first though (i.e. explicitly call the `ctx_setup` block).

I tested an alternative that also works:

```jinja
{% set request_title = notification.request.title %}
{% set recipient_name = recipient.profile.full_name %}

{% block subject %}
Request for {{ request_title }}
{% endblock subject %}
```

with this code:

```python
from flask import current_app

template = current_app.jinja_env.select_template(["zenodo_rdm/testing.jinja"])

data = {
    "notification": {"request": {"title": "This is the request title!"}},
    "recipient": {"id": 1234, "email": "test@mail.com", "profile": {"full_name": "Alex"}},
}
ctx = template.new_context(data)

print("before full render:")
print("".join(list(template.blocks["subject"](ctx))).strip())
# Prints: "Request for "  (i.e. `request_title` is not set)

# "Force" rendering the whole template (including global variables)
list(template.root_render_func(ctx))

print("after full render:")
print("".join(list(template.blocks["subject"](ctx))).strip())
# Prints: "Request for This is the request title!"  (i.e. now `request_title` is there)

print(ctx.vars)
# Prints:
#   {'request_title': 'This is the request title!', 'recipient_name': 'Alex'}
```

The unfortunate part is that we're rendering the whole template once, and then again each individual block... But that's not that expensive since it's just rendering :)

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
```diff
@@ -0,0 +1,35 @@
+{# notifications/community-submission.submitted.jinja #}
```
slint, line 1:
```suggestion
```

nit: no need to repeat the name of the template :)

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
slint, line 1: minor: since we're establishing a naming convention here, maybe we match it with the service/config ID naming:

```diff
-community-submission.submitted.jinja
+community-submission.submit.jinja
```

i.e. `<request_type_id>.<request_action>.jinja`

slint (commented):
One minor comment on the new context variables mechanism. I feel we're ending up with something that requires some convention or prior knowledge, so we should try to make it as simple/intuitive as possible.

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
ntarocco: You are right. I think that this is entire point of `self_html`, and any other `_html` link in `links`: they should point to the UI that displays that object.
It is indeed complicated in case of requests, because it depends on the logged in user that will see it (e.g. invitations).

@rekt-hard @slint Does it sound crazy if we redirect, depending on the logging user? 

E.g. for invitation, the `self_html` will be `/requests/<request_uuid>` and we redirect to:
- if I am the invited user `/me/requests/<request_uuid>`
- if I am the creator (community owner/manager/curator), then `/communities/<community_slug>/requests/<request_uuid>`
This would be a way to centralize this login, and avoid to re-implement it every time we need to use the `self_html`: the `self_html` cannot know in advance what is the URL.

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
rekt-hard: Thank you! The less knowledge needed the better!
Updated it in the [notifications pr](https://github.com/inveniosoftware/invenio-notifications/pull/11/files#diff-0de3f101a171677955e25bf6a31171835a6e2d194b4740a8009ca109d9667e2dR49)

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
rekt-hard: This is an interesting approach. Would the responsibility for deciding the final redirect destination be on the specific request?
For current request types, can we make sure that the context for the current user will be indisputable? I.e. there is no possibility that two destinations would make sense? Because then we end up in a similar position where we have to choose one (but each will be correct at least)

slint: I think taking into account the user identity context for the final link at the `/requests/<request_id>` endpoint is probably much easier to encapsulate centrally in some sort of "user-friendly UI page resolver" vs. doing it when generating links... 

On the other hand, this design would call for some sort of additional logic on each request type which defines the link resolution based on the user identity... Which sounds like something that could also just be baked in the whole "link context expand logic", that [already has access to the user identity](https://github.com/inveniosoftware/invenio-records-resources/blob/58503e2a6fdc7bd110f7d3bf3a51e7ca05fd63f0/invenio_records_resources/services/base/links.py#L88).

Not sure what's best, especially without having a first go at coding it :)

kpsherva: I have referenced this discussion in a new issue https://github.com/inveniosoftware/invenio-rdm-records/issues/1327 I think for now it is good to merge in the current state, since the links issue it is deeper and seems to me out of scope. to be addressed globally

rekt-hard: The link is still not correct though. So it should be fixed before merging. Otherwise users will get directed to a 404

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submitted.jinja
rekt-hard: This also allows to access dictionaries with the dot notation instead of having to use quotations marks :)

PR #994, opened 2022-07-05
custom fields: CLI POC

ppanero:
- closes https://github.com/inveniosoftware/invenio-cli/issues/302#issuecomment-1173918959

Example CLI call, for all fields:

```
invenio custom-fields create
```

For one or more specific fields (by name)

```
invenio custom-fields create -f experiment -f department
```

### DIFF
--- a/invenio_rdm_records/cli.py
+++ b/invenio_rdm_records/cli.py
@@ -278,3 +278,59 @@ def rebuild_index():
     affs_service.rebuild_index(identity=system_identity)
 
     click.secho("Reindexed records and vocabularies!", fg="green")
+
+
+# CUSTOM FIELDS
+
+@click.group()
+def custom_fields():
+    """InvenioRDM custom fields commands."""
+
+
+
+@custom_fields.command()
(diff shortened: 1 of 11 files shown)


ppanero:
# Open Questions:

- CF registration: blueprint vs config loading (post app, e.g. in the `views` - a bit dirty tho).
- Who handles the _prefix_. If we plan to store the custom fields inside a `custom` field as we do for `metadata` we need a place that takes care of that without the user knowing that (e.g. for `exists` command). 
    - Would this be the registry? then exists command might have issues since it does not pass by registry (unless we also check that, see code TODO/FIXME for more info).
    - what about the `BaseCF`? Then we need a way to pass the prefix to it, unless its not configurable (hardcoded). Not that bad since it affects also UI displays and all the data model flow (e.g. field name for Marshmallow)
-

ppanero:
@zzacharo closing this, since we have https://github.com/inveniosoftware/invenio-rdm-records/pull/997 ? then we can pass to a feature branch

Issue #819, opened 2021-10-11
publisher required for DataCite DOI registration


rekt-hard:
Package version (if known):

## Describe the bug
DataCite requires the record to have a publisher defined. When acquiring a DOI through the "Get DOI" button, the publisher field is not checked. If left empty, no DOI can be acquired and no error is presented.

## Steps to Reproduce
1. Create a new upload
2. Select 'No' on "Do you already have a DOI for this upload?"
3. Fill out all required fields
4. Remove text from publisher field
5. Save draft
6. Publish record

## Expected behavior
An error message is shown, that the publisher has to be defined for the DOI registration.

## Screenshots (if applicable)
![Screenshot from 2021-10-11 11-06-01](https://user-images.githubusercontent.com/72449192/136764414-b9985292-9338-4ff2-b283-bc9d4d818949.png)

fenekku:
The publisher-less DOI record is then marked as "Draft" in Datacite's dashboard which makes sense but adds to the confusion for unfamiliar user.

fenekku:
Closed via https://github.com/inveniosoftware/invenio-rdm-records/pull/1146

PR #1018, opened 2022-08-11
metadata: require publisher field since Datacite requires it

fenekku:
Datacite requires the publisher, but we didn't for some reason. This leads to trouble when a record is not created with a publisher.
Another PR will be made to move the publisher field in invenio-app-rdm.

- part of closing #819
- includes a little drive-by refactoring

### DIFF
--- a/invenio_rdm_records/services/pids/manager.py
+++ b/invenio_rdm_records/services/pids/manager.py
@@ -47,6 +47,36 @@ class PIDManager:
         if unknown_schemes:
             raise PIDSchemeNotSupportedError(unknown_schemes)
 
+    def _validate_identifiers(self, pids, errors):
+        """Validate and normalize identifiers."""
+        # TODO: Refactor to get it injected instead.
+        configured_identifiers = current_app.config["RDM_PERSISTENT_IDENTIFIERS"]
+
+        scheme_ids = []
+        for scheme, pids_attrs in pids.items():
+            id_ = pids_attrs.get("identifier")
+            if not id_:
+                continue
+
+            configured_identifier = configured_identifiers.get(scheme, {})
+            validator = configured_identifier.get("validator", lambda x: True)
+            normalizer = configured_identifier.get("normalizer", lambda x: x)
+            label = configured_identifier.get("label", scheme)
+
+            if not validator(id_):
+                errors.append(
+                    {
+                        "field": f"pids.{scheme}",
+                        "messages": [_("Invalid {scheme}").format(scheme=label)],
+                    }
+                )
+            else:
+                scheme_ids.append((scheme, normalizer(id_)))
+
+        # Modify outside of iteration - updates the pids globally
+        for scheme, id_ in scheme_ids:
+            pids[scheme]["identifier"] = id_
+
     def _validate_pids(self, pids, record, errors):
         """Validate an iterator of PIDs.
 
@@ -57,35 +87,6 @@ class PIDManager:
             success, val_errors = provider.validate(record=record, **pid)
             if not success:
                 errors.append({"field": f"pids.{scheme}", "messages": val_errors})
-
-    def _validate_identifiers(self, pids, errors):
-        """Validate and normalize identifiers."""
-        # TODO: Refactor to get it injected instead.
-        conf = current_app.config["RDM_PERSISTENT_IDENTIFIERS"]
-
-        identifiers = []
-        for scheme, pids_attrs in pids.items():
-            identifier = pids_attrs.get("identifier")
-
-            validator = conf.get(scheme, {}).get("validator", lambda x: True)
-            normalizer = conf.get(scheme, {}).get("normalizer")
-            label = conf.get(scheme, {}).get("label", scheme)
-
-            if identifier:
-                if not validator(identifier):
-                    errors.append(
-                        {
-                            "field": f"pids.{scheme}",
-                            "messages": [_("Invalid {scheme}").format(scheme=label)],
-                        }
-                    )
-                else:
-                    if normalizer is not None:
-                        identifiers.append((scheme, normalizer(identifier)))
-
-        # Modify outside of iteration - updates the pids globally
-        for scheme, id_ in identifiers:
-            pids[scheme]["identifier"] = id_
 
     def validate(self, pids, record, errors=None, raise_errors=False):
         """Validate PIDs."""
--- a/invenio_rdm_records/services/schemas/metadata.py
+++ b/invenio_rdm_records/services/schemas/metadata.py
@@ -343,7 +343,7 @@ class MetadataSchema(Schema):
     )
     title = SanitizedUnicode(required=True, validate=validate.Length(min=3))
     additional_titles = fields.List(fields.Nested(TitleSchema))
-    publisher = SanitizedUnicode()
+    publisher = SanitizedUnicode(required=True)
     publication_date = EDTFDateString(required=True)
     subjects = fields.List(fields.Nested(SubjectRelationSchema))
     contributors = fields.List(fields.Nested(ContributorSchema))
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -521,6 +521,7 @@ def minimal_record():
         },
         "metadata": {
             "publication_date": "2020-06-01",
+            "publisher": "InvenioRDM",
             "resource_type": {"id": "image-photo"},
             "creators": [
                 {
--- a/tests/resources/serializers/test_dublincore_serializer.py
+++ b/tests/resources/serializers/test_dublincore_serializer.py
@@ -74,6 +74,7 @@ def test_dublincorejson_serializer_minimal(running_app, updated_minimal_record):
         "creators": ["Name", "Troy Inc."],
         "dates": ["2020-06-01"],
         "rights": ["info:eu-repo/semantics/openAccess"],
+        "publishers": ["InvenioRDM"],
     }
 
     serializer = DublinCoreJSONSerializer()


tmorrell (approved):
Looks good to me!

invenio_rdm_records/services/pids/manager.py
```diff
@@ -47,6 +47,36 @@ def _validate_pids_schemes(self, pids):
...
+                        "messages": [_("Invalid {scheme}").format(scheme=label)],
+                    }
+                )
+            else:
+                scheme_ids.append((scheme, normalizer(id_)))
+
+        # Modify outside of iteration - updates the pids globally
+        for scheme, id_ in scheme_ids:
+            pids[scheme]["identifier"] = id_
+
```
fenekku, line 79: I was sitting on this change for a little while, so might as well pass it in here when looking at how DOI minting was affected by missing publisher.

I think it makes the code more readable. 

_Aside_
Also, it reveals how we could go one-step further in my ideal world: 

- make the `"RDM_PERSISTENT_IDENTIFIERS"`config take a factory object instead of a dict 
- that returns `ConfiguredIdentifier` objects that have a `label`, `scheme`, `validate`, `normalize` interface.

Basically an extract class to solve the primitive obsession on `dict` and reduce the config interface size. Because more changes involved, I'd be happy to submit a separate PR, it that sounds interesting.

invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -343,7 +343,7 @@ class MetadataSchema(Schema):
     )
     title = SanitizedUnicode(required=True, validate=validate.Length(min=3))
     additional_titles = fields.List(fields.Nested(TitleSchema))
-    publisher = SanitizedUnicode()
+    publisher = SanitizedUnicode(required=True)
```
fenekku, line 346: That's the real crux of this change.

tmorrell:
I just wanted to bump this up - it should really be merged for v10. We've had users working on integrations who have forgotten the publisher field, since it's not required, and DOI minting breaks in very unclear ways. The logging messages are really cryptic, and I basically had to manually copy over the metadata and test it in DataCite's system to figure out that we were dealing with a missing publisher field.

invenio_rdm_records/services/schemas/metadata.py
kpsherva: should publisher be globally required for any record of RDM instance? I am not sure about the implications of this change, especially since it is done to fix DOI, but affecting any future record addition - is there a way to simply validate if there is publisher just before requesting the DOI, instead adding it as required for every record?

fenekku: > should publisher be globally required for any record of RDM instance?

Yes. Publication date, title, creators... are all required for bibliographic records of RDM instance and so should publisher. Those fields are required to mint a public DOI. We shouldn't add a split case here (validating for a publisher). What is the issue you see with future record addition?

zzacharo: What about validating the publisher field in the [datacite provider interface](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/pids/providers/datacite.py#L196)? Like that I think requiring the publisher is totally in line with the DOI registration problem...

fenekku: Circling back to this:

Here is my take in hopefully clearer terms, so it can be moved along :smile: :

The choices we made in what is required or not in `MetadataSchema` were based on what fields are required or not by Datacite in order to mint a DOI. For instance, `title = SanitizedUnicode(required=True, validate=validate.Length(min=3))` is required because a title is required in Datacite (https://support.datacite.org/docs/schema-mandatory-properties-v43#3-title). A [publisher is required](https://support.datacite.org/docs/schema-mandatory-properties-v43#4-publisher) in Datacite, so I think it should be required on our end too. 

We are not having the DataCitePIDProvider check for title (or other required DOI fields), so we shouldn't have it check for publisher.  

The related PRs in the other modules that are related to this change are:

- https://github.com/inveniosoftware/invenio-app-rdm/pull/1833
- https://github.com/inveniosoftware/react-invenio-deposit/pull/569

This does indeed change the UI and future record addition... for the better :wink: . Without this change, as @tmorrell, myself and others have noticed, DOI minting fails silently and we show on the record landing page that the record has a public DOI when in fact it doesn't. Already ingested records are not affected, but will require a publisher on future edit and then will be able to have a public DOI.

The silver lining though, is that this field is given a value by default: https://github.com/inveniosoftware/cookiecutter-invenio-rdm/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/invenio.cfg#L156 in instances. But for instances that remove this default, or even for those where the default is still there, but the user removes it ([the user is drunk](https://theuserisdrunk.com/)), the problem occurs.

Maybe there was a reason to not have this field required but I don't know it. Maybe @lnielsen knows?

fenekku: Counterpoints first, then defeated acceptance, and then longer term solution? :wink: :

**Counterpoints**
(I am just representing my stakeholders here.)

> DOIs are an optional feature.

True enough but they are the reason why we picked the DataCite schema in the first place; it's the easiest schema choice to pick to make sure there will be compatibility with minting DOI. If there is an additional level of picking and choosing of fields on top of that, it opens some difficult discussions (like this one!) I've been asked "Why this field is not required but this other field is?" alot and I am sure you have too. Stakeholders and higher ups have been much more willing to accept the current fields knowing we were just following DataCite. If we are actually not strictly following DataCite... (things are going to get bogged down real fast is my experience)    

> Publisher is not displayed prominently anywhere (title, resource type, publication date and creators on the other hand are prominently displayed most places).

:rotating_light: : Publisher shows up on the citation which is a _very_ prominent place for our librarians  :laughing: 

> Publisher: I think most users will actually not have a clue for what to put in the field 

Which is why there is [a default by default](https://github.com/inveniosoftware/cookiecutter-invenio-rdm/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/invenio.cfg#L156) . It would be interesting to better understand the use cases where it is removed though. 

>  specially cultural heritage where you won't actually know the publisher,

Yes, we have special collections and that's what our librarians do: "Unknown publisher" or some such. It's fine by them.

**Acceptance**
Having a default publisher sent to DataCite when there isn't one on the record would probably be ok. The added setting is not ideal (another corner case to cover) but fine. The moment librarians realize there is a publisher showing up on DataCite citations but not on our citations, they will probably raise this issue again :shrug: 

**Longer term solution**
This is perhaps pie-in-the-sky and mostly motivated by the years of debate among our librarians/stakeholders about what fields to use.

I don't see there ever being consensus about fields to capture among institutions. And I don't see any value in striving for that consensus. Different institutions should capture the different fields they care about. The only restrictions should be that that instances can convert between schemas.

In an ideal InvenioRDM , **every** field is implemented in a "custom field"-fashion (now that we have that experience) and placed in the cookiecutter for an instance giving a default data model. Then, an institution can change those fields completely as they see fit using the library of programmatic fields provided by InvenioRDM. Most of the functionality of InvenioRDM isn't specific to fields (or only concerned with them when it comes to serialization to other schemas) anyway.


Anyway, food for thought. The proposed serialization of default publisher is fine if we don't want to require publishers like Datacite does.

lnielsen: Just to be sure its explicit: I'm fine with bringing up the discussion if the publisher field should be required, as long as it's covered in one of the telecons so the full community have a chance to provide feedback and we can validate assumptions. I know it's fine by you and Caltech and probably will be fine by most that uses DataCite DOIs, but I'm not sure about the rest hence the need to confirm with all partners to understand the impact and if it's feasible. 

> Stakeholders and higher ups have been much more willing to accept the current fields knowing we were just following DataCite. If we are actually not strictly following DataCite... (things are going to get bogged down real fast is my experience)

DataCite schema made sure we didn't have too many metadata discussions compared to what we could have had. That said, we never were strictly DataCite. We have exceptions like the role on creator when valid use cases were presented. We allowed not to mint/require DOIs because many partners had use cases that required it. Thus, DataCite has always been about following best practices (related to e.g. persistent identifiers), and has helped frame discussions, but we were never able to strictly follow DataCite. 

> I don't see there ever being consensus about fields to capture among institutions. And I don't see any value in striving for that consensus. 

FAIR, Next Generation Repositories, COUNTER Code of Practice + a lot of the community work is all about consensus among different institutions so that we can enable interoperability. Like any diplomacy this is a slow process, because many stakeholders provide their input and have to agree. Similarly, I find consensus is what brings value to InvenioRDM - understanding how we can design features so it's not custom build to a single institution, but works across many of them. For metadata, it's about findability and interoperability. RDM provides a core set of metadata according to best practices to describe scholarly outputs and make them discoverable *across* domains and *across* the entire open science ecosystem.  At a system-level, the common set of metadata further is needed to reduce complexity - e.g. allowing fully custom metadata would mean everyone would have to build their own BibTeX exporter (something that's notoriously difficult to get correct).

chriz-uniba:
Is there a date scheduled for bringing this up on a partners-meeting?

We also experienced strange situations where invenioRDM and fabrica didn't match and except for some hidden 'warnings' in the workers-logs (DOI-minting went wrong) nobody realized that minting went wrong.

tmorrell:
I think it's on the agenda for tomorrow. Any more details on what didn't match? I know the error message propagation could be improved, but we haven't seen many failures other than publisher.

fenekku:
Closed in favor of https://github.com/galterlibrary/galter-ir-importer/pull/137

PR #1021, opened 2022-08-17
Feature/custom fields

zzacharo:
- requires https://github.com/inveniosoftware/invenio-records/pull/294
- requires https://github.com/inveniosoftware/invenio-communities/pull/774


invenio_rdm_records/cli.py
```diff
@@ -278,3 +280,176 @@ def rebuild_index():
...
+    fields = []
+    if fields_names:
+        for field_name in fields_names:
+            fields.append(available_fields.get(field_name))
+    else:
+        fields = available_fields.values()
+
+    properties = {}
+    for field in fields:
+        properties[f"custom_fields.{field.name}"] = field.mapping
```
ntarocco, line 311: usual question about objects with inner fields...

invenio_rdm_records/cli.py
```diff
@@ -278,3 +280,176 @@ def rebuild_index():
...
+    for part in parts:
+        mapping = mapping["properties"]  # here to avoid last field access
+        if part not in mapping.keys():
+            return False
+        mapping = mapping[part]
+
+    return True
+
+
+@records.command("create")
```
ntarocco, line 330: Silly question: given that we have `index init`, shall we use `init` cmd name instead?

invenio_rdm_records/cli.py
```diff
@@ -278,3 +280,176 @@ def rebuild_index():
...
+    click.secho("Creating all custom fields...", fg="green")
+    # multiple=True makes it an iterable
+    properties = _prepare_mapping(field_name, available_fields)
+
+    try:
+        record_index = current_rdm_records.records_service.config.record_cls.index
+        draft_index = current_rdm_records.records_service.config.draft_cls.index
+
+        record_index.put_mapping(body={"properties": properties})
+        draft_index.put_mapping(body={"properties": properties})
```
ntarocco, line 359: I guess that if the second fails, we cannot rollback the first...

invenio_rdm_records/cli.py
```diff
@@ -278,3 +280,176 @@ def rebuild_index():
...
+        click.secho("Created all custom fields!", fg="green")
+    except RequestError as e:
+        click.secho("An error occured while creating custom fields.", fg="red")
```
ntarocco, line 362: What about suggesting what to check?
```suggestion
        click.secho("An error occured while creating search mappings for custom fields. Is the search service healthy?", fg="red")
```

invenio_rdm_records/config.py
```diff
@@ -421,3 +337,144 @@ def make_doi(prefix, record):
...
+                icon: "pencil",
+                description: "You should fill this field with one of the experiments e.g LHC, ATLAS etc.",
+            }
+        },
+        ...
+    }]
+"""
+
+
+COMMUNITIES_CUSTOM_FIELDS = {}
```
ntarocco, line 414: move this and below to `invenio-communities` and then it will be part of `cookiecutter`/`invenio.cfg`?

invenio_rdm_records/ext.py
```diff
@@ -88,10 +87,6 @@ def __init__(self, app=None):
     def init_app(self, app):
         """Flask application initialization."""
         self.init_config(app)
-        self.metadata_extensions = MetadataExtensions(
```
ntarocco, line 91: Config is always loaded [before](https://github.com/inveniosoftware/invenio-base/blob/master/invenio_base/app.py#L105) loading extensions, so if we want to do some validation check, the `ext/init_app` is probably a good place.

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -161,6 +162,12 @@ class UIRecordSchema(BaseObjectSchema):
         fields.Nested(AdditionalTitlesSchema), attribute="metadata.additional_titles"
     )
 
+    # Custom fields
+    # TODO: take into account vocabularies and localization of titles
```
ntarocco, line 166: Will this TODO be forgotten? :)

invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -53,7 +56,11 @@ class Meta:
         values=fields.Nested(PIDSchema),
     )
     metadata = NestedAttribute(MetadataSchema)
-    # ext = fields.Method('dump_extensions', 'load_extensions')
+    # FIXME: with NestedAttribute it does not work
```
ntarocco, line 59: Not clear what should be fixed here? can we fix before merge?

requirements-feature.txt
```diff
@@ -0,0 +1,5 @@
+# This file is adding constraints to be used when a PR is targeting a feature/* branch
+-e git+https://github.com/inveniosoftware/invenio-records.git@feature/custom-fields#egg=invenio-records
```
ntarocco, line 2: Release and remove before merge?

tests/conftest.py
```diff
@@ -297,6 +298,14 @@ def app_config(app_config):
         ),
     ]
 
+    # Custom fields
+    app_config["RDM_CUSTOM_FIELDS"] = [
+        TextCF(name="cern:myfield", use_as_filter=True),
+    ]
+    app_config["COMMUNITIES_CUSTOM_FIELDS"] = [
```
ntarocco, line 305: move to `invneio-communities`?

invenio_rdm_records/records/jsonschemas/records/record-v5.0.0.json
```diff
@@ -11,89 +11,100 @@
     "id": {
       "$ref": "local://definitions-v1.0.0.json#/identifier"
     },
-    "pid": {"$ref": "local://definitions-v1.0.0.json#/internal-pid"},
-
+    "pid": {
```
ppanero, line 14: There are back incampatible changes, needs to be v6.

invenio_rdm_records/cli.py
ppanero: Agree, `invenio custom-fields init`, there is also `index create` but is to create a full index. Adding templates or mappings happens in `init` so it also makes more sense to call it `init` here

invenio_rdm_records/cli.py
ppanero: opened an issue https://github.com/inveniosoftware/invenio-rdm-records/issues/1030

invenio_rdm_records/cli.py
ppanero: IIRC, the way Zach made it work was by making a custom `CF` class that would return the nested object in `field.mapping`. Then it will work. That means that for now we only support primitive types and lists of those. Custom objects can be added but require a bit of coding (a `CF` class is like 20-30 lines that can go to the cfg file or to a custom module)

invenio_rdm_records/cli.py
ppanero: Not really needed because the following line will print the actual error (e.g. ES connection refused). If we add this comment, then it might be misleading (imagine this says healthy and then the error message something about duplicate or permissions)

invenio_rdm_records/config.py
ppanero: I'm going to leave it as is, not to break it now. But there is an issue already to do this in the board.

invenio_rdm_records/resources/serializers/ui/schema.py
ppanero: made into issue

invenio_rdm_records/services/schemas/__init__.py
ppanero: Actually nothing. It was fixed but the FIXME not removed. Nice catch!

tests/conftest.py
ppanero: same issue as above

PR #1034, opened 2022-09-06
add query parameters to search OAI sets

alejandromumo:
closes https://github.com/inveniosoftware/invenio-administration/issues/49

Note: set base branch to `feature/administration-oai` so it can be integrated with the most recent implementation of oai sets views. It can be changed later on when this PR is finished

### DIFF
--- a/invenio_rdm_records/oaiserver/services/services.py
+++ b/invenio_rdm_records/oaiserver/services/services.py
@@ -18,6 +18,7 @@ from invenio_records_resources.services.base import LinksTemplate
 from invenio_records_resources.services.records.schema import ServiceSchemaWrapper
 from invenio_records_resources.services.uow import unit_of_work
 from marshmallow import ValidationError
+from sqlalchemy import or_
 from sqlalchemy.orm.exc import NoResultFound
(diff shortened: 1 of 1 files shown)


invenio_rdm_records/oaiserver/services/services.py
```diff
@@ -130,7 +131,13 @@ def search(self, identity, params):
 
         search_params = self._get_search_params(params)
 
-        oai_sets = OAISet.query.order_by(
+        query_param = search_params["q"]
+        filters = []
+
+        if query_param:
+            filters.extend([OAISet.name.ilike(f"%{query_param}%"), OAISet.spec.ilike(f"%{query_param}%")])
```
alejandromumo, line 138: For now it's filtering based on `name` or `spec`. Do we want to have more fields? 
Should this be configurable?

kpsherva: I think we could merge and check if we receive feedback on that. Maybe we should include a tooltip for the searchbox informing about this limitation ?

kpsherva (approved):

PR #1050, opened 2022-09-16
mappings: add missing fields and make them strict

### LINKED ISSUE #1051: record query with content (e.g. gallery) returns 0

### LINKED ISSUE #1052: mappings: make `titles` for i18n strict using dynamic templates
similar to how it was done for `pids`, to be tested:

```json
"dynamic_templates": [
    {
      "i18n_title": {
        "path_match": "*.title.*",
        "match_mapping_type": "object",
        "mapping": {
            "type": "text",
            "fields": {
              "keyword": {
                "type": "keyword",
                "ignore_above": 256
              }
            }
          }
        }
    }
]
```

ppanero:
- closes #1051 
- closes #1052

# Why?

Before we did not state the `strict` property on the mappings, therefore they defaulted to *true*. This means that any field we were sending was being indexed (as text + keyword). By chance, it worked. We had this way many of them like formats, sizes, uuid, review data in the parent, etc. Even some very basic fields like "title" and "description" which were missing from the mapping 🤦.

# Why now?

I made the mappings strict as part of the custom fields development, because "why not, since that way we would be covered of users not abusing the mappings and only adding custom where allowed (i.e. under the`custom_fields.` field). Result: I could not find anything (search) by title, description, etc. (because they were not being indexed).

# Solution

Added all missing fields (hope I didn't forget any, but now ES will complain and we can add it).

# Improvement

Dynamic templates for pids and titles with i18n capabilities. Before they were catch-all fields, which could lead to abuse. Now they are strict in content (keys are still dynamic, meaning `en` `de` etc. are up to the user/app).

### DIFF
--- a/invenio_rdm_records/records/mappings/v7/rdmrecords/drafts/draft-v5.0.0.json
+++ b/invenio_rdm_records/records/mappings/v7/rdmrecords/drafts/draft-v5.0.0.json
@@ -590,8 +590,7 @@
           "related_identifiers": {
             "properties": {
               "identifier": {
-                "type": "keyword",
-                "copy_to": ["related.identifier"]
+                "type": "keyword"
               },
               "relation_type": {
                 "type": "object",
@@ -745,8 +744,7 @@
                 }
               },
               "title": {
-                "type": "text",
-                "copy_to": ["suggest_title"]
+                "type": "text"
               },
               "type": {
                 "type": "object",
@@ -820,21 +818,6 @@
             "type": "keyword"
           }
         }
-      },
-      "related": {
-        "type": "object",
-        "properties": {
-          "identifier": {
-            "type": "keyword"
-          }
-        }
-      },
-      "suggest_title": {
-        "type": "completion",
-        "analyzer": "simple",
-        "preserve_separators": true,
-        "preserve_position_increments": true,
-        "max_input_length": 50
       }
     }
   }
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/records/mappings/v7/rdmrecords/drafts/draft-v5.0.0.json
```diff
@@ -634,6 +733,72 @@
...
+          "enabled": {
+            "type": "boolean"
+          },
+          "default_preview": {
+            "type": "keyword"
+          },
+          "order": {
+            "type": "keyword"
+          },
+          "entries": {
```
ppanero, line 749: if entries are not indexed, removed. so we do it in the future when we reach this task

invenio_rdm_records/records/mappings/v7/rdmrecords/records/record-v5.0.0.json
```diff
@@ -634,6 +749,72 @@
...
+              }
+            }
+          },
+          "meta": {
+            "type": "object",
+            "dynamic": true
+          }
+        }
+      },
+      "related": {
```
ppanero, line 804: Whats is for

invenio_rdm_records/records/mappings/v7/rdmrecords/drafts/draft-v5.0.0.json
```diff
@@ -1,12 +1,68 @@
 {
   "mappings": {
-    "dynamic": false,
+    "dynamic_templates": [
+      {
+        "pids": {
```
ppanero, line 5: pids mapping now is defined, so it nos a "catch all" field.

PR #1058, opened 2022-09-19
tests: adds extract_messages dry command

mb-wali:
This is to make sure all marked strings are actually extractable.

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -53,6 +53,7 @@ fi
 
 export LC_TIME=en_US.UTF-8
 python -m check_manifest
+python -m setup extract_messages --dry-run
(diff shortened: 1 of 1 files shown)


ppanero (commented):
LGTM, but I'm not very familiar with the transifex/string extraction workflows. Is this something we should add to other modules? i.e. something that should be generic in the `run-tests.sh`?

mb-wali:
> LGTM, but I'm not very familiar with the transifex/string extraction workflows. Is this something we should add to other modules? i.e. something that should be generic in the `run-tests.sh`?

Yes, my suggestion is to add this line to all the repositories which has the translations to it.

Issue #1078, opened 2022-09-22
OAI metada prefix formats should be obtained from the backend


alejandromumo:
Metadata prefixes for the OAI links table are hardcoded. Formats can be retrieved from the backend (https://inveniordm.docs.cern.ch/reference/rest_api_oaipmh_sets/#get-metadata-formats) 

However, one must think how to display them in the Dropdown. For now I have this code as a guideline:

``` javascript
import { http } from "@js/invenio_administration/src/api/config.js";

function LinksTable({ data }) {
  const [links, setLinks] = useState(data.links);
  const [formats, setFormats] = useState([]); // save known formats as a state

  useEffect(async () => {
    const response = await http.get(
      '/oaipmh/formats'
    );
  
    const formats = response.data.hits;
    setFormats(formats); // update the known formats only when the component is rendered for the first time (note the empty dependency list of the hook)
    // TODO each format might need to be converted to a {key, text, value} object before saving the state 
  }, []);

 return(
  ...
  <Dropdown
    options={formats}
  /> 
);
}
```

PR #1079, opened 2022-09-22
Links table: add metadata prefix dropdown

### LINKED ISSUE #1064: administration: Add formatting options to oai-set
Screenshot 2022-09-20 at 10 53 04

In the links section we need a dropdown with a set of different exporting formats (check docs for extensive list) which will replace the metadataPrefix URL parameter of the List records and list identifiers links depending on the chosen format.
Default format option is oai_dc as currently visible.

alejandromumo:
closes #1064 #1078


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
```diff
@@ -5,32 +5,136 @@
...
+import React, { useEffect, useState } from "react";
 import { i18next } from "@translations/invenio_administration/i18next";
 import CopyButton from "./CopyButton";
-import { Table, Header } from "semantic-ui-react";
+import { Table, Header, Dropdown, Grid } from "semantic-ui-react";
+import { http } from "@js/invenio_administration/src/api/config.js";
+import { capitalize } from "lodash";
+
+/** Map of known formats and their name. */
+const knownFormats = {
```
alejandromumo, line 16: Used to have readable names for currently known OAI formats. 

If a format is added later, it won't be "known". To support it nonetheless, this component will display the unknown format in the dropdown.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
```diff
@@ -5,32 +5,136 @@
...
+            value: formt.id,
+            text: knownFormats[formt.id] ?? formatKeyToName(formt.id)
+          }
+        });
+        setFormats(serialized);
+      }
+    } catch (error) {
+      console.error(error);
+    }
+  }, []);
```
alejandromumo, line 45: Empty dependencies means that this hook is only used once, when the component is first rendered.

This hook will retrieve OAI formats from the backend and store them in the components state.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
```diff
@@ -5,32 +5,136 @@
...
+      const response = await http.get(
+        '/api/oaipmh/formats'
+      );
+      const formats = response.data?.hits?.hits;
+      if (Array.isArray(formats) && formats.length > 0) {
+        const serialized = formats.map((formt) => {
+          return {
+            key: formt.id,
+            value: formt.id,
+            text: knownFormats[formt.id] ?? formatKeyToName(formt.id)
```
alejandromumo, line 37: if the format is known, `text` is already hardcoded in the component. Otherwise, try to make it more readable (e.g remove `_` and capitalise the key)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
```diff
@@ -5,32 +5,136 @@
...
+    } catch (error) {
+      console.error(error);
+    }
+  }, []);
+
+  /**
+   * Replaces the metadata prefix in the link.
+   * Used to update links when the metadata prefix is changed.
+   */
+  const replaceLinkPrefix = (link, newPrefix) => {
```
alejandromumo, line 51: Replaces a format for a new one. E.g.:

`&metadataPrefix=oai_dc&` is replaced for `&meatadaPrefix=datacite&`

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
```diff
@@ -5,32 +5,136 @@
...
+   * Uses a regex to retrieve the prefix, matching "&metadataPrefix=...&".
    *
-   * @param {string} url
-   * @returns {string} stripped url
+   * Returns null if pattern not found.
    */
-  const sanitizeLink = (url) => {
-    return url;
+  const getPrefixFromLink = (link) => {
+    const prefixRegex = /\&metadataPrefix\=(.*)\&/;
```
alejandromumo, line 66: Decided to use a regex so the component only has one state that triggers re-renders (`links`).

Initially was using a state `currentPrefix` to keep track of the prefix currently being used but that required two states that, after being updated, would trigger a re-render each.

The downside of this approach is that it's tied to the backend's `&metadataPrefix&` parameter. If that changes in the future, this will not work properly.

If I am mistaken and having two states would be better / cleaner, please let me know and this can be modified.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
```diff
@@ -5,32 +5,136 @@
...
+      setLinks(newLinks);
+    }
+  };
+
+  const listRecords = links["oai-listrecords"];
+  const listIdentifiers = links["oai-listidentifiers"];
+
+  // Set default prefix based on backend link, use dublin core if unknown.
+  let defaultPrefix = getPrefixFromLink(listRecords)
+  if (!Object.prototype.hasOwnProperty.call(formats, defaultPrefix)) {
```
alejandromumo, line 110: The dropdown will have as a default the format currently in use (e.g. the one retrieved from the backend).
However, if that format is not known, it will use OAI DublinCore by default.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
```diff
@@ -5,32 +5,136 @@
...
+          return {
+            key: formt.id,
+            value: formt.id,
+            text: knownFormats[formt.id] ?? formatKeyToName(formt.id)
+          }
+        });
+        setFormats(serialized);
+      }
+    } catch (error) {
+      console.error(error);
```
kpsherva, line 43: should we display an error to the user?

jennur:
## Updated screenshot
Screenshot 2022-09-23 at 15 59 46

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
jrcastro2: I might lack some context about this, but I think that we shouldn't change forcefully the format to simply adapt it to "our" frontend needs. This is connected to the previous comment, I think it can be addressed in a separate issue.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
jrcastro2: Thanks for the clarification! I personally tend to choose the option that is more readable, so I would probably go for the `currentPrefix` option, but this is just personal choice, this looks good to me. :+1:

jrcastro2 (approved):
LGTM!

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
ntarocco: The fact that we have to add a comment in the PR to explain this makes me very scared about future changes/refactorings, especially because we don't have frontend integration tests. Please consider in the future to avoid using `useEffect` or, eventually, we need to make it very clear and readable what this code is doing and why. :)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
ntarocco: Future change:
agree, please go with the state to make this component easier to understand and maintain. We are not very worried here about re-rendering, given that this will happen after a user action.
Let's parse the URL with [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) and manipulate query args with [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams), avoiding `RegEx`.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
```diff
@@ -5,53 +5,166 @@
...
+
+/** Map of known formats and their name. */
+const knownFormats = {
+  "oai_dc": "OAI Dublin Core",
+  "datacite": "DataCite",
+  "oai_datacite": "OAI DataCite",
+}
 
 function LinksTable({ data }) {
+  const [links, setLinks] = useState(data.links);
```
ntarocco, line 23: This func is massive! :) I would suggest splitting it. Does it make sense to keep a functional cmp here?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/oaipmh/details/LinksTable.js
```diff
@@ -5,53 +5,166 @@
...
+      setLinks(newLinks);
+    }
+  };
+
+  const listRecords = links["oai-listrecords"];
+  const listIdentifiers = links["oai-listidentifiers"];
+
+  // Set default prefix based on backend link, use dublin core if unknown.
+  let defaultPrefix = getPrefixFromLink(listRecords)
+  if (!Object.prototype.hasOwnProperty.call(formats, defaultPrefix)) {
```
ntarocco, line 114: Suggestion: for more complex code, prefer assigning expressions to a `const` with easy-to-understand names.
```suggestion
  const hasDefaultPrefix = Object.prototype.hasOwnProperty.call(formats, defaultPrefix);
  if (!hasDefaultPrefix) {
```

ntarocco (approved):
Nice work! For me, it can be merged.
I added a few suggestions because it took me quite some time to understand what is happening here, with the hope that it can make the code even more readable.

PR #1094, opened 2022-09-29
datacite: fixed award id removal.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1904

this needs to be properly analysed. Right now, export fails because award's id is stripped and when the award is resolved (for export) it's not found in the DB.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1038,7 +1038,7 @@ def awards_v(app, funders_v):
     award = awards_service.create(
         system_identity,
         {
-            "id": "755021",
+            "id": "00k4n6c32::755021",
             "identifiers": [
                 {
                     "identifier": "https://cordis.europa.eu/project/id/755021",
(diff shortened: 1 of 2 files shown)


ppanero:
I seem to recall that we prepended the funder to the actual award id just in case different funders use the same award id (e.g. an int counter). See [here](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/contrib/awards/datastreams.py#L53). I think is safe to assume that the content of the PR is okay and we simply have to fix the tests.

ppanero (commented):
I seem to recall that we prepended the funder to the actual award id just in case different funders use the same award id (e.g. an int counter). See [here](https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/contrib/awards/datastreams.py#L53). I think is safe to assume that the content of the PR is okay and we simply have to fix the tests.

ppanero:
**we will need to backport this to go in v9 too. **

tests/conftest.py
```diff
@@ -457,7 +457,7 @@ def full_record(users):
                     "funder": {
                         "id": "00k4n6c32",
                     },
-                    "award": {"id": "00k4n6c32::755021"},
```
ppanero, line 460: we should change how the data goes in (i.e. the awards_v fixture) instead of the dump/expected data. To match https://github.com/inveniosoftware/invenio-vocabularies/blob/master/invenio_vocabularies/contrib/awards/datastreams.py#L53

ppanero (approved):

PR #1105, opened 2022-10-06
oai-set permissions: remove deprecated Admin generator

kpsherva:


tests/conftest.py
```diff
@@ -48,7 +48,7 @@
 from invenio_access.permissions import superuser_access, system_identity
 from invenio_accounts.models import Role
 from invenio_accounts.testutils import login_user_via_session
-from invenio_admin.permissions import action_admin_access
+from invenio_administration.permissions import administration_access_action
```
ntarocco, line 51: I think that in the tests, we wanted to test the `superuser-access` action, and not `Invenio-Administration` access.

ntarocco: I would suggest to use the same naming as here: https://github.com/inveniosoftware/invenio-records-permissions/pull/82

Can you please fix any other repo having this? Thanks!

kpsherva: in this particular ones (oai-sets resource) we wanted to grant the administration access (not the superuser) and I thought it makes sense to leave the admin access in tests as well.

PR #1113, opened 2022-10-18
add communities fixture loading

anikachurilova:
* create the communities fixture with all the required fields
* add fixture loading to fixtures mechanism
* add the yaml file with fake data for all the required fields
* closes https://github.com/zenodo/zenodo-rdm/issues/52


invenio_rdm_records/fixtures/communities.py
```diff
@@ -0,0 +1,40 @@
...
+            },
+            "slug": slug,
+            "metadata": {
+                "title": metadata_title,
+            },
+        }
+
+        service = current_communities.service
+        try:
+            service.create(data=community_data, identity=system_identity)
```
ppanero, line 38: moderate: could this be done asynchronously to avoid blocking during creation? i.e. in a celery task?

Also, note that there exists a similar functionality in [invenio-communities.cli](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/cli.py#L45), and a [task](https://github.com/inveniosoftware/invenio-communities/blob/master/invenio_communities/fixtures/tasks.py#L29) fot its creation already exists.

While I do agree that maybe we can have the data in this module (tha `.yaml` file), maybe we can use the existing task from communities? all in all just to consolidate the code and avoid duplicates 😇

invenio_rdm_records/fixtures/communities.py
```diff
@@ -0,0 +1,34 @@
...
+        self.logo_path = logo_path
+
+    def create(self, entry):
+        """Load a single community."""
+        logo = entry.pop("logo", None)
+
+        if logo:
+            logo_file_path = self.logo_path / logo
+            with open(Path(logo_file_path), "rb") as filestream:
+                create_demo_community(entry, filestream)
```
anikachurilova, line 32: Logo uploading moved to the create_demo_community task in order to reuse existing code and avoid code duplicates.
See [pull request](https://github.com/inveniosoftware/invenio-communities/pull/823).

invenio_rdm_records/fixtures/communities.py
```diff
@@ -0,0 +1,34 @@
...
+
+from invenio_communities.fixtures.tasks import create_demo_community
+
+from .fixture import FixtureMixin
+
+
+class CommunitiesFixture(FixtureMixin):
+    """Communities fixture."""
+
+    def __init__(self, logo_path, search_paths, filename):
```
ppanero, line 20: minor: Can "logo path" be moved to the end of the params list? to keep consistency of API with other `Fixture` classes.

invenio_rdm_records/fixtures/communities.py
ppanero, line 32: major: celery tasks can be called in 2 different ways:

a) as a function, like in this case and in line 34. That means they will run "eagerly" or "on the spot".
b) delayed or "asynchronously", for that we have to call them as `create_demo_community.delay(....)`. That is for example why the `Fixtures` class has a `delay` parameter. See more [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/fixtures/vocabularies.py#L346).

Extra information linking this and my previous comment. What would happen if you open the descriptor here (as done in line 31, but then call the task with `.delay`? Would it work?

ppanero (commented):
Nice work 🚀 ! We are almost there! Just some comments on asynchronous task handling 😇

invenio_rdm_records/fixtures/__init__.py
```diff
@@ -41,5 +42,11 @@ def run(self):
             "users.yaml",
         ).load()
 
+        CommunitiesFixture(
+            Path("./app_data/logos"),
+            [Path("./app_data"), dir_ / "data"],
```
slint, line 47: minor/shelve: something I noticed in other fixtures as well, the `Path("./app_data")` path assumes that the user is running the command from the root cookiecutted directory (or any directory with an `app_data` folder...

The correct approach would be to use `Path(current_app.instance_path) / "app_data"`. This can be addressed separately though, so feel free to shelve.

ppanero: @anikachurilova could you open an issue about this in invenio-rdm-records?

invenio_rdm_records/fixtures/__init__.py
```diff
@@ -28,18 +31,25 @@ def __init__(self, identity):
...
             self._identity,
-            app_data_folder=Path("./app_data"),
+            app_data_folder=app_data_folder,
             pkg_data_folder=dir_ / "data",
             filename="vocabularies.yaml",
         ).load()
 
         UsersFixture(
-            [Path("./app_data"), dir_ / "data"],
+            [app_data_folder, dir_ / "data"],
```
ppanero, line 44: very minor: create `data_folder` variable as you did with `app_data_folder`?

invenio_rdm_records/fixtures/communities.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, search_paths, filename, logo_path):
+        """Initialize the communities fixture."""
+        super().__init__(search_paths, filename)
+        self.logo_path = logo_path
+
+    def create(self, entry):
+        """Load a single community."""
+        logo = entry.pop("logo", None)
+
+        if logo:
```
ppanero, line 29: very minor, no need to change just as information. It could be done as having the parameters defined and then just do one function call (which would happen in any of the if/else clauses):


```python
kwargs = {}

if logo:
    kwargs["logo_path"] = logo_file_path

create_demo_community(entry, **kwargs)
```

But again, this is a personal bias/code style of mine. So no need to change 😅

ppanero (commented):
It LGTM! comments are minor, so no need to act on them. Mainly information. The only thing that is missing for it to go in are tests. You can check how the `UserFixture` is tested [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/tests/fixtures/test_fixtures.py#L176).

tests/fixtures/test_fixtures.py
```diff
@@ -189,6 +192,44 @@ def test_load_users(app, db, admin_role):
...
+    communities = CommunitiesFixture(
+        [dir_ / "app_data", dir_.parent.parent / "invenio_rdm_records/fixtures/data"],
+        "communities.yaml",
+        dir_ / "app_data" / "img",
+    )
+
+    # mandatory for files bucket creation
+    location = Location(name="local", uri=str(dir_), default=True)
+    db.session.add(location)
+    db.session.commit()
```
slint, line 207:
```suggestion
def test_load_communities(app, db, location):
    dir_ = Path(__file__).parent
    service = current_communities.service
    communities = CommunitiesFixture(
        [dir_ / "app_data", dir_.parent.parent / "invenio_rdm_records/fixtures/data"],
        "communities.yaml",
        dir_ / "app_data" / "img",
    )
```

minor: instead of defining a `Location` here, you can re-use the `location` fixture (coming from [`pytest-invenio`](https://github.com/inveniosoftware/pytest-invenio/blob/12aafb80acc61fe54e1575ecc125e0c854fdb100/pytest_invenio/fixtures.py#L643)).

slint (approved):
LGTM! One minor comment for using an existing test fixture (if possible and works :) )

PR #1123, opened 2022-11-04
Files fetching too

### LINKED ISSUE #1119: external files: apply new permission generator
see [comment](https://github.com/inveniosoftware/invenio-records-resources/pull/416#discussion_r1001581606) and `invenio-records-resources/tests/mock_module/permissions.py`

ppanero:
- closes #1119 

Tests are going to fail until the release chain is fixed (administration, drafts-resources and communities missing IIRC).

Screenshot 2022-11-04 at 16 38 11

### DIFF
--- a/invenio_rdm_records/services/generators.py
+++ b/invenio_rdm_records/services/generators.py
@@ -10,6 +10,7 @@
 """Invenio-RDM-Records Permissions Generators."""
 
 import operator
+from abc import abstractmethod
 from functools import reduce
 from itertools import chain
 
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/services/permissions.py
```diff
@@ -72,6 +73,10 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     can_read_files = [
         IfRestricted("files", then_=can_view, else_=can_all),
     ]
+    can_get_content_files = [
+        # preview is same as read_files
```
ppanero, line 77: This equivalences between preview/review and the new actions are taken from here <https://github.com/inveniosoftware/invenio-records-resources/commit/fd65e834e5426a789dd91e2544bdc52ae1ec1994#diff-13fd70056deb6848af5488fc2db9e690a05806d34b2b9333218e37bc60168d4b>. This means, we apply "same as before" but with the local files condition, resulting in:
```python
if local:
    same as before
else:
    only system process / superuser
```

invenio_rdm_records/services/generators.py
```diff
@@ -86,6 +116,44 @@ def query_filter(self, **kwargs):
             return q_public
 
 
+class IfDraft(ConditionalGenerator):
```
ppanero, line 119: I've asked in rdm-internal and nobody seems to be using this, plus this is the only usage in the code base. History tells that it was a temporary generator, replaced later by splitting the actions into `can_..._files` (for records) and `can_draft_..._files).

Should we remove it?

kpsherva: maybe deprecating it would be safer, especially because it goes as far as the latest LTS?

invenio_rdm_records/services/generators.py
```diff
@@ -18,12 +19,52 @@
...
+    )
+    """
+
+    def __init__(self, then_, else_):
+        """Constructor."""
+        self.then_ = then_
+        self.else_ = else_
+
+    @abstractmethod
+    def _condition(self, **kwargs):
```
zzacharo, line 43: I think this can be 
```suggestion
    def condition(self, **kwargs):
```

as it belongs to the public API of the class

zzacharo: Discussed IRL

invenio_rdm_records/services/generators.py
ppanero: I can add a deprecation warning. But IMO, we can just merge as is and do the deprecation in another PR if needed since it is not the actual goal of it, but more of an aftermath.

zzacharo (approved):

slint (approved):

PR #1131, opened 2022-11-18
Fix missing index prefix during creation of demo inclusion requests & invites

max-moser:
Discussion on Discord: https://discord.com/channels/692989811736182844/704629170470125578/1042794935717478400

### DIFF
--- a/invenio_rdm_records/fixtures/tasks.py
+++ b/invenio_rdm_records/fixtures/tasks.py
@@ -16,7 +16,6 @@ from flask_principal import Identity, UserNeed
 from invenio_access.permissions import any_user, authenticated_user, system_identity
 from invenio_communities.generators import CommunityRoleNeed
 from invenio_communities.members.errors import AlreadyMemberError
-from invenio_communities.members.records.api import Member
 from invenio_communities.proxies import current_communities
 from invenio_records_resources.proxies import current_service_registry
 from invenio_requests import current_events_service, current_requests_service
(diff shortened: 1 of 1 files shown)


kpsherva (approved):

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -79,7 +78,7 @@ def _get_random_community(communities):
     id = communities[r]["id"]
     # create community owner identity
     members = current_communities.service.members
-    Member.index.refresh()
+    members.indexer.refresh()
```
ntarocco, line 81: Wow, thanks for this fix 👀 
Could you please double-check if [this](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/tests/fixtures/test_cli.py#L108) test needs the same fix? Thanks!

max-moser: There's a bunch of test cases across many modules where this change could be applied.

I can't remember when exactly and where, but I remember briefly discussing fixes for tests with @ppanero.
IIRC, we concluded that it's okay for the tests to not have knowledge about configured custom search prefixes, because such configuration just isn't set there – and if it caused issues, tests would likely be failing anyway.

ppanero: That is correct. We could try applying the change everywhere but there was no easy way of getting all the occurences. Then, as Max mentioned, we use the `indexer` due to prefix knowledge.

PR #1143, opened 2022-12-15
providers: add validate_record to validate record according to provider [+]

### LINKED ISSUE #819: publisher required for DataCite DOI registration

### LINKED ISSUE #1137: Validate Publisher via DataCiteProvider

fenekku:
- closes #1137 ; closes #819

Replaces:
- closes https://github.com/inveniosoftware/react-invenio-deposit/pull/569
- closes https://github.com/inveniosoftware/invenio-rdm-records/pull/1018
- closes https://github.com/inveniosoftware/invenio-app-rdm/pull/1833

### DIFF
--- a/invenio_rdm_records/services/errors.py
+++ b/invenio_rdm_records/services/errors.py
@@ -52,3 +52,44 @@ class ReviewInconsistentAccessRestrictions(ReviewException):
             *args,
             **kwargs,
         )
+
+
+class ValidationErrorWithMessageAsList(Exception):
+    """Record Validation error where the messages are already a list.
+
+    There is a large context around this to understand. Field errors are
+    sent to the frontend by either:
+
+    - raising a marshmallow ValidationError that is then caught by an error
+      handler and converted to a list of
+      `{field: <fieldpath>, messages: <error msg array>}` in the returned JSON.
+      (See invenio_records_resources/resources/errors.py)
+    - not raising an error but filling out an `errors` list that is
+      serialzed out in JSON normally.
+
+    The conversion process in the first case expects a dict of error messages
+    inside the ValidationError. But these dict errors are converted to list
+    as soon as the schema is loaded. The rest of the codebase works with the
+    resulting list of errors. So for an error *raised* after schema
+    validation, one would have to construct it by converting the list back to
+    a dict, only for the error handler to reconvert it back again to that list
+    in the end. (Note: passing the list of errors directly would result in
+    `errors: {"_schema": <list of errors>}` passed to the frontend which
+    isn't what it expects.)
+
+    This error along with accompanying error_handler foregoes the need for
+    that.
+    """
+
+    def __init__(self, message):
+        """Constructor.
+
+        Note that we keep the `message` parameter as interface to look like
+        a marshmallow ValidationError even though it is annoying that the field
+        is then called by `self.messages`.
+
+        :param message: list of dicts in the shape:
+                        `{"<fieldA>": ["<msgA1>", ...], ...}`
+        """
+        assert isinstance(message, list)
+        self.messages = message
--- a/invenio_rdm_records/services/pids/manager.py
+++ b/invenio_rdm_records/services/pids/manager.py
@@ -13,15 +13,17 @@ from invenio_pidstore.errors import PIDDoesNotExistError
 from invenio_pidstore.models import PIDStatus
 from marshmallow import ValidationError
 
+from ..errors import ValidationErrorWithMessageAsList
 from .errors import PIDSchemeNotSupportedError, ProviderNotSupportedError
 
 
 class PIDManager:
     """RDM PIDs Manager."""
 
-    def __init__(self, providers):
+    def __init__(self, providers, required_schemes=None):
         """Constructor for RecordService."""
         self._providers = providers
+        self._required_schemes = required_schemes
 
     def _get_provider(self, scheme, provider_name=None):
         """Get a provider."""
@@ -46,17 +48,6 @@ class PIDManager:
         unknown_schemes = input_schemes - supported_schemes
         if unknown_schemes:
             raise PIDSchemeNotSupportedError(unknown_schemes)
-
-    def _validate_pids(self, pids, record, errors):
-        """Validate an iterator of PIDs.
-
-        This function assumes all pid schemes are supported by the system.
-        """
-        for scheme, pid in pids.items():
-            provider = self._get_provider(scheme, pid.get("provider"))
-            success, val_errors = provider.validate(record=record, **pid)
-            if not success:
-                errors.append({"field": f"pids.{scheme}", "messages": val_errors})
 
     def _validate_identifiers(self, pids, errors):
         """Validate and normalize identifiers."""
@@ -87,15 +78,56 @@ class PIDManager:
         for scheme, id_ in identifiers:
             pids[scheme]["identifier"] = id_
 
+    def _validate_pids(self, pids, record, errors):
+        """Validate an iterator of PIDs.
+
+        This function assumes all pid schemes are supported by the system.
+        """
+        for scheme, pid in pids.items():
+            provider = self._get_provider(scheme, pid.get("provider"))
+            success, val_errors = provider.validate(record=record, **pid)
+            if not success:
+                errors.append({"field": f"pids.{scheme}", "messages": val_errors})
+
+    def _validate_record(self, pids, record, errors):
+        """Validate the record according to the PIDs' requirements.
+
+        Here we check if the record is compatible from the point of
+        view of the pids...
+            - ... it contains
+            - ... it would contain according to configured required pids
+
+        The responsibility lies with each provider since they are the ones
+        that know their criteria for a record that is complete enough to get
+        a PID.
+        """
+        schemes = set(pids.keys()) | set(self._required_schemes)
+        scheme_provider_names = [
+            # provider_name for an absent-but-required pid will be None
+            (scheme, pids.get(scheme, {}).get("provider"))
+            for scheme in schemes
+        ]
+        providers = [
+            self._get_provider(scheme, provider_name)
+            for scheme, provider_name in scheme_provider_names
+        ]
+
+        for provider in providers:
+            success, provider_errors = provider.validate_record(record)
+            if not success:
+                errors.extend(provider_errors)
+
     def validate(self, pids, record, errors=None, raise_errors=False):
         """Validate PIDs."""
+        # if errors is [] we have to use it
         errors = [] if errors is None else errors
         self._validate_pids_schemes(pids)
         self._validate_identifiers(pids, errors)
         self._validate_pids(pids, record, errors)
+        self._validate_record(pids, record, errors)
 
         if raise_errors and errors:
-            raise ValidationError(message=errors)
+            raise ValidationErrorWithMessageAsList(message=errors)
 
     def read(self, scheme, identifier, provider_name):
         """Read a pid."""
--- a/invenio_rdm_records/services/pids/manager.py
+++ b/invenio_rdm_records/services/pids/manager.py
@@ -85,7 +85,7 @@ class PIDManager:
         """
         for scheme, pid in pids.items():
             provider = self._get_provider(scheme, pid.get("provider"))
-            success, val_errors = provider.validate(record=record, **pid)
+            success, val_errors = provider.validate_pid(record=record, **pid)
             if not success:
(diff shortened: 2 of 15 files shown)


invenio_rdm_records/services/errors.py
```diff
@@ -52,3 +52,44 @@ def __init__(self, *args, **kwargs):
             *args,
             **kwargs,
         )
+
+
+class ValidationErrorWithMessageAsList(Exception):
```
fenekku, line 57: Start here to understand the error handling motivation and see the test to see it in action.

tests/resources/test_resources_pids.py
```diff
@@ -147,3 +147,23 @@ def test_external_doi_required(
         {"field": "pids.doi", "messages": ["Missing DOI for required field."]}
     ]
     assert draft.json["pids"] == {"doi": {"provider": "external", "identifier": ""}}
+
+
+def test_pids_publish_validation_error(
```
fenekku, line 152: Test that checks we solved the error handling issue.

invenio_rdm_records/services/pids/manager.py
```diff
@@ -87,15 +78,56 @@ def _validate_identifiers(self, pids, errors):
...
+        Here we check if the record is compatible from the point of
+        view of the pids...
+            - ... it contains
+            - ... it would contain according to configured required pids
+
+        The responsibility lies with each provider since they are the ones
+        that know their criteria for a record that is complete enough to get
+        a PID.
+        """
+        schemes = set(pids.keys()) | set(self._required_schemes)
```
fenekku, line 104: Solves a subtle potential issue compared to previous PR where 2 different providers could have been listed for the same scheme (the default and the selected) or same provider listed twice.

invenio_rdm_records/services/pids/manager.py
```diff
@@ -87,15 +78,56 @@ def _validate_identifiers(self, pids, errors):
...
+                errors.extend(provider_errors)
+
     def validate(self, pids, record, errors=None, raise_errors=False):
         """Validate PIDs."""
+        # if errors is [] we have to use it
         errors = [] if errors is None else errors
         self._validate_pids_schemes(pids)
         self._validate_identifiers(pids, errors)
         self._validate_pids(pids, record, errors)
+        self._validate_record(pids, record, errors)
```
fenekku, line 127: Make the validate_record call private (_validate_record) and call it here with `record`.

I buy the argument that `record` is what is the closest to what is published and therefore makes sense to validate (although the reader may be surprised like I originally was). (I don't buy so much the argument about "another component, added in-between in the future, that could change the record metadata. You cannot predict that." - that could be said about anything in which case any validation we make would be useless because it could be undone by another component.)

invenio_rdm_records/services/pids/providers/base.py
```diff
@@ -164,3 +164,17 @@ def validate(self, record, identifier=None, provider=None, **kwargs):
             pass
 
         return True, []
+
+    def validate_record(self, record):
```
ntarocco, line 168: As discussed in the previous PR, IMO this creates more confusion than clarity.
The `validate` method above is enough to validate PID, record and anything else. It was designed exactly for these use cases.
@zzacharo shared the same concerns about duplicating this method.

It would be nice to have a comment from other reviewers too (@tmorrell maybe, given that you have worked extensively around this).

invenio_rdm_records/services/pids/manager.py
ntarocco: It is not a question of what arguments you buy. The validation makes sense on what will be used by the provider (the record). There is no guarantee that the `draft` will not be potentially changed by another component, before or after.

invenio_rdm_records/services/pids/manager.py
ntarocco: I guess that if a required scheme is not in `pids`, we should fail?
Why not validating this before, very explicitly, and probably in `_validate_pids`?

If this is a subtle use case, it should be made very clear and very explicit, e.g. documenting it in the code or via a named var/func.
The code with the extra comments `# provider_name for an absent-but-required pid will be None` basically hides all this.

invenio_rdm_records/services/pids/providers/base.py
fenekku: I am so sorry, but I think there was a misunderstanding. I understood [the discussion](https://github.com/inveniosoftware/invenio-rdm-records/pull/1140/files#diff-82789e81eafd9ec8071709d9a32ecd6b52c83cb1b9bf4847924ddbd54727fa7b) to be on `validate_record` at the pids manager level in `manager.py` as a public method. My bad!

Here is what I understand you are suggesting:

Merge `_validate_record` into `_validate_pids` at the pids manager level and merge `validate_record` add `validate` at the provider level.

I think this is problematic. Let me try to break up my argument in 3 pieces so it is very clear to any reader:

#### 1- There is value in putting different things in different places

`validate` (I actually renamed it to `validate_pid` to address confusion problem) and `validate_record` on the provider do 2 different things. One of those things is to validate a **given** PID (an identifier/provider/client) and the *other* is to validate a record (actually the metadata) that potentially doesn't have a PID (e.g. when creating a new draft and letting InvenioRDM mint the DOI). Indeed, provider validation is also done on draft creation/save not just on publish. Passing the record for `validate_pid` is fine because we need to be able to tell if the pid is a duplicate.

For the confusion problem: if we walk through 2 scenarios in the shoes of a developer, things look like the following:

Scenario 1: Add a new provider

As a developer, I want to mint a new kind of PID in my instance. I therefore create a new `MyProvider` class that inherits from `PIDProvider`. Among the questions I need to answer are: what is a valid `identifier` (and potentially `provider`/`client`) for my PID? and what caracteristics must a record have in order to get minted my new kind of PID? Lo and behold two separate methods then exist to answer these 2 separate questions: `validate_pid` answers the first and `validate_record` answers the second.

Scenario 2: Maintain the codebase

I am a developer (of any experience level) and I need to fix a bug in how a Provider validates their identifier. `validate_pid` is the only method I need to know about and fix. I don't need to also understand the code related to what is a valid record according to this provider. You would need to understand that unrelated code if it were in the same method.

---

#### 2- The problem of very complex logic around "emptiness"

Conflating `validate_pid` and `validate_record` would also mean conflating the 2 loops that go through providers. One of those loops was only looking at given pids and the other was looking at given pids UNION required pids. In the case of given pids, an empty value is typically bad, while in the case of required pids an empty value is normal because the pid hasn't been minted yet (remember that validation happens at draft creation too). If both are funnelled in the same method we need to distinguish the cases where the PID (`identifier`/`provider`) is empty but shouldn't be, from the ones where it's empty but it's fine. That is incredibly error-prone. 

Pids are already very complex: we have to keep the draft/publish scenarios in mind, the pids that can change and those that can't, the external ones and the internal ones... Keeping the methods separate saves us from some complexity.

---

#### 3- The time and unintended consequences of extensive code changes

This goes back to the previous discussion we had. Conflating in one method means changing the interface of the errors outputted by the provider from a list of strings to a list of dicts. This involves changing a number of files that would not have been touched otherwise. This takes more time and opens up the risk of introducing other bugs.

---

To help move this along/identify what I am missing, I'd be happy to 

- have someone else's opinion on these points  
- take a look at actual alternative code suggestions
- have a video discussion about it in January. I am not sure I can be more precise than I am right now though :laughing: 

I also hope that it is clear that my intent is to better the codebase and the project, not to save myself from necessary work or for any other reason.

See you in 2023!

ntarocco: Given that the returned errors are now the same type as before, there is no need to break the interface and rename `validate` to `validate_pid`.
The interface was very clear: `validate(pid, record)`. The developer will decide **internally** if it makes sense to split in that provider's implementation.

Example:
```
def validate(pid, record):
    _validate_pid(pid)
    _validate_record(record)
```

There is no point to over-engineer this, given that we don't know upfront if all provider's implementation will need to validate both pid and record metadata.
If there was no change in the interface, we could probably go with your solution, but now I don't see the advantage.

Finally, about the for-loops, I agree it is very complex, but let's answer in the comment above: it is not clear at all to me why we are checking for the required_schemes and why we do that in the `validate_record` method.

invenio_rdm_records/services/pids/manager.py
```diff
@@ -87,15 +78,56 @@ def _validate_identifiers(self, pids, errors):
...
+
+        This function assumes all pid schemes are supported by the system.
+        """
+        for scheme, pid in pids.items():
+            provider = self._get_provider(scheme, pid.get("provider"))
+            success, val_errors = provider.validate(record=record, **pid)
+            if not success:
+                errors.append({"field": f"pids.{scheme}", "messages": val_errors})
+
+    def _validate_record(self, pids, record, errors):
```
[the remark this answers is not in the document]
fenekku: Compared to previous `_validate_pids`, 

- `_validate_record` checks if the record is valid according to minters: `provider.validate_record` is called for schemes in the draft union required schemes (those "required" pids may not be on the record at this point and that's fine) 
- `_validate_record` extends the list of errors with a different error dict per different specific incorrect fields if there are errors (i.e. multiple error dicts for a single pid). 

`_validate_pids` checked/checks if the provided `"pids"` (pid/client/provider) (i.e. schemes in the passed draft) themselves were/are valid and it created/creates a single error dict per pid.

Those are the differences.

invenio_rdm_records/services/pids/manager.py
fenekku: > 

No: if a required scheme is not in pids it should not necessarily fail. We may be in the case where a new draft is made/published and doesn't have the required PID yet (e.g. doesn't have `"doi": {...}` in its "pids" fields yet). The test for this case already exists here:
https://github.com/inveniosoftware/invenio-rdm-records/blob/87b986241e53f8741b4ab9a457b98b26bfb6e2a7/tests/services/components/test_pids_component.py#L351 .

For the `provider_name` being `None`, that's because in that case the `"provider"` value is not given. By passing `None` we get the default provider for the scheme to be chosen.

fenekku:
I just got back from break today and I know CERN is back only on the 5th. Let's arrange a meeting to discuss this on Friday or next week: I've replied to the questions above and would be interested in what alternative you had in mind.

See you soon!

invenio_rdm_records/services/pids/manager.py
ntarocco: >  if a required scheme is not in pids it should not necessarily fail... a new draft is published and doesn't have the required PID yet...
If published, it must have it, no?

In any case, honestly, the fact that we need to align on the implemented behaviour with PR comments is a hint that we should improve existing or new implementations, independently who coded it, to make it more readable/understandable/immediate: no one wants to spend hours trying to figure out what a few lines of code are solving.

invenio_rdm_records/services/pids/providers/base.py
kpsherva: I agree with @ntarocco.
In my opinion, interface of `validate(pid, record)` is flexible enough to accommodate any needed validations inside. In addition, the interface is clean and easily understandable - it does not introduce the separation for pid and record, which is valuable in cases where it is simply not needed, providing a more generic solution - exactly what you would expect from a framework - and leaves it to the developer to decide about validating their specific use case.

The required schemes check is also not clear to me.

fenekku: Ok... 3rd time's the charm :slightly_smiling_face: https://github.com/inveniosoftware/invenio-rdm-records/pull/1146/files ?  I *think* that new PR combines `validate_pid` and `validate_record` into `validate()` as you want (right?). Perhaps it makes clearer some of the consequences of that too.

ntarocco:
https://github.com/inveniosoftware/invenio-rdm-records/pull/1146 merged 🚀, shall we close this one now?

fenekku:
Closed in favor of https://github.com/galterlibrary/galter-ir-importer/pull/137

PR #1146, opened 2023-01-09
providers: validate record according to provider [+]

### LINKED ISSUE #1137: Validate Publisher via DataCiteProvider

fenekku:
- closes #1137

### DIFF
--- a/invenio_rdm_records/services/pids/manager.py
+++ b/invenio_rdm_records/services/pids/manager.py
@@ -13,15 +13,17 @@ from invenio_pidstore.errors import PIDDoesNotExistError
 from invenio_pidstore.models import PIDStatus
 from marshmallow import ValidationError
 
+from ..errors import ValidationErrorWithMessageAsList
 from .errors import PIDSchemeNotSupportedError, ProviderNotSupportedError
 
 
 class PIDManager:
     """RDM PIDs Manager."""
 
-    def __init__(self, providers):
+    def __init__(self, providers, required_schemes=None):
         """Constructor for RecordService."""
         self._providers = providers
+        self._required_schemes = required_schemes
 
     def _get_provider(self, scheme, provider_name=None):
         """Get a provider."""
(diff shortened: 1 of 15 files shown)


invenio_rdm_records/services/pids/manager.py
```diff
@@ -87,15 +78,50 @@ def _validate_identifiers(self, pids, errors):
...
+            - ... it would contain according to configured required pids
+
+        The responsibility lies with each provider since they are the ones
+        that know their criteria for a record that is complete enough to get
+        a PID.
+        """
+        # Validate according to the schemes that the draft has and the schemes that
+        # the draft would be given. _required_schemes are schemes that would be given
+        # (if not already on the draft).
+        schemes = set(pids.keys()) | set(self._required_schemes)
```
fenekku, line 98: Hopefully that clarifies the "required_schemes" bit a little for you.

fenekku (commented):
This is the combined pid and record validation approach.

Compared to the previous PR, it's slightly more code that needs more prudence and it modifies more pre-existing code instead of adding separate code. 

share your thoughts.

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -196,16 +197,36 @@ def delete(self, pid, **kwargs):
...
+        # Checking if the identifier is not None is crucial because not all records at
+        # this point will have a DOI identifier (and that is fine in the case of initial
+        # creation)
+        if identifier is not None:
+            # Format check
+            try:
+                self.client.api.check_doi(identifier)
+            except ValueError as e:
+                # modifies the error in errors in-place
+                error = self._get_or_append_error_dict(errors)
```
ntarocco, line 218: Nice the `check_doi`!

From a first look, I thought you had forgotten the `errors.append(error)`, but then, if I understand correctly, what is happening is:
1. `errors` is modified inside the func `_get_or_append_error_dict`, and `error` appended
2. the appended `error` is then return and modified

Wouldn't it better to pass the `message` inside and avoid this?
E.g.
```
errors = self._get_or_append_error_dict(errors, error_msg)
```
or no return value:
```
self._get_or_append_error_dict(errors, error_msg)
```
and rename the func `append_error_dict`? It looks like below that `error_msg` might a single item or a list.

fenekku: The above is indeed what was happening: returning an error_dict that had been appended so that modifications to it would modify the error_dict in `errors`.

The reason behind my approach for `get_or_append_error_dict` was just reflexive generality - the method provides a more generic functionality (retrieving an error dict guaranteed to be in given `errors`) that can be re-used in more cases and those cases are responsible for their own quirks (like appending or extending when one or multiple messages). I had been able to give a crisp name to the method (`get_or_append_error_dict` is crisp) which was a sign for me that it was a nice encapsulation. 

Anyway, I've replaced it with `_insert_pid_type_error_msg` which isn't as generic, but does the trick for what we need as you suggested.

zzacharo (approved):

ntarocco:
Thanks a lot for all your work and changes!

PR #1148, opened 2023-01-13
 communities: returns 404 when not found

TLGINO:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1965


kpsherva (approved):

ntarocco (approved):
LGTM! Thank you for this fix and 👏🏼 for your first PR!

To optimize the code even further, it would be nice to use an `exists` method instead of `read`. We can probably add this `invenio-records-resources` and then use it here.

PR #1156, opened 2023-01-20
datacite: fix reversion in affiliation ROR handling and cleanup

tmorrell:
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1157

### Description

This PR fixes a reversion in the DataCite serializer. In v9 ROR affiliations were serialized with the whole url https://ror.org/identifier. That changed in v10 to just the identifier portion due to other identifier cleanup. This is an issue because DataCite expects ROR affiliations to be in their url form and unceremoniously deletes them if not. I've asked DataCite to change this, but we can also fix this directly in our serializer.

I also did a bit of cleanup by using the same approach as funders for determining which identifier to use.

### Checklist

- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).

**Third-party code**

**Reminder**

### DIFF
--- a/invenio_rdm_records/resources/serializers/datacite/schema.py
+++ b/invenio_rdm_records/resources/serializers/datacite/schema.py
@@ -94,20 +94,36 @@ class PersonOrOrgSchema43(Schema):
                 }
                 identifiers = affiliation.get("identifiers")
                 if identifiers:
-                    # PIDS-FIXME: DataCite accepts only one, how to decide
-                    identifier = identifiers[0]
+                    # FIXME: Make configurable
+                    DATACITE_AFFILIATION_IDENTIFIER_TYPES_PREFERENCE = (
+                        "ror",
+                        "isni",
+                        "gnd",
+                    )
+                    identifier = get_preferred_identifier(
+                        DATACITE_AFFILIATION_IDENTIFIER_TYPES_PREFERENCE, identifiers
+                    )
+                    if not identifier:
+                        identifier = identifiers[0]
+                        identifier["scheme"] = "Other"
+
                     id_scheme = get_scheme_datacite(
                         identifier["scheme"],
                         "VOCABULARIES_AFFILIATION_SCHEMES",
-                        default=identifier["scheme"],
-                    )
-
-                    if id_scheme:
-                        aff["affiliationIdentifier"] = identifier["identifier"]
-                        aff["affiliationIdentifierScheme"] = id_scheme.upper()
+                        default=identifier["scheme"].upper(),
                         # upper() is fine since this field is free text. It
                         # saves us from having to modify invenio-vocabularies
                         # or do config overrides.
+                    )
+
+                    if id_scheme == "ROR":
+                        identifier_value = "https://ror.org/" + identifier["identifier"]
+                    else:
+                        identifier_value = identifier["identifier"]
+
+                    if id_scheme:
+                        aff["affiliationIdentifier"] = identifier_value
+                        aff["affiliationIdentifierScheme"] = id_scheme
 
                 serialized_affiliations.append(aff)
 
@@ -462,8 +478,14 @@ class DataCite43Schema(Schema):
         """Get funding references."""
         # constants
         # FIXME: Make configurable
-        FUNDER_IDENTIFIER_TYPES_PREFERENCE = ("ror", "grid", "doi", "isni", "gnd")
-        AWARD_IDENTIFIER_TYPES_PREFERENCE = ("doi", "url")
+        DATACITE_FUNDER_IDENTIFIER_TYPES_PREFERENCE = (
+            "ror",
+            "grid",
+            "doi",
+            "isni",
+            "gnd",
+        )
+        DATACITE_AWARD_IDENTIFIER_TYPES_PREFERENCE = ("doi", "url")
         TO_FUNDER_IDENTIFIER_TYPES = {
             "isni": "ISNI",
             "gnd": "GND",
@@ -486,7 +508,7 @@ class DataCite43Schema(Schema):
             identifiers = funder.get("identifiers", [])
             if identifiers:
                 identifier = get_preferred_identifier(
-                    FUNDER_IDENTIFIER_TYPES_PREFERENCE, identifiers
+                    DATACITE_FUNDER_IDENTIFIER_TYPES_PREFERENCE, identifiers
                 )
                 if not identifier:
                     identifier = identifiers[0]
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -94,20 +94,32 @@ def get_affiliation(self, obj):
...
-                    if id_scheme:
-                        aff["affiliationIdentifier"] = identifier["identifier"]
-                        aff["affiliationIdentifierScheme"] = id_scheme.upper()
+                        default=identifier["scheme"].upper(),
                         # upper() is fine since this field is free text. It
                         # saves us from having to modify invenio-vocabularies
                         # or do config overrides.
+                    )
+
+                    if id_scheme == "ROR":
```
ppanero, line 115: I'm wondering if this is the case for more than one scheme, e.g. we also had ORCID URLs [here](https://github.com/inveniosoftware/invenio-rdm-records/commit/2cb3ecb5ae7a164a403442a64ce569fff8ec9b42#diff-e35eefcbf00ccd39e06e792ade38b37c8c0be576e979678bf099d1741e61ca0a), in which case we might need something more generic 🤔

ppanero, line 115: I'm wondering if this is the case for more than one scheme, e.g. we also had ORCID URLs [here](https://github.com/inveniosoftware/invenio-rdm-records/commit/2cb3ecb5ae7a164a403442a64ce569fff8ec9b42#diff-e35eefcbf00ccd39e06e792ade38b37c8c0be576e979678bf099d1741e61ca0a), in which case we might need something more generic 🤔

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -94,20 +94,32 @@ def get_affiliation(self, obj):
                 }
                 identifiers = affiliation.get("identifiers")
                 if identifiers:
-                    # PIDS-FIXME: DataCite accepts only one, how to decide
-                    identifier = identifiers[0]
+                    # FIXME: Make configurable
+                    AFFILIATION_IDENTIFIER_TYPES_PREFERENCE = ("ror", "isni", "gnd")
```
ppanero, line 98: naming 😅: would we use this variable somewhere else? It has no relation to "datacite" nor "serialization", but it indicates the preference in that specific case.

ppanero (commented):
Thanks for the PR! 

Note: when this is released, we will need to ship a script to retrigger datacite updates for all records 😮 Othewise, as reported in Discord, records are not properly linked.

invenio_rdm_records/resources/serializers/datacite/schema.py
tmorrell: I've tested ORCIDS (as well as funder RORs) and DataCite handles them fine if they don't have the url. So I think this is isolated just to affiliation RORs.

slint (approved):

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -94,20 +94,36 @@ def get_affiliation(self, obj):
                 }
                 identifiers = affiliation.get("identifiers")
                 if identifiers:
-                    # PIDS-FIXME: DataCite accepts only one, how to decide
-                    identifier = identifiers[0]
+                    # FIXME: Make configurable
+                    DATACITE_AFFILIATION_IDENTIFIER_TYPES_PREFERENCE = (
```
zzacharo, line 98: Does this make sense to be configurable? Does this list get updated? If not, I would remove the FIXME otherwise I would export this to the module's config. What do you think @slint @tmorrell ?

slint: This should be configurable, in the sense that it decides what is the preferred identifier for sending to DataCite, which is an instance preference at the end (since there's no recommendation from the DataCite schema on what is "the" canonical identifier type for various entities like authors, orgs, funders, etc.).

I think the difficulty with extracting into config at the moment, might be that we do unit testing without assuming an application context pushed, so it might turn into a much bigger task of refactoring tests.

tmorrell: I agree with @slint that this would be good to make configurable in the future, but the complexity of modifying the config and tests is too much for the moment. We haven't had anyone request to have this configuration, and if we don't merge this for v11 we're going to continue to be sending missing metadata to DataCite.

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -94,20 +94,36 @@ def get_affiliation(self, obj):
...
-                    if id_scheme:
-                        aff["affiliationIdentifier"] = identifier["identifier"]
-                        aff["affiliationIdentifierScheme"] = id_scheme.upper()
+                        default=identifier["scheme"].upper(),
                         # upper() is fine since this field is free text. It
                         # saves us from having to modify invenio-vocabularies
                         # or do config overrides.
+                    )
+
+                    if id_scheme == "ROR":
```
ppanero, line 119: @tmorrell I'm sorry for coming back to this again. As you mentioned DataCite seems to work with ORCID even if this is not a URL, however, does it work if it is a URL? If the answer is yes, then we can simplify the code into:

```python
 if id_scheme:
                         aff["affiliationIdentifier"] = idutils.to_url(identifier_value)
                         aff["affiliationIdentifierScheme"] = id_scheme
```

This would remove the need to tread schemes differently, simplifies code, and makes it more maintainable IMO.

Once again, sorry for the loop hehe

tmorrell: That would be an elegant solution if all identifiers were supported in to_url. ISNI is on the RDM metadata scheme, but not in to_url. I might be able to add an ISNI url format to idutils, but it seems a bit beyond the scope of this PR.

ppanero: ahhhh you are right, I had assumed that `to_url` would return the value if the URL was not configured, but it seems it returns `""`. See [here](https://github.com/inveniosoftware/idutils/blob/master/idutils/__init__.py#L1008). 

Then I'd suggest:
1. We merge this to get the fix in
2. We open an issue to re-work it properly, and evaluate the change on default behaviour of idutils

Thanks Tom!

zzacharo (approved):

zzacharo (approved):

PR #1162, opened 2023-01-27
records: remove from community

ppanero:
- closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1163

### DIFF
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -91,6 +91,8 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
     routes["item-review"] = "/<pid_value>/draft/review"
     routes["item-actions-review"] = "/<pid_value>/draft/actions/submit-review"
     # Community records
+    routes["item-remove-community"] = "/<pid_value>/communities"
+    routes["item-remove-community-list"] = "/<pid_value>/communities/<community_id>"
     routes["community-records"] = "/communities/<pid_value>/records"
 
     request_view_args = {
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/resources/config.py
```diff
@@ -91,6 +91,8 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
     routes["item-review"] = "/<pid_value>/draft/review"
     routes["item-actions-review"] = "/<pid_value>/draft/actions/submit-review"
     # Community records
+    routes["item-remove-community"] = "/<pid_value>/communities"
```
ppanero, line 94: remove bulk, this functionality has been out-scoped

invenio_rdm_records/resources/config.py
```diff
@@ -91,6 +91,8 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
...
+    routes["item-remove-community-list"] = "/<pid_value>/communities/<community_id>"
```
ppanero, line 95: move to an `/actions` endpoint, something similar to what Zenodo does:

```shell
POST /communities/<community_id>/curaterecord/

Payload:

{
    "recid": "1133576",
    "action":"remove"
}
```

This would enable support for the community addition, without having to send a "PUT" operation on the record. Therefore, we would be treating communities in a similar way we are treating PIDs.

tests/services/test_rdm_service.py
```diff
@@ -130,34 +130,39 @@ def test_embargo_lift_with_error(running_app, search_clear, minimal_record):
         service.lift_embargo(_id=record["id"], identity=superuser_identity)
 
 
+#
+# Communities
+#
+def _create_record(db, data):
```
ppanero, line 136: These two functions might be better moved to a fixture

jrcastro2:
superseed by https://github.com/inveniosoftware/invenio-rdm-records/pull/1178

Issue #1165, opened 2023-01-30
add export format for MarcXML


TLGINO:

ppanero:
Missing merging app-rdm

PR #1166, opened 2023-01-30
serializer: marcxml + testing

TLGINO:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1165

### DIFF
--- a/invenio_rdm_records/resources/serializers/marcxml/__init__.py
+++ b/invenio_rdm_records/resources/serializers/marcxml/__init__.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""MARCXML Serializer for Invenio RDM Records."""
+
+
+import traceback
+from copy import deepcopy
+
+from dojson.contrib.to_marc21.utils import dumps_etree
+from flask_resources.serializers import SerializerMixin
+from lxml import etree
+
+from .schema import MARCXMLSchema
+
+
+class MARCXMLSerializer(SerializerMixin):
+    """Marshmallow based MARCXML serializer for records.
+
+    Note: This serializer is not suitable for serializing large number of
+    records.
+    """
+
+    def __init__(self, **options):
+        """Constructor."""
+        self._schema_cls = MARCXMLSchema
+
+    def serialize_object(self, obj):
+        """Serialize a single record and persistent identifier to etree.
+
+        :param obj: Record instance
+        """
+        json = self._schema_cls().dump(obj)
+
+        def changeKeysToTags(dict):
+            """Changes the key name to the corresponding MARCXML tag (number)."""
+            # [!] The string in the first array corresponds to the tag[0:4] + ind1[4] + ind2[5]
+            # [!] The first string needs to be length *5* (this is to do with the dojson parser)
+            # [!] The second string corresponds to the subfield code
+
+            # Example: "creators" : ["100a ", "a"]
+
+            #   <datafield tag="100" ind1="a" ind2=" ">
+            #       <subfield code="a">Tarocco, Nicola</subfield>
+            #   </datafield>
+
+            # To add support for more tags, use the corresponding codes from here
+            # https://scoap3.org/scoap3-repository/marcxml/
+
+            changes = {
+                "contributors": ["700a ", "u"],  # Abstract
+                "titles": ["245a ", "a"],  # Title
+                "creators": ["100a ", "a"],  # First author
+                "identifiers": ["024  ", "a"],  # DOI
+                "relations": ["856 2", "a"],  # Related Ressource
+                "rights": ["540  ", "a"],  # License
+                "dates": [
+                    "260c ",
+                    "c",
+                ],  # Publication Information - Date of Publication
+                "subjects": ["653  ", "a"],  # Keywords
+                "descriptions": ["520  ", "a"],  # Abstract
+                "publishers": [
+                    "260b ",
+                    "a",
+                ],  # Publication Information - Publisher Name
+                "types": ["901  ", "u"],  # Local Implementation
+                "sources": ["246i ", "x"],  # Source
+                "coverage": ["510  ", "a"],  # Location
+                "formats": ["520 1", "a"],  # Abstract
+                "parent_id": ["024 1", "a"],  # ID
+                "community_id": ["024 2", "a"],  # ID
+                "last_updated": [
+                    "260ca",
+                    "c",
+                ],  # Publication Information - Date of Publication
+                "sizes": ["520 2", "a"],  # Abstract
+                "version": ["024 3", "a"],  # ID
+                "funding": ["856 1", "a"],  # Related Ressource
+            }
+
+            dict_copy = deepcopy(dict)
+            for key in dict_copy:
+                if key in changes:
+                    dict[changes[key][0]] = {changes[key][1]: dict.pop(key)}
+
+            return dict
+
+        json = changeKeysToTags(json)
+        return etree.tostring(dumps_etree(json))
--- a/invenio_rdm_records/resources/serializers/marcxml/schema.py
+++ b/invenio_rdm_records/resources/serializers/marcxml/schema.py
@@ -65,19 +65,21 @@ class MARCXMLSchema(Schema):
         for funder in all_funders:
             funder_string = ""
 
-            identifiers = funder.get("award", {}).get("identifiers", [{}])[0]
-            scheme = identifiers.get("scheme", "null")
-            identifier = identifiers.get("identifier", "null")
+            award = funder.get("award", {})
 
-            title = funder.get("award", {}).get("title", {})
-            title = list(title.values())[0] if title != {} else "null"
+            identifier = award.get("identifiers", [{}])[0]
+            scheme = identifier.get("scheme", "null")
+            identifier_value = identifier.get("identifier", "null")
 
-            number = funder.get("award", {}).get("number", "null")
+            title = award.get("title", {})
+            title = list(title.values())[0] if title else "null"
+
+            number = award.get("number", "null")
             id = funder["funder"]["id"]
             name = funder["funder"].get("name", "null")
 
             funder_string += f"award_identifiers_scheme={scheme}; "
-            funder_string += f"award_identifiers_identifier={identifier}; "
+            funder_string += f"award_identifiers_identifier={identifier_value}; "
             funder_string += f"award_title={title}; "
             funder_string += f"award_number={number}; "
             funder_string += f"funder_id={id}; "
@@ -94,11 +96,11 @@ class MARCXMLSchema(Schema):
         """Get locations."""
         locations = []
 
-        access_location = obj["metadata"].get("locations", {})
-        if access_location == {}:
+        access_location = obj["metadata"].get("locations", [])
+        if not access_location:
             return missing
 
-        for location in obj["metadata"].get("locations", []).get("features", []):
+        for location in access_location.get("features", []):
             location_string = ""
 
             place = location.get("place")
--- a/invenio_rdm_records/resources/serializers/marcxml/schema.py
+++ b/invenio_rdm_records/resources/serializers/marcxml/schema.py
@@ -1,0 +1,286 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""MARCXML based Schema for Invenio RDM Records."""
+
+import bleach
+from invenio_access.permissions import system_identity
+from invenio_vocabularies.proxies import current_service as vocabulary_service
+from marshmallow import Schema, fields, missing
+
+from ..ui.schema import current_default_locale
+from ..utils import get_vocabulary_props
+
+
+class MARCXMLSchema(Schema):
+    """Schema for records in MARC."""
+
+    contributors = fields.Method("get_contributors")
+    titles = fields.Method("get_titles")
+    creators = fields.Method("get_creators")
+    identifiers = fields.Method("get_identifiers")
+    relations = fields.Method("get_relations")
+    rights = fields.Method("get_rights")
+    dates = fields.Method("get_dates")
+    subjects = fields.Method("get_subjects")
+    descriptions = fields.Method("get_descriptions")
+    publishers = fields.Method("get_publishers")
+    types = fields.Method(
+        "get_types"
+    )  # Corresponds to ressource_type in the metadata schema
+    sources = fields.Method(
+        "get_sources"
+    )  # Corresponds to references in the metadata schema
+    coverage = fields.Method(
+        "get_locations"
+    )  # Corresponds to locations in the metadata schema
+    formats = fields.Method("get_formats")
+    parent_id = fields.Method("get_id")  # No GOOD
+    community_id = fields.Method("get_id_community")
+    last_updated = fields.Method("last_updated")
+    sizes = fields.Method("get_sizes")
+    version = fields.Method("get_version")
+    funding = fields.Method("get_funding")
+
+    def get_version(self, obj):
+        """Get the version of the record."""
+        return obj["metadata"].get("version", missing)
+
+    def get_sizes(self, obj):
+        """Get the size of the record."""
+        return obj["metadata"].get("sizes", missing)
+
+    def get_funding(self, obj):
+        """Get funding information."""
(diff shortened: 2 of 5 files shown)


invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -0,0 +1,193 @@
...
+from marshmallow import Schema, fields, missing
+
+from ..ui.schema import current_default_locale
+from ..utils import get_vocabulary_props
+
+
+class MARCXMLSchema(Schema):
+    """Schema for records in MARC."""
+
+    contributors = fields.Method("get_contributors")
```
zzacharo, line 24: How was this list gathered? From zenodo? Is there something missing still from InvenioRDM model?

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -0,0 +1,193 @@
...
+    def get_creators(self, obj):
+        """Get creators."""
+        return [c["person_or_org"]["name"] for c in obj["metadata"].get("creators", [])]
+
+    def get_relations(self, obj):
+        """Get relations."""
+        rels = []
+
+        # Fundings
+        # FIXME: Add after UI support is there
```
zzacharo, line 60: leftover? why is not possible now?

tests/resources/serializers/test_marcxml_serializer.py
```diff
@@ -0,0 +1,93 @@
...
+def test_marcxml_serializer_minimal_record(running_app, updated_minimal_record):
+    """Test serializer for MARCXML"""
+    serialized_record = {}
+    expected_data = {}
+
+    serializer = MARCXMLSerializer()
+    serialized_record = create_record(
+        serializer.serialize_object(updated_minimal_record)
+    )
+    expected_data = '<record xmlns="http://www.loc.gov/MARC21/slim"><datafield tag="245" ind1="a" ind2=" "><subfield>A Romans story</subfield></datafield><controlfield tag="540">info:eu-repo/semantics/openAccess</controlfield><datafield tag="260" ind1="c" ind2=" "><subfield>2020-06-01</subfield></datafield><datafield tag="100" ind1="a" ind2=" "><subfield>Name</subfield></datafield><datafield tag="100" ind1="a" ind2=" "><subfield>Troy Inc.</subfield></datafield><controlfield tag="901">info:eu-repo/semantic/other</controlfield><datafield tag="260" ind1="b" ind2=" "><subfield>Acme Inc</subfield></datafield></record>'
```
zzacharo, line 47: dump question: how did the `controlfield` end up to the serialized version? I don't see it in the schema you defined above....

tests/resources/serializers/test_marcxml_serializer.py
```diff
@@ -0,0 +1,93 @@
...
+    if isinstance(elements, Iterable) and not isinstance(elements, str):
+        return [a for i in elements for a in flatten(i)]
+    else:
+        return [elements]
+
+
+def test_marcxml_serializer_full_record(running_app, updated_full_record):
+    """Test serializer for MARCXML"""
+    serializer = MARCXMLSerializer()
+    serialized_record = create_record(serializer.serialize_object(updated_full_record))
```
zzacharo, line 82: Do we need the `create_record` call here? Is that function being called also when we serialize it via the REST API? Could we test exactly the output that the serializer is giving back from the REST API point of view? i.e when we call `/records/<pid_value>/export/marcxml`?

zzacharo (commented):
Great job! Bear with me and my dump questions :)

invenio_rdm_records/resources/serializers/marcxml/schema.py
TLGINO: Ok so most of the schema is taken from
https://github.com/TLGINO/invenio-rdm-records/blob/d1052925f1106be65c80d4c94d91d3b31572837d/invenio_rdm_records/resources/serializers/dublincore/schema.py#L19
I have adapted/omited some to better correspond to the marcxml standard

tests/resources/serializers/test_marcxml_serializer.py
utnapischtim, line 47: The subfield tag is not correct. The subfields are missing the attributes (subfield codes). As an example from loc sandburg.

invenio_rdm_records/resources/serializers/marcxml/schema.py
TLGINO: I assumed this was not possible as it isn't implemented in
https://github.com/TLGINO/invenio-rdm-records/blob/d1052925f1106be65c80d4c94d91d3b31572837d/invenio_rdm_records/resources/serializers/dublincore/schema.py#L19
If it is, then I can change it in both implementations

tests/resources/serializers/test_marcxml_serializer.py
TLGINO: all the xml tags get automatically assigned using the ```dojson``` library
```return etree.tostring(dumps_etree(json))```

https://github.com/TLGINO/invenio-rdm-records/blob/d1052925f1106be65c80d4c94d91d3b31572837d/invenio_rdm_records/resources/serializers/marcxml/__init__.py#L68

File: tests/resources/serializers/test_marcxml_serializer.py (hunk 4 above)
Comment (TLGINO) [reply tests/resources/serializers/test_marcxml_serializer.py:82]: _Do we need the create_record call here?_
I will try testing without and let you know!

 _Is that function being called also when we serialize it via the REST API?_
 I don't think so, so it's quite probable I can remove it in the test
 
 _Could we test exactly the output that the serializer is giving back from the REST API point of view? i.e when we call /records/<pid_value>/export/marcxml?_
 Do you mean making a GET request to the running server instance?
 In which case I think it should be possible, although the current test already tests effectively the same (by calling the function that serializes the record and that is then returned by the request)

File: tests/resources/serializers/test_marcxml_serializer.py (hunk 3 above)
Comment (TLGINO) [reply tests/resources/serializers/test_marcxml_serializer.py:47]: Hmm you're right, I'll modify this thx

Comment (TLGINO) [reply tests/resources/serializers/test_marcxml_serializer.py:47]: Ok so I need to use the ```create_record``` function as 2 exports to MARCXML doesn't garantee the exact same output (the order of the tags may be different).
So I need to parse it as a ```GroupableOrderedDict``` object in order to parse it.

File: invenio_rdm_records/resources/serializers/marcxml/schema.py [hunk 5]
```diff
@@ -0,0 +1,234 @@
...
+    identifiers = fields.Method("get_identifiers")
+    relations = fields.Method("get_relations")
+    rights = fields.Method("get_rights")
+    dates = fields.Method("get_dates")
+    subjects = fields.Method("get_subjects")
+    descriptions = fields.Method("get_descriptions")
+    publishers = fields.Method("get_publishers")
+    types = fields.Method("get_types")
+    sources = fields.Method("get_sources")
+    coverage = fields.Method("get_locations")
```
zzacharo, line 36: I think that's not listed https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/schemas/metadata.py#L330 . Can you also that we don't miss any other fields?

TLGINO: So I've done a sort of splitting of some fields for clarity / reusing existing schema names
"publisher" is present on the schema (hence publishers is just a list of them)
"types" is present as "resource_type" on the metadata schema
"sources" is present as "references"  on the metadata schema
"coverage" is present as "locations"  on the metadata schema

should these variables be renamed to their corresponding metadata schema counter-part or should i leave them as is (IE like in the other schema.py files)?

Also i have added the missing fields from the metadata schema

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -0,0 +1,287 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""MARCXML based Schema for Invenio RDM Records."""
+
+from __future__ import absolute_import, print_function, unicode_literals
```
ppanero, line 10: Is this needed? Aren't this available in python 3.7+?

invenio_rdm_records/resources/serializers/marcxml/schema.py
ppanero: I think is good, if you want for clarity you can put it in the docstring of the function. For example, "locations corresponds to coverage in the standard".

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -0,0 +1,287 @@
...
+        """Get funding information."""
+        if "funding" not in obj["metadata"]:
+            return missing
+
+        funders = []
+
+        for funder in obj["metadata"].get("funding", []):
+            funder_string = ""
+
+            scheme = (
```
ppanero, line 65: minor extract it outside so you have to calculate it only once, even if access to dict or position in lists is O(1), for readability, would help in many lines below too.

```python
award = funder.get("award", {}).get("identifiers", [{}])[0]

scheme = award.get("scheme", "null")
identifier = award.get("identifier", "null")
```

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -0,0 +1,287 @@
...
+        return funders
+
+    def last_updated(self, obj):
+        """Get date last updated."""
+        return [obj["updated"]]
+
+    def get_locations(self, obj):
+        """Get locations."""
+        locations = []
+        if not ("metadata" in obj and "locations" in obj["metadata"]):
```
ppanero, line 100: why do we check for `metadata in obj` here, but in all the other fields we do `obj["metadata"]`? My question: this gives the impression that metadata could not be there, however if that was the case it would throw a `KeyError` in any of the other functions. Therefore, is the first condition of the `if` needed?

ppanero (commented):
Nice job! 🚀 

Note that I'm not proficient in MARC XML, so if I said something that does not make sense please ignore (e.g. I am assuming that `null` is part of the standard, in other serializers instead of returning null, we ignore the field). Therefore, I tried to keep focus on the actual python code.

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -0,0 +1,287 @@
...
+    coverage = fields.Method("get_locations")
+    formats = fields.Method("get_formats")
+    parent_id = fields.Method("get_id")  # No GOOD
+    community_id = fields.Method("get_id_community")
+    last_updated = fields.Method("last_updated")
+    sizes = fields.Method("get_sizes")
+    version = fields.Method("get_version")
+    funding = fields.Method("get_funding")
+
+    def get_version(self, obj):
```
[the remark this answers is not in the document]
TLGINO: I think for consistency reasons it's probably best to keep obj["metadata"] as this is used throughout this schema, but also in all the other serializer schemas, let me know your opinion on this :)

invenio_rdm_records/resources/serializers/marcxml/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+from copy import deepcopy
+
+from dojson.contrib.to_marc21.utils import dumps_etree
+from flask_resources.serializers import SerializerMixin
+from lxml import etree
+
+from .schema import MARCXMLSchema
+
+
+class MARCXMLSerializer(SerializerMixin):
```
ppanero, line 21: Would it make sense to inherit from `MarshmallowSerializer`? or does it complicate a lot the rest of the logic?
https://github.com/inveniosoftware/flask-resources/blob/master/flask_resources/serializers/base.py#L31

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -0,0 +1,287 @@
...
+        for a in obj["metadata"].get("related_identifiers", []):
+            rels.append(f"{a['scheme']}:{a['identifier']}")
+
+        return rels or missing
+
+    def get_rights(self, obj):
+        """Get rights."""
+        rights = []
+
+        if "access" in obj and "status" in obj["access"]:
```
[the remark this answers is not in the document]
TLGINO: Much cleaner you're right thx!

invenio_rdm_records/resources/serializers/marcxml/__init__.py
TLGINO: The choice to inherit from `SerializerMixin` was made because I noticed a pattern on the other serializers
if return type is `JSON` --> inherit from `MarshmallowJSONSerializer` (which in turn inherits `MarshmallowSerializer`)
if return type is `XML` --> inherit from `SerializerMixin`
Nothing seems to directly inherit `MarshmallowSerializer`
Note : `MarshmallowSerializer` inherits `SerializerMixin`

invenio_rdm_records/resources/serializers/marcxml/schema.py
TLGINO: you're correct this is redundant, will fix

invenio_rdm_records/services/services.py
```diff
@@ -129,6 +129,7 @@ def search_community_records(
     ):
         """Search for records published in the given community."""
         self.require_permission(identity, "read")
+
```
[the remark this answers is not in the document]
TLGINO: Just checked and this is still in master so not sure :/

invenio_rdm_records/resources/serializers/marcxml/__init__.py
ppanero: I see your point, for example with the Dublin Core serializer. However, I think we made a mistake there. Analyzing a bit the inheritance in the serializers:

- We have to base classes: a raw serializer `SerializerMixin` which is nothing more than an interface (could just be an abstract class instead of a Mixing since it does not bring functionality, but that is another story). Then we have [MarshmallowSerializer](https://github.com/inveniosoftware/flask-resources/blob/master/flask_resources/serializers/base.py#L31), which is meant for those serializers where we use a Marshmallow schema (which is the case here and also in DublinCore).
- Then we have specific ones for JSON, i.e. `MarshmallowJSONSerializer`, which basically takes the `MarshmallowSerializer` but [specifies an output format](https://github.com/inveniosoftware/flask-resources/blob/master/flask_resources/serializers/json.py#L85): JSON.

In my head the ideal case scenario would be:
1. We create a `MarshmallowXMLSerializer` that takes the schema and then outputs XML.
2. We migrate DublinCore to use this

However, your code is already functional and I do not want to block you, meaning we could potentially merge and do more changes in another PR. Point number 2 is definetly outside the scope of this PR, I'm wondering about point 1... WDYT? cc: @ntarocco

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -0,0 +1,286 @@
...
+    def last_updated(self, obj):
+        """Get date last updated."""
+        return [obj["updated"]]
+
+    def get_locations(self, obj):
+        """Get locations."""
+        locations = []
+
+        access_location = obj["metadata"].get("locations", {})
+        if access_location == {}:
```
ppanero, line 98:
```suggestion
        if not access_location:
```

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -0,0 +1,286 @@
...
+            rels.append(f"{a['scheme']}:{a['identifier']}")
+
+        return rels or missing
+
+    def get_rights(self, obj):
+        """Get rights."""
+        rights = []
+        access_right = obj.get("access", {}).get("status", {})
+
+        if access_right != {}:
```
ppanero, line 171:
```suggestion
        if access_right:
```

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -0,0 +1,286 @@
...
+        # Related identifiers
+        for a in obj["metadata"].get("related_identifiers", []):
+            rels.append(f"{a['scheme']}:{a['identifier']}")
+
+        return rels or missing
+
+    def get_rights(self, obj):
+        """Get rights."""
+        rights = []
+        access_right = obj.get("access", {}).get("status", {})
```
ppanero, line 169: isn't `status` a string, not a dict? So we would just expect a string value or None?
```suggestion
        access_right = obj.get("access", {}).get("status")
```

ppanero (commented):
Nice job, now is quite more simple. I've added a few more minor comments mostly on data types to use and simplify `if` statements. I will merge tomorrow morning once these have been fixed.

invenio_rdm_records/resources/serializers/marcxml/__init__.py
ppanero: Opened an issue about this https://github.com/inveniosoftware/invenio-rdm-records/issues/1184

PR #1169, opened 2023-02-03
implement direct publish

zzacharo:
- closes https://github.com/inveniosoftware/invenio-communities/issues/871


invenio_rdm_records/services/review/service.py
```diff
@@ -180,3 +173,42 @@ def submit(self, identity, id_, data=None, revision_id=None, uow=None):
         uow.register(RecordIndexOp(draft, indexer=self.indexer))
 
         return request_item
+
+    @unit_of_work()
+    def submit(self, identity, id_, data=None, revision_id=None, uow=None):
```
ntarocco, line 178: Why splitting these 2 methods? In my mind there should be only one, no option: if I can direct publish, it should just happen, and not by choosing another endpoint?

alejandromumo (approved):

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -203,3 +206,21 @@ class UIRecordSchema(BaseObjectSchema):
...
     )
+
+    @pre_dump
+    def add_communities_permissions_and_roles(self, obj, **kwargs):
+        """Inject current user's permission to community receiver."""
+        receiver = (
+            obj.get("expanded", {}).get("parent", {}).get("review", {}).get("receiver")
+        )
+        if receiver:
+            can_direct_publish = _community_permission_check(
```
alejandromumo, line 217: question: I am a bit out of the loop on this subject and this catched my attention. Is it common to check permissions on serialization?

invenio_rdm_records/resources/resources.py
```diff
@@ -116,13 +116,16 @@ def review_delete(self):
     @request_headers
     @request_view_args
     @request_data
-    @response_handler()  # TODO: probably needs to change
+    @response_handler()
     def review_submit(self):
-        """Submit a draft for review."""
+        """Submit a draft for review or directly publish it."""
+        # TODO: check if we can parse it in the body parser or in a better place?
+        direct_publish = resource_requestctx.data.pop("direct_publish", False)
```
kpsherva, line 123: why do we pop it?

zzacharo: because is not part of the review payload. I thought is more clear to pop it and indicate its purpose i.e specify if we should try to direct publish or create a review request.

invenio_rdm_records/services/config.py
```diff
@@ -85,6 +85,11 @@ def archive_download_enabled(record, ctx):
     return current_app.config["RDM_ARCHIVE_DOWNLOAD_ENABLED"]
 
 
+def is_draft_and_can_direct_publish(record, ctx):
+    """Determine if submit review link should be included."""
+    return is_draft(record, ctx) and record.parent.review is not None
```
ntarocco, line 90: Not checking if `can_direct_publish` here? Is the func name misleading?

invenio_rdm_records/services/config.py
```diff
@@ -248,6 +253,10 @@ class RDMRecordServiceConfig(RecordServiceConfig, ConfiguratorMixin):
             "{+api}/records/{id}/draft/actions/submit-review",
             when=is_draft_and_has_review,
         ),
+        "submit-and-publish-review": RecordLink(
```
ntarocco, line 256: remove?

invenio_rdm_records/services/review/service.py
```diff
@@ -143,18 +144,44 @@ def delete(self, identity, id_, revision_id=None, uow=None):
         return True
 
     @unit_of_work()
-    def submit(self, identity, id_, data=None, revision_id=None, uow=None):
-        """Submit record for review."""
-        # Get record and check permission
+    def submit(
+        self, identity, id_, data=None, direct_publish=None, revision_id=None, uow=None
```
ntarocco, line 148:
```suggestion
        self, identity, id_, data=None, direct_publish=False, revision_id=None, uow=None
```

invenio_rdm_records/resources/resources.py
```diff
@@ -116,13 +116,17 @@ def review_delete(self):
...
     @request_view_args
     @request_data
-    @response_handler()  # TODO: probably needs to change
+    @response_handler()
     def review_submit(self):
-        """Submit a draft for review."""
+        """Submit a draft for review or directly publish it."""
+        # we pop the value from the payload as it is used only as a service argument
+        # and it is not part of the request payload
+        direct_publish = resource_requestctx.data.pop("direct_publish", False)
```
ntarocco, line 124: Any validation here? E.g. optional field, boolean?

invenio_rdm_records/services/review/service.py
```diff
@@ -143,18 +144,44 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
         # since it is submit review action, assume the receiver is community
         resolved_community = draft.parent.review.receiver.resolve()
+        if direct_publish:
+            # Direct publish to community workflow
+            current_communities.service.require_permission(
+                identity, "direct_publish", record=resolved_community
+            )
+
+            # Submit draft for review
+            submitted_review = self._submit(
```
ntarocco, line 165: @slint after our chat of yesterday, for example, here, we might not want to trigger: an e-mail for the creation of the request, to owner and community curators, and then immediately after a second e-mail that it has been accepted.
It might make sense to move the `send_notif` at the service level, and not the at the action level.

invenio_rdm_records/services/review/service.py
```diff
@@ -143,18 +144,44 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
+                identity, draft, resolved_community, data, revision_id, uow=uow
+            )
+            # Accept draft in community
+            request_item = current_requests_service.execute_action(
+                identity, submitted_review.data["id"], "accept", data=data, uow=uow
+            )
+            return request_item
```
ntarocco, line 172: is the parent record already committed/indexed here?

invenio_rdm_records/services/review/service.py
```diff
@@ -143,18 +144,44 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
         draft = self.draft_cls.pid.resolve(id_, registered_only=False)
-        self.require_permission(identity, "update_draft", record=draft)
-
         # Preconditions
         if draft.parent.review is None:
             raise ReviewNotFoundError()
 
         # since it is submit review action, assume the receiver is community
         resolved_community = draft.parent.review.receiver.resolve()
+        if direct_publish:
```
ntarocco, line 158: Shouldn't we change this to:
```
if direct_publish or can_direct_publish:
```

Cases:
1. the user forces the direct_publish
2. the user does not force it, but (s)he can direct publish

invenio_rdm_records/resources/resources.py
```diff
@@ -116,13 +116,20 @@ def review_delete(self):
...
     @request_view_args
     @request_data
-    @response_handler()  # TODO: probably needs to change
+    @response_handler()
     def review_submit(self):
-        """Submit a draft for review."""
+        """Submit a draft for review or directly publish it."""
+        # we pop the value from the payload as it is used only as a service argument
+        # and it is not part of the request payload
+        require_review = False
```
ntarocco, line 124: we can replace these 3 lines with:
`require_review = resource_requestctx.data.pop("require_review", False)`

invenio_rdm_records/services/review/service.py
```diff
@@ -143,19 +145,51 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
-        self.require_permission(identity, "update_draft", record=draft)
-
         # Preconditions
         if draft.parent.review is None:
             raise ReviewNotFoundError()
 
         # since it is submit review action, assume the receiver is community
         resolved_community = draft.parent.review.receiver.resolve()
 
+        if require_review:
```
ntarocco, line 160: suggestion: to remove one nesting level and makes things more readable, we could:
```
try:
    current_communities.service.require_permission(
                    identity, "direct_publish", record=resolved_community
                )
except...
    require_review = True

if require_review:
    ...
else:
    ...

```

invenio_rdm_records/resources/serializers/ui/schema.py
ntarocco: we do it when we want to return the permissions that the user making the request has, it is a way to expose them, and allow building UI correctly

ntarocco (commented):
LGTM! There are still a couple of comments unanswered, we will take care of fixing and merging.

invenio_rdm_records/services/review/service.py
```diff
@@ -143,18 +144,44 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
-        # Get record and check permission
+    def submit(
+        self, identity, id_, data=None, direct_publish=None, revision_id=None, uow=None
+    ):
+        """Submit record for review or direct publish to the community."""
         draft = self.draft_cls.pid.resolve(id_, registered_only=False)
-        self.require_permission(identity, "update_draft", record=draft)
-
         # Preconditions
         if draft.parent.review is None:
```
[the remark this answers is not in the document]
ntarocco: Yes, the review is a 2-step process:
1. create a review
2. submit it
See https://inveniordm.docs.cern.ch/reference/rest_api_reviews/#createupdate-a-review-request

You cannot submit without having created it first

invenio_rdm_records/services/review/service.py
```diff
@@ -143,19 +145,53 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
-        # Get record and check permission
-        draft = self.draft_cls.pid.resolve(id_, registered_only=False)
-        self.require_permission(identity, "update_draft", record=draft)
+    def submit(
+        self, identity, id_, data=None, require_review=False, revision_id=None, uow=None
+    ):
+        """Submit record for review or direct publish to the community."""
+        if not isinstance(require_review, bool):
+            raise ValidationError(
+                _("The require_review param must be true or false."),
```
ppanero, line 154: `require_review` is already in the `field_name`, and _param_ seems very code oriented.
```suggestion
                _("Must be a boolean, True or False"),
```

tests/conftest.py
```diff
@@ -362,7 +363,7 @@ def search_clear(search):
     _search_create_indexes(current_search, current_search_client)
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture()
```
ppanero, line 366: minor: since they are dictionaries, can we make them scope `session` or even `module`? to avoid recreating them all the time (unless they are modified inside tests, but we use copy IIRC).

ppanero (approved):

tests/conftest.py
ppanero: https://github.com/inveniosoftware/invenio-rdm-records/issues/1196

invenio_rdm_records/services/review/service.py
ppanero: "No" but yes. The `_submit` method registers a `RecordCommitOp`. So it will be committed by the UoW and fail otherwise.

PR #1171, opened 2023-02-03
Update `get_content_files` permissions

max-moser:
The permissions for getting the file content were a bit too restrictive, requiring `can_preview` (i.e. "can preview draft") permissions rather than `can_read_files` (i.e. "can read files") permissions.
This PR fixes that by setting the conditional generator's `then_` value to the same as `can_read_files`.

That being said, I have the feeling that the given check for the file having a `L`ocal storage class and giving out access based on that could be reworked into business logic rather than permissions...
At least that's what it feels like to me – WDYT?

### DIFF
--- a/invenio_rdm_records/services/permissions.py
+++ b/invenio_rdm_records/services/permissions.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2019 CERN.
 # Copyright (C) 2019 Northwestern University.
+# Copyright (C) 2023 TU Wien.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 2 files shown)


max-moser:
Quick initial testing showed that the `view`-permission shared links work.

Further testing should be done to validate the permissions with shared links of different permissions and the "download all files" feature.
Also, as @ppanero suggested, we should probably add test cases for downloading public/public records' files as anonymous user.

invenio_rdm_records/services/permissions.py
```diff
@@ -74,8 +75,8 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
         IfRestricted("files", then_=can_view, else_=can_all),
     ]
     can_get_content_files = [
-        # preview is same as read_files
-        IfFileIsLocal(then_=can_preview, else_=[SystemProcess()])
+        # TODO: is this business logic, rather than permissions?
```
slint, line 78: It is closer to business logic, but is difficult to delegate to the service/storage... I think @ppanero might remember a bit more on what the discussion was to leave it here.

slint (approved):
LGTM, I guess one small clarification left to see for what we open a new issue.

invenio_rdm_records/services/permissions.py
ppanero: It was highly complex/coupling to do this. We had to do it in the service component and there were use cases that could not be easily abstracted and it was coupling transfer classes, data layer and services. And there was a corner case that was a deadlock (i.e. needed some info to continue but the info needed what was being calculated or it wasn't available at that point in time). However, I do not recall the specifics. #shouldvedocumented

Issue #1173, opened 2023-02-06
oai-pmh: add export format for MarcXML


zzacharo:

TLGINO:
This is a WIP commit for setup purposes

ppanero:
missing app-rdm pr

PR #1176, opened 2023-02-06
oai-pmh: export for marcxml

TLGINO:
* added export for oai-pmh to marcxml
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1173


invenio_rdm_records/services/services.py
```diff
@@ -129,6 +129,7 @@ def search_community_records(
     ):
         """Search for records published in the given community."""
         self.require_permission(identity, "read")
+
```
ppanero, line 132: minor: this is not in master <https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/services.py#L132> I think there is something weird going on the rebase.

ppanero (approved):
Nice job! Very minor comment, if you fix that and the tests (failing on black linting) I can merge.

invenio_rdm_records/services/services.py
TLGINO: I'm confused as I see that in master and also in the link you sent :/

PR #1187, opened 2023-02-14
contrib: added codemeta custom fields.

### LINKED ISSUE #1186: Implement Software Heritage (Code Meta) fields as custom fields
Implement the following fields from [Code Meta:](https://codemeta.github.io/terms/)

- [x] codeRepository
- [x] programmingLanguage
- [x] runtimePlatform
- [x] developmentStatus
- [x] operatingSystem

alejandromumo:
closes #1186

### DIFF
--- a/invenio_rdm_records/contrib/codemeta/__init__.py
+++ b/invenio_rdm_records/contrib/codemeta/__init__.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""CodeMeta contrib module."""
+from .custom_fields import (
+    CODEMETA_CUSTOM_FIELDS,
+    CODEMETA_CUSTOM_FIELDS_UI,
+    CODEMETA_FACETS,
+    CODEMETA_NAMESPACE,
+)
+
+__all__ = (
+    CODEMETA_FACETS,
+    CODEMETA_CUSTOM_FIELDS,
+    CODEMETA_CUSTOM_FIELDS_UI,
+    CODEMETA_NAMESPACE,
+)
--- a/invenio_rdm_records/contrib/codemeta/custom_fields.py
+++ b/invenio_rdm_records/contrib/codemeta/custom_fields.py
@@ -1,0 +1,127 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""CodeMeta specific custom fields.
+
+Implements the following fields:
+- Repository URL
+- Programming language
+- Runtime platform
+- Operating system
+- Development status
+"""
+
+from flask_babelex import _
+from invenio_records_resources.services.custom_fields import KeywordCF
+from invenio_records_resources.services.records.facets import CFTermsFacet
+from invenio_vocabularies.services.custom_fields import VocabularyCF
+from invenio_vocabularies.services.facets import VocabularyLabels
+from marshmallow import validate
+
+CODEMETA_NAMESPACE = {
+    # CodeMeta
+    "code": "https://codemeta.github.io",
+}
+
+
+CODEMETA_CUSTOM_FIELDS = [
+    KeywordCF(
+        name="code:codeRepository",
+        field_args={
+            "validate": validate.URL(),
+            "error_messages": {"validate": "You must provide a valid URL."},
+        },
+    ),
+    KeywordCF(name="code:programmingLanguage"),
+    KeywordCF(name="code:runtimePlatform"),
+    KeywordCF(name="code:operatingSystem"),
+    VocabularyCF(
+        name="code:developmentStatus",
+        vocabulary_id="code:developmentStatus",
+        dump_options=True,
+        multiple=False,
+    ),
+]
+
+CODEMETA_CUSTOM_FIELDS_UI = {
+    "section": _("CodeMeta"),
+    "fields": [
+        dict(
+            field="code:codeRepository",
+            ui_widget="Input",
+            props=dict(
+                label="Repository URL",
+                icon="linkify",
+                description="URL or link where the code repository is hosted.",
+            ),
+        ),
+        dict(
+            field="code:programmingLanguage",
+            ui_widget="Input",
+            props=dict(
+                label="Programming language",
+                icon="code",
+                description="Repository's programming language.",
+                placeholder="Python ...",
+            ),
+        ),
+        dict(
+            field="code:runtimePlatform",
+            ui_widget="Input",
+            props=dict(
+                label="Runtime platform",
+                icon="cog",
+                description="Repository runtime platform or script interpreter dependencies.",
+                placeholder="Java v1, Python2.3, .Net Framework 3.0 ...",
+            ),
+        ),
+        dict(
+            field="code:operatingSystem",
+            ui_widget="Input",
+            props=dict(
+                label="Supported operating system",
+                icon="desktop",
+                description="Supported operating systems.",
+                placeholder="Windows 7, OSX 10.6, Android 1.6 ...",
+            ),
+        ),
+        dict(
+            field="code:developmentStatus",
+            ui_widget="Dropdown",
+            props=dict(
+                label="Development Status",
+                placeholder="Repository status",
+                icon="heartbeat",
+                description="Repository current status.",
+                search=False,
+                multiple=False,
+                clearable=True,
+            ),
+        ),
+    ],
+}
+
+CODEMETA_FACETS = {
+    "developmentStatus": {
+        "facet": CFTermsFacet(
+            field="code:developmentStatus.id",
+            label=_("Development status"),
+            value_labels=VocabularyLabels("code:developmentStatus"),
+        ),
+        "ui": {  # ui display
+            "field": CFTermsFacet.field("code:developmentStatus.id"),
+        },
+    },
+    "programmingLanguage": {
+        "facet": CFTermsFacet(
+            field="code:programmingLanguage",
+            label=_("Programming language"),
+        ),
+        "ui": {  # ui display
+            "field": CFTermsFacet.field("code:programmingLanguage"),
+        },
+    },
+}
--- a/invenio_rdm_records/fixtures/data/vocabularies.yaml
+++ b/invenio_rdm_records/fixtures/data/vocabularies.yaml
@@ -49,3 +49,7 @@ funders:
 awards:
   pid-type: awa
   data-file: vocabularies/awards.yaml
+# code meta controlled vocabularies
+code:developmentStatus:
+  pid-type: ds
+  data-file: vocabularies/contrib/codemeta/development_status.yaml
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1431,3 +1431,13 @@ def curator(UserFixture, community, app, db):
         system_identity, community.id, invitation_data
     )
     return curator
+
+
+@pytest.fixture(scope="module")
+def cli_runner(base_app):
+    """Create a CLI runner for testing a CLI command."""
+
+    def cli_invoke(command, *args, input=None):
+        return base_app.test_cli_runner().invoke(command, args, input=input)
+
+    return cli_invoke
--- a/tests/contrib/codemeta/conftest.py
+++ b/tests/contrib/codemeta/conftest.py
@@ -1,0 +1,159 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""CodeMeta specific test configs."""
+import pytest
+from invenio_access.permissions import system_identity
+from invenio_vocabularies.proxies import current_service as vocabulary_service
+from invenio_vocabularies.records.api import Vocabulary
+
+from invenio_rdm_records.cli import (
+    create_records_custom_field,
+    custom_field_exists_in_records,
+)
+from invenio_rdm_records.contrib.codemeta.custom_fields import (
+    CODEMETA_CUSTOM_FIELDS,
+    CODEMETA_FACETS,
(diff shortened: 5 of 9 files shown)


invenio_rdm_records/contrib/codemeta/custom_fields.py
```diff
@@ -0,0 +1,118 @@
+# -*- coding: utf-8 -*-
```
alejandromumo, line 1: Please challenge the placement of these files.

We considered the options:

- `invenio_rdm_records/services/custom_fields/codemeta.py`
- `invenio_rdm_records/services/custom_fields.py`
- `invenio_rdm_records/contrib/codemeta/custom_fields.py` <- implemented option

tests/fixtures/conftest.py
```diff
@@ -47,16 +47,6 @@ def create_app(instance_path, entry_points):
     return _create_app
 
 
-@pytest.fixture(scope="module")
```
alejandromumo, line 50: Moved one level up in the hierarchy to `/tests/conftest.py`

tests/contrib/codemeta/test_codemeta_custom_fields.py
```diff
@@ -0,0 +1,141 @@
...
+def codemeta_fields_record(minimal_codemeta_record):
+    """Creates a record containing CodeMeta fields."""
+    draft = create_draft(minimal_codemeta_record)
+    record = create_record(draft.id)
+    return record
+
+
+def test_custom_fields_loading(running_app, cli_runner, codemeta_cf_name):
+    """Tests whether custom fields are correctly loaded on demand."""
+    # If custom fields are not initialised, "custom_field_exists_in_records" fails with KeyError['properties']
```
alejandromumo, line 47: Is this a bug? It breaks here https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/custom_fields/mappings.py#L49

The function `field_exists` assumes that `mapping['properties']` exists.

invenio_rdm_records/contrib/codemeta/custom_fields.py
```diff
@@ -0,0 +1,127 @@
...
+CODE_CUSTOM_FIELDS_UI = {
+    "section": _("CodeMeta"),
+    "fields": [
+        dict(
+            field="code:codeRepository",
+            ui_widget="Input",
+            props=dict(
+                label="Repository URL",
+                icon="linkify",
+                description="You should fill this field with the repository URL.",
```
ppanero, line 58: Fields in the RDM deposit form have a explanatory nature, with no verb or "order" to the user. For consistency I would change descriptions to something like:
```suggestion
                description="URL or link where the code repository is hosted.",
```

Now the challenge comes from writing a sentence to define a field without repeating the name of the field 😂 Feel free to change the provided suggestion.

invenio_rdm_records/fixtures/data/vocabularies.yaml
```diff
@@ -49,3 +49,7 @@ funders:
 awards:
   pid-type: awa
   data-file: vocabularies/awards.yaml
+# code meta controlled vocabularies
+developmentstatus:
```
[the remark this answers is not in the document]
ppanero: If we go for contrib I would say yes, but (IIRC) the `FixturesEngine` does not support looking into folders that are not `/app_data` or `/data` (root under fixtures)

tests/contrib/codemeta/conftest.py
```diff
@@ -0,0 +1,103 @@
...
+    loaded_search = running_app.app.config["RDM_SEARCH"]
+    running_app.app.config["RDM_SEARCH"] = {
+        **loaded_search,
+        "facets": loaded_search["facets"] + [codemeta_cf_facetted_term],
+    }
+    result = cli_runner(create_records_custom_field)
+    assert result.exit_code == 0
+
+
+@pytest.fixture(scope="function")
```
ppanero, line 79: this fixtures will be recreated per test function there are mere constants, using `session` would improve performance. You can also just declare a global variable.

tests/contrib/codemeta/conftest.py
```diff
@@ -0,0 +1,103 @@
...
+def init_codemeta_custom_fields(
+    running_app,
+    cli_runner,
+    codemeta_development_status_vocabulary,
+    codemeta_cf_facetted_term,
+):
+    """Initializes codemeta custom fields."""
+    # Add code meta custom fields
+    running_app.app.config["RDM_CUSTOM_FIELDS"] = CODE_CUSTOM_FIELDS
```
ppanero, line 66: Can we overwrite config using the `app_config` fixture?

tests/contrib/codemeta/conftest.py
```diff
@@ -0,0 +1,103 @@
...
+    running_app.app.config["RDM_NAMESPACES"] = CODE_NAMESPACE
+    loaded_facets = running_app.app.config["RDM_FACETS"]
+    running_app.app.config["RDM_FACETS"] = {**loaded_facets, **CODE_FACETS}
+    loaded_search = running_app.app.config["RDM_SEARCH"]
+    running_app.app.config["RDM_SEARCH"] = {
+        **loaded_search,
+        "facets": loaded_search["facets"] + [codemeta_cf_facetted_term],
+    }
+    result = cli_runner(create_records_custom_field)
```
ppanero, line 75: I don't like using a CLI inside a test (it will spawn a full new app context to run the command?) but I do not see any major drawback either nor how to solve it without complex logic (register mappings and custom record class). Do we have other occurrences of this in the codebase? if not, how did we solve similar cases? Could this affect other tests?

tests/contrib/codemeta/test_codemeta_custom_fields.py
ppanero: It assumes `properties` exists because is how the lower `Mapping` class is implemented, see <https://github.com/opensearch-project/opensearch-py/blob/main/opensearchpy/helpers/mapping.py#L106>

If `properties` is not there, something is wrong on our side.

tests/contrib/codemeta/test_codemeta_custom_fields.py
```diff
@@ -0,0 +1,141 @@
...
+
+@pytest.fixture(scope="function")
+def codemeta_fields_record(minimal_codemeta_record):
+    """Creates a record containing CodeMeta fields."""
+    draft = create_draft(minimal_codemeta_record)
+    record = create_record(draft.id)
+    return record
+
+
+def test_custom_fields_loading(running_app, cli_runner, codemeta_cf_name):
```
ppanero, line 45: Should this be responsability of this module? The logic behind CF loading is in the cli (not contrib/codemeta), I think this module should assume it works (this is what triggered me to think if we could avoid using the CLI command in tests)

ppanero (commented):
Content LGTM! nice job! Questions are more on code placement and responsabilities.

tests/contrib/codemeta/conftest.py
alejandromumo: I went for this approach as it was the only way I've found to add custom fields programatically in tests, without replicating the code from the cli command. 
If, however, that does not sound OK then I can create function to initialise custom fields (it would likely be a copy of the CLI command).

tests/contrib/codemeta/test_codemeta_custom_fields.py
alejandromumo: My idea here was to test whether loading these specific custom fields do not yield any error. 
I do agree that it shouldn't be this module's responsability to test whether loading custom fields works or not, however I think that it's useful to have a test that validates the loading of CodeMeta CF.
In your opinion, would it be enough to simply create a draft and see if it fails? If so, I can remove this test.

tests/contrib/codemeta/conftest.py
alejandromumo: You mean using the fixture `app_config` or moving these config's loading to the fixture itself? If the latter, I'd prefer to load them on demand since that's how the end user will load these custom fields. They must be explicitely loaded when they are to be used.

ppanero: I meant, on demand. You can overwrite the `app_config` and then all gets injected automatically, you can also control the scope, etc. e.g. https://github.com/inveniosoftware/invenio-rdm-records/blob/d2cede0b1a4544e18175904b9b7dbe5de209d17c/tests/conftest.py#L214

tests/contrib/codemeta/conftest.py
ppanero: I don't think is optimal, but changing it would require investigation and work that we cannot afford now. So is okay!

tests/contrib/codemeta/test_codemeta_custom_fields.py
ppanero: I'm convinced thanks! could you add this explanation in the test docstring or a comment? I see myself coming back to this again 😂

alejandromumo:
Tests are failing but feel free to discuss, namely [this comment](https://github.com/inveniosoftware/invenio-rdm-records/pull/1187#discussion_r1105991504)

invenio_rdm_records/fixtures/data/vocabularies.yaml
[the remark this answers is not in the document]
slint: Also not a big fan of this as a top-level term. I don't know how much flexibility we have with generic vocabulary names, but could we do something like:

- `code:developmentStatus`, i.e. mapping directly to the custom field name(space)
- `codemeta/developmentStatus`

Path-wise it could be `vocabularies/contrib/development_status.yaml` or similar (and would still work based on the 2 prefixes we have).

tests/conftest.py
```diff
@@ -1329,3 +1329,13 @@ def admin(UserFixture, app, db, admin_role_need):
     datastore.add_role_to_user(u.user, role)
     db.session.commit()
     return u
+
+
+@pytest.fixture(scope="module")
+def cli_runner(base_app):
```
slint, line 1335: comment/minor: isn't this fixture already available [from `pytest-invenio`](https://github.com/inveniosoftware/pytest-invenio/blob/779fab4eb1ed899ff3445c075705776782920b64/pytest_invenio/fixtures.py#L363)?

slint (approved):
LGTM! Some naming nits, and a possible fixture removal 🙌

invenio_rdm_records/contrib/codemeta/custom_fields.py
kpsherva: It was suggested to go with the contrib, since extracting the fields might be easier with more structural separation

tests/conftest.py
alejandromumo: I think it is, I moved it upper in the hierarchy but I think it can in fact be removed.

alejandromumo: Actually it's not 😃 

Their signature is slightly different:

- [pytest](https://github.com/inveniosoftware/pytest-invenio/blob/779fab4eb1ed899ff3445c075705776782920b64/pytest_invenio/fixtures.py#L375):
  ```python
  def cli_invoke(command, input=None, *args):
       return base_app.test_cli_runner().invoke(command, args, input=input)
   # Usage
   cli_invoke(command, 'test', '-f')
  ```
  This version does not work because `args` is a tuple ('-f') and `input` is not picked by flask cli runner. In reality this works for commands without any args, as in [app-rdm](https://github.com/inveniosoftware/invenio-app-rdm/blob/f228f9d593118261fde0f9b80dad0d52088e8ad7/tests/fixtures/test_cli.py#L18)

- invenio_rdm_records and [invenio-communities](https://github.com/inveniosoftware/invenio-communities/blob/9de35e431b14f9a558e44721e8db0c2faea1f5ef/tests/conftest.py#L421)
  ```python
  def cli_invoke(command, *args, input=None):
        return base_app.test_cli_runner().invoke(command, args, input=input)
  
  # Usage
  cli_invoke(command, '-f', input='test')
  ```
  This version works because `args` is a tuple : `('-f', 'test')` and [flask test cli](https://click.palletsprojects.com/en/8.1.x/api/#click.testing.CliRunner.invoke) likes it. None that `input` is not used either by the cli runner 😄 

That sneaky difference makes all the difference (pun intended).

TL;DR
I'd argue that this needs to be analysed in another issue so we can do some code cleanup. It shouldn't be too hard but I'd prefer to tackle it afterwards so this task isn't blocked further.

slint: Nice one spotting this!

Agree, we can shelve this and address separately, since if we fix this it needs coordination across many modules.

tests/contrib/codemeta/conftest.py
alejandromumo: I ended up doing something different, sorry. I was having quite some trouble to make it work with `app_config`, as it gets picked up by `base_app` when the app is initialised.
What I am doing now is simply injecting the configs in the app, initialising the codemeta custom fields and rolling the config back afterwards.

I like your idea better, however I ran into multiple issues (some not even related) and this is how I made it work.

tests/contrib/codemeta/test_codemeta_custom_fields.py
alejandromumo: I added a small comment in the test. I also ended up splitting this test into smaller ones.

invenio_rdm_records/fixtures/data/vocabularies.yaml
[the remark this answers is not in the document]
alejandromumo: Sounds much better, thanks for the suggestion!

tests/conftest.py
alejandromumo: Created a separate [issue](https://github.com/inveniosoftware/pytest-invenio/issues/94) to adress it.

kpsherva (commented):
LGTM! but as I am not expert on custom fields maybe someone else should take a look

invenio_rdm_records/contrib/codemeta/__init__.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""Custom fields implementation for CodeMeta."""
```
slint, line 7:
```suggestion
"""CodeMeta contrib module."""
```

nit: since there might be more things added in the future.

slint (commented):
Some minor comments on mappings/types, and naming nits (I should start paying more attention in the first reviews 😏 )

zzacharo (approved):
Needs rebasing and addressing some minor comments :)

zzacharo (approved):

PR #1200, opened 2023-02-21
contrib: added journal custom field.

### LINKED ISSUE #1191: Implement journal field
Implement the field `journal` considering its nested attributes:

- `journal.pages`
- `journal.title`
- `journal.volume`
- `journal.year`

Take into account the following requirements when implementing this field:

- It is needed in zenodo-rdm. 
- It must be a custom field. 
- It must be searchable. Search must be compatible with the current [Zenodo search guide](https://help.zenodo.org/guides/search/)
- If needed, update DataCite serializer taking into account [Datacite Schema 4.4 docs](https://support.datacite.org/docs/datacite-metadata-schema-44). 
  - A recommended field [`RelatedItem` ](https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#20-relateditem) can be used to serialize journal's information. 
- The upload form should be similar to zenodo (e.g. `Journal` has its own section)

alejandromumo:
closes #1191 

Javascript widget was implemented in https://github.com/inveniosoftware/react-invenio-forms/pull/192


ppanero (approved):

kpsherva (approved):

lnielsen:
Why is ``year`` part of the journal field? It's not on the current Zenodo form, and normally it's calculated in the citation from the publication date.

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -0,0 +1,114 @@
...
+    "fields": [
+        {
+            "field": "journal:journal",
+            "ui_widget": "Journal",
+            "template": "journal.html",
+            "props": {
+                "label": _("Journal"),
+                "title": {
+                    "label": _("Title"),
+                    "placeholder": _("Add the journal title..."),
```
lnielsen, line 77: In most other deposit form fields we don't add placeholders, and the text doesn't say anything beyond what the label says, so I wouldn't add placeholders here.

alejandromumo:
> Why is `year` part of the journal field? It's not on the current Zenodo form, and normally it's calculated in the citation from the publication date.

This was migrated based on the [search guide](https://help.zenodo.org/guides/search/) and the [zenodo jsonschema](https://github.com/zenodo/zenodo/blob/a05df8aa6b0ae1a1b88bd8d8bc5cc75250ded14f/zenodo/modules/records/jsonschemas/records/record-v1.0.0.json#L744) of `journal`

jrcastro2 (commented):
As I am joining the sprint again, I have a stupid question:
This looks like specific to Zenodo, why do we implement this in `invenio-rdm-records`?

kpsherva:
> As I am joining the sprint again, I have a stupid question: This looks like specific to Zenodo, why do we implement this in `invenio-rdm-records`?

this is not only zenodo specific, we want to have it in rdm as a feature. CDS for example will need the journal and other fields. They are implemented as custom fields since we want to have an option of not including them in the instance if they are not needed

PR #1205, opened 2023-02-22
resources: test record access and search for different content types

### LINKED ISSUE #1204: test search endpoint for different export formats
- [ ] https://github.com/inveniosoftware/invenio-app-rdm/issues/2031#issuecomment-1438654086
- check https://github.com/front-matter/commonmeta-py

ppanero:
- requires https://github.com/inveniosoftware/flask-resources/pull/107
- closes #1204

### DIFF
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -67,9 +67,11 @@ record_serializers = {
     "application/marcxml+xml": ResponseHandler(MARCXMLSerializer()),
     "application/vnd.inveniordm.v1+json": ResponseHandler(UIJSONSerializer()),
     "application/vnd.citationstyles.csl+json": ResponseHandler(CSLJSONSerializer()),
-    "application/vnd.datacite.v43+json": ResponseHandler(DataCite43JSONSerializer()),
+    "application/vnd.datacite.datacite+json": ResponseHandler(
+        DataCite43JSONSerializer()
+    ),
     "application/vnd.geo+json": ResponseHandler(GeoJSONSerializer()),
-    "application/vnd.datacite.v43+xml": ResponseHandler(DataCite43XMLSerializer()),
+    "application/vnd.datacite.datacite+xml": ResponseHandler(DataCite43XMLSerializer()),
     "application/x-dc+xml": ResponseHandler(DublinCoreXMLSerializer()),
     "text/x-bibliography": ResponseHandler(
         StringCitationSerializer(url_args_retriever=csl_url_args_retriever),
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -46,6 +46,7 @@ from .serializers import (
     DataCite43XMLSerializer,
     DublinCoreXMLSerializer,
     GeoJSONSerializer,
+    MARCXMLSerializer,
     StringCitationSerializer,
     UIJSONSerializer,
 )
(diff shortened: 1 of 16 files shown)


invenio_rdm_records/resources/config.py
```diff
@@ -63,13 +64,12 @@ def csl_url_args_retriever():
...
+    "application/marcxml+xml": ResponseHandler(MARCXMLSerializer()),
     "application/vnd.inveniordm.v1+json": ResponseHandler(UIJSONSerializer()),
     "application/vnd.citationstyles.csl+json": ResponseHandler(CSLJSONSerializer()),
-    "application/vnd.datacite.datacite+json": ResponseHandler(
-        DataCite43JSONSerializer()
-    ),
+    "application/vnd.datacite.v43+json": ResponseHandler(DataCite43JSONSerializer()),
     "application/vnd.geo+json": ResponseHandler(GeoJSONSerializer()),
-    "application/vnd.datacite.datacite+xml": ResponseHandler(DataCite43XMLSerializer()),
+    "application/vnd.datacite.v43+xml": ResponseHandler(DataCite43XMLSerializer()),
```
ppanero, line 72: @slint should this be:

a) vnd.datacite.datacite+xml (as it was before, no versioning)
b) vnd.datacite.v43+xml
c) vnd.datacite.datacite.v43+xml

I'm leaning towards b (the one implemented), but I'm not sure and I cannot find concrete info. Also, why is this a vendor (vnd) and others like DublinCore are not?

tests/resources/test_resources.py
```diff
@@ -727,6 +727,39 @@ def test_link_update(
     assert link_result.json["permission"] == "preview"
 
 
+def test_response_handlers(running_app, minimal_record, client_with_login):
+    # Check that the response handlers are correctly configured.
+    # The correctness of its serialization is tested in /tests/serializers
+
+    accept = [
+        "application/json",
+        # "application/marcxml+xml",  # Fails on search (not implemented)
```
ppanero, line 736: major: the purpose of this test is to showcase that we are missing serialization, and they are not being thoroughly tested in the serializer tests.

invenio_rdm_records/resources/config.py
TLGINO: Disclaimer, I'm not @slint ;)
Going through Zenodo code it seems that that what is used for v31 is
`application/x-datacite+xml`
See [here](https://github.com/zenodo/zenodo/blob/b27daab5825f310739bc46ca49c9454f65b72ed0/zenodo/modules/records/serializers/__init__.py#L122)

ppanero: I'm not @slint  either (keeping headers for consistency 😂)

Thanks @tmorrell, I think in that case it makes sense to keep `application/vnd.datacite.datacite+xml`. I'm wondering what would happen then if we support both v4.3 and v4.4 or so? 🤔

slint: As the real @slint, even though much has already been said, I guess it's my time to comment here :)

Agree that the double `datacite` sounds weird, but maybe there's some reasoning behind it (single doesn't seem work, so it's not a typo in the docs). 

For Zenodo we do have the `application/x-datacite+xml`, which could be from older documentation, but I couldn't honestly track where it came from...

I think at this point specifying the version doesn't make much difference also, since it's just additional complexity, and also the versions are backward compatible (if someone can submit/parse v4.3, the same payload is valid in v4.4). 

So my vote is also for `application/vnd.datacite.datacite+xml`.

ppanero: _Would the real @slint shady please stand up, please stand up_

Thanks for the discussion, I'll revert back to `application/vnd.datacite.datacite+xml` and `application/vnd.datacite.datacite+json`

ppanero:
Tests will not pass until flask resources is released

kpsherva (approved):

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -286,3 +288,58 @@ def get_formats(self, obj):
         """Get data formats."""
         formats = obj["metadata"].get("formats", missing)
         return formats
+
+    @post_dump()
+    def changeKeysToTags(self, data, many, **kwargs):
```
slint, line 293: nit: I know it's out of scope for the change, but is there a reason this is camelCased?

tests/resources/test_resources.py
```diff
@@ -727,6 +727,39 @@ def test_link_update(
     assert link_result.json["permission"] == "preview"
 
 
+def test_response_handlers(running_app, minimal_record, client_with_login):
```
slint, line 730: nit: missing docstring here (thought `pytest` would complain and fail).

slint (approved):
LGTM, some minor nits.

PR #1206, opened 2023-02-22
serializers: fixing / updates to dublincore

TLGINO:
* added location to export + tests
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1201

### DIFF
--- a/invenio_rdm_records/resources/serializers/dublincore/schema.py
+++ b/invenio_rdm_records/resources/serializers/dublincore/schema.py
@@ -12,11 +12,12 @@ from invenio_access.permissions import system_identity
 from invenio_vocabularies.proxies import current_service as vocabulary_service
 from marshmallow import Schema, fields, missing
 
+from ..schemas import BaseSchema
 from ..ui.schema import current_default_locale
 from ..utils import get_vocabulary_props
 
 
-class DublinCoreSchema(Schema):
+class DublinCoreSchema(BaseSchema):
     """Schema for Dublin Core in JSON."""
 
     contributors = fields.Method("get_contributors")
@@ -32,7 +33,7 @@ class DublinCoreSchema(Schema):
     types = fields.Method("get_types")
     sources = fields.Method("get_sources")
     languages = fields.Method("get_languages")
-    coverage = fields.Method("get_locations")
+    locations = fields.Method("get_locations")
     formats = fields.Method("get_formats")
 
     def get_titles(self, obj):
@@ -189,26 +190,6 @@ class DublinCoreSchema(Schema):
         if languages:
             return [language["id"] for language in languages]
 
-        return missing
-
-    def get_locations(self, obj):
-        """Get locations."""
-        # FIXME: Add after UI support is there
-        # locations = []
-        # for location in obj['metadata'].get('locations', []):
-        #     location_string = ""
-        #     place = location.get('place')
-        #     if place:
-        #         location_string += f"name={place}"
-
-        #     geometry = location.get('geometry')
-        #     if geometry:
-        #         geo_type = geometry['type']
-        #         if geo_type == "Point":
-        #             coords = geometry['coordinates']
-        #             location_string += f"; lat={coords[0]}; lon={coords[1]}"
-
-        #     locations.append(location_string)
         return missing
 
     def get_formats(self, obj):
--- a/invenio_rdm_records/resources/serializers/marcxml/schema.py
+++ b/invenio_rdm_records/resources/serializers/marcxml/schema.py
@@ -37,7 +37,7 @@ class MARCXMLSchema(Schema):
     sources = fields.Method(
         "get_sources"
     )  # Corresponds to references in the metadata schema
-    coverage = fields.Method(
+    locations = fields.Method(
         "get_locations"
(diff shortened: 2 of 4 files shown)


invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -32,7 +32,7 @@ class DublinCoreSchema(Schema):
     types = fields.Method("get_types")
     sources = fields.Method("get_sources")
     languages = fields.Method("get_languages")
-    coverage = fields.Method("get_locations")
+    location = fields.Method("get_locations")
```
zzacharo, line 35: this should `locations` I think...

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -193,23 +193,31 @@ def get_languages(self, obj):
...
-        #     geometry = location.get('geometry')
-        #     if geometry:
-        #         geo_type = geometry['type']
-        #         if geo_type == "Point":
-        #             coords = geometry['coordinates']
-        #             location_string += f"; lat={coords[0]}; lon={coords[1]}"
-
-        #     locations.append(location_string)
-        return missing
+        locations = []
```
zzacharo, line 196: Isn't that similar if not the same with the marcxml schema method?

TLGINO: It is, should I abstract it/make it one common method for all?

zzacharo: I think that would be nice :) Less error-prone approach!

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -96,31 +97,7 @@ def last_updated(self, obj):
...
-            if geometry:
-                geo_type = geometry["type"]
-                if geo_type == "Point":
-                    coords = geometry["coordinates"]
-                    location_string += f"; lat={coords[0]}; lon={coords[1]}"
-
-            locations.append(location_string)
-
-        return locations
+        return DublinCoreSchema.get_locations(obj)
```
ppanero, line 100: major: I see the fact that the logic inside is the same (very nice catch!), however I would try avoid making schemas dependent of each other. If needed, there could be a `schemas.py` or `fields.py` in the `serializers/` root folder (i.e. one level up than currently). 

cc @zzacharo

TLGINO: sure that seems nicer
I'll get to work on it thx :)

TLGINO: Something like this?
Note: some functions found in `schemas.py` are only used once
This is because the BibTex serializer which I'm working on would also call these functions (hence 2 calls)

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -14,6 +14,8 @@
 from invenio_vocabularies.proxies import current_service as vocabulary_service
 from marshmallow import Schema, fields, missing, post_dump
 
+from ..dublincore.schema import DublinCoreSchema
+from ..schemas import CommmonSchema
```
ppanero, line 18: This import could be removed, and make `MARCXMLSchema(CommmonSchema):` to make use of inheritance. WDYT?

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -50,7 +52,7 @@ class MARCXMLSchema(Schema):
 
     def get_version(self, obj):
         """Get the version of the record."""
-        return obj["metadata"].get("version", missing)
+        return CommmonSchema.get_version(obj)
```
ppanero, line 55: after the inheritance here you can just make `self...`

ppanero (commented):
Just one minor change to use inheritance, then is good to go!

invenio_rdm_records/resources/serializers/marcxml/schema.py
jrcastro2: If we go with inheritance I would rename it to `BaseSchema`.

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -14,6 +14,8 @@
 from invenio_vocabularies.proxies import current_service as vocabulary_service
 from marshmallow import Schema, fields, missing, post_dump
 
+from ..dublincore.schema import DublinCoreSchema
```
jrcastro2, line 17: Not needed
```suggestion
```

invenio_rdm_records/resources/serializers/marcxml/schema.py
TLGINO: Hadn't though of inheritance, have now implemented both comments :)
GH CI tests in progress

ppanero (approved):

kpsherva:
closing, already reviewed and merged in https://github.com/inveniosoftware/invenio-rdm-records/pull/1212

Issue #1211, opened 2023-02-23
Implement meeting field


alejandromumo:
Implement the field `meeting` considering its nested attributes:

- `meeting.acronym`
- `meeting.dates`
- `meeting.place`
- `meeting.session_part`
- `meeting.session`
- `meeting.title`
- `meeting.url`

Take into account the following requirements when implementing this field:

- It is needed in zenodo-rdm. 
- It must be a custom field. 
- It must be searchable. Search must be compatible with the current [Zenodo search guide](https://help.zenodo.org/guides/search/)
- If needed, update DataCite serializer taking into account [Datacite Schema 4.4 docs](https://support.datacite.org/docs/datacite-metadata-schema-44). 
  - A recommended field [`RelatedItem` ](https://support.datacite.org/docs/datacite-metadata-schema-v44-recommended-and-optional-properties#20-relateditem) can be used to serialize journal's information. 
- The upload form should be similar to zenodo (e.g. meeting section is called `Conference `)

PR #1213, opened 2023-02-23
contrib: added meeting custom field

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1211


invenio_rdm_records/contrib/meeting/custom_fields.py
```diff
@@ -0,0 +1,119 @@
...
+                    validate=validate.URL(error="You must provide a valid URL."),
+                ),
+            }
+        )
+
+    @property
+    def mapping(self):
+        """Meeting search mappings."""
+        return {
+            "type": "nested",
```
ppanero, line 49: is this `nested` or `object` in the end? We might get issues with `nested` (they also have performance implications). So unless is needed for a specific use case (i.e. query patterns) I'd change to `object`.

invenio_rdm_records/contrib/meeting/custom_fields.py
```diff
@@ -0,0 +1,119 @@
...
+        {
+            "field": "meeting:meeting",
+            "ui_widget": "Meeting",
+            "template": "meeting.html",
+            "props": {
+                "label": _("Conference"),
+                "title": {
+                    "label": _("Conference title"),
+                    "placeholder": _(""),
+                    "description": _("Add the conference title"),
```
ppanero, line 84: This was also commented in another PR. We need a consensus on how we do descriptions:
a) do we put "actionables" (i.e. add, write, fill in, etc.)
b) do we put mere descriptive text (e.g. "Title of the conference")

I think in the current deposit form we have b). I'm not against changing it, just to be consistent.

invenio_rdm_records/contrib/meeting/custom_fields.py
alejandromumo: You're right, it should be an `object` since we won't have an array of objects (meetings) in the same document.

invenio_rdm_records/contrib/meeting/custom_fields.py
alejandromumo: I agree on keeping b) , for consistency.

invenio_rdm_records/contrib/meeting/custom_fields.py
```diff
@@ -0,0 +1,119 @@
...
+        return fields.Nested(
+            {
+                "acronym": SanitizedUnicode(),
+                "dates": SanitizedUnicode(),
+                "place": SanitizedUnicode(),
+                "session_part": SanitizedUnicode(),
+                "session": SanitizedUnicode(),
+                "title": SanitizedUnicode(),
+                "url": SanitizedUnicode(
+                    validate=validate.URL(error="You must provide a valid URL."),
```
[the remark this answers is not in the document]
ppanero: what about marshmallow-utils? (can be moved to another issue)

ppanero (approved):
Just open an issue to be tackled separately about the URL validation.

invenio_rdm_records/contrib/meeting/custom_fields.py
ppanero: About the RelationshipCF, Alex mentioned we cannot do with object. I do not recall why (but there is also a comment mentioning this in the code 😅 we just miss the reason. @slint ?

slint: This one is `nested` on purpose :)

It's arbitrary key/value list pairs that are represented, which wouldn't work with object.

invenio_rdm_records/contrib/meeting/custom_fields.py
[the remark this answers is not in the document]
ppanero: https://github.com/inveniosoftware/invenio-rdm-records/issues/1220

PR #1222, opened 2023-02-28
records: fix permissions when removing records

ntarocco:
* adds an extra permission check to ensure that a community curator cannot remove records from communities where he is not a curator.

### DIFF
--- a/invenio_rdm_records/services/communities/service.py
+++ b/invenio_rdm_records/services/communities/service.py
@@ -11,6 +11,7 @@ from invenio_records_resources.services import (
     Service,
     ServiceSchemaWrapper,
 )
+from invenio_records_resources.services.errors import PermissionDeniedError
 from invenio_records_resources.services.uow import (
     RecordCommitOp,
     RecordIndexOp,
@@ -41,19 +42,23 @@ class RecordCommunitiesService(Service, RecordIndexerMixin):
         """Include the record in the given communities."""
         pass
 
-    def _remove(self, community_id, record):
+    def _remove(self, identity, community_id, record):
         """Remove a community from the record."""
         if community_id not in record.parent.communities.ids:
             raise RecordCommunityMissing(record.id, community_id)
+
+        # check permission here, per community: curator cannot remove another community
+        self.require_permission(
+            identity, "remove_community", record=record, community_id=community_id
+        )
 
         # Default community is deleted when the exact same community is removed from the record
         record.parent.communities.remove(community_id)
 
     @unit_of_work()
-    def delete(self, identity, record_id, data, revision_id=None, uow=None):
+    def remove(self, identity, record_id, data, revision_id=None, uow=None):
         """Remove communities from the record."""
         record = self.record_cls.pid.resolve(record_id)
-        self.require_permission(identity, "remove_community", record=record)
 
         valid_data, errors = self.schema.load(
             data,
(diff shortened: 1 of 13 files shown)


invenio_rdm_records/services/communities/service.py
```diff
@@ -41,19 +41,23 @@ def create(self, identity, data, record, uow=None):
...
+        self.require_permission(
+            identity, "remove_community", record=record, community_id=community_id
+        )
+
         # Default community is deleted when the exact same community is removed from the record
         record.parent.communities.remove(community_id)
 
     @unit_of_work()
-    def delete(self, identity, record_id, data, revision_id=None, uow=None):
+    def remove(self, identity, record_id, data, revision_id=None, uow=None):
```
ppanero, line 58: we might need to have a `delete` overwritten and `NotImplemented(use remove instead)`. Otherwise programatically we might reach https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/records/service.py#L468 and have unknwon/undesired effect.

ntarocco: Here we have changed approach and we completely create a new service without inheriting from `RecordService`. We inherit from the [base class](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/services/base/service.py#L15).
IMO, when create a service to CRUD records, than we should stick to that interface. For custom ad-hoc services, I would instead create APIs with more meaningful names then REST APIs verbs: in this case `add` and `remove`, instead of `update` and `delete`.

ppanero: LGTM, I missed the inheritance change

tests/conftest.py
```diff
@@ -1408,31 +1408,6 @@ def admin(UserFixture, app, db, admin_role_need):
     return u
 
 
-@pytest.fixture()
-def curator(UserFixture, community, app, db):
```
ntarocco, line 1412: removed because duplicated

tests/resources/conftest.py
```diff
@@ -32,9 +32,7 @@ def rdm_record_service():
 
 
 @pytest.fixture()
-def community_record(
-    uploader, minimal_record, community, rdm_record_service, running_app, db
-):
+def record_community(uploader, minimal_record, community, rdm_record_service, db):
```
ntarocco, line 35: renamed to have the same name as the same fixture in `services/conftest.py`

tests/services/test_service_community_records.py
```diff
@@ -26,22 +26,20 @@ def service():
 
 
 def test_remove_record_from_community_success(
-    running_app, community, record_community, service
+    curator, community, record_community, service
```
ntarocco, line 29: change tests to use `curator` instead of `superuser`

tests/services/test_service_communities.py
```diff
@@ -8,155 +8,159 @@
...
 @pytest.fixture()
-def record_communities_service():
+def service():
     """Get the current records communities service."""
     return current_record_communities_service
 
 
 def test_remove_community_from_record_success(
-    running_app,
+    curator,
```
ntarocco, line 25: change tests to use `curator` instead of `superuser`

invenio_rdm_records/services/communities/service.py
```diff
@@ -68,14 +73,21 @@ def delete(self, identity, record_id, data, revision_id=None, uow=None):
...
                         "community": community_id,
-                        "message": f"The record {record_id} does not belong to the community {community_id}",
+                        "message": "The record does not belong to the community.",
+                    }
+                )
+            except PermissionDeniedError:
+                errors.append(
+                    {
+                        "community": community_id,
+                        "message": "Permission denied.",
```
kpsherva, line 90: maybe we could have a more informative message fe. `Your permissions are not sufficient to remove a record from this community`

ntarocco: We honestly took what is there everywhere, for consistency. The operation is to remove, to give a meaningful msg we should really explain why the permission is denied e.g. `you are not the owner` or `you are not a curator`

tests/resources/test_resources_community_records.py
```diff
@@ -90,7 +90,7 @@ def test_removal_of_multiple_communities_success(
 
 
 def test_exceeded_max_number_of_records(
-    client, curator, community_record, headers, community, service
+    client, curator, record_community, headers, community, service
```
jrcastro2, line 93: Very minor: A leftover I fortgot to remove, my bad ....
```suggestion
```

jrcastro2 (approved):
LGTM!

PR #1225, opened 2023-02-28
serializer: update tests for date intervals

ntarocco:
* changes the UI serializer test for date interval to apply the unicode space char change as in Babel. See https://github.com/python-babel/babel/commit/d76db13414a6104019edc0dbafa82410e8d08e63

Babel 2.12 upgrades CLDR to v42, which introduces this change to the unicode `space` char.

### DIFF
--- a/tests/resources/serializers/test_ui_serializer.py
+++ b/tests/resources/serializers/test_ui_serializer.py
@@ -144,8 +144,8 @@ def test_ui_serializer(app, full_to_dict_record):
                 }
             ],
         },
-        "publication_date_l10n_long": "January 2018 – September 2020",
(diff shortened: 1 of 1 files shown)


kpsherva:
> How would this affect the final UI? I'm wondering if we should decode it to utf-8 here https://github.com/inveniosoftware/marshmallow-utils/blob/master/marshmallow_utils/fields/babel.py#L75
> 
> That would be the lowest module where the value is produced.

there will be no noticeable impact to the UI, except date ranges not jumping to the next line inside the text - which is even better. u2009 is a "thin space", better visually.

kpsherva (approved):

Issue #1202, opened 2023-02-21
export: add DCAT-AP format


anikachurilova:
Add possibility to export file in DCAT-AP format on landing page

PR #1233, opened 2023-03-07
serializers: add datacite to DCAT-AP serializer

anikachurilova:
* fix subjects mapping
* cover serializer with test
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1202

### DIFF
--- a/invenio_rdm_records/resources/serializers/datacite/__init__.py
+++ b/invenio_rdm_records/resources/serializers/datacite/__init__.py
@@ -31,10 +31,7 @@ class DataCite43XMLSerializer(MarshmallowSerializer):
 
     def __init__(self, **options):
         """Constructor."""
-        if options:
-            encoder = options.get("encoder", schema43.tostring)
-        else:
-            encoder = schema43.tostring
+        encoder = options.get("encoder", schema43.tostring)
         super().__init__(
             format_serializer_cls=SimpleSerializer,
             object_schema_cls=DataCite43Schema,
--- a/tests/resources/serializers/test_dcat_serializer.py
+++ b/tests/resources/serializers/test_dcat_serializer.py
@@ -7,115 +7,10 @@
 
 """Resources serializers tests."""
 
-import pytest
-
 from invenio_rdm_records.resources.serializers import DCATSerializer
 
 
-@pytest.fixture
-def full_modified_record(full_record):
-    full_record["metadata"]["additional_descriptions"][0]["lang"]["title"] = {
-        "en": "English",
-    }
-    full_record["metadata"]["additional_descriptions"][0]["type"]["title"] = {
-        "de": "Technische Informationen",
-        "en": "Technical info",
-    }
-    full_record["metadata"]["additional_titles"][0]["lang"]["title"] = {
-        "en": "English",
-    }
-    full_record["metadata"]["additional_titles"][0]["type"]["title"] = {
-        "de": "Alternativer Titel",
-        "en": "Alternative title",
-    }
-    full_record["metadata"]["contributors"][0]["affiliations"].append(
-        {
-            "name": "TU Wien",
-        }
-    )
-    full_record["metadata"]["contributors"][0]["role"]["title"] = {
-        "de": "DatenmanagerIn",
-        "en": "Data manager",
-    }
-    # additional contributor without ORCID
-    full_record["metadata"]["contributors"].append(
-        {
-            "person_or_org": {
-                "family_name": "Dirk",
-                "given_name": "Dirkin",
-                "name": "Dirk, Dirkin",
-                "type": "personal",
-            },
-            "role": {
-                "id": "other",
-                "title": {
-                    "de": "VerteilerIn",
-                    "en": "Other",
-                },
-            },
-        }
-    )
-    # additional creator without ORCID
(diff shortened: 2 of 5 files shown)


invenio_rdm_records/resources/serializers/dcat/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+
+from datacite import schema43
+from lxml import etree as ET
+from pkg_resources import resource_stream
+from werkzeug.utils import cached_property
+
+from invenio_rdm_records.resources.serializers import DataCite43XMLSerializer
+
+
+class DCATSerializer(DataCite43XMLSerializer):
```
ppanero, line 18: This serializer is missing the serialized list (or does the inherited one works fine?), it should also be added in the tests [here](https://github.com/ppanero/invenio-rdm-records/blob/c88cbde27c432d732a4e49571725b27f959c25df/tests/resources/test_resources.py#L736) to make sure it works.

invenio_rdm_records/resources/serializers/dcat/__init__.py
```diff
@@ -0,0 +1,63 @@
...
```
ppanero, line 18: major: IMO inheritance is being broken here. The parent class is defined with `SimpleSerializer` and also the `BaseListSerializer`. The later one is allowing the proper serialization at the level of the first. This leads me to two questions:
- Why is `serialize_object_list` needed here if the implementation is identical to the formatter class (`SimpleSerializer`), see [here](https://github.com/inveniosoftware/flask-resources/blob/master/flask_resources/serializers/simple.py#L24)
- Could it be refactored so we also do not need `serialize_object`? If we move the call to `self.transform_with_xslt` inside `_etree_tostring` we could simply pass this function as encoder. This would require overwritting the constructor passing `encoder=self._etree_tostring`.

pseudocode:

```python
def __init__(self, **options):
        """Constructor."""
        super().__init__(encoder=self. _etree_tostring)

def _etree_tostring(self, record, **kwargs):
         root = self.transform_with_xslt(record, **kwargs)
         return ET.tostring(
             root,
             pretty_print=True,
             xml_declaration=True,
             encoding="utf-8",
         ).decode("utf-8")
```

Note that this will require a change in [`DataCite43XMLSerializer`](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/datacite/__init__.py#L29) to use options:

```python

def __init__(self, **options):
        """Constructor."""
        options = options or {"encoder": schema43.tostring}
        super().__init__(
            **options,
             ...
        )
```

It is a bit tedious (I'm sorry) but it will make use of inheritance and leave the code much cleaner/clear.

invenio_rdm_records/resources/serializers/datacite/__init__.py
```diff
@@ -31,9 +31,10 @@ class DataCite43XMLSerializer(MarshmallowSerializer):
 
     def __init__(self, **options):
         """Constructor."""
+        encoder = options["encoder"] if "encoder" in options else schema43.tostring
```
ppanero, line 34: This might fail if options is passed as `None` (i.e. no `kwargs`):
```python
>>> options = None
>>> encoder = options["encoder"] if "encoder" in options else "test"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'NoneType' is not iterable
```

In addition you can use `get` with a default value:

```suggestion
if options:
        encoder = options.get("encoder", schema43.tostring)
```

tests/resources/test_resources.py
```diff
@@ -742,6 +742,7 @@ def test_response_handlers(running_app, minimal_record, client_with_login):
         "application/vnd.datacite.datacite+xml",
         "application/x-dc+xml",
         "text/x-bibliography",
+        "text/turtle",
```
ppanero, line 745: missing discussion today at 14 about turtle vs rdf or both.

ppanero (approved):

tests/resources/serializers/test_dcat_serializer.py
```diff
@@ -0,0 +1,261 @@
...
+"""Resources serializers tests."""
+
+import pytest
+
+from invenio_rdm_records.resources.serializers import DCATSerializer
+
+
+@pytest.fixture
+def full_modified_record(full_record):
+    full_record["metadata"]["additional_descriptions"][0]["lang"]["title"] = {
```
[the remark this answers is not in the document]
anikachurilova: Additional fields do not only test DCAT schema, there are just all the fields that can be in the record, and are absent in the `full_record`. Some of them are not even serialized by DCAT, but still good to have them in test in case the serializer changes - we would know.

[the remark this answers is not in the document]
slint: I feel that hints towards a need to update `full_record` with all the fields (so that it is truly "full"). But I get that this might actually cause other tests to fail (which would be good to discover though :) ).

A silly alternative is to create a new `fullest_record` fixture or similar and open an issue to swap it later on.

ppanero (approved):
I've opened https://github.com/inveniosoftware/invenio-rdm-records/issues/1254 to consolidate the fixtures as a diff task.

Issue #1234, opened 2023-03-08
oai-pmh: add export format for DCAT-AP


anikachurilova:
epic: https://github.com/inveniosoftware/invenio-app-rdm/issues/2031

PR #1239, opened 2023-03-08
OAI-PMH: export for DCAT-AP

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1234

### DIFF
--- a/invenio_rdm_records/oai.py
+++ b/invenio_rdm_records/oai.py
@@ -21,6 +21,7 @@ from lxml import etree
 
 from .proxies import current_rdm_records
 from .resources.serializers.datacite import DataCite43XMLSerializer
+from .resources.serializers.dcat import DCATSerializer
 from .resources.serializers.dublincore import DublinCoreXMLSerializer
 from .resources.serializers.marcxml import MARCXMLSerializer
 from .services.pids.providers.oai import OAIPIDProvider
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -409,11 +409,11 @@ def get_subjects(self, obj):
...
         for subject in subjects:
-            sub_text = subject.get("subject")
-            if sub_text:
-                serialized_subjects.append({"subject": sub_text})
+            _id = subject.get("id")
+            if _id:
+                ids.append(_id)
             else:
-                ids.append(subject.get("id"))
+                serialized_subjects.append({"subject": subject.get("subject")})
```
rekt-hard, line 416: Will this access always evaluate to something to be used or should there be a check/default value?

rekt-hard (approved):
LGTM :)

invenio_rdm_records/resources/serializers/datacite/schema.py
anikachurilova: The subjects list can be missing (which is handled on line 406). Bus as long as subjects is present, the subject inside is mandatory (it's either we add subjects or we don't)

jrcastro2 (approved):

invenio_rdm_records/oai.py
```diff
@@ -40,6 +41,16 @@ def oai_marcxml_etree(pid, record):
     return etree.fromstring(MARCXMLSerializer().serialize_object(record["_source"]))
 
 
+def oai_dcat_etree(pid, record):
+    """OAI DCAT-AP format for OAI-PMH.
+
+    It assumes that record is a search result.
+    """
+    return etree.fromstring(
+        DCATSerializer().serialize_object(record["_source"]).encode(encoding="utf-8")
```
ppanero, line 50: 🚀 very nice, I opened and issue to do this in other functions that don't 💪  https://github.com/inveniosoftware/invenio-rdm-records/issues/1255

ppanero (approved):

Issue #1232, opened 2023-03-07
custom fields: update deposit form


kpsherva:
- [x] Custom field imprint is missing "book" field - compare with zenodo
- [x] Implement the agreed solution based on mockups
- [x] Update the deposit form
- [x] Update the landing page
- [x] Update the record's listing.   

**Updated mockups**

## Deposit form
![Image](https://user-images.githubusercontent.com/38131488/223471736-b3c59354-ad16-4afe-aefc-92cc9cbda338.png)

## Landing page
![image](https://user-images.githubusercontent.com/21204744/224312905-e6557d80-bc05-4f9e-9bff-fbc9209110fb.png)

## Record's listing
![image](https://user-images.githubusercontent.com/21204744/224312974-3bd94390-d8ef-4cd0-a534-b9dc8c6b0ad1.png)

alejandromumo:
The field `imprint.title` is already implemented: https://github.com/inveniosoftware/invenio-rdm-records/blob/fd17d8d236312db85dd8316035e28d210175636f/invenio_rdm_records/contrib/imprint/custom_fields.py#L31

Is this issue about the UI element (e.g. label)?

alejandromumo:
## Deposit form

![Screenshot at Mar 08 16-48-10](https://user-images.githubusercontent.com/21204744/223761902-346cbaba-de83-44c7-85dc-5e650894eb08.png)


![Screenshot at Mar 08 16-48-01](https://user-images.githubusercontent.com/21204744/223761912-0b3935e5-cf9b-4c1c-b46b-ff74e83e744a.png)

## Landing page


![Screenshot at Mar 10 11-20-00](https://user-images.githubusercontent.com/21204744/224290889-06f7986e-0b5f-452d-8673-1b4e00ca973e.png)


## Record listing

![Screenshot at Mar 10 11-11-56](https://user-images.githubusercontent.com/21204744/224289861-eefd77c8-46b2-49f8-bb9c-cc80f56949ff.png)

![Screenshot at Mar 10 11-12-23](https://user-images.githubusercontent.com/21204744/224289854-ed9d401a-0510-4075-b2d6-351f1e433cae.png)

alejandromumo:
The current implementation creates a duplication in the UI:

![image](https://user-images.githubusercontent.com/21204744/224331518-70c4ab0d-8bdb-4fda-ad0e-c1baed89c93c.png)

[...]

PR #1212, opened 2023-02-23
serializers: export for Bibtex

TLGINO:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1208

PR #1240, opened 2023-03-08
contrib: updated custom fields UI widgets.

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1232


alejandromumo:
Tests are failing until https://github.com/inveniosoftware/invenio-rdm-records/pull/1212 is merged

invenio_rdm_records/contrib/imprint/custom_fields.py
```diff
@@ -10,7 +10,6 @@
 - imprint.isbn
 - imprint.pages
 - imprint.place
-- imprint.publisher
```
alejandromumo, line 13: Publisher was redundant, the record already has a publisher. Therefore, it was removed.

invenio_rdm_records/contrib/thesis/custom_fields.py
```diff
@@ -25,15 +25,19 @@
...
-            props=dict(
-                label="Awarding university",
-                icon="building",
-                description="University name",
-                placeholder="",
-            ),
-        )
+        {
+            "field": "thesis:university",
+            "ui_widget": "Thesis",
```
alejandromumo, line 30: Had to implement a new widget so the UI is consistent (compared to other custom fields e.g. `journal`, `imprint`)

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -143,6 +143,69 @@ class FundingSchema(Schema):
...
+        )
+        result.update({"imprint": imprint_string})
+
+    if thesis:
+        result.update({"thesis": thesis})
+
+    return result
+
+
+def compute_conference_information(obj, dummyctx):
```
alejandromumo, line 195: This simply "copies" the conference custom field to another field:

```python
record = {
    'custom_fields': {
        'meeting:meeting': {
            'title': 'xyz'
        }
    },
    'conference': {
        'title': 'xyz'
    }
}
```

and then gets rendered in the UI. 

I didn't create a single string (like in `publishing_information`) because the UI for conference also has other elements (e.g. `url`)

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -143,6 +143,69 @@ class FundingSchema(Schema):
...
+
+    result = {}
+    if journal:
+        title = journal.get("title", "")
+        if title:
+            volume = journal.get("volume", "")
+            issn = journal.get("issn", "")
+            issue = journal.get("issue", "")
+            pages = journal.get("pages", "")
+            journal_string = (
```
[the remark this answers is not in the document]
tmorrell: I think it makes sense to have in the ui serializer, since it will be used in both the record landing page and search results.

I don't think we want the ISSN to show up here. Makes more sense under the "Imprint" field, if it's going in the top box.

One solution to avoid having to think too much about the implementation is to pick a citation style, and have this string match what is automatically generated in the citation box. There are lots of ways for this string formatting to go wrong. 

An example bug in the current Zenodo implementation. An article could have a page range, or an article number. In the current implementation those are both dropped into the pages field, but they need to be rendered differently (p. for an article number, pp. for page range). You can see an example of this at https://zenodo.org/record/1155339. If you change the automatic citation to 'IEEE', you'll see how the automatic citation treats the pages field.

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -171,6 +234,10 @@ class UIRecordSchema(BaseObjectSchema):
         partial(CustomFieldsSchemaUI, fields_var="RDM_CUSTOM_FIELDS")
     )
 
+    publishing_information = fields.Function(compute_publishing_information)
+
+    conference = fields.Function(compute_conference_information)
```
jrcastro2, line 239: Comment: Maybe I am saying something stupid here, but couldn't this be replaced by:
```suggestion
    conference = fields.Nested(<your_field_class>, attribute="custom_fields.meeting:meeting")
```

invenio_rdm_records/resources/serializers/ui/schema.py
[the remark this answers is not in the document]
kpsherva: to be discussed internally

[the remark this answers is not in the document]
kpsherva: > I think it makes sense to have in the ui serializer, since it will be used in both the record landing page and search results.
> 
> I don't think we want the ISSN to show up here. Makes more sense under the "Imprint" field, if it's going in the top box.
> 
> One solution to avoid having to think too much about the implementation is to pick a citation style, and have this string match what is automatically generated in the citation box. There are lots of ways for this string formatting to go wrong.
> 
> An example bug in the current Zenodo implementation. An article could have a page range, or an article number. In the current implementation those are both dropped into the pages field, but they need to be rendered differently (p. for an article number, pp. for page range). You can see an example of this at https://zenodo.org/record/1155339. If you change the automatic citation to 'IEEE', you'll see how the automatic citation treats the pages field.

@tmorrell to follow up on this: we agreed we need to keep the Zenodo citation style for the time being, to keep the backwards compatibility with the format which was discussed and agreed previously. If you would like to suggest a change here, I think it should be discussed on the telecon.

PR #1241, opened 2023-03-09
serializers: replace fields.Method by attribute key with dotted access

ppanero:
We could even move actual fields to the base schema, but that was more complex and error prone work.

### DIFF
--- a/invenio_rdm_records/resources/serializers/bibtex/schema.py
+++ b/invenio_rdm_records/resources/serializers/bibtex/schema.py
@@ -9,7 +9,7 @@
 
 import datetime
 
-from marshmallow import Schema, fields, missing, post_dump
+from marshmallow import fields, missing, post_dump
 from slugify import slugify
 
 from ..schemas import BaseSchema
@@ -21,7 +21,7 @@ class BibTexSchema(BaseSchema):
 
     id = fields.Str()
     resource_id = fields.Str(attribute="metadata.resource_type.id")
-    version = fields.Method("get_version")
+    version = fields.Str(attribute="metadata.version")
     date_created = fields.Method("get_date_created")
     locations = fields.Method("get_locations")
     titles = fields.Method("get_titles")
--- a/invenio_rdm_records/resources/serializers/csl/schema.py
+++ b/invenio_rdm_records/resources/serializers/csl/schema.py
@@ -57,7 +57,7 @@ class CSLJSONSchema(Schema):
     language = fields.Method("get_language")
     version = SanitizedUnicode(attribute="metadata.version")
     note = fields.Method("get_note")
-    doi = fields.Method("get_doi", data_key="DOI")
+    doi = fields.Str(attribute="pids.doi.identifier", data_key="DOI")
     isbn = fields.Method("get_isbn", data_key="ISBN")
     issn = fields.Method("get_issn", data_key="ISSN")
     publisher = SanitizedUnicode(attribute="metadata.publisher")
@@ -105,10 +105,6 @@ class CSLJSONSchema(Schema):
         languages = metadata.get("languages")
 
         return languages[0]["id"] if languages else missing
-
-    def get_doi(self, obj):
-        """Get DOI."""
-        return obj["pids"].get("doi", {}).get("identifier", missing)
 
     def get_isbn(self, obj):
         """Get ISBN."""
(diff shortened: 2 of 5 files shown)


invenio_rdm_records/resources/serializers/bibtex/schema.py
```diff
@@ -21,7 +21,7 @@ class BibTexSchema(BaseSchema):
 
     id = fields.Str()
     resource_id = fields.Str(attribute="metadata.resource_type.id")
-    version = fields.Method("get_version")
+    version = fields.Str(attribute="metadata.version")
```
TLGINO, line 24: This is a cool way to do it, will use it in the future!

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -36,10 +36,6 @@ class DublinCoreSchema(BaseSchema):
     locations = fields.Method("get_locations")
     formats = fields.Method("get_formats")
```
anikachurilova, line 37: Can be changed the same way as you did in marcxml schema? Also for the `sources` field.

invenio_rdm_records/resources/serializers/csl/schema.py
```diff
@@ -57,7 +57,7 @@ class CSLJSONSchema(Schema):
     language = fields.Method("get_language")
     version = SanitizedUnicode(attribute="metadata.version")
     note = fields.Method("get_note")
-    doi = fields.Method("get_doi", data_key="DOI")
+    doi = fields.Str(attribute="pids.doi.identifier", data_key="DOI")
```
anikachurilova, line 60: Small question: are we sure that the logic of the `missing` field or something similar stays if we use it like this?

anikachurilova (commented):
Thank you for getting rid of so much code! Good job! 
Left some small questions/comments.

invenio_rdm_records/resources/serializers/dublincore/schema.py
ppanero: You are correct, thanks!

invenio_rdm_records/resources/serializers/csl/schema.py
ppanero: Yes, the need for `missing` is because if a `Method` field returns `None` it will be use as its value, so there is a need to specify "ignore this field". For example:

```python
>>> from marshmallow import Schema, fields, missing

>>> class A(Schema):
...     test = fields.Str()
...     another = fields.Str()

>>> data = {"test": 1}
>>> A().dump(data)
{'test': '1'}       < --- See here the key "another" is not present
>>> A().dump({})
{}                     < --- Both keys are ignored
```

Now if we replace test by a method:

```python
>>> class B(Schema):
...     test = fields.Method("get_test")
...     another = fields.Str()
...     def get_test(self, obj):
...             return obj.get("test")
...
>>> B().dump(data)
{'test': 1}
>>> B().dump({})
{'test': None}     <--- we got `None` as value
```

Now we if return missing we go back to the same behaviour than in the first case:

```python
>>> class C(Schema):
...     test = fields.Method("get_test")
...     another = fields.Str()

...     def get_test(self, obj):
...             return obj.get("test", missing)
...
>>> C().dump(data)
{'test': 1}
>>> C().dump({})
{}
```

anikachurilova (approved):

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -31,15 +31,14 @@ class DublinCoreSchema(BaseSchema):
     descriptions = fields.Method("get_descriptions")
     publishers = fields.Method("get_publishers")
     types = fields.Method("get_types")
-    sources = fields.Method("get_sources")
+    # TODO: sources = fields.List(fields.Str(), attribute="metadata.references")
+    sources = fields.Constant(
```
kpsherva, line 35: why do we have this as TODO? what is missing?

ppanero: That was there before the code changes. i.e. not part of this PR, I just made it more clear.

IMO, I think the [`references`](https://github.com/inveniosoftware/invenio-rdm-records/blob/a02775d565795a2574bd6d8812143b64917de6fd/invenio_rdm_records/services/schemas/metadata.py#L367) is just not used at the moment (i.e. not exposed in the UI). We are defaulting to missing (as it was before, this behaviour didn't change) to avoid data migration afterwards, so right now is just ignored.

PR #1243, opened 2023-03-09
serializers: fix legacy marcxml issue

TLGINO:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2128

### DIFF
--- a/invenio_rdm_records/resources/serializers/marcxml/schema.py
+++ b/invenio_rdm_records/resources/serializers/marcxml/schema.py
@@ -22,10 +22,11 @@ from ..utils import get_vocabulary_props
 class MARCXMLSchema(BaseSchema):
     """Schema for records in MARC."""
 
+    doi = fields.Method("get_doi")
+    oai = fields.Str(attribute="pids.oai.identifier")
     contributors = fields.Method("get_contributors")
     titles = fields.Method("get_titles")
     creators = fields.Method("get_creators")
-    identifiers = fields.Method("get_identifiers")
     relations = fields.Method("get_relations")
     rights = fields.Method("get_rights")
     dates = fields.Method("get_dates")
@@ -39,7 +40,6 @@ class MARCXMLSchema(BaseSchema):
     sources = fields.Constant(
         missing
     )  # Corresponds to references in the metadata schema
-    locations = fields.Method("get_locations")
     formats = fields.List(fields.Str(), attribute="metadata.formats")
     parent_id = fields.Str(attribute="parent.id")
     community_id = fields.List(fields.Str(), attribute="parent.communities.ids")
@@ -222,7 +222,7 @@ class MARCXMLSchema(BaseSchema):
             "contributors": ["700a ", "u"],  # Abstract
             "titles": ["245a ", "a"],  # Title
             "creators": ["100a ", "a"],  # First author
-            "identifiers": ["024  ", "a"],  # DOI
+            "oai": ["909CO", "o"],  # OAI
             "relations": ["856 2", "a"],  # Related Ressource
             "rights": ["540  ", "a"],  # License
             "dates": [
@@ -254,5 +254,9 @@ class MARCXMLSchema(BaseSchema):
         for key in data_copy:
             if key in changes:
                 data[changes[key][0]] = {changes[key][1]: data.pop(key)}
+
+        doi = data.pop("doi", None)
+        if doi:
+            data["024  "] = {"a": doi, "2": "doi"}
 
         return data
--- a/invenio_rdm_records/resources/serializers/schemas.py
+++ b/invenio_rdm_records/resources/serializers/schemas.py
@@ -12,6 +12,17 @@ from marshmallow import Schema, fields, missing
 class BaseSchema(Schema):
     """BaseSchema for serializers in JSON."""
 
+    def get_doi(self, obj):
+        """Get DOI."""
+        if "doi" in obj["pids"]:
+            return obj["pids"]["doi"]["identifier"]
+
+        for identifier in obj["metadata"].get("identifiers", []):
+            if identifier["scheme"] == "doi":
+                return identifier["identifier"]
+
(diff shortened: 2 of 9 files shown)


invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -22,10 +21,11 @@
 class MARCXMLSchema(BaseSchema):
     """Schema for records in MARC."""
 
+    doi = fields.Method("get_doi")
+    oai = fields.Method("get_oai")
```
ppanero, line 25: Could this be obtained using the `attribute`?
```suggestion
    oai = fields.Str(attibute="pids.oai.identifier")
```

invenio_rdm_records/resources/serializers/schemas.py
```diff
@@ -12,6 +12,15 @@
 class BaseSchema(Schema):
     """BaseSchema for serializers in JSON."""
 
+    def get_doi(self, obj):
```
ppanero, line 15: Why do we favor related identifiers instead of `pids`?  see https://inveniordm.docs.cern.ch/reference/metadata/#external-pids for the `pids` field vs identifiers https://inveniordm.docs.cern.ch/reference/metadata/#alternate-identifiers-0-n

I see two options:
- Use just the `pids.doi.identifier` field
- Use both, meaning if the value is in `pids.doi.identifier` use that one, otherwise fallback to the `identifiers` field
WDYT?

tests/resources/serializers/test_marcxml_serializer.py
```diff
@@ -100,72 +103,67 @@ def test_marcxml_serializer_full_record(running_app, updated_full_record):
...
+            '    <subfield code="u">Nielsen, Lars Holm</subfield>',
+            "  </datafield>",
+            '  <datafield tag="540" ind1=" " ind2=" ">',
+            '    <subfield code="a">info:eu-repo/semantics/embargoedAccess</subfield>',
+            '    <subfield code="a">A custom license</subfield>',
+            '    <subfield code="a">https://customlicense.org/licenses/by/4.0/</subfield>',
+            '    <subfield code="a">Creative Commons Attribution 4.0 International</subfield>',
+            '    <subfield code="a">https://creativecommons.org/licenses/by/4.0/legalcode</subfield>',
+            "  </datafield>",
+            '  <datafield tag="024" ind1=" " ind2="3">',
```
ppanero, line 119: according to the issue the pids should be splited. However, I do not see the doi of the record (see [fixture](https://github.com/inveniosoftware/invenio-rdm-records/blob/5bdf56cf30115207648f61661e5e0378cb0a4c6f/tests/conftest.py#L371)) in the serialization. I **think** (not 100% sure) that the result should be something like:

```
<datafield tag="024" ind1=" " ind2=" ">
        <subfield code="a">10.5281/inveniordm.1234</subfield>
        <subfield code="2">doi</subfield>
</datafield>
```

invenio_rdm_records/resources/serializers/schemas.py
ppanero: It is a system field, but it is configurable per instance (i.e. `DOI_ENABLED=False`) so there might be instances that do not have DOIs. While this is a big issue (e.g. datacite serialization should not work, I actually wonder how gracefully we fail in that case - bug 🤔? ), serializers such as bibtex should still work?

TLGINO: so regarding BibTex, DOI is not needed for serialization (it is considered an optional param)
Regarding the rest of your comments I'm a bit lost :sweat_smile:

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -255,4 +255,9 @@ def keys_to_tags(self, data, many, **kwargs):
             if key in changes:
                 data[changes[key][0]] = {changes[key][1]: data.pop(key)}
 
+        doi = data.get("doi", None)
```
ppanero, line 258: you can merge line 258 and 260. If you pass a default value to `pop` it will not throw an exception.

```python
>>> a = {"test": 1}
>>> a.pop("no", None)
>>> a.pop("test")
1
```

so you could do something like:

```python
doi = data.pop("doi", None)
if doi:
    data[...] = ...
```

invenio_rdm_records/resources/serializers/schemas.py
```diff
@@ -12,6 +12,19 @@
...
+        """Get DOI."""
+        item = [
```
ppanero, line 17: I think the idea is the opposite:

1. Try `pids`. This field has more checks and constraints so it is more "trust worthy" and should be taken into account first (right @kpsherva?)
2. Then default to identifiers

In addition, I think early return could be use to simplify the logic of this function. Creating lists to then return `iten[0]` is a bit counter intuitive. In addtion, it means that we would be looping through all identifiers when in reality we only need the first one (i.e. we could save processing time). For example, as follows (pseudocode):

```python
if "doi" in ...:
    return obj[...]

for identifier in obj[...]:
    if identifier["scheme"] == "doi":
        return ...

return missing
```

Since we are using returns we can also avoid using nested if-else.

tests/resources/serializers/test_marcxml_serializer.py
ppanero: I think I'm missing something. Are DOI and OAI being tested in `024` fields, I think it should be tested. I can only find the `v1` identifiers in the tests. We can chat or have a quick call to clarify this if needed.

invenio_rdm_records/resources/serializers/marcxml/schema.py
TLGINO: Oh didn't know that was a thing thx!

ppanero (approved):
LGTM, nice job!

Needs clarification about `ind1 = 7`, can be tackled separately.

PR #1256, opened 2023-03-16
WIP: index stats with records to enable sorting by them

max-moser:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2143

After a bit of digging through which pieces of code are related to indexing records in the search indices, I found that the most appropriate spot would be via a `SearchDumper` extension.

As such, this PR introduces the `StatisticsDumperExt` which queries the current record's statistics from the aggregations search indices and puts them into a new field for the dumped/indexed JSON.
Further, it adds the `stats` field and its values to the latest record search index mappings.
This will require a rebuild of the indices in the version upgrade guide.
Given that the changes are backwards-compatible (just additions), I don't think that a new version for the mappings is required?

**Idea:** A (read-only) system field for the record could be beneficial to enable easy fetching of statistics for a record in `invenio shell`, e.g.:
```python
In [0]: record.stats
Out[0]:
{'this_version': {'views': 1,
  'unique_views': 1,
  'downloads': 0,
  'unique_downloads': 0,
  'data_volume': 0.0},
 'all_versions': {'views': 1,
  'unique_views': 1,
  'downloads': 0,
  'unique_downloads': 0,
  'data_volume': 0.0}}
```

For now (just to test out a prototype implementation), the logic for fetching the queries was copied from [Invenio-App-RDM](https://github.com/inveniosoftware/invenio-app-rdm/blob/master/invenio_app_rdm/records_ui/views/records.py#L79-L131).
This duplication should be eliminated and a proper place to live for the code should be found.
The potentially ugly part about it is that the query requires knowledge about the structure of the query results, which is configured in Invenio-App-RDM.

To do:
* [ ] Check if we can utilize some short-lived caching for the statistics
* [ ] Wire up the front-end to enable sorting by statistics there
* [ ] De-duplicate the query logic
* [ ] Explore the system-field idea


invenio_rdm_records/config.py
```diff
@@ -197,6 +197,10 @@ def always_valid(identifier):
         title=_("Least recently updated"),
         fields=["updated"],
     ),
+    "mostviewed": dict(title=_("Most viewed"), fields=["-stats.all_versions.views"]),
+    "mostdownloaded": dict(
+        title=_("Most downloaded"), fields=["-stats.all_versions.downloads"]
+    ),
```
max-moser, line 203: this works if the statistics are only dumped into the record JSON before indexing (e.g. via a dumper extension), and does not require the statistics to be part of the record otherwise

invenio_rdm_records/records/dumpers/statistics.py
```diff
@@ -0,0 +1,113 @@
...
+        """Dump the download & view statistics to the data dictionary."""
+        if record.is_draft:
+            return
+
+        try:
+            recid = record.pid.pid_value
+            parent_recid = record.parent.pid.pid_value
+
+            parent_data = dict_lookup(data, self.keys, parent=True)
+            parent_data[self.key] = _get_stats(recid=recid, parent_recid=parent_recid)
```
ntarocco, line 107: Move the `try` here?

invenio_rdm_records/records/dumpers/statistics.py
```diff
@@ -0,0 +1,113 @@
...
+        except KeyError as e:
+            current_app.logger.warning(e)
+
+    def load(self, data, record_cls):
+        """Keep the download & view statistics in the data dictionary."""
+        pass
```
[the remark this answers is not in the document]
ntarocco: After internal chat, yes the REST API should return the stats, this is what is happening on Zenodo, e.g. https://zenodo.org/api/records/7753101 (ping @kpsherva)

invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -70,7 +70,7 @@ class Meta:
     is_published = fields.Boolean(dump_only=True)
     status = fields.String(dump_only=True)
 
-    # stats = NestedAttribute(StatsSchema, dump_only=True)
+    stats = fields.Dict(attribute="stats", dump_only=True)
```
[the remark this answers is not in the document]
ntarocco: What happens if I create a record, and I pass a `stats` field? It should be ignored/removed because of the `dump_only` and also because of the `pass` in the `load` method.

max-moser:
One issue that I've observed stems from the fact that the display of statistics in search results follows a different logic path than the landing page's statistics:
* The landing page gets the statistics directly from the aggregations indices and is thus quite up to date [1]
* The search gets their statistics from the indexed records

So what can happen is that the statistics displayed on the landing page are _more up to date_ (and higher) than the search results, which is very weird from a user perspective.
We should find a way to keep the record indices more up to date...

Maybe as part of the statistics aggregator?
We could add some logic that checks which records' statistics have been updated by the last aggregation, and then trigger reindexing of the affected records?

---

[1] at least, as up to date it reasonably can be – question about regularity of aggregation task execution and a little bit of potential caching is out of scope here

invenio_rdm_records/records/systemfields/statistics.py
```diff
@@ -0,0 +1,52 @@
...
+
+        return stats
+
+    def pre_commit(self, record, **kwargs):
+        """Ensure that the statistics stay transient."""
+        record.pop("stats", None)
+
+    def pre_dump(self, record, data, **kwargs):
+        """Do nothing in particular before a record is dumped."""
+        # TODO this could be an alternative to the search dumper extension
```
max-moser, line 48: what do you think :question: :grey_question: :question:

invenio_rdm_records/services/schemas/__init__.py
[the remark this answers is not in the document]
max-moser: yes, exactly.
if a `stats` field is provided, it will be ignored.

max-moser:
:information_source:
Also, please note that this PR only provides the `stats` field to records, but not to drafts.
That's because we're not tracking any statistics for drafts, and querying for record stats in drafts would mix stuff up.
One consequence is that there's no stats being displayed in the search results in "my dashboard", and neither do we offer the possibility to sort by views/downloads there.

invenio_rdm_records/records/systemfields/statistics.py
```diff
@@ -0,0 +1,52 @@
...
+            # returns the field itself.
+            return self
+
+        stats = record.get("stats", None)
+        if stats:
+            return stats
+
+        recid = record["id"]
+        parent_recid = record.parent["id"]
+        stats = get_record_stats(recid=recid, parent_recid=parent_recid)
```
slint, line 37: On the point of the out-of-sync stats on the search vs. record landing page: on Zenodo we actually [pull the stats on the record landing page from ES](https://github.com/zenodo/zenodo/blob/513e1fc8882c80e4568ab66bae60ebc658646356/zenodo/modules/stats/utils.py#L119-L124), buy fetching only the `_stats` key from the indexed document. That way we have the same numbers, and ES acts as a "cache" as well. We could maybe use the same logic here?

invenio_rdm_records/stats/permissions.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+
+from invenio_stats import current_stats
+
+DenyAllPermission = type(
+    "Deny",
+    (),
+    {"can": lambda self: False, "allows": lambda *args: False},
+)()
+"""The permission factory that denies without even blinking."""
```
kpsherva, line 21: 😅

invenio_rdm_records/stats/permissions.py
```diff
@@ -0,0 +1,33 @@
...
+
+"""Permission factories for Invenio-Stats.
+
+In contrast to the very liberal defaults provided by Invenio-Stats, these permission
+factories deny access unless otherwise specified.
+"""
+
+from invenio_stats import current_stats
+
+DenyAllPermission = type(
```
kpsherva, line 16: I think we already have the DenyAll, I have a feeling this is a duplicate

invenio_rdm_records/stats/templates/aggregations/aggr_file_download/os-v1/aggr-file-download-v1.0.0.json
```diff
@@ -0,0 +1,69 @@
...
+    "properties": {
+      "timestamp": {
+        "type": "date",
+        "format": "date_optional_time"
+      },
+      "count": {
+        "type": "integer"
+      },
+      "unique_count": {
+        "type": "integer"
```
kpsherva, line 32: will we ever max out the integer?

invenio_rdm_records/stats/templates/events/record_view/os-v2/record-view-v1.0.0.json
```diff
@@ -0,0 +1,70 @@
+{
+  "index_patterns": ["__SEARCH_INDEX_PREFIX__events-stats-record-view-*"],
```
kpsherva, line 2: silly question: aren't these mappings already defined in invenio stats?

invenio_rdm_records/records/systemfields/statistics.py
max-moser: I've updated the statistics system field s.t. it will first try to get the record's stats from the records search index (`rdmrecords-records-record-v5.0.0`).
Only as a fallback will the system field try to get the most up-to-date statistics via the `invenio-stats` queries.

The code isn't exactly pretty, but I don't think that the required piece of logic has been implemented already (e.g. the `service.indexer` doesn't have a `.get()`, etc.).

invenio_rdm_records/stats/templates/events/record_view/os-v2/record-view-v1.0.0.json
max-moser: You mean the [`contrib` mappings](https://github.com/inveniosoftware/invenio-stats/blob/master/invenio_stats/contrib/record_view/os-v2/record-view-v1.json), right?

Unfortuantely, they don't match our use-case exactly – for example, they're lacking the `parent_recid` mapping, which we're using for statistics of a single record version vs. all vesrions.
I'm not 100% sure about the actual usage of the `contrib` mappings (probably they're purely examples), so I left them as-is.

invenio_rdm_records/stats/templates/aggregations/aggr_file_download/os-v1/aggr-file-download-v1.0.0.json
max-moser: In the long run, definitely – if we get to experience it in our life times however is a different question :smile: 
Zenodo is [using integers as well](https://github.com/zenodo/zenodo/blob/master/zenodo/modules/stats/templates/aggregations/v7/aggr-record-download-v1.0.0.json#L29), so I suspect the chances are relatively low.

The [official recommendations](https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html#_which_type_should_i_use) on which integer type to pick are relatively vague.
That all being said, I wouldn't mind upgrading to `long`.
It seems like it wouldn't make a difference for storage anyway, and I doubt that this would become a performance bottleneck for us.

invenio_rdm_records/services/permissions.py
```diff
@@ -63,6 +63,14 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     can_authenticated = [AuthenticatedUser(), SystemProcess()]
     can_all = [AnyUser(), SystemProcess()]
 
+    #
+    # Miscellaneous
+    #
+    # Allow for querying of statistics
+    # - This is currently disabled because it's not needed and could potentially
+    #   open up surface for denial of service attacks
+    can_query_stats = [Disable()]
```
max-moser, line 72: @kpsherva the new translation layer between query API and service-layer permissions uses this new permission per default.
It should be relatively easy to adapt in a way s.t. the statistics API endpoint actually checks for record permissions based on the query parameters given by the user.

This (or a similar) permission could also be used for `stats` field-level permissions when dumping the records' stats with the marshmallow schema.

invenio_rdm_records/stats/permissions.py
```diff
@@ -0,0 +1,59 @@
...
+class StatsPermissionTranslator:
+    """Translator class between the statistics endpoint and the service layer."""
+
+    def __init__(self, action_name, service=None, **params):
+        """Constructor."""
+        self.service = service or records_service
+        self.action_name = action_name
+        self.params = params
+
+    def translate_params(self, identity):
```
max-moser, line 29: The `identity` argument is accepted but not used here because this `StatsPermissionTranslator` class is expected to be subclassed to match various needs.
If we're not passing the identity here, subclasses would also have to access `flask.g.identity`, which feels a bit weird and requires an additional import.

For example, I can easily imagine a use-case where the record needs to be resolved from the DB (e.g. via the service) before further checks.

This could be done by code like:
```python
def translate_params(self, identity):
  params = {**self.params}
  if "recid" in params:
    rec = self.service.read(id_=params["recid"], identity=identity)._record
    params["record"] = rec  # for the permission generators

  return params
```

invenio_rdm_records/stats/templates/aggregations/aggr_file_download/os-v1/aggr-file-download-v1.0.0.json
slint: `integer` supports up to 2^32 (so ~4 billion), which I think is more than fine (the most viewed record we have right now on Zenodo is at ~1.5 million).

invenio_rdm_records/records/mappings/os-v1/rdmrecords/records/record-v5.0.0.json
```diff
@@ -769,6 +769,48 @@
             "type": "keyword"
           }
         }
+      },
+      "stats": {
```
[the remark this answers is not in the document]
slint: This is a backwards-compatible change, so no version bump. But in this case, and with the current state of index management commands, it has to be added to the migration recipe as a PUT mapping request.

invenio_rdm_records/records/systemfields/statistics.py
slint: I agree that the logic is a bit scattered at the moment, and I also don't know of any other place where we have a strong opinion/reasoning on `pre_dump` vs. search dumper extension... I guess the latter could be customized more easily in case someone wants to disable e.g. statistics?

On the other hand, I have a stronger feeling on having it done here, since I think `pre_dump` applies to both item/DB and list/Search results. It's very likely that I'm wrong, so not sure if it's worth going down this rabbit-hole.

max-moser: Yeah, unless I'm misunderstanding, that was my thinking as well... The system field's `pre_dump()` might be called from a number of spots that we would need to check exhaustively.
I'm just gonna update the comments here (system field) and at the search dumper extension to point at each other, and leave it at that for now.

invenio_rdm_records/stats/templates/events/record_view/v7/record-view-v1.0.0.json
```diff
@@ -0,0 +1,70 @@
+{
+  "index_patterns": ["__SEARCH_INDEX_PREFIX__events-stats-record-view-*"],
+  "settings": {
+    "index": {
+      "refresh_interval": "5s"
+    }
+  },
+  "mappings": {
+    "date_detection": false,
+    "dynamic": false,
```
ppanero, line 10: why `false` and not `strict`?

With `false` new fields are ignored. These fields will not be indexed or searchable, but will still appear in the `_source` field of returned hits. These fields will not be added to the mapping, and new fields must be added explicitly.
With `strict` If new fields are detected, an exception is thrown and the document is rejected. New fields must be explicitly aded to the mapping.

ppanero, line 10: why `false` and not `strict`?

With `false` new fields are ignored. These fields will not be indexed or searchable, but will still appear in the `_source` field of returned hits. These fields will not be added to the mapping, and new fields must be added explicitly.
With `strict` If new fields are detected, an exception is thrown and the document is rejected. New fields must be explicitly aded to the mapping.

In the later we would basically be notified if something went wrong.

invenio_rdm_records/records/systemfields/statistics.py
```diff
@@ -0,0 +1,75 @@
...
+"""Cached transient field for record statistics."""
+
+from invenio_records.systemfields import SystemField
+from invenio_search.proxies import current_search_client
+from invenio_search.utils import build_alias_name
+
+from ...stats.utils import get_record_stats
+
+
+class RecordStatistics(SystemField):
```
ppanero, line 17: consistency with other sysfields:

```suggestion
class RecordStatisticsField(SystemField):
```

invenio_rdm_records/records/systemfields/statistics.py
```diff
@@ -0,0 +1,75 @@
...
+    """Field for lazy fetching and caching (but not storing) of record statistics."""
+
+    def __init__(self, key=None):
```
ppanero, line 20: Not needed? it's like the base class (parent) default, see https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/systemfields/base.py#L93

invenio_rdm_records/services/schemas/stats.py
```diff
@@ -2,17 +2,28 @@
...
 
-"""RDM record schemas."""
+"""Schema for record statistics."""
 
 from marshmallow import Schema, fields
 
 
-class StatsSchema(Schema):
-    """Files metadata schema."""
+class PartialStatsSchema(Schema):
```
ppanero, line 15: This would be the schema for 1 version right? What about `RecordStatsSchema`? No need for name change, otherwise, just add in the docstring to "what" is it partial.

invenio_rdm_records/stats/event_builders.py
```diff
@@ -0,0 +1,112 @@
...
+    event.update(
+        {
+            # When:
+            "timestamp": datetime.utcnow().isoformat(),
+            # What:
+            "bucket_id": str(obj.bucket_id),
+            "file_id": str(obj.file_id),
+            "file_key": obj.key,
+            "size": obj.file.size,
+            "recid": record.get("id", None),
```
ppanero, line 43: could there be a case of both recird and parent being "None"? Or even just one of them? Even if it is a file related event there would be an associated record? I'm just wondering if we would check if one or both exists actually... It affect other event builders in this file.

ppanero (commented):
I have a concern regarding the file structure and a root `stats` folder. IMO it should not exist and be splited in the correct layers (e.g. `services/stats/permissions.py`). We have other modules (e.g. communities that became quite bad https://github.com/inveniosoftware/invenio-communities/issues/922) I'd like to avoid this here...

invenio_rdm_records/services/schemas/stats.py
max-moser: yes, it's the shared schema between "this version" and "all versions" because both look the same...
i was also toying with the idea of `RecordStatsSchema` but figured it might be confusing which one to pick (`StatsSchema` or `RecordStatsSchema`?) if somebody wanted to import it :sweat_smile:

invenio_rdm_records/stats/event_builders.py
max-moser: tbh, i don't think that either of them being `None` could ever be the case during normal operation...
if it were, chances are that accessing `record.parent` would also fail.
should we change the accessor to `record["id"]` and `record.parent["id"]`?

invenio_rdm_records/stats/templates/events/record_view/v7/record-view-v1.0.0.json
max-moser: tbh, there's no better reason other than me not thinking about it :sweat_smile: 
thanks a lot for the suggestion, i think `strict` is much better!

invenio_rdm_records/services/schemas/stats.py
max-moser: updated the docstring

ntarocco (approved):
LGTM! Just wondering if most of this (but the data model changes) should belong to `invenio-records-resources` instead: is someone else (e.g. `InvenioILS`) in the future wants to re-use this, it will have to be copy/pasted.

max-moser:
@ppanero is it okay for you if we create a follow-up issue for the restructuring and merge this one for now?

@ntarocco let's see how far up the ladder this code can still climb :smile:

ppanero:
@max-moser yes, it is okay for me. But I would really like to make emphasis on that new issue to be tackled in the next sprint. Otherwise things start to diverge, become backwards incompatible and then we cannot refactor the module without biiig change (this is a message for facilitators hehe)

Issue #1231, opened 2023-03-07
Serialize new custom fields


anikachurilova:
Support new custom fields: [journal](https://github.com/inveniosoftware/invenio-rdm-records/issues/1191), [thesis](https://github.com/inveniosoftware/invenio-rdm-records/issues/1219), [notes](https://github.com/inveniosoftware/invenio-rdm-records/issues/1217), [meeting](https://github.com/inveniosoftware/invenio-rdm-records/issues/1211), [imprint](https://github.com/inveniosoftware/invenio-rdm-records/issues/1215) in existing serializers:
- [ ] csl
- [ ] datacite
- [ ] dcat
- [ ] dublincore
- [ ] geojson
- [ ] iiif
- [ ] marcxml
- [ ] bibtex

alejandromumo:
We have two options in the [RFC](https://github.com/inveniosoftware/rfcs/blob/master/rfcs/rdm-0064-custom-fields.md#unresolved-questions):

- Per field
- Per export format 

This topic needs a discussion before the implementation

alejandromumo:
We decided to prototype the option `per field`, using `journal` as an example.

The prototype should also contain multiple formats to have a grasp on how this can grow in terms of code

alejandromumo:
For the prototype, we decided to add a generic post dump pipeline (can also be implemented for other stages of the serialization, e.g. `pre_dump`). 

Each serializer then adds the steps it sees fit for its own needs (e.g. `DataciteJsonSerializer` adds custom fields processors, each defining its own `post_dump` implementation). Everything is then "merged" together (with many caveats)

alejandromumo:
To showcase the implemented mechanism I implemented `journal` serialization into `DataCite`, `MarcXML` and `DublinCore`. Also added `DublinCore` serializartion for `meeting`. 

Since this will scale up (e.g. X formats * Y custom fields), I will first wait until the post_dump/pre_dump mechanism is fully approved before implementing all the combinations of custom fields and formats.

Thus, I am moving the issue to review

PR #1257, opened 2023-03-16
Dublin Core serializer: Transform identifiers to URLs if possible

max-moser:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/1976

Adding prefixes to the `oai_dc` metadata format for OAI-PMH is a requirement we got from [BASE](https://www.base-search.net/).
This PR only touches the Dublin Core serializer in order to be minimally invasive.
It transforms the identifiers by trying to:
* Turn the identifier into a URL if possible (e.g. for DOIs and handles)
* Prefix the identifier with its scheme (e.g. the EAN-13 value `9780521425575` will become `ean13:9780521425575`) if that's not already the case (e.g. `ark:/123/456` will stay the same)


invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -56,11 +77,11 @@ def get_relations(self, obj):
 
         # Alternate identifiers
         for a in obj["metadata"].get("alternate_identifiers", []):
-            rels.append(f"{a['scheme']}:{a['identifier']}")
```
max-moser, line 59: The old logic here would unconditionally append the scheme as a prefix, which looks weird in some cases.
For example, `ark:/123/456` would be turned into `ark:ark:/123/456` which seems redundant to me.

An EAN-13 value of e.g. `9780521425575` would be turned into `ean13:9780521425575` both before and after the PR.

invenio_rdm_records/resources/serializers/dublincore/schema.py
```diff
@@ -39,11 +40,31 @@ class DublinCoreSchema(BaseSchema):
...
+
+        return result
+
     def get_identifiers(self, obj):
         """Get identifiers."""
         items = []
-        items.extend(i["identifier"] for i in obj["metadata"].get("identifiers", []))
-        items.extend(p["identifier"] for p in obj.get("pids", {}).values())
+
+        for id_ in obj["metadata"].get("identifiers", []):
```
ppanero, line 63: Does the order have any impact on the output? In general we try to give priority to `pids` since those fields have much more validation and logic rather than a simple format validation.  However, if the consumer of dc data does not have any implications due to order, then it is all the same. Otherwise, I'd inver the order (first pids, then identifiers).

I know it was potentially wrong before, but we can profit not and fix it hehe

ppanero (approved):

invenio_rdm_records/resources/serializers/dublincore/schema.py
max-moser: sure thing, done :smile:

max-moser:
Updated the PR to control the behavior of `schema._transform_identifier()` via the `schema.context`.
Added a possibility to set the schema's context via the configuration, like:
```python
OAISERVER_METADATA_FORMATS = {
    "oai_dc": {
        "serializer": (
            "invenio_rdm_records.oai:dublincore_etree",
            {
                "schema_context": {
                    "prefix_identifier_schemes": True,
                    "urlize_identifiers": True,
                }
            },
        ),
        "schema": "http://www.openarchives.org/OAI/2.0/oai_dc.xsd",
        "namespace": "http://www.openarchives.org/OAI/2.0/oai_dc/",
    },
}
```
The relevant part here is that the value for `serializer` is a tuple with two entries.
The second entry is used as keyword argument for the function referenced in the first entry.

max-moser:
Ping @alejandromumo 
I couldn't quickly identify code changes related to https://github.com/inveniosoftware/invenio-rdm-records/issues/1231 but saw that you're working on it; could you please check if any changes here (e.g. building & passing the `schema_context`) interfere with that task?

invenio_rdm_records/oai.py
```diff
@@ -26,9 +26,9 @@
 from .services.pids.providers.oai import OAIPIDProvider
 
 
-def dublincore_etree(pid, record):
+def dublincore_etree(pid, record, **serializer_kwargs):
```
max-moser, line 29: In case the `serializer` value for an OAI metadata format in the `OAISERVER_METADATA_FORMATS` config item happens to be a tuple, the second part of that tuple is being passed as a keyword argument to the function referenced by the first part. See: https://github.com/inveniosoftware/invenio-oaiserver/blob/master/invenio_oaiserver/utils.py#L53-L54

invenio_rdm_records/oai.py
```diff
@@ -26,9 +26,9 @@
...
     """Get DublinCore XML etree for OAI-PMH."""
-    json = DublinCoreXMLSerializer().dump_obj(record["_source"])
+    json = DublinCoreXMLSerializer(**serializer_kwargs).dump_obj(record["_source"])
```
max-moser, line 31: So if the second part of the tuple has the shape `{"schema_context": {...}}`, it will be passed to the schema as `context` here via the serializer (see the added passing of `**options` in `invenio_rdm_records/resources/serializers/dublincore/__init__.py` below).

alejandromumo:
> Ping @alejandromumo I couldn't quickly identify code changes related to #1231 but saw that you're working on it; could you please check if any changes here (e.g. building & passing the `schema_context`) interfere with that task?

#1231 is still a WIP, but I don't think the changes you implemented will interfere with that.

max-moser:
ping @slint thoughts & prayers?

slint (approved):
LGTM, and pretty sane for customizing

PR #1258, opened 2023-03-20
serializers: added schema processors

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1231


invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,13 +12,79 @@
...
+from abc import ABC
 from invenio_i18n import lazy_gettext as _
 from invenio_records_resources.services.custom_fields import BaseCF
-from invenio_records_resources.services.records.facets import CFTermsFacet
-from marshmallow import fields
+from marshmallow import fields, missing
 from marshmallow_utils.fields import SanitizedUnicode
 
 
+class PostDumpMixin(ABC):
```
alejandromumo, line 22: To be moved to a more relevant place when the prototype is accepted

invenio_rdm_records/resources/serializers/datacite/__init__.py
```diff
@@ -22,6 +29,7 @@ def __init__(self, **options):
             format_serializer_cls=JSONSerializer,
             object_schema_cls=DataCite43Schema,
             list_schema_cls=BaseListSchema,
+            processors=datacite_post_dump_processors,
```
alejandromumo, line 32: `processors` can be passed directly to the schema, instead of delegating that responsability to `MarshmallowSerializer`. However, that requires a bigger refactor (e.g. in `flask-resources`) 

```suggestion
            format_serializer_cls=JSONSerializer,
            object_schema_cls=DataCite43Schema(processors=datacite_post_dump_processors),
            list_schema_cls=BaseListSchema,
```

invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -168,7 +169,7 @@ class SubjectSchema43(Schema):
     subjectScheme = fields.Str(attribute="scheme")
 
 
-class DataCite43Schema(Schema):
+class DataCite43Schema(BaseSchema):
```
alejandromumo, line 172: In order to have a `post_dump` pipeline, other schemas must inherit from `BaseSchema` as well.

Question: shouldn't `BaseSchema` be higher in the hierarchy? E.g. `flask-resources`.

ping @ppanero  @slint

invenio_rdm_records/resources/serializers/schemas.py
```diff
@@ -79,3 +86,17 @@ def get_contributors(self, obj):
...
+
+    @post_dump(pass_original=True)
+    def post_dump_pipeline(self, data, original, many, **kwargs):
+        """Defines sequence of post dump steps."""
+        result = deepcopy(data)
+        for processor in self.processors:
+            if not hasattr(processor, "post_dump"):
+                raise TypeError("Post process processors must be callables")
+            # TODO does it need to be instantiated?
+            dumped_data = processor().post_dump(result, original)
```
alejandromumo, line 98: Question: how to avoid needing to instantiate a `processor` here? Imo, `processor().post_dump(result, original)` is uglier than `processor.post_dump(result, original)`

invenio_rdm_records/resources/serializers/schemas.py
```diff
@@ -79,3 +86,17 @@ def get_contributors(self, obj):
...
+            merged = _merge_dictionaries(result, dumped_data)
```
alejandromumo, line 99: This is a bit tricky, since we're applying steps in sequence to an already serialized record. 
E.g. order matters, different steps might try to "override" something that was already computed. 
For now, `_merge_dictionaries` only accepts new keys or keys whose value is a list. For the latter, values are simply appended.

invenio_rdm_records/resources/serializers/datacite/schema.py
slint: I'm a bit against the current use/naming of `BaseSchema`. It currently contains "common metadata helper methods" that are used in sub-classing serializer schemas for tansforming/formatting fields.

Now, we want to introduce our "plugin" mixin which adds `post_dump`/`pre_dump` methods, which is more of a configuration/functional/plugin "utility" rather than centralizing common metadata rules. See my comment on the class below on how we could maybe proceed.

---

(separate note for the "common metadata helper methods")

I'm certain also that as we build more format serializers the common metadata rules will diverge a lot and become a bit messy... So in this case I would wait until we first build them, and afterwards deduplicate/centralize logic.

invenio_rdm_records/resources/serializers/schemas.py
slint: I think `.processors` should be a list of already instantiated classes. If we start adding parameters to them, we'll have a hard time passing them individually as e.g. `processor_params`.

invenio_rdm_records/resources/serializers/schemas.py
```diff
@@ -79,3 +86,17 @@ def get_contributors(self, obj):
         return [
             c["person_or_org"]["name"] for c in obj["metadata"].get("contributors", [])
         ] or missing
+
+    @post_dump(pass_original=True)
+    def post_dump_pipeline(self, data, original, many, **kwargs):
+        """Defines sequence of post dump steps."""
+        result = deepcopy(data)
```
slint, line 93: I'm not a big fan of side-effects, but the `post_dump` API for marshmallow kind of assumes it, i.e. `data` is modified and returned from the function. In that sense, we can follow the same principle and just re-assign `data` on each processor call.

invenio_rdm_records/resources/serializers/schemas.py
```diff
@@ -79,3 +86,17 @@ def get_contributors(self, obj):
...
+        for processor in self.processors:
+            if not hasattr(processor, "post_dump"):
+                raise TypeError("Post process processors must be callables")
```
slint, line 96: I would instead make a check in the `__init__` like:

```python
    def __init__(self, processors=None, **kwargs):
        super().__init__()
        assert all(isinstance(p, DumpProcessorMixin) for p in processors)
        self.processors = processors or []
```

Then by default I would make the `DumpProcessorMixin` implement an "identity" `post_dump`/`pre_dump` (i.e. `lambda x: x`)

invenio_rdm_records/resources/serializers/schemas.py
slint: I would avoid this merging logic by assuming that the processors receive the full object and are responsible for resolving order/key conflicts in the data.

slint (commented):
Great job, some initial comments, but it's going in the right direction IMHO

invenio_rdm_records/contrib/journal/custom_fields.py
alejandromumo: Suggestions for the placement of `PostDumpMixin`?

Also, to keep the file clean should we move the processors to their own file? E.g. `invenio_rdm_records/contrib/journal/serializer.py` or similar?

invenio_rdm_records/resources/serializers/datacite/schema.py
alejandromumo: > I'm a bit against the current use/naming of `BaseSchema`. It currently contains "common metadata helper methods" that are used in sub-classing serializer schemas for tansforming/formatting fields.
> ## Now, we want to introduce our "plugin" mixin which adds `post_dump`/`pre_dump` methods, which is more of a configuration/functional/plugin "utility" rather than centralizing common metadata rules.
> 
> (separate note for the "common metadata helper methods")
> 
> I'm certain also that as we build more format serializers the common metadata rules will diverge a lot and become a bit messy... So in this case I would wait until we first build them, and afterwards deduplicate/centralize logic.

In that case, are you suggesting to duplicate the logic in all the schemas we currently have and then deduplicate it when we figure out a better structure for schemas/serializers?

invenio_rdm_records/resources/serializers/schemas.py
```diff
@@ -6,12 +6,25 @@
...
+    def __init__(self, processors=None, **kwargs):
+        """Initialize a new instance of the BaseSchema class.
+
+        :param processors: A list of `DumpProcessorMixin` instances used to modify the serialized output, defaults to None.
+        :type processors: list, optional
+        :raises AssertionError: If any item in the `processors` list is not an instance of `DumpProcessorMixin`.
+        """
+        super().__init__()
+        self.processors = processors or []
+        assert all(isinstance(p, DumpProcessorMixin) for p in self.processors)
```
ppanero, line 26: this is more of a moralistic quesiton: where is the threshold on using asserts instead of typing? my point being, if we do not use typing we assume the contract/programmatic API is respected/implemented. Then we do not have assertions. Otherwise, we should be consistent and assert all types of constuctor attributes (i.e. use typing)

In some very specific cases it makes sense to do assertions, what would be the reasoning behind this case?

invenio_rdm_records/resources/serializers/datacite/schema.py
ppanero: While I do agree on the `BaseClass` naming, we did this to deduplicate code from 4 serializers. If in the future the deduplicate we can re-think it. There can be other ways (mixins, etc.) I'll leave the choice of method to you. But I'd not go to duplicate code...

invenio_rdm_records/resources/serializers/schemas.py
ntarocco: Nice question!

Just my 2 cents, feel free to disagree: I use `assert` when I expect an internal object to be as expected. If that is not the case, it means something is terribly broken in the system, internally.
For example, when I fetch a published record from the DB, I can assert that it does not have a `review`, see [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/61d72f0e8f6c9e1c864548b61313a791e490b406/invenio_rdm_records/requests/community_inclusion.py#L42).

For value/type checking, I would prefer clear error msgs (you can do `assert <condition>, msg`), with an `if` with a `raise TypeError`, or even better, a specific excp. **Why?** Because a specific raised exception can then bubble up and can be caught by a `try/except`, and the code could recover in certain cases or behave/raise differently.

The other very minor difference is that assertions [might be disabled in Python if the code is optimized](https://discuss.python.org/t/stop-ignoring-asserts-when-running-in-optimized-mode/13132/41?page=3).

Honestly, in this specific case, I think it is fine to use `assert`: if the type is wrong, you cannot do much, it is an unrecoverable Internal Server Error.

invenio_rdm_records/resources/serializers/datacite/schema.py
ntarocco: > In order to have a `post_dump` pipeline, other schemas must inherit from `BaseSchema` as well.

Why is that? The [BaseSchema](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/schemas.py#L12) class only contains a few helper methods at the moment. Joining what @slint said, it now feels like more a helper, utility, a plugin, than a base class (which maybe made sense before!).

If I interpret well what Alex suggests, that should be converted to a Mixin or injected via contructor (if possible) to compose functionalities, and the convert the children to:
```
class ASchema(Schema, CommonFieldsMixin)
```

> Question: shouldn't `BaseSchema` be higher in the hierarchy? E.g. `flask-resources`.
Maybe I am not looking at the right BaseSchema class, that's why I have doubts with my question above, but it contains a few fields of the RDM data model, so it should stay here.

ppanero: +1 to CommonFieldsMixin

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,13 +12,151 @@
...
-from invenio_records_resources.services.records.facets import CFTermsFacet
-from marshmallow import fields
+from marshmallow import fields, missing
 from marshmallow_utils.fields import SanitizedUnicode
 
 
+class DumpProcessorMixin(ABC):
+    """Mixin class that defines pre_dump and post_dump methods to extend records serialization."""
+
+    def post_dump(self, data, original={}, **kwargs):
```
ppanero, line 26: major: do not use mutable objects in func parameters

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,13 +12,151 @@
...
+        """Hook called after the marshmallow serialization of the record.
+
+        :param data: The dumped record data.
+        :param original: The original record data.
+        :param kwargs: Additional keyword arguments.
+        :returns: The serialized record data.
+        """
+        return data
+
+    def pre_dump(self, data, original={}, **kwargs):
```
ppanero, line 36: same as above

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,13 +12,151 @@
...
+        :param kwargs: Additional keyword arguments.
+        :returns: The data to dump.
+        """
+        return data
+
+
+class JournalDataciteProcessor(DumpProcessorMixin):
+    """Dump processor for datacite serialization of 'Journal' custom field."""
+
+    def post_dump(self, data, original={}, **kwargs):
```
ppanero, line 51: same, check all ocurrences of `={}` there might be more

invenio_rdm_records/resources/serializers/schemas.py
ppanero: I'm consistently a pain regarding consistency haha

While I agree the it is nice to have a check to see if they implement the correct classes, the `processors` list comes form config (i.e. is not like a system check). Thus we are doing "parameter type checking", if we want that we should use Python's typing... Otherwise, implement asserts everywhere, for example:

- We should check custom fields implement BaseCF
- We should check PIDProviders implement PIDProvider
- a looong etc of classes or list of them that come from config (invenio.cfg) and can be changed by the user

I'm not against leaving the `assert`. I'm having a hard time understanding why in this case is needed but not in all the other one where a class or list of them come from config, so I would not know how to proceed if I had to code something similar.

>  The other very minor difference is that assertions might be disabled in Python if the code is optimized.

Nice, didnt know! Thanks for sharing.

invenio_rdm_records/contrib/journal/custom_fields.py
alejandromumo: In this case, the original might (or not) be passed by the schema dumper. In marshmallow that is controlled with the `pass_original=True` parameter. That's why I added a default there, to at least make sure it was an object if not passed (and `original.get()` does not fail).

invenio_rdm_records/resources/serializers/bibtex/schema_formats.py
```diff
@@ -81,7 +81,7 @@ def publication_article():
         name = "article"
         req_fields = ["author", "title", "journal", "year"]
         opt_fields = ["volume", "number", "pages", "month", "note"]
-        return format(name, req_fields, opt_fields)
+        return BibTexFormatter.format(name, req_fields, opt_fields)
```
alejandromumo, line 84: Fixed a small issue where `format` was the in-built format method and not `BibTexFormatter.format`

invenio_rdm_records/resources/serializers/schemas.py
```diff
@@ -79,3 +79,100 @@ def get_contributors(self, obj):
...
+
+        :returns: The result of the pipeline processing on the serialized data.
+        """
+        for dumper in self.dumpers:
+            # Data is assumed to be modified and returned by the dumper
+            data = dumper.pre_dump(data)
+        return data
+
+
+class LoaderMixin(Schema):
```
alejandromumo, line 135: Not used by now but since it's the same as the Dumper counter part I decided to add it straightaway

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+    def pre_load(data, many=False, **kwargs):
+        raise NotImplementedError()
+
+
+class JournalDataciteDumper(FieldDumper):
+    """Dump processor for datacite serialization of 'Journal' custom field."""
+
+    def post_dump(self, data, original={}, **kwargs):
+        """Adds the journal information as a related identifier in the DataCite metadata."""
+        custom_fields = original.get("custom_fields", {})
```
max-moser, line 62:
```suggestion
        custom_fields = (original or {}).get("custom_fields", {})
```
This would enable you to set `original=None` in the method's argument list above.

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+
+        return data
+
+
+class JournalDublinCoreDumper(FieldDumper):
+    """Dump processor for dublin core serialization of 'Journal' custom field."""
+
+    def post_dump(self, data, original={}, **kwargs):
+        """Adds serialized journal data to the input data under the 'sources' key."""
+        custom_fields = original.get("custom_fields", {})
```
max-moser, line 95:
```suggestion
        custom_fields = (original or {}).get("custom_fields", {})
```

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+        journal_data = custom_fields.get("journal:journal", {})
+        title = journal_data.get("title")
+        volume = journal_data.get("volume")
+        issue = journal_data.get("issue")
+        pages = journal_data.get("pages")
+        year = original.get("metadata", {}).get("publication_date")
```
max-moser, line 101:
```suggestion
        year = (original or {}).get("metadata", {}).get("publication_date")
```

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+
+        return data
+
+
+class JournalMarcXMLDumper(FieldDumper):
+    """Dump processor for MarcXML serialization of 'Journal' custom field."""
+
+    def post_dump(self, data, original={}, **kwargs):
+        """Adds serialized journal data to the input data under the '773' key."""
+        custom_fields = original.get("custom_fields", {})
```
max-moser, line 130:
```suggestion
        custom_fields = (original or {}).get("custom_fields", {})
```

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+
+        issue = journal_data.get("issue")
+        if issue:
+            items_dict["n"] = issue
+
+        pages = journal_data.get("pages")
+        if pages:
+            items_dict["c"] = pages
+
+        year = original.get("metadata", {}).get("publication_date")
```
max-moser, line 154:
```suggestion
        year = (original or {}).get("metadata", {}).get("publication_date")
```

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+        code = "773  "
+
+        data[code] = items_dict
+        return data
+
+
+class JournalCSLDumper(FieldDumper):
+    def post_dump(self, data, original={}, **kwargs):
+        # TODO
+        custom_fields = original.get("custom_fields", {})
```
max-moser, line 172:
```suggestion
        custom_fields = (original {}).get("custom_fields", {})
```

invenio_rdm_records/contrib/meeting/custom_fields.py
```diff
@@ -18,12 +18,37 @@
...
+from invenio_rdm_records.contrib.journal.custom_fields import FieldDumper
 from invenio_rdm_records.services.schemas.metadata import _valid_url
 
 
+class MeetingDublinCoreDumper(FieldDumper):
+    """Dump processor for dublin core serialization of 'Meeting' custom field."""
+
+    def post_dump(self, data, original={}, **kwargs):
+        """Adds serialized meeting data to the input data under the 'sources' key."""
+        custom_fields = original.get("custom_fields", {})
```
max-moser, line 33:
```suggestion
        custom_fields = (original or {}).get("custom_fields", {})
```

invenio_rdm_records/contrib/journal/custom_fields.py
max-moser:
```suggestion
    def post_dump(self, data, original=None, **kwargs):
```

invenio_rdm_records/contrib/journal/custom_fields.py
max-moser:
```suggestion
    def pre_dump(self, data, original=None, **kwargs):
```

invenio_rdm_records/contrib/journal/custom_fields.py
max-moser:
```suggestion
    def post_dump(self, data, original=None, **kwargs):
```

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+        related_identifiers.append(serialized_journal)
+        data["relatedIdentifiers"] = related_identifiers
+
+        return data
+
+
+class JournalDublinCoreDumper(FieldDumper):
+    """Dump processor for dublin core serialization of 'Journal' custom field."""
+
+    def post_dump(self, data, original={}, **kwargs):
```
max-moser, line 93:
```suggestion
    def post_dump(self, data, original=None, **kwargs):
```

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+        sources.append(serialized_data)
+        data["sources"] = sources
+
+        return data
+
+
+class JournalMarcXMLDumper(FieldDumper):
+    """Dump processor for MarcXML serialization of 'Journal' custom field."""
+
+    def post_dump(self, data, original={}, **kwargs):
```
max-moser, line 128:
```suggestion
    def post_dump(self, data, original=None, **kwargs):
```

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+        # TODO in zenodo, journal field serializes to
+        # TODO code 909, C, 4 but that's not in the specification
+        code = "773  "
+
+        data[code] = items_dict
+        return data
+
+
+class JournalCSLDumper(FieldDumper):
+    def post_dump(self, data, original={}, **kwargs):
```
max-moser, line 170:
```suggestion
    def post_dump(self, data, original=None, **kwargs):
```

invenio_rdm_records/contrib/meeting/custom_fields.py
```diff
@@ -18,12 +18,37 @@
...
 from marshmallow_utils.fields import SanitizedUnicode
 
+from invenio_rdm_records.contrib.journal.custom_fields import FieldDumper
 from invenio_rdm_records.services.schemas.metadata import _valid_url
 
 
+class MeetingDublinCoreDumper(FieldDumper):
+    """Dump processor for dublin core serialization of 'Meeting' custom field."""
+
+    def post_dump(self, data, original={}, **kwargs):
```
max-moser, line 31:
```suggestion
    def post_dump(self, data, original=None, **kwargs):
```

invenio_rdm_records/resources/serializers/bibtex/__init__.py
```diff
@@ -22,6 +25,7 @@ def __init__(self, **options):
             format_serializer_cls=SimpleSerializer,
             object_schema_cls=BibTexSchema,
             list_schema_cls=BaseListSchema,
+            schema_kwargs={"dumpers": bibtex_dumpers},
```
max-moser, line 28: note: the mutable values here are fine, because they're used as arguments for _calling_ a function; the issue is only when mutable objects are being used as the default values for function parameters (i.e. when defining a function)

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+
+        :param data: The record data to dump.
+        :param original: The original record data.
+        :param kwargs: Additional keyword arguments.
+        :returns: The data to dump.
+        """
+        return data
+
+
+class FieldLoader(ABC):
```
max-moser, line 49: Missing docstrings

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+            volume = f"{volume}({issue})"
+
+        if not volume:
+            volume = issue
+
+        parts = [
+            title,
+            volume,
+            pages,
+            "({0})".format(year) if year else None,
```
max-moser, line 113:
```suggestion
            f"({year})" if year else None,
```
Absolutely not necessary, just a personal preference because I think it's a bit easier on the eye

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,181 @@
...
+
+        # TODO in zenodo, journal field serializes to
+        # TODO code 909, C, 4 but that's not in the specification
+        code = "773  "
+
+        data[code] = items_dict
+        return data
+
+
+class JournalCSLDumper(FieldDumper):
```
max-moser, line 169: Missing docstrings

invenio_rdm_records/resources/serializers/schemas.py
max-moser: Just commenting to say that I heard typing and got interested.
I understand that this is/was something that was avoided so far, because the system was evolving too fast to be worth the effort... And if they get outdated regularly, they become unreliable and actually more harmful than beneficial.
Anyway, given that most of InvenioRDM's code is getting more and more stable (or at least that's what it feels like to me), would it make sense to reevaluate typing annotations?

invenio_rdm_records/contrib/meeting/custom_fields.py
ntarocco: This needs to be fixed :)
and add below:
```
_original = original or {}
```

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,199 @@
...
+        _original = original or {}
+        custom_fields = _original.get("custom_fields", {})
+        journal_data = custom_fields.get("journal:journal", {})
+
+        if not journal_data:
+            return data
+
+        items_dict = {}
+
+        title = journal_data.get("title")
```
ppanero, line 154: can we collapse the 4 ifs? 😇 

```python
field_keys = {p: title, v: volume, n: issue, c: pages}
for key, field in field_keys.items():
    value = journal_data.get(field)
    if value:
        items_dict[key] = value
```

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,199 @@
...
+        """Adds serialized journal data to the input data."""
+        _original = original or {}
+        custom_fields = _original.get("custom_fields", {})
+        journal_data = custom_fields.get("journal:journal", {})
+        title = journal_data.get("title")
+        volume = journal_data.get("volume")
+        issue = journal_data.get("issue")
+        pages = journal_data.get("pages")
+
+        if not journal_data:
```
ppanero, line 197: can return earlier (line 192) and save a few ops.

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,199 @@
...
+            return data
+
+        data["container_title"] = title
```
ppanero, line 200: what if they are `None`. Do we want those fields there? or only those with values?

invenio_rdm_records/resources/serializers/schemas.py
```diff
@@ -79,3 +79,100 @@ def get_contributors(self, obj):
...
+    def __init__(self, dumpers=None, **kwargs):
+        """Initialize a new instance of the `DumperMixin` class.
+
+        :param dumpers: A list of `DumpProcessorMixin` instances used to modify the serialized output, defaults to None.
+        :type dumpers: list, optional
+        :raises AssertionError: If any item in the `dumpers` list is not an instance of `DumpProcessorMixin`.
+        """
+        super().__init__(**kwargs)
+        self.dumpers = dumpers or []
+        assert all(isinstance(p, FieldDumper) for p in self.dumpers)
```
ppanero, line 99: I've already commented on this and is not clear to me how do we want to proceed. This is parameter field checking, not business logic. What was the consensus? I'd like someone to tell me if I should stop bringing this up 😇 

ref https://github.com/inveniosoftware/invenio-rdm-records/pull/1258#discussion_r1143114019

cc @kpsherva @zzacharo  this might be worth a discussion

invenio_rdm_records/resources/serializers/schemas.py
```diff
@@ -79,3 +79,100 @@ def get_contributors(self, obj):
         return [
             c["person_or_org"]["name"] for c in obj["metadata"].get("contributors", [])
         ] or missing
+
+
+class DumperMixin(Schema):
```
ppanero, line 84: Seeing this I think the `FieldDumper` and `FieldLoader` classes should be ABC (interfaces). i.e. remove the _Mixin_ word from docstring and have methods be `@abstractmethod` when it makes sense.

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,199 @@
...
+        issue = journal_data.get("issue")
+        pages = journal_data.get("pages")
+        year = _original.get("metadata", {}).get("publication_date")
+
+        if volume and issue:
+            volume = f"{volume}({issue})"
+
+        if not volume:
+            volume = issue
+
```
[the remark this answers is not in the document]
alejandromumo: Yes `if x` does the trick there. Same happens for the year part.

invenio_rdm_records/contrib/journal/custom_fields.py
```diff
@@ -12,14 +12,199 @@
...
+class FieldLoader(ABC):
+    """Mixin class that defines pre_load and post_load methods to extend records deserialization."""
+    def post_load(data, **kwargs):
+        """Hook called after the marshmallow deserialization of the record.
+
+        :param data: The loaded record data, already deserialized.
+        :param kwargs: Additional keyword arguments.
+        :returns: The deserialized record data.
+        """
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
alejandromumo: You're right, it's inconsistent.

Both should be a mixin that return the data by default.

invenio_rdm_records/resources/serializers/schemas.py
alejandromumo: I keeped it after that discussion. As @ntarocco mentioned, that's an unrecoverable error for the system.

I am not a big fan of asserts in the code either but seemed ok in this case.

ppanero: Discussed IRL with Karolina and Nico. We should remove this checks.

While I do agree that it is _an unrecoverable error for the system_, it is an error due to the type/class of the object, not due to the business logic. It would be the same if we pass an object that is not a custom field in the config, or an object that is not a serializer (long etc.).

A different case is the one mentioned in the transformer/queryparser PR. There we need to make sure that a series of preconditions/business logic is fulfilled (i.e. two parameters that are dependant are provided). In those cases we should put an `assert`.

ppanero:
I went through all the comments, there are many and it could be that some slipped through (they are not marked as _outdated_, nor resolved, and some are still present).

TL;DR:
- Read through the non-resolved comments
- Remove the `assert`s for type checking
- Move `LoaderMixin` and `DumperMixing` to the root of custom fields (i.e. services) or even to invenio-records-resources since communities might need this too (check for other potential generic classes, there might be more)

tests/conftest.py
```diff
@@ -500,14 +500,6 @@ def full_record(users):
                 }
             ],
         },
-        "ext": {
```
ppanero, line 503: old content from ext (old custom fields)

tests/services/pids/providers/test_datacite_pid_provider.py
```diff
@@ -47,7 +47,7 @@ def record_w_links(running_app, minimal_record):
     draft = service.create(system_identity, minimal_record)
     record = service.publish(system_identity, draft.id)
 
-    return record
+    return record.to_dict()
```
ppanero, line 50: We were passing wrong types to the service provider. This was discovered due to the `_original` object of the post_dump processors

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -202,6 +203,8 @@ def get_types(self, obj):
         t = props.get("eurepo")
         return [t] if t else missing
 
+    # FIXME should this be a processor? that's somehow the post_dump place
```
ppanero, line 206: WDYT?

invenio_rdm_records/resources/serializers/csl/__init__.py
```diff
@@ -36,6 +37,7 @@ def __init__(self, **options):
             format_serializer_cls=JSONSerializer,
             object_schema_cls=CSLJSONSchema,
             list_schema_cls=BaseListSchema,
+            schema_kwargs={"dumpers": [JournalCSLDumper()]},  # Order matters
```
ppanero, line 40: When tackling the refactoring this would be made configurable, because the dumpers should only be present if the fields are enabled. At the moment this fields are enabled in ZenodoRDM.

ppanero:
Tests are passing with the latest flask-resources

Screenshot 2023-04-05 at 11 13 06

PR #1260, opened 2023-03-21
preferences: add support for preferences.locale

TLGINO:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2069

### DIFF
--- a/invenio_rdm_records/utils.py
+++ b/invenio_rdm_records/utils.py
@@ -7,6 +7,7 @@
 
 """Utility functions."""
 
+from flask import current_app
 from flask_security.confirmable import confirm_user
 from flask_security.utils import hash_password
 from invenio_accounts.proxies import current_datastore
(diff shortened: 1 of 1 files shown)


invenio_rdm_records/utils.py
```diff
@@ -23,7 +23,7 @@ def get_or_create_user(email):
             username=email.split("@")[0],
             password=hash_password("123456"),
             active=True,
-            preferences=dict(visibility="public", email_visibility="public"),
+            preferences=dict(visibility="public", email_visibility="public", preferences_locale='ch'),
```
ntarocco, line 26: you could eventually omit this one, it is used in the fixtures: it will use the default.
```suggestion
            preferences=dict(visibility="public", email_visibility="public", locale="en_GB"),
```

invenio_rdm_records/utils.py
```diff
@@ -23,7 +23,12 @@ def get_or_create_user(email):
             username=email.split("@")[0],
             password=hash_password("123456"),
             active=True,
-            preferences=dict(visibility="public", email_visibility="public"),
+            preferences=dict(
+                visibility="public",
+                email_visibility="public",
+                locale="en_GB",
```
alejandromumo, line 29: I suppose this only used to create users via cli or the admin user, but the locale and timezone are supposed to be hardcoded? 
Locale is set to `en` in other places, is there any reason why it's set to `en_GB` here?

TLGINO: Fixed thx!

alejandromumo (commented):
LGTM :)

However tests are failing, I suppose due to changes in `invenio-db`?

TLGINO:
> LGTM :)
> 
> However tests are failing, I suppose due to changes in `invenio-db`?

They all run localy for me (need to import the other moduels though to pass tests etc)
Will follow with a pic of all tests passing soon

PR #1264, opened 2023-03-27
service: add records request search

kpsherva:
* resources: add records request search endpoint
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2075


invenio_rdm_records/resources/resources.py
```diff
@@ -222,6 +222,38 @@ def remove(self):
...
+    @request_view_args
+    @response_handler(many=True)
+    def search(self):
+        """Search for record's communities."""
+        items = self.service.search(
+            identity=g.identity,
+            params=resource_requestctx.args,
+            record_pid=resource_requestctx.view_args["record_pid"],
+            search_preference=search_preference(),
+            expand=resource_requestctx.args.get("expand", True),
```
jrcastro2, line 251: Moderate: I think that you are missing to add the expand to to the search decorators. In records resources we do add it with the [request_extra_args](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/resources/records/config.py#L40).

jrcastro2 (commented):
It would be nice if we could add tests for this (could be done in another PR)

invenio_rdm_records/services/requests/service.py
```diff
@@ -0,0 +1,51 @@
...
+        self,
+        identity,
+        record_pid,
+        params=None,
+        search_preference=None,
+        expand=False,
+        extra_filter=None,
+        **kwargs
+    ):
+        """Search for record's communities."""
```
kpsherva, line 35: missing record permission check

ntarocco: looks good!

For the future: now thinking about this, we might want to override all upper class method, and add a `raise NotImplementedError`...

Issue #1268, opened 2023-03-30
Statistics: Restructure the code for consistency


max-moser:
The integration of the statistics code in a new top-level module called `stats` does not necessarily adhere to common modern practice of Invenio modules, which usually follows a separation between the various layers (services, etc.).
This should be refactored before it gets out of hand.

A similar issue is already open over at [`Invenio-Communities`](https://github.com/inveniosoftware/invenio-communities/issues/922), where this refactoring hasn't taken place yet.

PR #1269, opened 2023-03-31
Restructuring of statistics integration code

max-moser:
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1268

Based on a brief discussion with @ppanero.
I think that covers most of the suggestions except for the `records/stats/api.py` with the pseudo API class (a very simplistic data class that just has a `get_record_stats()` method similar to `Record.get_record()`) as a replacement for `records/stats/utils.py:get_record_stats()`.
Sorry, but I'm out of steam for that today :sweat_smile:


invenio_rdm_records/services/stats/permissions.py
```diff
@@ -47,11 +43,12 @@ def can(self):
         )
 
 
-def default_deny_permission_factory(query_name, params):
-    """Default deny permission factory.
+def permissions_policy_lookup_factory(query_name, params):
```
max-moser, line 46: Actually, the name has been inaccurate ever since I've built in the permission translation class, sorry about that! :sweat_smile:

zzacharo (approved):
LGTM but @ppanero can you also have a look that all the points from your discussion with @max-moser have been addressed?

invenio_rdm_records/records/stats/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+# Copyright (C) 2023 TU Wien.
+#
+# Invenio RDM Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Statistics integration for RDM records."""
+
+from .api import Statistics
+
+__all__ = Statistics
```
kpsherva, line 14:
```suggestion
__all__ = "Statistics"
```

ppanero: While I do agree, it is inconsistent accross modules:
- In some we have a list [] in some a set ()
- In some we quote, in the latest we don't


I can made the change you request, but black formatting removes it when there is only one item 🤷‍♂️

kpsherva: nit, not a blocker

Issue #1276, opened 2023-04-17
Move deposit components


kpsherva:
details: https://codimd.web.cern.ch/v_uudxuhSMup3JDjzdXzzQ

PR #1277, opened 2023-04-17
assets: move deposit components from react-invenio-deposit

kpsherva:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1276


.github/workflows/tests.yml
```diff
@@ -83,3 +90,15 @@ jobs:
       - name: Run tests
         run: |
           ./run-tests.sh
+
+      - name: Install deps for frontend tests
+        working-directory: ./invenio_administration/assets/semantic-ui/js/invenio_administration
```
alejandromumo, line 95: Is this directory correct (`invenio_administration`)?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/index.js
```diff
@@ -0,0 +1,7 @@
+// This file is part of InvenioAdministration
```
alejandromumo, line 1:
```suggestion
// This file is part of Invenio-RDM-Records
```

There are other occurrences of this in the PR

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/customFields/CustomField.js
```diff
@@ -0,0 +1,68 @@
...
+    deserializedDefault = null,
+    serializedDefault = null,
+    allowEmpty = false,
+    vocabularyFields = [],
+  }) {
+    super({ fieldpath, deserializedDefault, serializedDefault, allowEmpty });
+    this.vocabularyFields = vocabularyFields;
+  }
+
+  #mapCustomFields(record, customFields, mapValue) {
```
alejandromumo, line 26: Not added in this PR but the name starting with `#` triggered my curiosity.

alejandromumo (commented):
LGTM.

minor: there are multiple headers outdated (e.g. still mentioning `react-invenio-deposit`). I also realised we do not have a lot of documentation (jsdocs) on many react components, however it is just a thought (it's completely out of scope from this issue).

major: tests are failing since the `MANIFEST.in` needs to be updated. 
major: gh actions is complaining, I suppose it's due to formatting / linting issues. But let's see if tests pass after the manifest is updated.

Note: I did not review every component's code, as that was already done in `react-invenio-deposit` (and we're juts moving code). I checked however that the file structure is the correct one to be picked by the alias `@js/invenio_rdm_records` and other minor changes to `invenio-rdm-records` itself (e.g. gh workflows)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/customFields/CustomField.js
kpsherva: its a private class attr (if that is what you are asking about): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

.github/workflows/tests.yml
```diff
@@ -52,6 +52,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Use Node.js ${{ matrix.node-version }}
```
ntarocco, line 55: Same here missing `node-version`

ntarocco (approved):

Issue #1279, opened 2023-04-18
(TO BE DECLINED) Permissions: introduce a permission for restricted files


anikachurilova:
**NOTE**: will be moved to zenodo-rdm:  https://github.com/zenodo/zenodo-rdm/issues/271
- - `can_community_access = False`:
    - record owner - access
    - community owner - no access
    - random user - no access
- `can_community_access = True`:
    - record owner - access
    - community owner - access
    - random user - no access

Files not restricted for community - old permission logic.
Files restricted - only record owner can access.

See: https://github.com/inveniosoftware/invenio-rdm-records/issues/1274

PR #1280, opened 2023-04-18
permissions: manage permissions depending on new permission flag

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1279

Feel free to suggest better namings :heart:

### DIFF
--- a/invenio_rdm_records/services/generators.py
+++ b/invenio_rdm_records/services/generators.py
@@ -117,6 +117,25 @@ class IfRestricted(ConditionalGenerator):
             return q_public
 
 
+class IfFilesRestrictedForCommunity(IfRestricted):
+    """Conditional generator for files restriction for community."""
+
+    def __init__(self, then_, else_):
+        """Constructor."""
+        super().__init__("files", then_, else_)
+
+    def _condition(self, record, **kwargs):
+        """Check if community can access restricted files of the migrated record."""
+        # can_community_read_files = getattr(record.parent, permission_flags.can_community_read_files, True)
+        can_community_read_files = False
+        is_restricted = super()._condition(record, **kwargs)
+
+        if is_restricted:
+            return not can_community_read_files
+        else:
+            return False
+
+
 class IfDraft(ConditionalGenerator):
     """Generator that depends on whether the record is a draft or not.
 
--- a/invenio_rdm_records/services/permissions.py
+++ b/invenio_rdm_records/services/permissions.py
@@ -22,6 +22,7 @@ from invenio_records_permissions.policies.records import RecordPermissionPolicy
 from .generators import (
     IfConfig,
     IfFileIsLocal,
+    IfFilesRestrictedForCommunity,
     IfRestricted,
     RecordCommunitiesAction,
     RecordOwners,
@@ -82,8 +83,16 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
         IfRestricted("record", then_=can_view, else_=can_all),
     ]
     # Allow reading the files of a record
-    can_read_files = [
+    # Hide record's files from community
+    can_read_files_owner = [RecordOwners(), SecretLinks("view"), SystemProcess()]
+    # Show record's files to community
+    can_read_files_community = [
         IfRestricted("files", then_=can_view, else_=can_all),
+    ]
+    can_read_files = [
+        IfFilesRestrictedForCommunity(
+            then_=can_read_files_owner, else_=can_read_files_community
+        ),
     ]
     can_get_content_files = [
         # note: even though this is closer to business logic than permissions,
--- a/tests/resources/test_record_file_permissions.py
+++ b/tests/resources/test_record_file_permissions.py
@@ -12,11 +12,11 @@
 Not every case is tested, but enough high-level ones for it to be useful.
 """
 
-
 from io import BytesIO
 
 import pytest
 
+from invenio_rdm_records.proxies import current_rdm_records_service
 from tests.helpers import login_user, logout_user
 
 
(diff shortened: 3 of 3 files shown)


invenio_rdm_records/services/generators.py
```diff
@@ -117,6 +117,21 @@ def query_filter(self, **kwargs):
             return q_public
 
 
+class IfRestrictedForCommunity(IfRestricted):
```
alejandromumo, line 120: minor: the name is a bit ambiguous, it is not clear what is restricted for the community.

```suggestion
class IfFilesRestrictedForCommunity(IfRestricted):
```

However I also understand that we're using like `IfFilesRestrictedForCommunity("files")` and "files" becomes redundant.

invenio_rdm_records/services/generators.py
```diff
@@ -117,6 +117,21 @@ def query_filter(self, **kwargs):
...
+    """Conditional generator for files restriction for community."""
+
+    def _condition(self, record, **kwargs):
+        """Check if community can access restricted files of the migrated record."""
+        # can_community_read_files = getattr(record.parent, permission_flags.can_community_read_files, True)
```
alejandromumo, line 125: is this a leftover?

invenio_rdm_records/services/permissions.py
```diff
@@ -82,9 +83,15 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
         IfRestricted("record", then_=can_view, else_=can_all),
     ]
     # Allow reading the files of a record
-    can_read_files = [
+    can_read_files_owner = [RecordOwners(), SecretLinks("view"), SystemProcess()]
+    can_read_files_community = [
```
alejandromumo, line 87: I am a bit confused by reading the code.

`can_read_files_community` -> `IfRestricted("files", ...)`
`can_read_files` -> `IfRestrictedForCommunity`("files", ...)

The logic seems to be correct based on the issue's description though.
Maybe a small code comment could increase the readabilty.

invenio_rdm_records/services/permissions.py
```diff
@@ -82,9 +83,15 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
         IfRestricted("record", then_=can_view, else_=can_all),
     ]
     # Allow reading the files of a record
-    can_read_files = [
+    can_read_files_owner = [RecordOwners(), SecretLinks("view"), SystemProcess()]
```
alejandromumo, line 86: question: why do we add `SecretLinks("view")`? 

I lack a lot of context on permissions so I am just curious 😄

tests/resources/test_record_file_permissions.py
```diff
@@ -72,6 +73,105 @@ def record_w_restricted_file(client, headers, running_app, minimal_record, users
...
+    client,
+    uploader,
+    record_community,
+    full_record,
+    community_owner,
+    headers,
+    community,
+    minimal_record,
+    db,
+):
```
alejandromumo, line 87: missing docs for the test and its methods.

alejandromumo (commented):
Left some minor comments / questions, mostly naming. 

Thanks for the tests 🚀 Nice job!

invenio_rdm_records/services/permissions.py
anikachurilova: `can_read_files_community` keeps the same permission logic as it was before and is called in `else_` of `can_read_files` which will be valuated in case the community has access to the files of the record.
I understand can be confusing, will add the comment!

invenio_rdm_records/services/permissions.py
alejandromumo: Ah this is about secret links 😅 that's what I wasn't understanding.

My question was poorly formulated, sorry!  I understand now what this is about :)

invenio_rdm_records/services/permissions.py
```diff
@@ -82,9 +83,17 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
...
     # Allow reading the files of a record
-    can_read_files = [
+    # Hide record's files from community
+    can_read_files_owner = [RecordOwners(), SecretLinks("view"), SystemProcess()]
+    # Show record's files to community
+    can_read_files_community = [
         IfRestricted("files", then_=can_view, else_=can_all),
     ]
+    can_read_files = [
+        IfRestrictedForCommunity(
```
ntarocco, line 93: open question for reviewers (we discussed with it with Anika):

Should all this go to ZenodoRDM only? it is possible thanks to the config var [RDM_PERMISSION_POLICY](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/config.py#L105). However, you will have to main the list of permissions updated in ZenodoRDM if changes.

@slint do you prefer to have this only in ZenodoRDM?

tests/resources/test_record_file_permissions.py
```diff
@@ -72,6 +72,99 @@ def record_w_restricted_file(client, headers, running_app, minimal_record, users
...
+    """Test the access to restricted files of a record for record owner, community owner and a simple user ."""
+    service = current_rdm_records_service
+    record_owner = uploader.login(client)
+
+    # publish a record with restricted files
+    minimal_record["access"]["files"] = "restricted"
+    recid = create_record_w_file(record_owner, minimal_record, headers)
+
+    req_item = service.read(uploader.identity, recid)
+    _add_to_community(req_item._record, community, service, db)
```
ntarocco, line 106: why not using `record = record_community.create_record()`, which will already create a record in a community, without duplicating the code above `_add_to_community`?

ntarocco (approved):
LGTM as a first step, we only have to decide where this should go.

tests/resources/test_record_file_permissions.py
anikachurilova: Because it creates and publishes a new record without files inside(I'm adding files in `create_record_w_file`)

invenio_rdm_records/services/permissions.py
slint: Yes, this should go in ZenodoRDM only, so that we keep the base permission policy class minimal. We can put it in `invenio.cfg` like so:

```python
# in zenodo-rdm/invenio.cfg

class ZenodoRDMRecordPermissionPolicy(RDMRecordPermissionPolicy):

    # Allow reading the files of a record
    # Hide record's files from community
    can_read_files_owner = [RecordOwners(), SecretLinks("view"), SystemProcess()]
    # Show record's files to community
    can_read_files_community = [
        IfRestricted(
            "files", 
            then_=RDMRecordPermissionPolicy.can_view, 
            else_=RDMRecordPermissionPolicy.can_all
        ),
    ]
    can_read_files = [
        IfFilesRestrictedForCommunity(
            then_=can_read_files_owner, else_=can_read_files_community
        ),
    ]
    can_get_content_files = [
        IfFileIsLocal(then_=can_read_files, else_=[SystemProcess()])
    ]

RDM_PERMISSION_POLICY = ZenodoRDMRecordPermissionPolicy
```

The inheritance/reusability is a bit broken/tricky in here, e.g. some attributes like `can_get_content_files` have to be redefined also since their dependent attributes are redefined in the class... Mangling the parent class fields is probably the best we can do now in terms of maintainability, so let's go with this and see in the future if we need something smarter as discussed in the past.

PR #1283, opened 2023-04-19
services: insure one community exists when remove

Samk13:

### Description

This PR adds an option to prevent deleting all communities in a record except for the last one, as our org requires at least one community per record. This feature is off by default and can be activated in your instance by adding:
```python
RDM_ENSURE_RECORD_COMMUNITY_EXISTS = True
```
Then you will still be able to remove communities, but not the last one.

This PR is connected to [this package](https://pypi.org/project/invenio-config-kth/) we create, it prevents submitting any record without selecting a community from the API level. and prevents users from creating communities unless they have certain privileges that we set for the theme programmatically. 
Future plans are to implement this as well with the master branch.

### Checklist

- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).

**Third-party code**

**Reminder**


invenio_rdm_records/services/communities/service.py
```diff
@@ -166,6 +168,10 @@ def _remove(self, identity, community_id, record):
         self.require_permission(
             identity, "remove_community", record=record, community_id=community_id
         )
+        if current_app.config.get("RDM_ENSURE_RECORD_COMMUNITY_EXISTS"):
```
kpsherva, line 171: we have discussed this PR internally with my colleagues and I have some more comments to follow up (sorry about that!)

- this check will not ensure the record being part of at least one community, however you can still create a record via deposit form and select no community. Is this a desired behaviour for the feature? If yes, what is the use case for allowing to create a record without a community but not allowing to remove one, once the community is added at least once? Could you explain us a bit more in what cases is this desired behaviour?

- this part seems to have reflection only in the backend, it should also be reflected in the frontend if we would like to consider it as a full feature, meaning, for example if `RDM_ENSURE_RECORD_COMMUNITY_EXISTS` and there is only one community attached to a record, the `Remove` button should be blocked in the UI, with a toolltip informing about the reason of blocking the button. Otherwise, the user is not informed early about the limitations in the instance (only when the error appears after an attempt of removing the last community)

Samk13: Thank you for the valuable input @kpsherva.
- to document the answer to the first question, We are currently depending on [invenio-config-kth](https://pypi.org/project/invenio-config-kth/) that overrides permissions to prevent record publishing in the absence of a selected community.

After our discussion on Discord, we agreed to include this behavior in the current PR to make it a complete feature so that when this option is True, you as a user will be denied publishing a record without belonging to at least one community and denied the removal of the last record community.

- We have also decided to bring up this question during the next teleconference with the community to seek their feedback on what behavior we expect in edge cases like: 
> What happens if you have a restricted community attached and a public one and you try to remove the public one what should happen then?

> Only the community manager of the community in question can remove the record from a community. A manager of community A cannot remove the record from community B.
so if community B is restricted (and the community manager A is not part of community B = he cannot see it), but in the interface, the button to remove community A is not disabled, it means (from the point of the manager A) there is a hidden last community.

This can be related to the question if we should make the "remove community" button visible in the UI or not.

invenio_rdm_records/services/permissions.py
```diff
@@ -147,8 +148,20 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     can_delete_draft = can_curate
     # Allow creating a new version of an existing published record.
     can_new_version = can_curate
+    can_publish_restricted = (
+        can_manage[1:]
```
jrcastro2, line 152: IMHO I would suggest to explicitly set the permissions if they cannot be fully reused.
```suggestion
        RecordCommunitiesAction("curate"),
        SystemProcess(),
```

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/components/CommunityHeader/CommunityHeader.js
```diff
@@ -54,6 +56,10 @@ class CommunityHeaderComponent extends Component {
                   {community.metadata.title}
                 </div>
               </>
+            ) : permissions?.can_publish_restricted ? (
+              <Message visible warning className="mt-15" style={{ marginLeft: "15px" }}>
+                {i18next.t("Community is required to submit your record.")}
+              </Message>
             ) : (
```
Samk13, line 63: The rationale behind using the Message component is to draw the user's attention, encouraging them to select a community. However, I am open to alternative suggestions or approaches to achieve this objective.

Samk13:
Preview for the new changes:
disable-publish-btn

fenekku (commented):
I am just dropped into this from https://github.com/inveniosoftware/invenio-app-rdm/pull/2215/files#diff-a779205cb118ac986e91089f0dcc0ed648e4b7a32b252151438c76a8fab7f10fR365-R366, so a lot more background might help me understand. On the flip side, there might be ways to encode this background/help for someone else who will be dropped into this ;) . 

There are bigger-picture, conceptual things to address in my mind to get that complete feature into InvenioRDM. But it's a nice feature (at NU I don't think we would use it though).

invenio_rdm_records/services/communities/service.py
Samk13: Thank you for your suggestions. I see your perspectives on both approaches. However, before finalizing our decision, I would also like to hear what the CERN team suggests on this matter. Their input could potentially provide additional insights or even more suitable solutions.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/components/CommunityHeader/CommunityHeader.js
```diff
@@ -54,6 +56,10 @@ class CommunityHeaderComponent extends Component {
                   {community.metadata.title}
                 </div>
               </>
+            ) : permissions?.can_publish_restricted ? (
```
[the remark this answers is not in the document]
Samk13: Thanks for your thoughtful feedback. I was trying to avoid adding yet another prop to pass, but if the team at CERN sees the benefit in that approach, we can certainly do it!
As to your observation about spacing and the use of the ternary operator, it’s a classic case of what I call "developer's haiku" - brevity may be the soul of wit, but it doesn’t always make for easily digestible code! 😉
I decided to completely remove the change in the community header component, as it adds unnecessary complexity without real significant advantages.

Samk13:
Following the suggestions made by @fenekku, I've rolled back the UI modifications that were previously included in the app instance. Here's a screen recording demonstrating the current functionality:

insure_community_per_record_update_2
Regarding the variable names, they've been left as is for now, until we can reach a consensus on a suitable renaming strategy. Feel free to share your input.

Additionally, I'm interested in hearing from the CERN team on this. I believe their insight could be valuable in refining this further and making it more beneficial for the community.

ntarocco (commented):
Suggestion: we can take advantage of permissions and avoid any code change, similarly to what done with `can_manage_files`:

```
can_remove_community = [
        IfConfig(
            "RDM_RECORD_ALWAYS_IN_COMMUNITY",
            then_=[IfOneCommunity(SystemProcess())],
            else_=[... as now ...],
        ),
    ]

```

The permission class `IfOneCommunity` could simply count how many communities are in the record.
In this way, not only the config var controls if this should be enabled or not, the config is also easily overridable and the code does not need changes.


It is also not clear to me how we ensure that, when submitting, a community is always selected.

Samk13:
Thank you for your suggestion @ntarocco . I've now implemented the permission change as you've recommended, leveraging the permissions similar to how we did with `can_manage_files`.

Regarding your concern about ensuring that a community is always selected during submission: 
I've introduced `can_publish_via_community`. This will disable the publish button unless a community is selected. If a user attempts to bypass this through the API, they'll also receive a permission denied error.

Issue #1289, opened 2023-04-25
Add a permission flags field


anikachurilova:
Add a new dict field to record parent 
```
permission_flags = {
    "can_community_access": False
}
```
See: https://codimd.web.cern.ch/kvN6rZAhTrOCzzeOdzX7cQ#

PR #1290, opened 2023-04-25
Add permission_flags system field

anikachurilova:
* update jsonschema and mappings
* cover with tests
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1289

### DIFF
--- a/invenio_rdm_records/records/jsonschemas/records/parent-v2.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/parent-v2.0.0.json
@@ -77,6 +77,12 @@
 
     "review": {
       "$ref": "local://requests/request-v1.0.0.json"
+    },
+
+    "permission_flags": {
+      "type": "object",
+      "additionalProperties": true,
+      "description": "Storage of permission flag fields."
     }
   }
 }
--- a/tests/records/systemfields/test_permission_flags.py
+++ b/tests/records/systemfields/test_permission_flags.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Test permission flags system field."""
+import pytest
+from invenio_access.permissions import system_identity
+
+from invenio_rdm_records.proxies import current_rdm_records
+from invenio_rdm_records.records import RDMParent, RDMRecord
+
+
+@pytest.fixture()
+def service(running_app):
+    """The record service."""
+    return current_rdm_records.records_service
+
+
+def test_permission_flags_valid(running_app, minimal_record):
+    """Tests permission_flags get and set."""
+    parent_data = {"permission_flags": {"can_community_read_files": True}}
+    parent = RDMParent.create(parent_data)
+    record = RDMRecord.create(minimal_record, parent=parent)
+
+    # test get
+    permission_flags = record.parent.permission_flags
+    assert permission_flags == {"can_community_read_files": True}
+
+    # test set
+    new_data = {"new_permission_flag": False}
+    parent.permission_flags = new_data
+    assert parent.permission_flags == new_data
+
+    # test absence of a flag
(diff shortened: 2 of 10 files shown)


jrcastro2 (approved):
LGTM!

tests/records/systemfields/test_permission_flags.py
```diff
@@ -0,0 +1,82 @@
...
+def test_permission_flags_valid(running_app, minimal_record):
+    """Tests get and set of ParentPermissionFlagsField class."""
+    parent_data = {"permission_flags": {"can_community_read_files": True}}
+    parent = RDMParent.create(parent_data)
+    record = RDMRecord.create(minimal_record, parent=parent)
+
+    # test get
+    permission_flags = record.parent.permission_flags
+    assert permission_flags == {"can_community_read_files": True}
+    assert permission_flags["can_community_read_files"] is True
```
alejandromumo, line 31: I think this is doing the same check as the line above

alejandromumo (commented):
LGTM, left a comment to validate the implementation logic based on the CodiMD comments by Alex

invenio_rdm_records/records/jsonschemas/records/parent-v2.0.0.json
```diff
@@ -77,6 +77,12 @@
 
     "review": {
       "$ref": "local://requests/request-v1.0.0.json"
+    },
+
+    "permission_flags": {
```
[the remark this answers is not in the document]
slint: This is fine, what I meant in the CodiMd was about storing a computed field (in that case `is_migrated` in the record/parent JSON in the DB.

The `permission_flags` field has a specific purpose and stores "unique" information so it's perfectly good to store it in here.

invenio_rdm_records/records/systemfields/permission_flags.py
```diff
@@ -0,0 +1,31 @@
...
+            return self
+
+        try:
+            return dict_lookup(record_parent, "permission_flags")
+        except KeyError:
+            return None
+
+    def __set__(self, record_parent, value):
+        """Set the record parent's permission_flag object."""
+        record_parent["permission_flags"] = value
```
slint, line 31: minor/question: could we get away with a `invenio_records.systemfields.DictField`?

slint (approved):
LGTM! One small comment to see if we can get away with the plain `DictField`

ppanero (commented):
PR LGTM, but why is it only in v5 of the mappings? The latest is now v6!

To be crosschecked with @slint @kpsherva @zzacharo but IIRC the agreement was to not create new versions of mappings if backwards compatible (as it is). But changes should go to v6 not v5 (IMO).

@anikachurilova can we add 1 test that is from the service level instead of just the sysfield? that would have caught this issue since the default mapping is v6 (would have crashed due to the lack of the new field) 😇

slint:
> To be crosschecked with @slint @kpsherva @zzacharo but IIRC the agreement was to not create new versions of mappings if backwards compatible (as it is). But changes should go to v6 not v5 (IMO).

Oh my... I think we have a v6 because of the files indexing (though I thought it was backwards compatible 🤔 ). Well in any case now we're in the mud, so yeah, these have to go to the latest version, i.e. v6...

Adding the test in the service would indeed help to catch this. Though we've had the discussion in the past that some of tests we already have are overlapping between layers, and we should have more smoke tests :)

anikachurilova:
> PR LGTM, but why is it only in v5 of the mappings? The latest is now v6!
> 
> To be crosschecked with @slint @kpsherva @zzacharo but IIRC the agreement was to not create new versions of mappings if backwards compatible (as it is). But changes should go to v6 not v5 (IMO).
> 
> @anikachurilova can we add 1 test that is from the service level instead of just the sysfield? that would have caught this issue since the default mapping is v6 (would have crashed due to the lack of the new field) 😇

@ppanero Isn't [this](https://github.com/inveniosoftware/invenio-rdm-records/pull/1290/files#diff-4d6183545af3f8e1717bf36a684ce898ee3319f0a2e2d9cfd1f628e2dd03c537R69) the test that you mean? The field is not accessible from service layer as it's not present in marshmallow schema

PR #1294, opened 2023-04-28
community modals: accessibility improvements

jennur:
Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2182

- Adds accessible close button to `<CommunitySelectionModal/>`
- Enables focusing on first input element when modal opens

Screenshot 2023-04-28 at 16 26 59

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/controls/PublishButton/SubmitReviewModal.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/controls/PublishButton/SubmitReviewModal.js
@@ -10,11 +10,23 @@ import _get from "lodash/get";
 import PropTypes from "prop-types";
 import React, { Component } from "react";
 import { Trans } from "react-i18next";
-import { ErrorLabel, RadioField, TextAreaField } from "react-invenio-forms";
+import {
+  ErrorLabel,
+  RadioField,
+  TextAreaField,
+  ErrorMessage,
+} from "react-invenio-forms";
 import { Button, Checkbox, Form, Icon, Message, Modal } from "semantic-ui-react";
 import * as Yup from "yup";
 
 export class SubmitReviewModal extends Component {
+  componentDidMount() {
+    // A11y: Focus the first input field in the form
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/controls/PublishButton/SubmitReviewModal.js
```diff
@@ -83,15 +90,15 @@ export class SubmitReviewModal extends Component {
             >
               <Modal.Header>{headerTitle}</Modal.Header>
               <Modal.Content>
-                {errors}
+                <div role="alert">{errors}</div>
```
kpsherva, line 93: why not using the message error component from react-invenio-forms instead?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/controls/PublishButton/SubmitReviewModal.js
```diff
@@ -15,6 +15,13 @@ import { Button, Checkbox, Form, Icon, Message, Modal } from "semantic-ui-react"
 import * as Yup from "yup";
 
 export class SubmitReviewModal extends Component {
+  componentDidMount() {
+    // A11y: Focus the first input field in the form
+    const firstFormFieldWrap = document.getElementById("accept-access-checkbox");
+    const checkboxElem = firstFormFieldWrap.querySelector("input");
+    checkboxElem?.focus();
```
kpsherva, line 22: I think using .ref is more "react way" of doing this, instead of using vanilla js, but this is not a strong opinion

jennur: I had trouble accessing the element through ref which is why I did it like this, but I can give it another try if it's worth the time. Let me know.

kpsherva: I would try just for the sake of consistency ;)

PR #1297, opened 2023-05-03
deposit: fixed thesis field.

alejandromumo:
Thesis data is already loaded under key `custom_fields.thesis:university`, e.g:
```javascript
"initialValues" = {
    "custom_fields": {
        "thesis:university": "MIT"
    }
}
```

Therefore, it's a string and not an object (we were trying to access `custom_fields.thesis:university.university`)


jennur (approved):
LGTM

kpsherva (approved):

PR #1299, opened 2023-05-03
mappings: increase numeric type for totalbytes

kpsherva:


invenio_rdm_records/records/mappings/os-v1/rdmrecords/drafts/draft-v6.0.0.json
```diff
@@ -821,7 +821,7 @@
             "type": "integer"
           },
           "totalbytes": {
-            "type": "integer"
+            "type": "unsigned_long"
```
kpsherva, line 824: for reviewers, discussion: needed because int was overflowing
however, it seems that marshmallow field is not overflowing and the docs does not provide more usuitable type, so I didn't change it.

ppanero: I do not find any Marshmallow definition for this field (see [search](https://github.com/search?q=org%3Ainveniosoftware+totalbytes&type=code).?

jrcastro2: Apparently int objects in python do not overflow see here [discussion about it](https://stackoverflow.com/questions/52151647/integer-overflow-in-python3), but then I find it weird that they actually handle this error in marshmallow code [here](https://marshmallow.readthedocs.io/en/stable/_modules/marshmallow/fields.html#Number) (as it should never occur), but IMHO this is okay

jrcastro2 (approved):

ppanero (approved):
In zenodo we are also using `long`

PR #1300, opened 2023-05-03
don't create admin@inveniosoftware.org user by default

fenekku:
Although that User doesn't have any privileges, its presence is
initially alarming and it does create an account that can be used
by anyone. Given that it's not necessary, we might as well not have it.

### DIFF
--- a/invenio_rdm_records/fixtures/tasks.py
+++ b/invenio_rdm_records/fixtures/tasks.py
@@ -13,7 +13,12 @@ import random
 from celery import shared_task
 from flask import current_app
 from flask_principal import Identity, UserNeed
-from invenio_access.permissions import any_user, authenticated_user, system_identity
+from invenio_access.permissions import (
+    any_user,
+    authenticated_user,
+    system_identity,
+    system_user_id,
+)
 from invenio_communities.generators import CommunityRoleNeed
 from invenio_communities.members.errors import AlreadyMemberError
 from invenio_communities.proxies import current_communities
@@ -47,8 +52,10 @@ def create_vocabulary_record(service_str, data):
 def create_demo_record(user_id, data, publish=True):
     """Create demo record."""
     service = current_rdm_records_service
-    identity = get_authenticated_identity(user_id)
-    identity.provides.add(UserNeed(user_id))
+    if user_id == system_user_id:
+        identity = system_identity
+    else:
+        identity = get_authenticated_identity(user_id)
 
     draft = service.create(data=data, identity=identity)
     if publish:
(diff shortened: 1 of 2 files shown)


fenekku:
@chriz-uniba for out of CERN review.

kpsherva:
@slint this is a duplication, I have put both on the board of the sprint because it needs deciding which approach we prefer to go with.
I think it is good if the "system" is the owner of the fixture records, unless specified otherwise, however I am not sure if we are able to specify otherwise in the fixtures. It should be tested before merging this if we can have a specific owner provided by the fixture.

slint:
> I think it is good if the "system" is the owner of the fixture records, unless specified otherwise, however I am not sure if we are able to specify otherwise in the fixtures. It should be tested before merging this if we can have a specific owner provided by the fixture.

Looks like [the "system" is handled](https://github.com/inveniosoftware/invenio-rdm-records/blob/9d43c516b63d5dc90e77c3beb5cc3abe761303eb/invenio_rdm_records/services/components/access.py#L59) on a technical level, so fine with me. In https://github.com/inveniosoftware/invenio-rdm-records/pull/1298 I avoided it because I suspected that it would fail to create a record (since the system identity doesn't have a user ID).

Passing a user in the fixtures as an owner for a record needs a bit more thought though, since we need a new convention for the `records.yaml` (similar to what we did for loading community logos). I would say though that I'm not even sure if anyone is using this fixture in reality or knows about it, since by default it's empty.

I would merge this for now, since it provides some more functionality, close #1298, and revisit separately if we want to implement specifying a user in the fixtures.

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -44,11 +44,13 @@ def create_vocabulary_record(service_str, data):
 
 
 @shared_task
-def create_demo_record(user_id, data, publish=True):
+def create_demo_record(user_id_or_identity, data, publish=True):
```
slint, line 47: Since this is a Celery task, I'm not sure if `Identity` serializes properly... IIRC it's a simple namedtuple in the end, but it's always a hustle with Celery task parameters that need to be primitive/raw values.

kpsherva: @fenekku did you have a chance to test if the identity serializes? I can merge it either way if @slint agrees

slint: I think we're good, since the only usage I see that calls `create_demo_record.delay(...)`, uses a user ID (which is serializable)

fenekku: I'll test Identity serialization out this morning.

fenekku: Ugh it's indeed not. I'll pass the system_user_id instead and convert in the creation function.

fenekku:
updated PR to check if user_id is the system_user_id instead (side-steps serialization of Identity) and then do conversion in the record demo creation task.

fenekku:
Failing CI has to do with unrelated

```
 E       elasticsearch.exceptions.RequestError: RequestError(400, 'strict_dynamic_mapping_exception', 'mapping set to strict, dynamic introduction of [notifications] within [user.preferences] is not allowed')
```

I will just wait until that's resolved upstream.

PR #1303, opened 2023-05-10
fixtures: method to add entries to fixtures

tmorrell:

### Description

The documentation indicates that new fixtures can be added https://inveniordm.docs.cern.ch/customize/vocabularies/, but that's not the case. I've added a new command line option `add_to_fixture` which adds new fixtures to a given vocabulary.

This seems to work fine, but feedback is much appreciated because I'm sure there are better approaches.

### Checklist

**Third-party code**

**Reminder**

### DIFF
--- a/invenio_rdm_records/fixtures/tasks.py
+++ b/invenio_rdm_records/fixtures/tasks.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2021-2022 CERN.
 # Copyright (C) 2021-2022 Northwestern University.
+# Copyright (C) 2023 California Institute of Technology.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
@@ -22,9 +23,11 @@ from invenio_access.permissions import (
 from invenio_communities.generators import CommunityRoleNeed
 from invenio_communities.members.errors import AlreadyMemberError
 from invenio_communities.proxies import current_communities
+from invenio_pidstore.errors import PersistentIdentifierError
 from invenio_records_resources.proxies import current_service_registry
 from invenio_requests import current_events_service, current_requests_service
 from invenio_requests.customizations import CommentEventType
+from invenio_vocabularies.records.api import Vocabulary
 
 from ..proxies import current_oaipmh_server_service, current_rdm_records_service
 from ..requests import CommunitySubmission
@@ -43,9 +46,21 @@ def get_authenticated_identity(user_id):
 
 @shared_task
 def create_vocabulary_record(service_str, data):
-    """Create a vocabulary record."""
+    """Create or update a vocabulary record."""
     service = current_service_registry.get(service_str)
-    service.create(system_identity, data)
+    if "type" in data:
+        # We only check non-datastream vocabularies for updates
+        try:
+            pid = (data["type"], data["id"])
+            # If the entry hasn't been added, this will fail
+            record = Vocabulary.pid.resolve(pid)
+            service.update(system_identity, pid, data=data)
+            current_app.logger.info(f"updated existing fixture with {data}")
+        except PersistentIdentifierError:
+            service.create(system_identity, data)
+            current_app.logger.info(f"added new fixture with {data}")
+    else:
+        service.create(system_identity, data)
 
 
 @shared_task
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/fixtures/tasks.py
```diff
@@ -45,7 +46,10 @@ def get_authenticated_identity(user_id):
 def create_vocabulary_record(service_str, data):
     """Create a vocabulary record."""
     service = current_service_registry.get(service_str)
-    service.create(system_identity, data)
+    try:
+        service.create(system_identity, data)
+    except IntegrityError:
+        current_app.logger.info(f"skipping creation of {data}, already existing")
```
ppanero, line 52: Minor: If I understand correctly the new functionality supports 2 scenarios:
1. Add new entries to already existing vocabularies (i.e. those whose vocabulary type is already in db)
2. Recreate (i.e. "reload"), which would delete the entry and allow recreation. 

For case number 2 I wonder if it would be fixed by instead of skipping, making a `service.update` call?

ppanero (commented):
LGTM! just one concern, we are relying heavily on `IntegrityErrors` which means that we are forcing the DB to error and recover (not sure how expensive it is, in addition this command is only a setup one so the impact would be minimized). However, this is something that would be fixed if datastreams supported generic vocabularies.

... And profiting to raise a topic hehe IMO, it can be merged for now. However, there should be a discussion if we should accept more work on old fixtures or put it on datastream. Otherwise, give up on the second. Maintaining two methods, none of them complete, starts to create divergence. cc @zzacharo @kpsherva @slint

invenio_rdm_records/fixtures/tasks.py
tmorrell: Thanks! I've reorganized the code so it now uses `service.update`

tmorrell:
I've been able to refactor the code so it relies less on database errors, and added the ability to update existing records.

I'd like to be able to shift more of the vocabularies (particularly affiliations) to datastreams, but the old fixtures are really useful for custom vocabularies. One just need to create a couple easy files and the custom vocabulary is up. I'm not really sure how to replicate that with datastreams.

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -43,9 +46,19 @@ def get_authenticated_identity(user_id):
...
+    """Create or update a vocabulary record."""
     service = current_service_registry.get(service_str)
-    service.create(system_identity, data)
+    pid_type = data['type']
+    pid_value = data['id']
+    try: 
+        pid = (pid_type,pid_value)
+        # If the entry hasn't been added, this will fail
+        record = Vocabulary.pid.resolve(pid)
+        service.update(system_identity,pid, data=data)
```
ppanero, line 57: minor: this might "re-activate" deleted entries, but I guess we assume that if they are in the data file and one runs the command we *want to reactive it*. So I'd say all good.

ppanero (approved):
Nice work! LGTM 🚀 

Regarding your question about switching more vocabularies to datastreams:

- For _affiliations_ it should be relatively easy since they are contrib. IIRC only the transformer part needs to be implemented. The service writer already exists and problably just need some configuration tuning, and there are plenty of loaders (this depends on the original format of the dump).
- Implementing loading with generic vocabularies, if the data path is only one (as with contrib vocabularies such as names), it would be quite straight forward. Probably just a transform and service config. However, the main issue comes on the layered approach, i.e. module `/data` and `app_data` path inspection which would be either lost of hard to migrate. In any case, I think a first step is to enable datastreams to import generic vocabularies, and a second step (if needed) to migrate the per-module loading. 

I'm happy to have a call and discuss this in more detail if needed.

tmorrell:
This PR now adds and updates fixtures, and I've also been able to rely less on integrity errors. There has also been an independent use of the new code #1348,  so hopefully this can be reviewed and merged soon.

invenio_rdm_records/fixtures/tasks.py
```diff
@@ -43,9 +46,21 @@ def get_authenticated_identity(user_id):
 
 @shared_task
 def create_vocabulary_record(service_str, data):
-    """Create a vocabulary record."""
+    """Create or update a vocabulary record."""
     service = current_service_registry.get(service_str)
-    service.create(system_identity, data)
+    if "type" in data:
```
ppanero, line 51: this might fail if the custom vocabulary has a `type` field, but at this point we do not have any other more sensible way of checking this.

PR #1306, opened 2023-05-12
Access requests

max-moser:
# Context

Closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2180 (related to the [access management RFC](https://codimd.web.cern.ch/U6_FMlwPS9ShGVtYDBjIUQ?view#Managing-access)).
Requires ~~https://github.com/inveniosoftware/invenio-records-resources/pull/472 to work (for skipping system role entities during expansion of result lists for grants), as well as~~ https://github.com/inveniosoftware/invenio-requests/pull/318 (to make the `self_html` links in requests work).

# Disclaimer

This PR changes some models structurally, which will require new versions of some JSON Schemas and search mappings.
For the sake of easier reviews (and because I'm a bit lazy), I'll move the changes to new version files after the reviews though.

# Implemented features

## The following items from the RFC are covered
* Rework of permission grants as part of the record parents
  * Structural changes to the subject
  * Addition of a new (purely informational) "origin" field
  * Enabled dumping them into the search engine
* Taking the grants into account for permissions
  * This allows records to be shared with _users_, _groups_, and _system roles_ (e.g. IP ranges, ...)
* Service and resource (i.e. API) for managing access grants
  * Including expansion of grants, to include useful information for the frontend
* Description and origin field for secret links
* Record/metadata visibility settings backed by an enum now
* Access request workflow for authenticated _users_
  * Including request UI
* Reworking record owners to be a single value rather than a list
* Database models for access request tokens, to be used in the guest-based request flow
* Email address as entity (sender/receiver of requests)
* Access requests workflow for unauthenticated _guests_ (via email/access request tokens)
  * Including request UI
* Creation of duplicate access requests returns a `409 CONFLICT` with a pointer to the already existing request

## Additional changes
* Fixed "edit"-level grants not being able to preview and "preview"-level grants not being able to view

## Removed from this PR to make reviews & merging easier

Some code for these removed pieces already exists on a [backup branch](https://github.com/max-moser/invenio-rdm-records/tree/mm/access-requests-orig-backup), so the work won't start from zero.

* Made the permissions available for secret links and grants configurable
  * Somebody might want to create a custom permission level like "access to files but not metadata" (which would be a step towards double-blind peer reviews)
* Access (request) settings in the parent record's JSON
  * This shouldn't be hard, but does require some design work

## Left for future developers
* Bulk operations for managing grants
* Explaining access to a record
* Endpoint that searches over both _users_ and _groups_ (relevant for access grant creation)
* Editing the payload of an existing access request (e.g. setting a different permission level to be granted)
* Editing the access settings for a record
* Spam prevention measures for guest-based requests
  * rate limiting for how often one email address can be used, to prevent an inbox from being spammed?
  * rate limiting for an IP address, to prevent system DOS
  * captchas?
* Changing the receiver of the access requests based on the record's access settings
  * This will likely require the workflow logic to be aware that receivers can be communities _or_ users

# Notes

* Some frontend additions for this functionality are available here: https://github.com/inveniosoftware/invenio-app-rdm/pull/2242

## Version migration

The access grant changes in this PR shouldn't require any changes, because the grants were unused before.
However, the changes made to record ownership requires something like the following (after the model migration):
```python
from invenio_rdm_records.records.api import RDMDraft, RDMRecord

def update_parents(records_or_drafts):
  for rec in records_or_drafts:
    if rec.parent["$schema"] and rec.parent["$schema"].endswith("parent-v3.0.0.json"):
      continue

    old_owners = rec.parent["access"]["owned_by"]
    if not old_owners:
      rec.parent.access.owner = None
    else:
      rec.parent.access.owner = old_owners[0]
      for old_owner in old_owners[1:]:
        click.secho(f"warning: {old_owner} was removed as owner of record '{rec.pid.pid_value}', they get an access grant with 'edit' permissions instead", fg="yellow")
        rec.parent.access.grants.create("user", old_owner["user"], "edit", "migration")

    record.parent.pop("$schema", None)
    record.pop("$schema", None)
    record.parent.commit()
    record.commit()

all_records = [RDMRecord(rm.data, model=rm) for rm in RDMRecord.model_cls.query.all()]
update_parents(all_records)

all_drafts = [RDMDraft(dm.data, model=dm) for dm in RDMDraft.model_cls.query.all()]
update_parents(all_drafts)
```


invenio_rdm_records/records/systemfields/access/grants.py
```diff
@@ -76,7 +89,7 @@ def subject_type(self):
         else:
             # if it's nothing else, it's gonna be a system role
             # (which doesn't have a persisted subject entity)
-            return "sysrole"
+            return "system_role"
```
max-moser, line 92: renamed `sysrole` to `system_role`, to better match the [`SystemRole` need](https://github.com/inveniosoftware/invenio-access/blob/master/invenio_access/permissions.py#L31)

invenio_rdm_records/records/jsonschemas/records/parent-v1.0.0.json
```diff
@@ -40,7 +40,7 @@
             "properties": {
               "subject": {
                 "type": "string",
-                "enum": ["user", "role", "sysrole"]
+                "enum": ["user", "role", "system_role"]
```
max-moser, line 43: because up until now the grants were actually not dumped, this property was actually unused... so i guess this is a backwards-compatible change? :eyes:

invenio_rdm_records/records/systemfields/access/grants.py
```diff
@@ -86,22 +99,23 @@ def subject_id(self):
...
+        """Create the need that this grant provides.
+
+        Note: This operation doesn't resolve the subject, and thus won't cause any
+        database lookups even if the. On the flipside, it also won't check if the
+        subject exists.
+        """
         if self.subject_type == "user":
-            need = UserNeed(self.subject.id)
+            # NOTE: the `UserNeed` must have an integer as ID, otherwise it won't match
+            need = UserNeed(int(self.subject_id))
```
max-moser, line 111: in contrast to e.g. the `datastore`, we _need_ the user's ID to be numeric in the `UserNeed` – otherwise, the permissions will never be granted

invenio_rdm_records/records/systemfields/access/grants.py
```diff
@@ -86,22 +99,23 @@ def subject_id(self):
 
         return self.subject.id
 
-    def covers(self, permission):
```
max-moser, line 89: this functionality is covered in the permission policy

invenio_rdm_records/records/jsonschemas/records/parent-v1.0.0.json
max-moser: this comment didn't age very well; there's so many changes now that i think a new version is justified.

invenio_rdm_records/records/systemfields/access/links.py
```diff
@@ -104,13 +104,13 @@ def remove(self, link):
     def create(
         self,
         permission_level,
-        extra_data=dict(),
```
max-moser, line 107: we never used this `extra_data`, but still getting rid of that bug

invenio_rdm_records/ext.py
```diff
@@ -80,6 +84,12 @@ def verify_token():
             session.pop("rdm-records-token", None)
             flash(_("Your shared link has expired."))
 
+    token = request.args.get("access-request-token", None)
```
max-moser, line 87: from @anikachurilova: is it safe to store the token's value in the session?

given that it's a server-side session, i think it should be okay; if server-side sessions get leaked, we're probably in more trouble than third parties being able to view access request details.
however, currently i'm letting the system create a comment including the generated secret link on the access request when it is being accepted – maybe we should remove this.

can you think of any potential attack vectors / security issues here?

invenio_rdm_records/services/services.py
```diff
@@ -131,3 +148,180 @@ def oai_result_item(self, identity, oai_record_source):
...
+
+        # TODO detect duplicates
+        access_token_data = access_token.to_dict()
+        record = self.record_cls.pid.resolve(access_token_data["record_pid"])
+        data = {
+            "payload": {
+                "permission": record.parent.access.settings.default_access_permission,
+                "email": access_token_data["email"],
+                "full_name": access_token_data["full_name"],
+                "token": access_token_data["token"],
```
max-moser, line 285: for good measure, we should probably create a new token here rather than re-using the existing one.

invenio_rdm_records/requests/access/models.py
```diff
@@ -0,0 +1,134 @@
...
+from secrets import token_urlsafe
+
+from flask import current_app
+from invenio_db import db
+from sqlalchemy_utils import UUIDType
+
+from .permissions import AccessRequestTokenNeed
+
+
+class AccessRequestToken(db.Model):
```
max-moser, line 21: should be renamed to something like `GuestAccessRequestToken`, in order to minimize confusion with RATs

invenio_rdm_records/requests/access/models.py
```diff
@@ -0,0 +1,134 @@
...
+            expires_at = datetime.combine(expires_at, datetime.min.time())
+
+        access_request_token = cls(
+            email=email,
+            full_name=full_name,
+            message=message,
+            record_pid=record_pid,
+            expires_at=expires_at,
+            token=token_urlsafe(
+                current_app.config.get("RDM_RECORDS_GUEST_ACCESS_TOKEN_LENGTH")
```
max-moser, line 129: probably we should factor in the `email` address and a `nonce` in the token generation rather than just generating a purely random token.
if we can extract these values from the (encrypted) token again, then we could use that to create more meaningful `Need`s from the token (like a `TokenNeed(nonce)`, and an `EmailNeed(email_address)`, which could help us with a few issues (like generating more sensible "comment author name" values on guest access requests).

invenio_rdm_records/records/jsonschemas/records/parent-v2.0.0.json
```diff
@@ -67,6 +73,38 @@
               }
             }
           }
+        },
+
+        "settings": {
```
max-moser, line 78: These access `settings` are intended to store configuration per record about who is able to request access (users/guests/both/nobody), the display text outlining the access conditions, and default values for the access request.

Right now, this PR does not include a REST API endpoint to change these values for a record.
IMO, that endpoint can come later, as it's not strictly necessary for the functioning of the access requests.

invenio_rdm_records/records/jsonschemas/records/parent-v2.0.0.json
```diff
@@ -67,6 +73,38 @@
...
+              "description": "The default permission to give out via access requests.",
+              "type": "string"
+            },
+            "default_link_expiration": {
+              "description": "The default time period before secret links expire.",
+              "type": ["string", "null"]
+            },
+            "default_request_receiver": {
+              "description": "The default receiver for access requests.",
+              "type": ["string", "null"]
```
max-moser, line 105: Per RFC, we want to have the possibilities to set **"record owner"** or **"managers/owners of community X"** (as long as the record is part of that community).

This could be encoded as a sort of **reference dictionary** (e.g. `{"user": "1337"}`, or `{"community": "XCOM"}`).
Of course, then we would need to ensure that this can't be set to just about any arbitrary user or community while setting the value (and we would need to re-validate whenever the record ownership or the record's communities change).

Alternatively, we could use a sort of simple string here (like `"record_owner"` or `"community:XCOM"`), but that's probably just strictly worse than the reference dictionary.

Ideas, thoughts, prayers?

invenio_rdm_records/records/jsonschemas/records/parent-v2.0.0.json
```diff
@@ -67,6 +73,38 @@
...
+              "description": "Custom description for access conditions to be displayed rather than the default text.",
+              "type": ["string", "null"]
+            },
+            "default_access_permission": {
+              "description": "The default permission to give out via access requests.",
+              "type": "string"
+            },
+            "default_link_expiration": {
+              "description": "The default time period before secret links expire.",
+              "type": ["string", "null"]
```
max-moser, line 101: String? Date?

invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
```diff
@@ -21,6 +21,9 @@
     "oneOf": [
```
max-moser, line 21: making the `agent` (used for record owners) nullable

invenio_rdm_records/services/services.py
```diff
@@ -131,3 +148,229 @@ def oai_result_item(self, identity, oai_record_source):
...
+        )
+
+        # immediately submit the request, unless it has errors
+        # TODO is this correct?
+        if request.errors:
+            return request
+
+        message = {
+            "payload": {
+                "content": data["payload"].get("message") or "No comment",
```
zzacharo, line 226:
```suggestion
                "content": data["payload"].get("message") or _("No comment"),
```

invenio_rdm_records/services/services.py
```diff
@@ -131,3 +148,229 @@ def oai_result_item(self, identity, oai_record_source):
...
+            }
+        }
+
+        # TODO can we pass the message here?
```
zzacharo, line 230: leftover?

invenio_rdm_records/services/services.py
```diff
@@ -131,3 +148,229 @@ def oai_result_item(self, identity, oai_record_source):
...
+        access_token = AccessRequestToken.create(
+            email=data["email"],
+            full_name=data["full_name"],
+            message=data["message"],
+            record_pid=id_,
+            shelf_life=timedelta(hours=6),
+        )
+
+        # TODO this doesn't work because it's creating an /api/ URL...
+        verify_url = url_for(
```
zzacharo, line 263: I guess we need to find a solution for this, although not sure how 😅  I

invenio_rdm_records/services/services.py
```diff
@@ -131,3 +148,229 @@ def oai_result_item(self, identity, oai_record_source):
...
+            return request
+
+        message = {
+            "payload": {
+                "content": data["payload"].get("message") or "No comment",
+            }
+        }
+
+        return current_requests_service.execute_action(
+            system_identity,  # TODO change
```
zzacharo, line 371: TODO to be addressed?

invenio_rdm_records/templates/semantic-ui/invenio_notifications/guest_access_request_accept.jinja
```diff
@@ -0,0 +1,23 @@
+{#
+  Copyright (C) 2023 TU Wien.
+
+  Invenio RDM Records is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+#}
+{% set access_request = notification.context["request"] %}
+{% set record = submission_request["topic"] %}
+{% set link = "https://localhost:5000/records/" + record.id + "?token=" + notification.context["token"] %}
```
zzacharo, line 9: To be fixed!

invenio_rdm_records/views.py
```diff
@@ -33,6 +48,74 @@ def init(state):
...
+    """
+    # TODO check how the email verification logic works and get inspiration from there
+    token = request.args.get("access-request-token")
+    access_request = current_service.create_guest_access_request(
+        identity=g.identity, token=token
+    )
+    if access_request is None:
+        abort(404)
+
+    # TODO if we replace the `url_for()` with the `access_request.links["self_html"]`
```
zzacharo, line 68: still valid?

invenio_rdm_records/ext.py
```diff
@@ -256,3 +277,16 @@ def fix_datacite_configs(self, app):
...
+        """Validate the configuration."""
+        token_length = app.config["RDM_RECORDS_GUEST_ACCESS_TOKEN_LENGTH"]
+        if token_length is not None and token_length <= 0:
+            raise Exception("invalid RDM_RECORDS_GUEST_ACCESS_TOKEN_LENGTH")
+
+        link_permission_levels = app.config["RDM_RECORDS_SECRET_LINK_PERMISSION_LEVELS"]
+        grant_permission_levels = app.config["RDM_RECORDS_GRANT_PERMISSION_LEVELS"]
+        if not set(link_permission_levels).intersection(set(grant_permission_levels)):
+            raise Exception(
+                "no intersection between RDM_RECORDS_SECRET_LINK_PERMISSION_LEVELS and RDM_RECORDS_GRANT_PERMISSION_LEVELS"
```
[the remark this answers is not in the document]
max-moser: Because the access requests generate **grants** for the user-based workflow and **secret links** for the guest-based workflow.
This is only really "necessary" if the access request settings apply to both workflows.
If we handle them separately (separate setting for the default permission level given by an accepted `guest-access-request` and `user-access-request`), there's no need for an overlap anymore.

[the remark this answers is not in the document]
max-moser: That being said, I've removed the configurability of the available permission levels from the PR for now, to reduce the scope.

invenio_rdm_records/records/jsonschemas/records/parent-v2.0.0.json
max-moser: Uhh... I interpreted the [RFC](https://codimd.web.cern.ch/U6_FMlwPS9ShGVtYDBjIUQ?view) in a way that you could select **a** community as alternative receiver, whose manager members would receive the requests.
Re-reading it, I still _think_ that's the intended semantics but I'm not 100% sure anymore.
Ping @lnielsen

invenio_rdm_records/records/systemfields/access/field/parent.py
```diff
@@ -8,61 +8,136 @@
...
-from ..owners import Owners
+from ..owners import Owner
+
+
+class AccessSettings:
+    """Access settings for a parent record."""
+
+    def __init__(self, settings_dict):
+        """Constructor."""
+        # TODO we need to make sure that the available permission levels for links & grants have some overlap
```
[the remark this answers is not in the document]
max-moser: True, that's a leftover... But for now, I've removed both the configurability of permission levels and the access from this PR anyway.

invenio_rdm_records/records/systemfields/access/field/parent.py
```diff
@@ -8,61 +8,136 @@
...
+            current_app.config["RDM_RECORDS_SECRET_LINK_PERMISSION_LEVELS"]
+        )
+        grants_permissions = set(
+            current_app.config["RDM_RECORDS_GRANT_PERMISSION_LEVELS"]
+        )
+        default_permission = list(links_permissions.intersection(grants_permissions))[0]
+
+        self.user_requests_enabled = settings_dict.get("user_requests_enabled", False)
+        self.guest_requests_enabled = settings_dict.get("guest_requests_enabled", False)
+        self.access_conditions_text = settings_dict.get("access_conditions_text", None)
```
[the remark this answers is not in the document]
max-moser: Actually yes :smile:

invenio_rdm_records/records/systemfields/access/links.py
max-moser: This one: https://github.com/inveniosoftware/flask-resources/issues/114

invenio_rdm_records/services/services.py
```diff
@@ -131,3 +148,229 @@ def oai_result_item(self, identity, oai_record_source):
...
+        ]
+
+        if any((req.is_open for req in requests)):
+            # TODO DuplicateRequestError
+            raise PermissionDeniedError()
+
+        record = self.record_cls.pid.resolve(id_)
+        data = {
+            "payload": {
+                "permission": record.parent.access.settings.default_access_permission,
```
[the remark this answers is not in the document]
max-moser: The intention is to allow the request's `receiver` to change the permission given out through the request as long as it's open (i.e. not yet accepted/declined/...) in the future.
So this setting would only be the "initial" value for new access requests, which can be changed during the access request's lifespan.

invenio_rdm_records/services/services.py
```diff
@@ -131,3 +148,229 @@ def oai_result_item(self, identity, oai_record_source):
...
+                "message": message,
+            }
+        }
+
+        # TODO might also be community managers
+        receiver = None
+        record_owner = record.parent.access.owner.resolve()
+        if record_owner:
+            receiver = record_owner
```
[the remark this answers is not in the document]
max-moser: Yes

invenio_rdm_records/services/services.py
```diff
@@ -131,3 +148,229 @@ def oai_result_item(self, identity, oai_record_source):
...
+            UserAccessRequest,
+            receiver,
+            topic=record,
+            expires_at=None,  # TODO expiration date
+            expand=expand,
+            uow=uow,
+        )
+
+        # immediately submit the request, unless it has errors
+        # TODO is this correct?
```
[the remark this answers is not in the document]
max-moser: Yes, that's the workflow for now.
I'm just wondering if it's "correct" do simply return the created (but not yet submitted) request in case the creation step already failed.
Probably this should be wrapped in an exception instead?

invenio_rdm_records/services/services.py
max-moser: Potentially via parts of the machinery behind the **links templates**?
Haven't had time to look deeper into this yet though...

invenio_rdm_records/services/services.py
max-moser: Yeah, this needs to be addressed as part of the issue regarding the guest's comments on the request.
In a nutshell, there's currently no good way to tell the services that _"the guest"_ (maybe with _"that email address"_) created the comment.
So for now, we're evading this by letting "the system" create the first comment on the request.

invenio_rdm_records/records/jsonschemas/records/parent-v3.0.0.json
```diff
@@ -0,0 +1,93 @@
...
+            "required": ["subject", "permission"],
+            "additionalProperties": false,
+            "properties": {
+              "subject": {
+                "type": "object",
+                "required": ["type", "id"],
+                "additionalProperties": false,
+                "properties": {
+                  "type": {
+                    "type": "string"
```
max-moser, line 45: removed the `enum` here, because we agreed that the content validation should be done by marshmallow rather than JSON Schema

invenio_rdm_records/records/mappings/os-v1/rdmrecords/drafts/draft-v6.0.0.json
max-moser, line 1: The JSON Schemas for records & drafts haven't changed, so their version numbers are still the same – as such, I think we should keep the mapping versions the same as well?
Especially given that the `grant` tokens haven't been dumped to search before anyway, so there's no incompatible data in the search index.
**However**, this will require destroying and rebuilding the record/draft indices because the mappings have changed.

invenio_rdm_records/requests/access/models.py
max-moser: after a brief discussion with @zzacharo on discord, we agreed to keep the name as is

invenio_rdm_records/requests/access/requests.py
```diff
@@ -0,0 +1,215 @@
...
+                "recipients": [self.receiver],
+                "sender": current_app.config["MAIL_DEFAULT_SENDER"],
+            }
+        )
+
+
+#
+# Actions
+#
+class UserSubmitAction(actions.SubmitAction):
```
jrcastro2, line 49: Question: Should we notify via email that an user/guest requested access?

invenio_rdm_records/requests/access/requests.py
```diff
@@ -0,0 +1,215 @@
...
+        message = f'Click <a href="{access_url}">here</a> to access the record.'
+        comment_data = {"payload": {"content": message}}
+        events_service.create(
+            system_identity, self.request.id, comment_data, CommentEventType, uow=uow
+        )
+
+        uow.register(RecordCommitOp(record._record.parent))
+        uow.register(
+            EmailOp(
+                receiver=payload["email"],
```
jrcastro2, line 109: Question: Not for now, would it make sense make the emails easily customizable? via a jinja template for instance

invenio_rdm_records/requests/access/requests.py
```diff
@@ -0,0 +1,215 @@
...
+
+        self.request["title"] = f"{name} requests access to {title}"
+        super().execute(identity, uow)
+
+
+class GuestAcceptAction(actions.AcceptAction):
+    """Accept action."""
+
+    def execute(self, identity, uow):
+        """Accept record into community."""
```
jrcastro2, line 84:
```suggestion
        """Accept guest access request."""
```

invenio_rdm_records/requests/access/requests.py
```diff
@@ -0,0 +1,215 @@
...
+        )
+
+        super().execute(identity, uow)
+
+
+class UserAcceptAction(actions.AcceptAction):
+    """Accept action."""
+
+    def execute(self, identity, uow):
+        """Accept record into community."""
```
jrcastro2, line 123:
```suggestion
        """Accept user access request."""
```

invenio_rdm_records/views.py
```diff
@@ -33,6 +49,80 @@ def init(state):
...
+    # TODO check how the email verification logic works and get inspiration from there
+    token = request.args.get("access_request_token")
+    try:
+        access_request = current_service.access.create_guest_access_request(
+            identity=g.identity, token=token
+        )
+    except DuplicateAccessRequestError:
+        # TODO should we redirect to the already existing request? would that leak data?
+        # TODO we should still delete the guest access request token
+        abort(409)
```
jrcastro2, line 70: Comment: I think we agreed on redirecting, there should not be any data leaking as the token is send via email.

jrcastro2 (commented):
Massive work! Thank you very much! :clap:  :rocket: 
Just left a few of comments and questions.

invenio_rdm_records/requests/access/requests.py
max-moser: Yes, I think email notifications are a massive improvement to the usefulness of requests.
It's one of the things that I'm gonna create follow-up issues for.

invenio_rdm_records/requests/access/requests.py
max-moser: Definitely!

invenio_rdm_records/records/systemfields/access/grants.py
```diff
@@ -113,75 +161,75 @@ def to_token(self):
...
-            raise ValueError("unknown subject type: {}".format(subject_type))
-
-        if fetch_subject:
-            if subject_type == "user":
-                subject = User.query.get(subject_id)
-            elif subject_type == "role":
-                subject = Role.query.get(subject_id)
-        else:
-            subject = None
+        if subject_type not in ["user", "role", "system_role"]:
```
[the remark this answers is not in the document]
max-moser: We can keep the `system_role`s in there; I actually came up with a fix to the issue at hand via the `DummyExpandingService`.
This eliminates the need for the Records-Resources PR (which I already closed), and still allows us to keep the `system_role` subjects around.

We don't offer a way through the UI to put such grants into the system, but if somebody manages to do that, nothing is going to break anymore thanks to the `DummyExpandingService`.

invenio_rdm_records/alembic/cfcb8cb78708_create_access_request_tokens_table.py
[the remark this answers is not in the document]
max-moser: Mostly just separation of concerns, because they're kind of unrelated changes to two different DB models that were introduced at different points during the development of this PR.

invenio_rdm_records/requests/access/models.py
```diff
@@ -0,0 +1,134 @@
...
+    expires_at = db.Column(db.DateTime, nullable=False)
+    """Expiration date."""
+
+    email = db.Column(db.String, nullable=False)
+    """E-mail address for which the access request was made."""
+
+    full_name = db.Column(db.String, nullable=False, default="")
+    """Full name of the requester."""
+
+    message = db.Column(db.Text, nullable=False, default="")
```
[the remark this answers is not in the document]
max-moser: Here, I would interpret an empty string as the absence of a customized text, without the potential nullpointer de-referencing issues.
If there's arguments for allowing a default value of null however (e.g. storage efficiency), then let's go for that.

invenio_rdm_records/requests/access/models.py
```diff
@@ -0,0 +1,134 @@
...
+    )
+    """Creation timestamp."""
+
+    expires_at = db.Column(db.DateTime, nullable=False)
+    """Expiration date."""
+
+    email = db.Column(db.String, nullable=False)
+    """E-mail address for which the access request was made."""
+
+    full_name = db.Column(db.String, nullable=False, default="")
```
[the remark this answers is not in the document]
max-moser: It should be fine – we're not allowing null values to be stored in the DB, and the default value for that field is set to the empty string rather than null.

invenio_rdm_records/views.py
max-moser: True, I knew I missed something :sweat_smile:

invenio_rdm_records/resources/resources.py
```diff
@@ -66,6 +68,16 @@ def p(route):
...
+            route(
+                "POST",
+                p(routes["user-access-request"]),
+                self.create_user_access_request,
+            ),
+            route(
+                "POST",
+                p(routes["guest-access-request"]),
+                self.create_guest_access_request_token,
+            ),
```
jrcastro2, line 80: Minor: IMHO I would move these 2 routes to the `RDMRecordRequestsResource`?

invenio_rdm_records/views.py
max-moser: Fixed!

invenio_rdm_records/resources/resources.py
max-moser: The `RDMRecordRequestsResource` uses the `RecordRequestsService`, which doesn't have the necessary methods (`create_user_access_request`, `create_guest_access_request_token`) available.
These are part of the `RecordAccessService`, which is available as sub-service `current_rdm_records_service.access`.

invenio_rdm_records/resources/errors.py
```diff
@@ -20,3 +23,33 @@ class HTTPJSONValidationWithMessageAsListException(HTTPJSONException):
     def __init__(self, exception):
         """Constructor."""
         super().__init__(code=400, errors=exception.messages)
+
+
+class ExtendedHTTPJSONException(HTTPJSONException):
```
zzacharo, line 28:
```suggestion
class HTTPJSONException(_HTTPJSONException):
```

invenio_rdm_records/services/permissions.py
```diff
@@ -188,3 +200,18 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
...
+
+
+class RDMRequestsPermissionPolicy(RequestPermissionPolicy):
+    """Permission policy for requets, adapted to the needs for RDM-Records."""
+
+    can_read = RequestPermissionPolicy.can_read + [guest_token]
+    can_update = RequestPermissionPolicy.can_update + [guest_token]
+    can_action_submit = RequestPermissionPolicy.can_action_submit + [guest_token]
+    can_action_cancel = RequestPermissionPolicy.can_action_cancel + [guest_token]
+    # can_create_comment = can_read  # TODO fix guest comment workflow & enable this
```
zzacharo, line 217: Issue to create and we can remove the todo I think :)

invenio_rdm_records/services/services.py
```diff
@@ -32,23 +29,25 @@ def __init__(
...
         pids_service=None,
         review_service=None,
     ):
         """Constructor for RecordService."""
         super().__init__(config, files_service, draft_files_service)
-        self._secret_links = secret_links_service
+        self._access = access_service
         self._pids = pids_service
         self._review = review_service
+        self._grants = grants_service
```
zzacharo, line 42: We miss a 
```python
@property
def grants(self):
   """Record grants service."""
    return self._grants
```
? Or I think that's not used anymore? Probably after you merged everything to access_service?

invenio_rdm_records/views.py
```diff
@@ -33,6 +51,78 @@ def init(state):
...
+@blueprint.route("/access-requests/verify")
+def verify_access_request_token():
+    """UI endpoint for verifying guest access request tokens.
+
+    When the token is verified successfully, a new guest access request will be created
+    and the token object will be deleted from the database.
+    The token value will be stored with the newly created request and grant access
+    permissions to the request details.
+    """
+    # TODO check how the email verification logic works and get inspiration from there
```
zzacharo, line 63: still valid?

invenio_rdm_records/services/access/service.py
```diff
@@ -0,0 +1,761 @@
...
+            record_pid=id_,
+            shelf_life=timedelta(hours=6),
+        )
+
+        # TODO `url_for()` doesn't work as is because it's creating an /api/ URL...
+        verify_url = url_for(
+            "invenio_rdm_records_ext.verify_access_request_token",
+            _external=True,
+            **{"access_request_token": access_token.token},
+        ).replace("/api/", "/")
```
ntarocco, line 645: are we 100% sure that the API and UI urls are the same, but the `/api/`? Probably, it just feels hackish... can we use the `_html` link instead?

invenio_rdm_records/services/access/service.py
```diff
@@ -0,0 +1,761 @@
...
+
+        uow.register(
+            EmailOp(
+                receiver=data["email"],
+                subject=_(
+                    "Access request for '%(record_title)s'",
```
ntarocco, line 651: Translations here too

ntarocco, line 651: @rekt-hard is the `subject` automatically prefixing something, e.g. the site name?
```
[ZENODO] bla bla bla
```
What about the footer?
We should make sure that there is coherence in the e-mail content.

invenio_rdm_records/views.py
```diff
@@ -33,6 +51,78 @@ def init(state):
...
+        abort(404)
+
+    url = f"{access_request.links['self_html']}?access_request_token={token}"
+
+    send_email(
+        {
+            "subject": _("Access request submitted successfully"),
+            "html_body": _(
+                (
+                    "Your access request was submitted successfully. "
```
ntarocco, line 89: Translations?

invenio_rdm_records/views.py
```diff
@@ -33,6 +51,78 @@ def init(state):
...
+            url = duplicate_request.links["self_html"]
+            token = duplicate_request.data["payload"]["token"]
+            return redirect(f"{url}?access_request_token={token}")
+
+    if access_request is None:
+        abort(404)
+
+    url = f"{access_request.links['self_html']}?access_request_token={token}"
+
+    send_email(
```
ntarocco, line 84: I would say that we might want to use the notification system instead, so that common login (templating, or maybe header/footer, etc...) will be used.

ntarocco (commented):
Massive work, thanks a lot! It looks clean!

I hope I didn't miss anything, because it is not easy to follow the flow of creation/validation of token in a PR :)
It would be great to have a test at resource level, to make sure that the entire flow works.

invenio_rdm_records/requests/access/models.py
```diff
@@ -0,0 +1,133 @@
...
+    the corresponding ``AccessRequestToken`` will be deleted, and a new
+    guest access request will be created with some of the token's information.
+    """
+
+    __tablename__ = "rdm_records_access_request_tokens"
+
+    id = db.Column(UUIDType, primary_key=True, default=uuid.uuid4)
+    """Access request token ID."""
+
+    token = db.Column(db.String, nullable=False)
```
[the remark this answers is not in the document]
zzacharo: That's true :) @max-moser can you add an explicit length on the string column so MySQL is happy please?

invenio_rdm_records/resources/config.py
```diff
@@ -168,6 +175,15 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
         ValidationErrorWithMessageAsList: create_error_handler(
             lambda e: HTTPJSONValidationWithMessageAsListException(e)
         ),
+        DuplicateAccessRequestError: create_error_handler(
+            # TODO nice suggestion at https://stackoverflow.com/questions/3825990/http-response-code-for-post-when-resource-already-exists
```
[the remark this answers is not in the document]
max-moser: Removed it, FTR it was the following text linking to [StackOverflow](https://stackoverflow.com/questions/3825990/http-response-code-for-post-when-resource-already-exists):
```
# TODO nice suggestion at https://stackoverflow.com/questions/3825990/http-response-code-for-post-when-resource-already-exists
# "I return HTTP 409 with a Location header pointing to the existing/conflicting resource." - Gili
```

invenio_rdm_records/services/services.py
max-moser: Good catch, that's a leftover!

invenio_rdm_records/config.py
```diff
@@ -498,3 +498,10 @@ def make_doi(prefix, record):
 
 RDM_RESOURCE_ACCESS_TOKEN_REQUEST_ARG = "resource_access_token"
 """URL argument to provide resource access token."""
+
+RDM_RECORDS_ACCESS_REQUEST_TOKEN_LENGTH = None
```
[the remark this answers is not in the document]
max-moser: Judging from the documentation of `secrets.token_urlsafe`, it should be okay to have the standard lib determine the size for us:
```
Help on function token_urlsafe in module secrets:

token_urlsafe(nbytes=None)
    Return a random URL-safe text string, in Base64 encoding.
    
    The string has *nbytes* random bytes.  If *nbytes* is ``None``
    or not supplied, a reasonable default is used.
    
    >>> token_urlsafe(16)  #doctest:+SKIP
    'Drmhze6EPcv0fN_81Bj-nA'
```

The idea was to give admins a simple switch to increase the token length if they don't trust the defaults, given how fast processing power can grow.
However, since there's gonna be a follow-up issue for guest comments on the access requests anyway (which may include encoding the external email address into the token), we can always re-add such a functionality later if it's deemed relevant.

invenio_rdm_records/records/systemfields/access/protection.py
```diff
@@ -8,48 +8,61 @@
...
 
+from enum import Enum
+
+
+class Visibility(Enum):
+    """Enum for the available visibility settings of a record's metadata and files."""
+
+    PUBLIC = "public"
+    RESTRICTED = "restricted"
+    PRIVATE = "restricted"
```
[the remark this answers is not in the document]
max-moser: No, it's just an alias for `restricted`, because it feels natural to me to have `private` be the opposite of `public`.
With auto-completion this shouldn't be too much of an issue though, and we can remove that alias if you prefer.

invenio_rdm_records/services/access/service.py
```diff
@@ -0,0 +1,761 @@
...
+        """Create a new instance of the resource unit."""
+        return self.config.link_result_item_cls(*args, **kwargs)
+
+    def link_result_list(self, *args, **kwargs):
+        """Create a new instance of the resource list."""
+        return self.config.link_result_list_cls(*args, **kwargs)
+
+    def grant_result_item(self, *args, **kwargs):
+        """Create a new instance of the resource unit."""
+        # TODO it seems like user profiles are expanded even if they are hidden
```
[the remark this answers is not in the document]
max-moser: As soon as you have a `grant` for a certain `user` in the record (i.e. "you shared access to the record with that user's account"), it seems like you will be able to see that user's profile information by **expanding** the grant (`?expand=1`), no matter if that user has set their profile visibility to public or restricted.
This may leak some information about the user to you (as the record owner) that the user wasn't comfortable with sharing.

Note that the current workflow doesn't allow you to share access to a record with a user unless you can read their profile though.

So this may circumvent a user's privacy settings if they decided to hide their profile only after you've shared access to a record with them!

invenio_rdm_records/services/access/service.py
```diff
@@ -0,0 +1,761 @@
...
+                    model_cls.json["created_by"] == {"user": str(identity.id)},
+                    model_cls.json["topic"] == {"record": id_},
+                )
+                if rm.data and rm.data["type"] == UserAccessRequest.type_id
+            )
+            if request.is_open
+        ]
+
+        if any((req.is_open for req in requests)):
+            raise DuplicateAccessRequestError([str(r.id) for r in requests])
```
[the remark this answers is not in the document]
max-moser: The DB doesn't have such elaborate checks for duplicates, and given that there's some checking of the request status[x] involved I'm not 100% sure it could even really have them.
Without this check, it was entirely possible to have multiple access requests open in parallel even though that doesn't really make sense here.
Maybe it makes sense for a different type of request (potentially even just a future one), but definitely not here :smile: 

[x] We're checking for `is_open`, which depends on the [request type](https://github.com/inveniosoftware/invenio-requests/blob/master/invenio_requests/customizations/request_types.py#L54-L62) in question.

> Might you can just check if requests is empty, given that you are already filtering above request.is_open?

Good catch, this was a leftover :smile:

invenio_rdm_records/services/access/service.py
```diff
@@ -0,0 +1,761 @@
...
+            data=message,
+            uow=uow,
+        )
+
+    @unit_of_work()
+    def create_guest_access_request_token(
+        self, identity, id_, data, expand=False, uow=None
+    ):
+        """Create an request token that can be used to create an access request."""
+        # TODO proper permission checks
```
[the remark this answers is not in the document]
max-moser: The permissions already work as expected, but the code for that could use a bit of face-lifting: Rather than having these awkward code constructs in the `access_service.create_*` methods right now, we should have new generators and new entries in the permission policy.

invenio_rdm_records/services/access/service.py
```diff
@@ -0,0 +1,761 @@
...
+        if authenticated_user in identity.provides:
+            raise PermissionDeniedError("request_guest_access")
+
+        record = self.record_cls.pid.resolve(id_)
+        if current_app.config.get("MAIL_SUPPRESS_SEND", False):
```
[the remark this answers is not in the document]
max-moser: I guess it _can_ be tested with some mocking magic :slightly_smiling_face: 

Right now, we're not preventing anything from happening – we're only logging a warning.
But I'm still a bit worried that the feature won't work as expected on any "real" instances that don't have email sending enabled.

Should we maybe add a feature flag for access requests in general, or just guest requests in particular?

invenio_rdm_records/services/access/service.py
max-moser: No, we're nowhere near 100% certainty about this :smile: 

I would've used `request.links["self_html"]` if we had a `request` result item available here, but unfortunately this isn't the case yet (we've only just created the token which will likely be used to create a request object in the future).
The machinery behind the link dumping would be appropriate to use here, I just didn't get around to look at how to use it without result items.

invenio_rdm_records/views.py
max-moser: Sure, but then we'll need to make sure that the email notification backend is always available because that's the basis for the entire guest-based access request workflow.

invenio_rdm_records/services/access/service.py
```diff
@@ -0,0 +1,761 @@
...
+        )
+
+        # immediately submit the request, unless it has errors
+        # TODO is this correct?
+        if request.errors:
+            return request
+
+        message = {
+            "payload": {
+                "content": data["payload"].get("message") or _("No comment"),
```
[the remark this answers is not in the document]
max-moser: I have no strong feelings about the value to display here, it was just the first thing that came to my mind.

invenio_rdm_records/requests/access/models.py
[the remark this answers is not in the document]
max-moser: Which value do you suggest for it?
On my machine, the generated `secrets.token_urlsafe()` tokens give me the following:
```python
In [2]: len(secrets.token_urlsafe())
Out[2]: 43
```

With a bit of leeway, maybe `80`?

invenio_rdm_records/services/access/service.py
```diff
@@ -0,0 +1,761 @@
...
+    #
+    # Access requests
+    #
+
+    @unit_of_work()
+    def create_user_access_request(
+        self, identity, id_, message, expand=False, uow=None
+    ):
+        """Create a user access request for the given record."""
+        # TODO hide behind "authenticated_user" permission wall
```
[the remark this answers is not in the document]
max-moser: Updated the comments to reflect the current state better

invenio_rdm_records/services/access/service.py
max-moser: Already marked for translation!

invenio_rdm_records/views.py
max-moser: Already marked!

invenio_rdm_records/requests/access/models.py
```diff
@@ -0,0 +1,133 @@
...
+        is_datetime = isinstance(expires_at, datetime)
+        if is_date and not is_datetime:
+            expires_at = datetime.combine(expires_at, datetime.min.time())
+
+        access_request_token = cls(
+            email=email,
+            full_name=full_name,
+            message=message,
+            record_pid=record_pid,
+            expires_at=expires_at,
```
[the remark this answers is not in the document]
max-moser: What should we do if the timezone is set to anything other than UTC? Calculate the UTC timestamp for it instead?
And what should we do if the timezone is **not** set? Assume that it's local time and convert it to UTC, or assume that it's UTC?

Regarding the first, I'm not sure if it makes a difference because the timezone is already set.
For the second question I believe that depends on whether the datetime was created via `datetime.now()` or `datetime.utcnow()` (neither add timezone info). AFAIK InvenioRDM uses `datetime.utcnow()` mostly, so the second assumption would be more reasonable?

invenio_rdm_records/requests/access/requests.py
```diff
@@ -0,0 +1,225 @@
...
+# Copyright (C) 2023 TU Wien.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Access requests for records."""
+
+from flask import current_app, g
+from invenio_access.permissions import authenticated_user, system_identity
+from invenio_i18n import gettext as _
```
max-moser, line 12: NOTE: We can't use `lazy_gettext` here because that will create a `flask_babel.speaklater.LazyString` which will cause issues with the JSONSchema.
In contrast, `gettext` will return `str` which is fine.

invenio_rdm_records/requests/access/requests.py
rekt-hard: If a notification is used, this would be possible out of the box (same goes for recipients and subject).

invenio_rdm_records/services/access/service.py
rekt-hard: As of now, there is nothing like that.
It sure would be nice to have a base template for emails. Maybe also a macro for the subject to prefix, as you mentioned.

invenio_rdm_records/services/access/service.py
rekt-hard: With a notification, it would resolve the request in a celery task and provide a `ResultItem` (at least the dump of it) to be used. In the UI, `me/requests/<request_id>` will resolve properly, if a user has access to a request.

The `self_html` would be usable in many places, if https://github.com/inveniosoftware/invenio-rdm-records/issues/1327 is solved (basically resolving/redirecting based on the user)

invenio_rdm_records/views.py
rekt-hard: I think it would be ok to use the notification system, as we now also rely on the correct configuration of the e-mail client for sending emails. Repository admins would have to actively remove the notification backend, to make this not work. If documented accordingly, this should be fine.

invenio_rdm_records/requests/access/models.py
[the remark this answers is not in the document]
max-moser: Note: For comparison, the secret link tokens in our DB seem to all have a length of `240`.

invenio_rdm_records/requests/access/requests.py
max-moser: Update: Simplified the fields stored in the DB s.t. translation is unnecessary, and removed the "btw here's the access link" comment after accepting the guest access request.
Only the email is still translated; now with `lazy_gettext`.

zzacharo (approved):

PR #1311, opened 2023-05-23
access: fix permissions check for managing access

slint:


invenio_rdm_records/services/schemas/__init__.py
```diff
@@ -43,10 +43,6 @@ class Meta:
 
         unknown = EXCLUDE
 
-    field_load_permissions = {
-        "files": "update_draft",
```
alejandromumo, line 47: question: why was this removed?

alejandromumo (approved):

anikachurilova (approved):
LGTM! Just small questions from us with Manuel.

invenio_rdm_records/services/schemas/__init__.py
slint: I know it looks spooky, but actually this check is redundant. There are only two places currently where someone can change the `files` field of a record (e.g. disable files): `records_service.create` and `records_service.update_draft`. Both of these places already check the exact same permission at the very beginning of the service call, so checking again here is unnecessary.

Besides that, the changes in https://github.com/inveniosoftware/invenio-drafts-resources/pull/230 also contain an even more specific permission check for changing any `files.*` fields.

PR #1323, opened 2023-05-31
resource-types: fix inconsistencies

zzacharo:


invenio_rdm_records/templates/semantic-ui/imprint.html
```diff
@@ -22,9 +22,3 @@
   <dt class="ui tiny header">{{ _("Title")}}</dt>
   <dd>{{field_value.get('title')}}</dd>
 {% endif %}
-
-{% if field_value.get('pages') %}
```
slint, line 26: minor: I think this is a separate fix the leaked in :)

slint (approved):
LGTM, I also verified that all the subtype IDs match their `props.subtype`

slint (approved):
LGTM, I also verified that all the subtype IDs match their `props.subtype`

Issue #1226, opened 2023-03-01
Move some generators to `records-pemissions`


slint:
Good candidates from https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/generators.py:

- `ConditionalGenerator`
- `IfConfig`

_Originally posted by @ppanero in https://github.com/inveniosoftware/invenio-rdm-records/pull/1203#discussion_r1121389534_

PR #1331, opened 2023-06-13
tokens: implement resource access token

anikachurilova:
Note: tests still failing, but the feature itself works, so can be re-reviewed already.

* add feature flag
* handle token validation
* add corresponding permission and generator
* add corresponding scope
* cover with tests
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2233

Get image as anonymous user using valid RAT:
valid pic

Get pdf file as anonymous user using valid RAT:
pdf ok valid

Some errors previews:
Screenshot 2023-06-13 at 17 59 03
Screenshot 2023-06-13 at 18 00 19

### DIFF
--- a/invenio_rdm_records/config.py
+++ b/invenio_rdm_records/config.py
@@ -10,6 +10,8 @@
 
 """DataCite-based data model for Invenio."""
 
+from datetime import timedelta
+
 import idutils
 from invenio_i18n import lazy_gettext as _
 
@@ -479,3 +481,17 @@ For example:
         ...
     }]
 """
+
+# Feature flag to enable/disable Resource access tokens (RATs)
+RDM_RESOURCE_ACCESS_TOKENS_ENABLED = False
+"""Flag to show whether RATs feature should be enabled."""
+
+# Configuration for Resource Access Tokens.
+RDM_RESOURCE_ACCESS_TOKENS_JWT_LIFETIME = timedelta(minutes=30)
+"""Maximum tokens lifetime."""
+
+RDM_RESOURCE_ACCESS_TOKENS_WHITELISTED_JWT_ALGORITHMS = ["HS256", "HS384", "HS512"]
+"""Accepted JWT algorithms for decoding the RAT."""
+
+RDM_RESOURCE_ACCESS_TOKEN_REQUEST_ARG = "resource_access_token"
+"""URL argument to provide resource access token."""
--- a/invenio_rdm_records/ext.py
+++ b/invenio_rdm_records/ext.py
@@ -9,14 +9,10 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 """DataCite-based data model for Invenio."""
-
-from flask import flash, g, request, session
 from flask_iiif import IIIF
 from flask_principal import identity_loaded
-from invenio_i18n import _
 from invenio_records_resources.resources.files import FileResource
 from invenio_records_resources.services import FileService
-from itsdangerous import SignatureExpired
 
 from invenio_rdm_records.oaiserver.resources.config import OAIPMHServerResourceConfig
 from invenio_rdm_records.oaiserver.resources.resources import OAIPMHServerResource
@@ -43,7 +39,6 @@ from .resources import (
     RDMRecordResourceConfig,
 )
 from .resources.resources import RDMRecordCommunitiesResource, RDMRecordRequestsResource
-from .secret_links import LinkNeed, SecretLink
 from .services import (
     CommunityRecordsService,
     IIIFService,
@@ -59,34 +54,13 @@ from .services import (
 )
 from .services.pids import PIDManager, PIDsService
 from .services.review.service import ReviewService
-
-
-def verify_token():
-    """Verify the token and store it in the session if it's valid."""
-    token = request.args.get("token", None)
-    if token:
-        try:
-            data = SecretLink.load_token(token)
-            if data:
-                session["rdm-records-token"] = data
-
-                # the identity is loaded before this handler is executed
-                # so if we want the initial request to be authorized,
-                # we need to add the LinkNeed here
-                if hasattr(g, "identity"):
-                    g.identity.provides.add(LinkNeed(data["id"]))
-
-        except SignatureExpired:
-            session.pop("rdm-records-token", None)
-            flash(_("Your shared link has expired."))
+from .utils import verify_token
 
 
 @identity_loaded.connect
-def on_identity_loaded(sender, identity):
-    """Add the secret link token need to the freshly loaded Identity."""
-    token_data = session.get("rdm-records-token")
-    if token_data:
-        identity.provides.add(LinkNeed(token_data["id"]))
+def on_identity_loaded(_, identity):
+    """Add secret link token or resource access token need to the freshly loaded Identity."""
+    verify_token(identity)
 
 
 from flask import Blueprint
@@ -112,7 +86,6 @@ class InvenioRDMRecords(object):
         self.init_config(app)
         self.init_services(app)
         self.init_resource(app)
-        app.before_request(verify_token)
         app.extensions["invenio-rdm-records"] = self
         app.register_blueprint(blueprint)
         # Load flask IIIF
--- a/invenio_rdm_records/services/generators.py
+++ b/invenio_rdm_records/services/generators.py
@@ -25,6 +25,8 @@ from invenio_search.engine import dsl
 
 from invenio_rdm_records.records import RDMDraft
 
+from ..tokens.permissions import RATNeed
+
 
 class ConditionalGenerator(Generator):
     """Generator that depends on whether a condition is true or not.
@@ -270,3 +272,21 @@ class IfConfig(ConditionalGenerator):
     def _condition(self, **_):
         """Check if the config value is truthy."""
         return current_app.config.get(self.config_key) in self.accept_values
+
+
+class ResourceAccessToken(Generator):
+    """Allow resource access tokens."""
+
+    def __init__(self, access):
+        """Constructor."""
+        self.access = access
+
+    def needs(self, record=None, file_key=None, **kwargs):
+        """Enabling Needs."""
+        record_owner = next(iter(record.parent.access.owners), None)
+
+        if record_owner and record and file_key:
+            signer_id = record_owner.owner_id
+            return [RATNeed(signer_id, record["id"], file_key, self.access)]
+
+        return []
--- a/invenio_rdm_records/services/permissions.py
+++ b/invenio_rdm_records/services/permissions.py
@@ -25,6 +25,7 @@ from .generators import (
     IfRestricted,
     RecordCommunitiesAction,
     RecordOwners,
+    ResourceAccessToken,
     SecretLinks,
     SubmissionReviewer,
 )
@@ -84,6 +85,7 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     # Allow reading the files of a record
     can_read_files = [
         IfRestricted("files", then_=can_view, else_=can_all),
+        ResourceAccessToken("read"),
     ]
     can_get_content_files = [
         # note: even though this is closer to business logic than permissions,
@@ -101,7 +103,7 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     # Allow reading metadata of a draft
     can_read_draft = can_preview
     # Allow reading files of a draft
-    can_draft_read_files = can_preview
+    can_draft_read_files = can_preview + [ResourceAccessToken("read")]
     # Allow updating metadata of a draft
     can_update_draft = can_review
     # Allow uploading, updating and deleting files in drafts
@@ -112,7 +114,7 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     ]
     can_draft_get_content_files = [
         # preview is same as read_files
-        IfFileIsLocal(then_=can_preview, else_=[SystemProcess()])
+        IfFileIsLocal(then_=can_draft_read_files, else_=[SystemProcess()])
     ]
     can_draft_commit_files = [
         # review is the same as create_files
--- a/invenio_rdm_records/tokens/errors.py
+++ b/invenio_rdm_records/tokens/errors.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Resource access tokens module errors."""
+from invenio_i18n import _
+from invenio_rest.errors import RESTException
+
+
+class ResourceAccessTokenError(RESTException):
+    """Resource access token base error class."""
+
+    code = 400
+
+
+class MissingTokenIDError(ResourceAccessTokenError):
+    """Resource access token error for missing token ID."""
+
+    description = _('Missing "kid" key with personal access token ID in JWT header.')
+
+
+class InvalidTokenIDError(ResourceAccessTokenError):
+    """Resource access token error for invalid token ID."""
+
+    description = _('"kid" JWT header value not a valid personal access token ID.')
+
+
+class InvalidTokenError(ResourceAccessTokenError):
+    """Resource access token error for invalid token."""
+
+    description = _("The token is invalid.")
+
+
+class ExpiredTokenError(InvalidTokenError):
+    """Resource access token error for expired token."""
+
+    description = _("The token is expired.")
+
+
+class RATFeatureDisabledError(ResourceAccessTokenError):
+    """Resource access token error for disabled feature."""
+
+    description = _("Resource Access Tokens feature is currently disabled.")
--- a/invenio_rdm_records/tokens/permissions.py
+++ b/invenio_rdm_records/tokens/permissions.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Permissions for Resource Access Token."""
+
+from collections import namedtuple
+from functools import partial
+
+_Need = namedtuple(
+    "Need", ["method", "signer_id", "pid_value", "file_key", "permission"]
+)
+RATNeed = partial(_Need, "rat")
+
+"""
+A Resource Access Tokens Need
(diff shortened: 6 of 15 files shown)


invenio_rdm_records/tokens/resource_access.py
```diff
@@ -0,0 +1,92 @@
...
+            raise MissingTokenIDError()
+        if not access_token_id.isdigit():
+            raise InvalidTokenIDError()
+
+        access_token = Token.query.get(int(access_token_id))
+        if (
+            not access_token
+            or tokens_generate_scope.id not in access_token.scopes
+            or (
+                access_token.expires is not None  # expired PAT check
```
anikachurilova, line 51: Added PAT expiration check here as it seems valid to me. Not present in Zenodo implementation though, so feel free to comment on removal

invenio_rdm_records/services/permissions.py
```diff
@@ -88,7 +89,9 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     can_get_content_files = [
         # note: even though this is closer to business logic than permissions,
         # it was simpler and less coupling to implement this as permission check
-        IfFileIsLocal(then_=can_read_files, else_=[SystemProcess()])
+        IfFileIsLocal(
+            then_=can_read_files + [ResourceAccessToken()], else_=[SystemProcess()]
```
anikachurilova, line 93: Note: added RAT permission only to the `then_` of `IfFileIsLocal`, as in other case the resource is available only for `SystemProcess`

Question: should we allow reading the resource with a RAT even for non-local files?

invenio_rdm_records/ext.py
```diff
@@ -75,7 +80,31 @@ def verify_token():
...
+                filename = (
+                    request.view_args.get("filename")
+                    or request.view_args.get("key")
+                    or None
+                )
+                pid_value = request.view_args.get("pid_value", None)
+
+                if pid_value and filename:
+                    try:
+                        record = RDMDraft.pid.resolve(pid_value, registered_only=False)
```
anikachurilova, line 97: Not proud of this being here and not in the validation method.... but it causes dependency loop in case of importing `RDMDraft` and `RDMRecord` in the `ResourceAccessToken` class

invenio_rdm_records/ext.py
```diff
@@ -75,7 +80,31 @@ def verify_token():
                 # we need to add the LinkNeed here
                 if hasattr(g, "identity"):
                     g.identity.provides.add(LinkNeed(data["id"]))
-
+            else:
```
anikachurilova, line 83: Note: in Zenodo token data is provided in the `token=` args of URL. In RDM it's "reserved" for SecretLinks. 

Suggestion/question: should we force users to put something like `resource-access-token=` to the URL args, of this `else` here is fine?

invenio_rdm_records/services/permissions.py
ppanero: I think is okay, at the moment I don't see more use cases for `else_` than the fetch files, which will become "local" once downloaded.

invenio_rdm_records/tokens/errors.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Resource access tokens module errors."""
+
+from __future__ import absolute_import, print_function
```
ppanero, line 10: this is for python 2.7, should be removed in all ocurrences 🔥  🚒

invenio_rdm_records/tokens/resource_access.py
```diff
@@ -0,0 +1,92 @@
...
+        except jwt.InvalidTokenError:
+            raise InvalidTokenError()
+
+        if not access_token_id:
+            raise MissingTokenIDError()
+        if not access_token_id.isdigit():
+            raise InvalidTokenIDError()
+
+        access_token = Token.query.get(int(access_token_id))
+        if (
```
ppanero, line 47: could this be split into semantically meaningful variables, e.g.:

```python
is_expired = access_token.expires is not None and access_token.expires < datetime.utcnow()
...
```

tests/tokens/test_resource_access.py
```diff
@@ -0,0 +1,699 @@
...
+    return recid
+
+
+def test_rat_validation_failed(
+    app, db, users, uploader, superuser_identity, minimal_record, oauth2_client
+):
+    """Test possible validation failings of Resource access tokens."""
+    parent = RDMParent.create({"access": {"owned_by": [{"user": users[0].id}]}})
+    record = RDMRecord.create(minimal_record, parent=parent)
+
```
ppanero, line 111: nit: add a comment why each case should fail, to ease reading (e.g. file does not exist, not owner, wrong kid, etc.). In addition, there are many cases in this one, when if one fails the subsequent wont' be tested. Might be worth spliting to one test per check.

ppanero (commented):
LGTM/Us, very nice job! 🚀 

Mostly minor or structural comments (no major on functionality). However, I'd like someone else's eyes on this since it affects permissions.

@kpsherva @slint structure discussion here (nice catch by @anikachurilova!)

In general we define `resources`, `services` and `records` modules at root level. Therefore, the contents of `/tokens` should be split. What we thought was:

- `records/tokens/scopes.py`
- `records/tokens/resource_access.py`
- `records/tokens/errors.py`
- `services/tokens/permissions.py`

My main doubt is `resource_access` since it imports `errors.py`. To this moment we had not. defined errors in any module at `records` level. Those have mostly been sqlalchemy ones (e.g. `NoResultFound`). But I think it is okay.

Extra: the root level folder for `secret_links` should also be moved inside `records/`

invenio_rdm_records/services/generators.py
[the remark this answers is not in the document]
slint: I guess you mean https://github.com/inveniosoftware/invenio-rdm-records/issues/1226 :)

It is worth it, but it's non-critical for now.

invenio_rdm_records/ext.py
slint: The name of the query string parameter could be configurable. The issue we currently have in Zenodo is that it is "hard-coded" and used by legacy clients, in which case we need to consider that it could collide with secret link tokens. So the `else` is unfortunately necessary for now. 

We might be able to mitigate this though since we're passing everything via the legacy REST API layer in ZenodoRDM, and thus we could perform the parameter "rename" on that level... The issue is that this handler runs in `before_request`, so there's little room to interject.

invenio_rdm_records/ext.py
```diff
@@ -75,7 +80,31 @@ def verify_token():
...
+                if not current_app.config.get("RDM_RESOURCE_ACCESS_TOKENS_ENABLED"):
+                    raise RATFeatureDisabledError()
+                # for records: "api/records/<pid_value>/files/<filename>/content"
+                # for drafts: "api/records/<pid_value>/draft/files/<key>/content"
+                filename = (
+                    request.view_args.get("filename")
+                    or request.view_args.get("key")
+                    or None
+                )
+                pid_value = request.view_args.get("pid_value", None)
```
slint, line 93: major: do we need to fetch and resolve these objects here? I would kind of "blindly" assume that the JWT payload includes the "claimed" subject and create a `RATNeed(**subject)`. Later in the permissions generator, these can be explicitly checked.

To translate into "human" terms, the flow would look something like:

- Client: *Hey, here's a RAT token, in which I claim that I can access as `user: 1234` on a record with `id: "m5ykg-ps142"`, the file with `key: "data.zip"`*
- Server: Ok... I verify that your token is indeed validly signed, so you can go on with your claims
- Client: *Noice*
- (...a few moments later in the permissions check of `get_content_files` of `/api/records/m5ykg-ps142/draft/files/data.zip`...)
- Server: Ok, from the permission policy it says I either need the **owner of the record**, an **admin role**, or someone with a RAT token that claims he can access as the owner of the record (`user: 1234`) this specific record (`id: "m5ykg-ps142"`) and file (`key: "data.zip"`)
- Client: *Oh boi, guess what I happen to have in my pocket* 😏
- Server: LGTM, here's your file contents 🫡

invenio_rdm_records/services/generators.py
```diff
@@ -262,3 +264,14 @@ def __init__(self, config_key, accept_values=[True], **kwargs):
...
         return current_app.config.get(self.config_key) in self.accept_values
+
+
+class ResourceAccessToken(Generator):
+    """Allow resource access tokens."""
+
+    def needs(self, record=None, file_key=None, **kwargs):
+        """Enabling Needs."""
+        if record and file_key:
+            return [RATNeed("get_content_files", record["id"], file_key)]
```
slint, line 275: minor: this could also be more generic than `get_content_files`, so something like `files`/`draft_files`, `record`/`draft_record`.

invenio_rdm_records/tokens/resource_access.py
```diff
@@ -0,0 +1,92 @@
...
+
+        rat_singer = access_token.user
+        payload = payload["sub"]
+
+        valid_claim = (
+            rat_singer.id == record.parent.access.owned_by[0].owner_id
+            and payload["record_id"] == record["id"]
+            and payload["file"] == filename
+            and payload["access"] == "read"
+        )
```
slint, line 87: I think this is not necessary here since we're only doing "formal" verification of the JWT signature, expiration, and structure. In that fashion, maybe you just verify that `access` is `read`/`write`, etc. and that at least a `record_id` key exists.

Thus the `record` and `filename` parameters are not needed in this function (see comment above about when/where to verify this).

slint (commented):
Overall looks very clean, thanks a lot for porting this from legacy Zenodo :)

My major gripe is with the extra validation/check when adding the `RATNeed` actually. I think though it would be easy to simplify, all the classes are there.

(Let me know if checking IRL would be easier)

invenio_rdm_records/services/generators.py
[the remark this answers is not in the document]
ppanero: hehe yes 😂 sorry for re-complaining

invenio_rdm_records/ext.py
anikachurilova: What do you think if I make both RAT url arg and SecretLink url arg configurable? Or is it not possible because can be used by legacy clients?

invenio_rdm_records/ext.py
```diff
@@ -75,7 +80,31 @@ def verify_token():
...
+
+                if pid_value and filename:
+                    try:
+                        record = RDMDraft.pid.resolve(pid_value, registered_only=False)
+                    except (PIDUnregistered, NoResultFound):
+                        record = RDMRecord.pid.resolve(pid_value)
```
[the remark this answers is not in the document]
anikachurilova: Tested and shown irl, the case will give the `PermissionDenied` because of the filename check in the validaion method.

invenio_rdm_records/tokens/resource_access.py
anikachurilova: If PATs never expire, this check indeed doesn't make sense, thanks for noticing!

invenio_rdm_records/ext.py
anikachurilova: Nevermind... discussed irl, decided to make separate `resource-access-token=` url arg and move it to config to avoid hard-code

invenio_rdm_records/config.py
```diff
@@ -479,3 +481,13 @@ def make_doi(prefix, record):
...
+
+# Feature flag to enable/disable Resource access tokens (RATs)
+RDM_RESOURCE_ACCESS_TOKENS_ENABLED = True
+
+# Configuration for Resource Access Tokens.
+RDM_RESOURCE_ACCESS_TOKENS_JWT_LIFETIME = timedelta(minutes=30)
+"""Maximum tokens lifetime."""
+
+RDM_RESOURCE_ACCESS_TOKENS_WHITELISTED_JWT_ALGORITHMS = ["HS256", "HS384", "HS512"]
+"""Accepted JWT algorithms for the ."""
```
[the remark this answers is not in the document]
anikachurilova: Top secret mystery

invenio_rdm_records/ext.py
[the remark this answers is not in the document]
anikachurilova: In the end moved it to `utils`, WTYD?

invenio_rdm_records/ext.py
```diff
@@ -112,7 +86,6 @@ def init_app(self, app):
         self.init_config(app)
         self.init_services(app)
         self.init_resource(app)
-        app.before_request(verify_token)
```
anikachurilova, line 115: Making changes to the identity from the `before_request` here and from the `on_identity_loaded` signal handler can cause issues because the order of execution is not controlled. The way to avoid this is to move identity logic to 1 place --> inside the signal handler `on_identity_loaded` (see above)

invenio_rdm_records/tokens/permissions.py
```diff
@@ -0,0 +1,18 @@
...
+
+from collections import namedtuple
+
+RATNeed = namedtuple(
+    "RATNeed", ["method", "signer_id", "pid_value", "file_key", "permission"]
+)
+"""
+A Resource Access Tokens Need
+ex: RATNeed("files", "3", "byy8e-7yz93", "file5.png", "edit")
+"""
```
slint, line 18:
```suggestion

_Need = namedtuple("Need", ["method", "signer_id", "pid_value", "file_key", "permission"]

RATNeed = partial(_Need, "rat")
"""
A Resource Access Tokens Need
ex: RATNeed("3", "byy8e-7yz93", "file5.png", "write_file")
"""
```

minor: two changes:

- Define a "base" `_Need` as done e.g. [for communtiies](https://github.com/inveniosoftware/invenio-communities/blob/49b69d4483db3a56c68a40dfb294966403512c45/invenio_communities/generators.py#L27-L30).
- the `permission` value can be a bit more explicit. imagine this could be used for all sorts of other resources/granularity so you could have values `read_record` for only viewing metadata, `write_file` for only writing a single file, `read` to read record + files. No need to change the current generators though (see relevant comment).

Don't worry too much about the Zenodo legacy values that are in the RFC/codebase, we can easily configure these to match things in legacy.

invenio_rdm_records/services/generators.py
```diff
@@ -262,3 +264,21 @@ def __init__(self, config_key, accept_values=[True], **kwargs):
...
+        """Constructor."""
+        self.access = access
+
+    def needs(self, record=None, file_key=None, **kwargs):
+        """Enabling Needs."""
+        record_owner = super().needs(record)
+
+        if record_owner and record and file_key:
+            signer_id = record_owner[0].value
+            return [RATNeed("files", signer_id, record["id"], file_key, self.access)]
```
slint, line 282:
```suggestion
            return [RATNeed(signer_id, record["id"], file_key, self.access)]
```

minor: so that it's not file-specific (see also the comment on defining `RATNeed`). The `self.access` should be enough to define on what level we want to apply the permission.

invenio_rdm_records/services/generators.py
```diff
@@ -262,3 +264,21 @@ def __init__(self, config_key, accept_values=[True], **kwargs):
...
+    def __init__(self, access):
+        """Constructor."""
+        self.access = access
+
+    def needs(self, record=None, file_key=None, **kwargs):
+        """Enabling Needs."""
+        record_owner = super().needs(record)
+
+        if record_owner and record and file_key:
+            signer_id = record_owner[0].value
```
slint, line 281:
```suggestion
class ResourceAccessToken(Generator):
    """Allow resource access tokens."""

    def __init__(self, access):
        """Constructor."""
        self.access = access

    def needs(self, record=None, file_key=None, **kwargs):
        """Enabling Needs."""
        record_owner = next(iter(record.parent.access.owners), None)

        if record_owner and record and file_key:
            signer_id = record_owner
```

nit: in the end, it looks a bit awkward to get the needs from the `RecordOwners` generator, and then "extract" the user ID from them to use again here, so duplicating might be simpler and more readable. 

The alternative would be to extract the logic in a `RecordOwners.get_owners(self, record)` method, but it's a one-liner, so it feels a bit too much.

invenio_rdm_records/config.py
```diff
@@ -479,3 +481,17 @@ def make_doi(prefix, record):
         ...
     }]
 """
+
+# Feature flag to enable/disable Resource access tokens (RATs)
+RDM_RESOURCE_ACCESS_TOKENS_ENABLED = True
```
slint, line 486:
```suggestion
RDM_RESOURCE_ACCESS_TOKENS_ENABLED = False
```

minor: maybe disable by default (it also looks like this is the assumption from the tests)

slint (commented):
Great job, looks super clean 💅

Some last comments on defaults and further clarity, and it's good to merge.

invenio_rdm_records/tokens/resource_access.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class ResourceAccessToken:
+    """Resource access tokens."""
+
+    def __init__(self, token):
+        """Initialize RAT token."""
+        self.token = token
+
+    def validate_rat(self):
```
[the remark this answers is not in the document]
anikachurilova: I agree, this can be a simple utility as it was in Zenodo. It became a class as i tried to map it to resource/service logic, but it didn't make sense here.

invenio_rdm_records/tokens/scopes.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""OAuth2 tokens scopes."""
+
+from invenio_i18n import lazy_gettext as _
+from invenio_oauth2server.models import Scope
+
+tokens_generate_scope = Scope(
```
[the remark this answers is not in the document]
anikachurilova: Is this necessary though? I saw that example, but since the scope is not visible in the UI, this is just a string that has to match.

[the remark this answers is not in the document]
slint: There are a couple of places besides the UI (e.g. `invenio tokens create ...`) that validate the scopes that are passed. The CLI specifically is actually the way we used to create these "hidden" tokens. I'm not sure also if the validation would silently fail (which would be bad in case of a typo), but in any case, it's better for consistency and testing as well (I missed actually in the tests that you're passing `_scopes=...` to the model, which bypasses the validation 😏 )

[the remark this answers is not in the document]
anikachurilova: Very good catch, thanks! Fixed both scopes and tests 🐀

slint (approved):
Ship it 💯 🥇 🍾

PR #1334, opened 2023-06-21
deposit form: enable file modifications depending on external DOi

anikachurilova:
* disable "New version" button on extrenal DOI
* enable file upload components on external DOI
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2259

Warning message preview:
Screenshot 2023-06-21 at 11 01 53

File upload box for records with external DOI:
Screenshot 2023-06-21 at 11 02 48

For the comparison: file upload box for records with our DOI (not touched):
Screenshot 2023-06-21 at 11 03 25

### DIFF
--- a/invenio_rdm_records/config.py
+++ b/invenio_rdm_records/config.py
@@ -479,3 +479,11 @@ For example:
         ...
     }]
 """
+
+# Feature flag to enable/disable locking file modifications for record with external DOIs on edit record action.
+RDM_LOCK_FILES_FOR_EXTERNAL_DOI = False
+"""Flag to show whether add/delete files of a record on edit action should be locked."""
+
+# Configuration for locking files for external DOIs.
+RDM_EXTERNAL_DOI_PROVIDER_KEY = "external"
+"""The keyword that represents external DOIs provider."""
--- a/invenio_rdm_records/services/generators.py
+++ b/invenio_rdm_records/services/generators.py
@@ -16,9 +16,10 @@ from itertools import chain
 
 from flask import current_app
 from flask_principal import UserNeed
-from invenio_access.permissions import authenticated_user
 from invenio_communities.generators import CommunityRoleNeed, CommunityRoles
 from invenio_communities.proxies import current_roles
+
+# from invenio_records.dictutils import dict_lookup
 from invenio_records_permissions.generators import Generator
 from invenio_records_resources.services.files.transfer import TransferType
 from invenio_search.engine import dsl
@@ -134,7 +135,7 @@ class IfDraft(ConditionalGenerator):
 
 
 class IfFileIsLocal(ConditionalGenerator):
-    """Conditional generator for file storage class."""
+    """Generator for file storage class."""
 
     def _condition(self, record, file_key=None, **kwargs):
         """Check if the record is a draft."""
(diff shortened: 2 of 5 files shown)


invenio_rdm_records/config.py
```diff
@@ -479,3 +479,11 @@ def make_doi(prefix, record):
         ...
     }]
 """
+
+# Feature flag to enable/disable locking file modifications for record with external DOIs on edit record action.
+RDM_LOCK_FILES_FOR_EXTERNAL_DOI = False
+"""Flag to show whether add/delete files of a record on edit action should be locked."""
+
+# Configuration for locking files for external DOIs.
+RDM_EXTERNAL_DOI_PROVIDER_KEY = "external"
```
anikachurilova, line 488: Put this to config, to not have this string hardcoded in app, but it is already hardcoded in many places, as well as the `datacite` provider key. Should use this config variable everywhere instead, WDYT?

invenio_rdm_records/services/generators.py
```diff
@@ -155,6 +155,26 @@ def _condition(self, record, file_key=None, **kwargs):
         return is_file_local
 
 
+# class IfRecordWithExternalDOI(ConditionalGenerator):
```
anikachurilova, line 158: Note: commented line through the PR - the approach using permissions. 
Question for the reviewers: which approach seems better?

kpsherva: commented in another PR: I think this approach with permissions is cleaner, however it would be nice to hear from @slint as well :)

tmorrell: I agree. Then you could add a bit more logic here to handle RDM_ENABLE_VERSIONS_FOR_EXTERNAL_DOI and pass permissions for both can_draft_modify_files and draft_show_version_button.

invenio_rdm_records/services/generators.py
```diff
@@ -155,6 +156,26 @@ def _condition(self, record, file_key=None, **kwargs):
...
+#     """Generator that depends on whether the record has external DOI or not."""
+#
+#     def _condition(self, record, **kwargs):
+#         """Check if the record has external DOI."""
+#         is_feature_enabled = not current_app.config["RDM_LOCK_FILES_FOR_EXTERNAL_DOI"]
```
tmorrell, line 164: I'm also wondering if RDM_UNLOCK_FILES_EXTERNAL_DOI might be a bit clearer, since we could avoid the `not` and map it clearly to the feature. Then the other one would be RDM_DISABLE_VERSIONS_EXTERNAL_DOI

invenio_rdm_records/services/generators.py
```diff
@@ -16,9 +16,10 @@
 
 from flask import current_app
 from flask_principal import UserNeed
-from invenio_access.permissions import authenticated_user
 from invenio_communities.generators import CommunityRoleNeed, CommunityRoles
 from invenio_communities.proxies import current_roles
+
+# from invenio_records.dictutils import dict_lookup
```
ppanero, line 22:
```suggestion
```

invenio_rdm_records/services/permissions.py
```diff
@@ -19,7 +19,7 @@
 )
 from invenio_records_permissions.policies.records import RecordPermissionPolicy
 
-from .generators import (
+from .generators import (  # IfRecordWithExternalDOI,
```
ppanero, line 22:
```suggestion
from .generators import (
```

ppanero (commented):
This PR LGTM in general, there are some discussions to have and as much as it sounds cliche I'd +1 almost every comment form Tom 😅

invenio_rdm_records/services/generators.py
anikachurilova: This is used in the commented part (permissions)

kpsherva:
closing as alternative approach was decided to follow

PR #1337, opened 2023-06-26
Add invenio GitHub

alejandromumo:
closes https://github.com/inveniosoftware/invenio-github/issues/98


alejandromumo:
Tests are failing because `invenio-github` version is still the old one and has some old dependencies. 

We need to:

- Merge https://github.com/inveniosoftware/invenio-webhooks/issues/65 ( `invenio-webhooks`)
- Release `invenio-webhooks`
- Bump `invenio-webhooks` in `invenio-github`(https://github.com/inveniosoftware/invenio-github/issues/98)
- Release https://github.com/inveniosoftware/invenio-github/issues/98
- Bump `invenio-github` in the current PR

invenio_rdm_records/github/release.py
```diff
@@ -0,0 +1,315 @@
...
+from invenio_pidstore.models import PersistentIdentifier
+from invenio_rdm_records.github.utils import get_extra_metadata, load_citation_metadata
+from invenio_rdm_records.proxies import current_rdm_records_service
+from invenio_records_resources.services.uow import UnitOfWork
+
+
+class RDMGithubRelease(GitHubRelease):
+    """Implement release API instance for RDM."""
+
+    def _related_identifiers(self):
```
ppanero, line 29: all the `_` method could be an API class with properties that extract from the actual GH release object, becoming a wapper API class with RDM-Records specific needs.

invenio_rdm_records/github/release.py
```diff
@@ -0,0 +1,315 @@
...
+        - The draft's ownership is set to the user's id via its parent.
+        - Upload files to the draft.
+        - Publish the draft.
+
+        In case of failure, the transaction is rolled back and the release status set to 'FAILED'
+
+        :raises ex: any exception generated by the records service (e.g. invalid metadata)
+        """
+        try:
+            self.release_object.status = ReleaseStatus.PROCESSING
```
ppanero, line 231: Does this reflect on the DB without any session.add or commits?

invenio_rdm_records/github/release.py
```diff
@@ -0,0 +1,315 @@
...
+
+from invenio_access.permissions import system_identity
+from invenio_db import db
+from invenio_pidstore.models import PersistentIdentifier
+from invenio_rdm_records.github.utils import get_extra_metadata, load_citation_metadata
+from invenio_rdm_records.proxies import current_rdm_records_service
+from invenio_records_resources.services.uow import UnitOfWork
+
+
+class RDMGithubRelease(GitHubRelease):
```
ppanero, line 26: This class seems to:

- Extract information from a release and create a dictionary of data via underscored methods (see comment below)
- Do record resolution manually, e.g. calls to `PersistentIdentifiers` and implement a pseudo resolver. Can we use existing resolvers/api classes?
- Process the release (missing validation?)
- Publish the release via rdm-records services

It seems to be a bit cluttered between an api class and a "service" (does not have to be a full fledge one, but split what's an object with information and data layer and what's management of it with service calls).

invenio_rdm_records/github/release.py
```diff
@@ -0,0 +1,315 @@
...
+                        self.user_identity, new_version_draft.id, data, uow=uow
+                    )
+
+                # Validate the release files are fetchable
+                self._test_zipball()
+
+                # Upload files to draft
+                draft_file_service = current_rdm_records_service.draft_files
+
+                draft_file_service.init_files(
```
ppanero, line 265: I know we discuss it, so this is just curiosity. Why not use the FETCH files? need to know when to comit?

setup.cfg
```diff
@@ -42,9 +42,11 @@ install_requires =
     invenio-i18n>=2.0.0,<3.0.0
     invenio-oaiserver>=2.0.0,<3.0.0
     invenio-vocabularies>=1.2.0,<2.0.0
+    nameparser>=1.1.1
     pytz>=2020.4
     pyyaml>=5.4.0
     python-slugify>=8.0.1
+    invenio-github>=1.0.0a8 # TODO actually bump it
```
ppanero, line 49: not ordered alphabetically

ppanero (commented):
Very nice job! massive including the invenio-github PR. Could I ask you to transform the `TODO` into issues? we know `TODO` die burried in the code.

*Needs tests* 😇

invenio_rdm_records/github/release.py
```diff
@@ -0,0 +1,315 @@
...
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""Github release API implementation."""
+
+import requests
+import yaml
+from flask import current_app
+from invenio_github.api import GitHubRelease
+from invenio_github.models import ReleaseStatus
```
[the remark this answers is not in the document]
alejandromumo: Do you mean not using `invenio_github.models.ReleaseStatus` and, instead, providing API access to it? E.g. in `invenio_github.api`?

invenio_rdm_records/github/release.py
```diff
@@ -0,0 +1,315 @@
...
+            return record
+        except Exception as ex:
+            # Flag release as FAILED and raise the exception
+            self.release_object.status = ReleaseStatus.FAILED
+            raise ex
+
+    def process_release(self):
+        """Processes a github release.
+
+        The release might be first validated, in terms of sender, and then published.
```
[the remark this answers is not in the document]
alejandromumo: Yes. Basically, I and @slint were discussing whether the public interface of `GithubRelease` should be just a `publish` method or also a `process_release`. But then we agreed that maybe the release "manager" (in this case in RDM) might want to do something before publishing the record - even though for now the release processing only publishes a record.

Conceptually, processing the release and publishing the record are distinct actions thus we separated them.

invenio_rdm_records/github/release.py
alejandromumo: Point-by-point replies:

1 - totally agree :)
2 - I ended up using `PersistentIdentifier.get_by_object_uuid` because we associate the record id (UUID) to the release as a weak reference to the record. Therefore, when we have a release with an associated record UUID, we can't use the PID resolver as it expects a `pid_value` and not `object_uuid`.  I did not find any resolver for such a case but maybe I missed it
3 - see the reply below. This method is part of the public interface of `invenio_github.api.GithubRelease` and, in this case, simply calls the record publishing part. Which validation do you refer to?

invenio_rdm_records/github/release.py
alejandromumo: I discussed it with @slint, iirc we went for the simpler option, for now, to integrate this as soon as possible. 

But if I recall correctly, the authentication part was the major block we thought of. Right now we authenticate the user through the github client and then we request the release file.

invenio_rdm_records/github/release.py
alejandromumo: You are right, we have to commit after changing the status

invenio_rdm_records/services/github/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+        """Extract description from a release.
+
+        If the relesae does not have any body, the repository description is used.
+        Falls back for "No description provided".
+        """
+        if self.rdm_release.release_payload.get("body"):
+            return markdown(self.rdm_release.release_payload["body"])
+        elif self.rdm_release.repository_payload.get("description"):
+            return self.rdm_release.repository_payload["description"]
+        return "No description provided."
```
zzacharo, line 58: translation?

invenio_rdm_records/services/github/metadata.py
```diff
@@ -0,0 +1,167 @@
...
+    def load_citation_metadata(self, citation_data):
+        """Get the metadata file."""
+        if not citation_data:
+            return {}
+
+        citation_schema = current_app.config.get("GITHUB_CITATION_METADATA_SCHEMA")
+
+        assert isinstance(
+            citation_schema, Schema
+        ), "Citation schema is needed to load citation metadata."
```
zzacharo, line 163: translation?

PR #1338, opened 2023-06-27
permissions: fix checks when new record

slint:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2262
* Fixes the permission check for cases where there is no record object to make checks against (e.g. when rendering the empty draft form).


jrcastro2 (approved):
LGTM!

kpsherva (approved):

PR #1385, opened 2023-07-26
access: add endpoint for updating parent access settings

anikachurilova:
* implement service layer method to update parent access settings
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1384

### DIFF
--- a/invenio_rdm_records/records/systemfields/access/field/parent.py
+++ b/invenio_rdm_records/records/systemfields/access/field/parent.py
@@ -8,39 +8,12 @@
 
 """Access system field."""
 
-from flask import current_app
 from invenio_records.systemfields import SystemField
 
+from ..access_settings import AccessSettings
 from ..grants import Grants
 from ..links import Links
 from ..owners import Owner
-
-
-class AccessSettings:
-    """Access settings for a parent record."""
-
-    def __init__(self, settings_dict):
-        """Constructor."""
-        self.allow_user_requests = settings_dict.get("allow_user_requests", False)
-        self.allow_guest_requests = settings_dict.get("allow_guest_requests", False)
-        self.accept_conditions_text = settings_dict.get("accept_conditions_text", None)
-
-    def dump(self):
-        """Dump the record as dictionary."""
-        return {
-            "allow_user_requests": self.allow_user_requests,
-            "allow_guest_requests": self.allow_guest_requests,
-            "accept_conditions_text": self.accept_conditions_text,
-        }
-
-    def __repr__(self):
-        """Return repr(self)."""
-        return "<{} requests: {}/{}, text: {}>".format(
-            type(self).__name__,
-            self.allow_guest_requests,
-            self.allow_user_requests,
-            bool(self.accept_conditions_text),
-        )
 
 
 class ParentRecordAccess:
@@ -84,7 +57,7 @@ class ParentRecordAccess:
         self._owned_by = owned_by if owned_by else owner_cls(None)
         self.grants = grants if grants else grants_cls()
         self.links = links if links else links_cls()
-        self.settings = settings if settings else settings_cls({})
+        self._settings = settings if settings else settings_cls({})
         self.errors = []
 
     @property
@@ -106,6 +79,16 @@ class ParentRecordAccess:
     def owner(self, value):
         """Setter for the owner property."""
         self._owned_by = self.owner_cls(value)
+
+    @property
+    def settings(self):
+        """An alias for the settings property."""
+        return self._settings
+
+    @settings.setter
+    def settings(self, value):
+        """Setter for the settings property."""
+        self._settings = self.settings_cls(value)
 
     def dump(self):
         """Dump the field values as dictionary."""
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -103,6 +103,7 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
     # Access requests
     routes["user-access-request"] = "/<pid_value>/access/request"
     routes["guest-access-request"] = "/<pid_value>/access/request/guest"
+    routes["access-request-settings"] = "/<pid_value>/access"
 
     request_view_args = {
         "pid_value": ma.fields.Str(),
--- a/invenio_rdm_records/services/access/service.py
+++ b/invenio_rdm_records/services/access/service.py
@@ -9,7 +9,6 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 """RDM record access settings service."""
-
 from datetime import datetime, timedelta
 
 import arrow
@@ -70,6 +69,11 @@ class RecordAccessService(RecordService):
     def schema_grant(self):
         """Schema for secret links."""
         return ServiceSchemaWrapper(self, schema=self.config.schema_grant)
+
+    @property
+    def schema_access_settings(self):
+        """Schema for record parent."""
+        return ServiceSchemaWrapper(self, schema=self.config.schema_access_settings)
 
     @property
     def expandable_fields(self):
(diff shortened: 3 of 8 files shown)


invenio_rdm_records/services/schemas/parent/access.py
```diff
@@ -70,12 +70,17 @@ class AccessSettingsSchema(Schema):
...
+    accept_conditions_text = SanitizedHTML(allow_none=True)
 
 
 class ParentAccessSchema(Schema, FieldPermissionsMixin):
     """Access schema."""
 
+    class Meta:
+        """Meta attributes for the schema."""
+
+        unknown = EXCLUDE
```
zzacharo, line 82: we should not need this? What was the case?

invenio_rdm_records/resources/config.py
```diff
@@ -103,6 +103,7 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
     # Access requests
     routes["user-access-request"] = "/<pid_value>/access/request"
     routes["guest-access-request"] = "/<pid_value>/access/request/guest"
+    routes["access-request-settings"] = "/<pid_value>/access"
```
zzacharo, line 106:
```suggestion
    routes["access-request-settings"] = "/<pid_value>/access/settings"
```
?

invenio_rdm_records/services/access/service.py
```diff
@@ -71,6 +71,11 @@ def schema_grant(self):
         """Schema for secret links."""
         return ServiceSchemaWrapper(self, schema=self.config.schema_grant)
 
+    @property
+    def schema_parent(self):
+        """Schema for record parent."""
+        return ServiceSchemaWrapper(self, schema=self.config.schema_parent)
```
zzacharo, line 77: we should not return the parent but rather the parent settings...shall we rename this schema to `schema_parent_access_settings` and be super explicit of what do we return?

invenio_rdm_records/services/schemas/parent/__init__.py
```diff
@@ -37,7 +37,7 @@ class RDMParentSchema(ParentSchema, FieldPermissionsMixin):
         "review": "review",
     }
 
-    access = fields.Nested(ParentAccessSchema, dump_only=True)
+    access = fields.Nested(ParentAccessSchema)
```
zzacharo, line 40: Here I think now we opened the pandora box of people being able to update the whole access schema of the parent and definitely we don;t want that 😅  I think you should ensure that the `ParentAccessSchema` allows only the settings to be updated and all the other fields should remain `dump_only`

anikachurilova: You're right! I made all the fields of `ParentAccessSchema`  `dump_only`, except for the settings.

invenio_rdm_records/services/schemas/parent/access.py
anikachurilova: The fields that are marked as `dump_only=True` are considered as the `unknown` fields during deserialisation and the default behaviour of the marshmallow schema validation is to raise the `Unknown field` error. 
The `unknown = EXCLUDE` prevents it from this and just excludes those fields. Same is happening on the record schema here https://github.com/inveniosoftware/invenio-rdm-records/blob/0ec37706559185d28899d0304ca090df2e93953c/invenio_rdm_records/services/schemas/record.py#L41
But it is not applied for the nested schemas by itlself unfortunately

zzacharo: we have a utility for this https://github.com/inveniosoftware/invenio-rdm-records/blob/d1f55f95bde32963357eba8127acd06b82ea7be9/invenio_rdm_records/services/schemas/parent/__init__.py#L51

anikachurilova: Since another schema (`AccessSettingsSchema`) was used for deserializing the record, this is not needed anymore

invenio_rdm_records/services/access/service.py
```diff
@@ -9,7 +9,7 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 """RDM record access settings service."""
-
+import json
```
zzacharo, line 12: unused?

invenio_rdm_records/services/access/service.py
```diff
@@ -749,3 +756,41 @@ def create_guest_access_request(self, identity, token, expand=False, uow=None):
...
+        self.require_permission(identity, "manage", record=record)
+
+        # Validation
+        data, __ = self.schema_parent_access_settings.load(
+            data, context=dict(identity=identity), raise_errors=True
+        )
+
+        # Update
+        for key in data:
+            setattr(parent.access.settings, key, data[key])
```
zzacharo, line 781: could't we do simply 
```suggestion
            setattr(parent.access, "settings", data)
```

invenio_rdm_records/services/schemas/parent/access.py
```diff
@@ -70,7 +70,7 @@ class AccessSettingsSchema(Schema):
     allow_user_requests = fields.Boolean()
     allow_guest_requests = fields.Boolean()
 
-    accept_conditions_text = SanitizedHTML()
+    accept_conditions_text = SanitizedHTML(allow_none=True)
 
 
 class ParentAccessSchema(Schema, FieldPermissionsMixin):
```
zzacharo, line 76: shouldn't we add the settings in the "field_dump_permissions"?

invenio_rdm_records/services/config.py
```diff
@@ -230,6 +231,7 @@ class RDMRecordServiceConfig(RecordServiceConfig, ConfiguratorMixin):
     # Schemas
     schema = RDMRecordSchema
     schema_parent = RDMParentSchema
+    schema_parent_access_settings = AccessSettingsSchema
```
zzacharo, line 234:
```suggestion
    schema_access_settings = AccessSettingsSchema
```
? similar to grant and link?

invenio_rdm_records/services/schemas/parent/access.py
anikachurilova: No, we need settings to be available for record serialization, for the request access form render. See here: https://github.com/inveniosoftware/invenio-app-rdm/blob/19b679f19d44623150333b1003b64e2710ca888a/invenio_app_rdm/records_ui/templates/semantic-ui/invenio_app_rdm/records/detail.html#L28C1-L28C1

invenio_rdm_records/services/access/service.py
```diff
@@ -749,3 +753,40 @@ def create_guest_access_request(self, identity, token, expand=False, uow=None):
...
+        data, __ = self.schema_access_settings.load(
+            data, context=dict(identity=identity), raise_errors=True
+        )
+
+        # Update
+        setattr(parent.access, "settings", data)
+
+        # Commit
+        uow.register(RecordCommitOp(parent))
+        if record:
```
zzacharo, line 781: from @ntarocco https://github.com/inveniosoftware/invenio-rdm-records/pull/1392/files/51716d8ab61b74ddaf667b6849a2467436bdcee1#r1280814271 . Do we dump the access settings in the record? Can you check if a record stores in the json the access information?

invenio_rdm_records/records/systemfields/access/field/parent.py
```diff
@@ -107,13 +80,23 @@ def owner(self, value):
...
+        self._settings = self.settings_cls(value)
+
     def dump(self):
         """Dump the field values as dictionary."""
         access = {
             "owned_by": self._owned_by.dump(),
             "links": self.links.dump(),
             "grants": self.grants.dump(),
-            "settings": self.settings.dump(),
+            "settings": self._settings.dump(),
```
zzacharo, line 99:
```suggestion
            "settings": self.settings.dump(),
```

zzacharo (approved):

PR #1386, opened 2023-07-27
Record deletion: Data model

max-moser:


invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
```diff
@@ -352,21 +352,21 @@
...
           "type": "string",
           "description": "Reason for removal."
         },
-        "category": {
+        "note": {
           "type": "string",
-          "description": "Category for the removal."
+          "description": "Public note about the removal."
         },
         "removed_by": {
```
lnielsen, line 366: Question: Probably this should be using the entity resolver?

invenio_rdm_records/records/models.py
```diff
@@ -54,6 +62,11 @@ class RDMRecordMetadata(db.Model, RecordMetadataBase, ParentRecordMixin):
     media_bucket_id = db.Column(UUIDType, db.ForeignKey(Bucket.id))
     media_bucket = db.relationship(Bucket, foreign_keys=[media_bucket_id])
 
+    # The deletion status is stored in the model so that we can use it in SQL queries
+    deletion_status = db.Column(
+        db.String(1), nullable=False, default=RecordDeletionStatusEnum.ACTIVE.value
+    )
```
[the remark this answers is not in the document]
lnielsen: The JSON querying in PostgreSQL needs to have indexes defined on the value you want to query on. In such case query performance is comparable to the column query. Without the db index, the query performance is horrible on JSON fields.

The column fields are think at the best still since when s**t hits the fan, you need to operate fast, and not have to remember how you query into a specific JSON field (at least I never remember the syntax exactly have to look it up :-).

invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
```diff
@@ -352,21 +352,21 @@
     },
     "tombstone": {
       "type": "object",
-      "description": "Tombstone for the record.",
+      "description": "Tombstone information for the record.",
```
lnielsen, line 355: Missing fields:
- Bibliographic citation (a text string)
- Related identifier: (relation, scheme, identifier) -> for relation we can likely use the existing relation entities

invenio_rdm_records/records/systemfields/record_deletion_status.py
```diff
@@ -0,0 +1,98 @@
...
+
+from invenio_records.systemfields import SystemField
+
+
+class RecordDeletionStatusEnum(enum.Enum):
+    """Enumeration of a record's possible deletion states."""
+
+    PUBLISHED = "P"
+    DELETED = "D"
+    MARKED = "X"
```
max-moser, line 20: The names here are taken from the [RFC](https://codimd.web.cern.ch/y9vYxSjPQ-ytxTZOzzinhA#Data-model).
I've taken the liberty of shortening `MARKED_FOR_PURGE` to `MARKED`, because the former seems a bit unwieldy.
However, just `MARKED` may be ambiguous, and I'm open for suggestions!

invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
max-moser: I had that thought too, but wouldn't that mean that we would be displaying something like `removed by Max Moser` rather than `removed by staff`?
Unless we want to go into some sort of heuristic magic with checking the roles of resolved entities at display time (which is bound to misfire from time to time)?

TBH, I wouldn't mind actually having the `removed by Max Moser` there, but that would differ from the RFC.

invenio_rdm_records/services/schemas/record.py
```diff
@@ -88,3 +90,21 @@ def default_nested(self, data, many, **kwargs):
...
         return data
+
+    def hide_tombstone_or_metadata(self, data):
+        """Hide tombstone info if the record isn't deleted and metadata if it is."""
+        if not data.get("deletion_status", {}).get("is_deleted", False):
+            data.pop("tombstone", None)
+        else:
+            data.pop("metadata", None)
+            data.pop("files", None)
+            data.pop("stats", None)
```
max-moser, line 101: not sure which fields we should actually hide, these were just a few that came to my mind.
probably we'll also need to adapt the generation of links as well, s.t. we don't generate links to something that won't work anyway (like `files`).

invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
```diff
@@ -406,7 +406,17 @@
     "additionalProperties": false,
     "properties": {
       "user": {
-        "type": "integer"
+        "type": ["string", "integer"]
```
max-moser, line 409: allowing `string` as well, in order to be able to represent `{"user": "system"}`

invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
```diff
@@ -406,7 +406,17 @@
...
+      }
+    }
+  },
+  "removal_reason": {
+    "type": "object",
+    "additionalProperties": false,
+    "description": "Reason for the removal of a record.",
+    "properties": {
+      "id": {
+        "$ref": "local://definitions-v1.0.0.json#/identifier"
```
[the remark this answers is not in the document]
lnielsen: Hm, yes, that's probably a good idea

invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
lnielsen: > Should the bibliographic citation text be initialized with the citation text that we can already export (/api/records/<recid>?locale=en-US&style=bibtex)?

Yes. Probably the repository should have a config var to say which citation style to use (APA, Harvard, etc.)

> Until we actually hard-delete the metadata (which only happens during purge if I understand correctly), we can actually always recreate this citation text on the fly... Or do we want to give the users/admins the possibility to modify the displayed text?

Right yes.....hmmm....hoenstly I'm not sure what's best....my thinking was that if it's just a text string, you can edit and remove sensitive data without having to edit the record title itself. Might be best to just have a string field, so that we don't need any advanced logic to render the tombstone.....just show whatever is in the field.

invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
lnielsen: Yes, I was thinking that we could probably simply handle this during rendering.....ie..e we store the specific person, but rendering we can decide what todo.

invenio_rdm_records/records/systemfields/record_deletion_status.py
lnielsen: I'm fine with that.....like the X :-) X for marked

invenio_rdm_records/services/schemas/record.py
lnielsen: Yep....and probably we shoudl remove all the other info from the record besides the tombstone......

lnielsen (commented):
Overall looks good to me.

We'll need to add the tombstone visibility, but that might be easier to do in a separate PR when also dealing with the views etc

invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
max-moser: perfect, we should be able to achieve this sort of behavior with the current implementation!

invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
max-moser: alright, i'll add a simple text field for the **citation text**, with a default value of the citation text in a configured flavor.

about the **identifier**; from the RFC i understood that this would simply be the DOI of the record, let's maybe go over that field in tomorrow's office hours

invenio_rdm_records/records/systemfields/tombstone.py
```diff
@@ -0,0 +1,157 @@
...
+        If the value is a string or integer, it is assumed that the
+        reference is a user and the value will be turned into a dictionary
+        of the shape ``{"user": value}``.
+        If the value is a referenceable entity (such as a ``User`` object),
+        the ``ResolverRegistry`` will be used to create a reference dict.
+        """
+        if value is None:
+            self._removed_by = None
+
+        elif (ref := ResolverRegistry.reference_entity(value)) is not None:
```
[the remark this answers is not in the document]
zzacharo: that's cool but if it's just that statement that forces us to move from 3.7 maybe is too aggressive now? ping @lnielsen @ntarocco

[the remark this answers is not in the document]
max-moser: just as a reminder, python 3.7 has reached end of life already: https://devguide.python.org/versions/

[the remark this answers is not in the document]
zzacharo: yes, we also started removing from tests but we were discussing not forcing people to upgrade their python version...so my comment was related to this topic...

invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
max-moser: update with summary of our brief discussion just now:

* we'll leave out the `identifier` for now, because that's part of the record already anyway.
* the `(related) identifiers` would be useful to make the tombstone more actionable, by supplying auxiliary information in a dedicated field – e.g. `deletion reason: duplicate`, and duplicate of what? of the record(s): `related identifier(s)`.
* i'll still add the `citation text` string field, which will be pre-filled with the record's citation text in a configurable style (e.g. APA)
* `tombstone visibility`: since it's only either public or restricted, can be a **boolean** like `is_visible: true/false`

invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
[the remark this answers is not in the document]
max-moser: update: it looks like i missed the prefix here...
it actually references `definitions-v1.0.0.json` (which is defined in [Records-Resources](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/records/jsonschemas/definitions-v1.0.0.json)), rather than `records/definitions-v1.0.0`, which would be the one defined here.

It seems like there's some duplication going on, like the `internal-pid` in [`definitions-v2.0.0`](https://github.com/inveniosoftware/invenio-records-resources/blob/master/invenio_records_resources/records/jsonschemas/definitions-v2.0.0.json#L139-L163) being the same as in [`records/definitions-v1.2.0`](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/jsonschemas/records/definitions-v1.2.0.json#L356-L389) (it has been removed from [`records/definitions-v2.0.0`](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json) though)

invenio_rdm_records/records/mappings/os-v1/rdmrecords/drafts/draft-v6.0.0.json
max-moser, line 1: Note that drafts shouldn't actually have tombstones, I've mostly added the info to the draft mappings as well for symmetry reasons.

invenio_rdm_records/services/schemas/record.py
max-moser: the current version right now pops most other fields, resulting in:
```python
In [5]: svc.read(system_identity, pid).data
Out[5]:
{'id': 'zm408-gyp73',
 'links': {'self': 'https://127.0.0.1/api/records/zm408-gyp73',
  'self_html': 'https://127.0.0.1/records/zm408-gyp73',
  'parent': 'https://127.0.0.1/api/records/806dg-1x022',
  'parent_html': 'https://127.0.0.1/records/806dg-1x022',
  'self_iiif_manifest': 'https://127.0.0.1/api/iiif/record:zm408-gyp73/manifest',
  'self_iiif_sequence': 'https://127.0.0.1/api/iiif/record:zm408-gyp73/sequence/default',
  'files': 'https://127.0.0.1/api/records/zm408-gyp73/files',
  'media_files': 'https://127.0.0.1/api/records/zm408-gyp73/media-files',
  'archive': 'https://127.0.0.1/api/records/zm408-gyp73/files-archive',
  'latest': 'https://127.0.0.1/api/records/zm408-gyp73/versions/latest',
  'latest_html': 'https://127.0.0.1/records/zm408-gyp73/latest',
  'draft': 'https://127.0.0.1/api/records/zm408-gyp73/draft',
  'versions': 'https://127.0.0.1/api/records/zm408-gyp73/versions',
  'access_links': 'https://127.0.0.1/api/records/zm408-gyp73/access/links',
  'reserve_doi': 'https://127.0.0.1/api/records/zm408-gyp73/draft/pids/doi',
  'communities': 'https://127.0.0.1/api/records/zm408-gyp73/communities',
  'communities-suggestions': 'https://127.0.0.1/api/records/zm408-gyp73/communities-suggestions',
  'requests': 'https://127.0.0.1/api/records/zm408-gyp73/requests'},
 'pids': {'oai': {'identifier': 'oai:my-site.com:zm408-gyp73',
   'provider': 'oai'}},
 'tombstone': {'removal_reason': {'id': 'spam',
   'title': {'de': 'Spam', 'en': 'Spam'}},
  'note': '',
  'removed_by': {'user': 'system'},
  'removal_date': '2023-08-04',
  'citation_text': '',
  'is_visible': True,
  'identifiers': [{'scheme': 'oai',
    'identifier': {'identifier': 'oai:my-site.com:zm408-gyp73',
     'provider': 'oai'}}]},
 'deletion_status': {'is_deleted': True, 'status': 'D'}}
```

note that `tombstone.identifiers` is effectively just a massaged clone of `pids`.

invenio_rdm_records/services/schemas/record.py
```diff
@@ -88,3 +90,37 @@ def default_nested(self, data, many, **kwargs):
             data["custom_fields"] = {}
 
         return data
+
+    def hide_tombstone_or_metadata(self, data):
+        """Hide tombstone info if the record isn't deleted and metadata if it is."""
+        if not data.get("deletion_status", {}).get("is_deleted", False):
```
max-moser, line 96: note that `deletion_status.is_deleted` is different from the top-level `is_deleted` (which has been around forever already)

max-moser:
Note for the `citation_text`:
AFAIK, the generation of citation texts in a specific flavor (e.g. APA, IEEE, BibTeX) goes through the service layer, so adding that to the system field itself would mix layers.
I think it should probably be added as a default value in the upcoming service function that's used to delete records/set tombstone information.

invenio_rdm_records/records/systemfields/tombstone.py
```diff
@@ -0,0 +1,192 @@
...
+        data = {
+            "note": self.note,
+            "removed_by": self.removed_by,
+            "removal_date": self.removal_date,
+            "citation_text": self.citation_text,
+            "is_visible": self.is_visible,
+        }
+
+        if self.removal_reason:
+            data["removal_reason"] = self.removal_reason
```
max-moser, line 139: If the `removal_reason` isn't set to anything, it's better to not have the field at all.
If it is set (e.g. to `removal_reason: None`, or `removal_reason: {"id": None}`) then the `relations` field will start complaining and prevent `record.commit()` from finishing.
Because the `relations` systemfield has the checks in `pre_commit()` and lexicographically comes before `tombstone`, the latter will have no chance of updating the record's data in `pre_commit()` to rectify the wrong data.

invenio_rdm_records/records/systemfields/tombstone.py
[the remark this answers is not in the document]
slint: As a general policy for Invenio, we want to try keeping the window open to allow older versions of libraries/runtimes running without completely failing (i.e. not "forcing" version upgrades). That's also why we still keep `python_requires = >=3.7` in `setup.cfg`. If other dependencies though start forcing us to drop v3.7, that's our cue to also officially do it, since the effort will become too much to maintain this "implicit" backwards compat.

I agree that in this case the walrus operator probably increases readability (vs the alternative of adding an extra level of nesting inside the `else:`). 

To play the devil's advocate though:

- if we're not testing this in CI, the only way to find out that Python 3.7 is broken is to wait for someone to report it...
- I see that e.g. even newer versions of Flask have [added `python_requires = >=3.8`](https://github.com/pallets/flask/blob/b1385919bea3238d49a6e2cfc8261771775cd6e9/pyproject.toml#L20). So we're currently in this limbo state where because we've upper-pinned Flask v2.3, we get a "free pass" for Python v3.7... I assume that once a security fix comes in Flask v2.3.x, and not v2.2.x, we must buckle up and make the move (and then be actually forced to drop Python 3.7).

At this point, I'm pro towards dropping Python v3.7 support and using Python v3.8+ features since if the 3.7 build is not failing already, it will happen anyways once we unpin Flask.

@ntarocco WDYT?

invenio_rdm_records/records/systemfields/deletion_status.py
```diff
@@ -0,0 +1,108 @@
...
+        return self.status
+
+
+class RecordDeletionStatusField(SystemField):
+    """System field for accessing a record's deletion status."""
+
+    def __init__(self, record_cls=None, key=None):
+        """Constructor."""
+        super().__init__(key=key)
+        self.record_cls = record_cls
```
slint, line 70:
```suggestion
    def __init__(self, key=None):
        """Constructor."""
        super().__init__(key=key)
```

minor: is `self.record_cls` used anywhere? I don't see it being passed in the API class where the systemfield is defined/used.

If that is the case, then maybe the constructor is not needed at all.

invenio_rdm_records/records/systemfields/tombstone.py
```diff
@@ -0,0 +1,192 @@
...
+        return repr(self.dump())
+
+
+class TombstoneField(SystemField):
+    """System field for accessing a record's deletion status."""
+
+    def __init__(self, draft_cls=None, key=None):
+        """Constructor."""
+        super().__init__(key=key)
+        self.draft_cls = draft_cls
```
slint, line 154: minor: ditto as my comment on the `self.record_cls` param/attribute in the `RecordDeletionStatusField`, if `self.draft_cls` is not used anywhere we can remove it (and the entire ctor as well possibly)

invenio_rdm_records/services/schemas/record.py
slint, line 96: nit/minor: for readability, I would add a variable and maybe swap the `if/else` branches to instead read the "positive" version of the statement (i.e. `if is_deleted: ... else: ...`).

invenio_rdm_records/services/schemas/record.py
```diff
@@ -88,3 +90,37 @@ def default_nested(self, data, many, **kwargs):
...
+            data.pop("is_draft", None)
+            data.pop("is_published", None)
+            data.pop("media_files", None)
+            data.pop("metadata", None)
+            data.pop("parent", None)
+            data.pop("revision_id", None)
+            data.pop("stats", None)
+            data.pop("status", None)
+            data.pop("updated", None)
+            data.pop("versions", None)
```
slint, line 113: minor: would it make sense to summarize all these into something like "pop all fields that are not `tombstone`"? I could imagine that in the future we add a new field in the schema, and then forget to add it here (probably tests will catch this, but you never know :) )

slint (commented):
Nice work! Looks in a good state overall, some minor comments/questions/nits but nothing too blocking.

invenio_rdm_records/records/models.py
```diff
@@ -54,6 +56,11 @@ class RDMRecordMetadata(db.Model, RecordMetadataBase, ParentRecordMixin):
     media_bucket_id = db.Column(UUIDType, db.ForeignKey(Bucket.id))
     media_bucket = db.relationship(Bucket, foreign_keys=[media_bucket_id])
 
+    # The deletion status is stored in the model so that we can use it in SQL queries
+    deletion_status = db.Column(
+        db.String(1), nullable=False, default=RecordDeletionStatusEnum.PUBLISHED.value
```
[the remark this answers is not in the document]
max-moser: I can't make general statements here, but in this particular case here I just didn't know that it existed :sweat_smile: 
At some point I should really take the time to _at least_ skim through the entire documentation...

Sounds sweet though, I'm gonna give it a shot and see what happens

invenio_rdm_records/services/schemas/record.py
max-moser: i've been toying with the idea too, but wasn't entirely sure about the set of fields to keep.
especially the `links` entries might be something worth keeping around`[1]`, so that the REST API stays machine-actionable.

the current set of keys that aren't thrown out here are:
`dict_keys(['id', 'links', 'pids', 'tombstone', 'deletion_status'])`

an example looks like:
```python
{'id': 'zm408-gyp73',
 'links': {'self': 'https://127.0.0.1/api/records/zm408-gyp73',
  'self_html': 'https://127.0.0.1/records/zm408-gyp73',
  'parent': 'https://127.0.0.1/api/records/806dg-1x022',
  'parent_html': 'https://127.0.0.1/records/806dg-1x022',
  'self_iiif_manifest': 'https://127.0.0.1/api/iiif/record:zm408-gyp73/manifest',
  'self_iiif_sequence': 'https://127.0.0.1/api/iiif/record:zm408-gyp73/sequence/default',
  'files': 'https://127.0.0.1/api/records/zm408-gyp73/files',
  'media_files': 'https://127.0.0.1/api/records/zm408-gyp73/media-files',
  'archive': 'https://127.0.0.1/api/records/zm408-gyp73/files-archive',
  'latest': 'https://127.0.0.1/api/records/zm408-gyp73/versions/latest',
  'latest_html': 'https://127.0.0.1/records/zm408-gyp73/latest',
  'draft': 'https://127.0.0.1/api/records/zm408-gyp73/draft',
  'versions': 'https://127.0.0.1/api/records/zm408-gyp73/versions',
  'access_links': 'https://127.0.0.1/api/records/zm408-gyp73/access/links',
  'reserve_doi': 'https://127.0.0.1/api/records/zm408-gyp73/draft/pids/doi',
  'communities': 'https://127.0.0.1/api/records/zm408-gyp73/communities',
  'communities-suggestions': 'https://127.0.0.1/api/records/zm408-gyp73/communities-suggestions',
  'requests': 'https://127.0.0.1/api/records/zm408-gyp73/requests'},
 'pids': {'oai': {'identifier': 'oai:my-site.com:zm408-gyp73',
   'provider': 'oai'}},
 'tombstone': {'note': '',
  'removed_by': None,
  'removal_date': '2023-08-04',
  'citation_text': '',
  'is_visible': True,
  'identifiers': [{'scheme': 'oai',
    'identifier': 'oai:my-site.com:zm408-gyp73'}]},
 'deletion_status': {'is_deleted': True, 'status': 'D'}}
```

are you okay with that set of fields?

`[1]` in a cut-down form of course, as e.g. the `files` links won't make much sense anymore given that they're not accessible anyway.
however, some links (like `self_html`) may still be useful to enable crawlers to navigate around.

invenio_rdm_records/services/schemas/record.py
```diff
@@ -88,3 +90,37 @@ def default_nested(self, data, many, **kwargs):
...
+
+            data["tombstone"]["identifiers"] = [
+                {"scheme": k, "identifier": v} for k, v in data.get("pids", {}).items()
+            ]
```
max-moser, line 117: this basically just duplicates the information already available in `record['identifiers']`, massaged into a slightly more digestible format.
should we keep this around and/or throw `record['identifiers']` out, if the record is deleted?

invenio_rdm_records/records/models.py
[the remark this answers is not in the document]
max-moser: From quick testing, it seems to be working just fine, thanks for the pointer!

Setting it to a valid value from the enumeration will work as expected, and setting it to something invalid won't let the transaction commit:
```python
StatementError: (builtins.ValueError) 'S' is not a valid RecordDeletionStatusEnum
[SQL: UPDATE rdm_records_metadata SET updated=%(updated)s, version_id=%(version_id)s, deletion_status=%(deletion_status)s WHERE rdm_records_metadata.id = %(rdm_records_metadata_id)s AND rdm_records_metadata.version_id = %(rdm_records_metadata_version_id)s]
[parameters: [{'deletion_status': 'S', 'updated': datetime.datetime(2023, 8, 8, 14, 20, 56, 946563), 'rdm_records_metadata_version_id': 29, 'version_id': 30, 'rdm_records_metadata_id': UUID('b94589e1-7d1f-4b7f-a0c3-a654d2913a5c')}]]
```

invenio_rdm_records/services/schemas/record.py
```diff
@@ -88,3 +90,28 @@ def default_nested(self, data, many, **kwargs):
...
+
+        if is_deleted:
+            keys_to_keep = ["tombstone", "deletion_status", "id", "pids", "links"]
+            for key in [k for k in data if k in keys_to_keep]:
+                data.pop(key, None)
+
+            pids = data.get("pids", {})
+            data["tombstone"]["identifiers"] = [
+                {"scheme": k, "identifier": v["identifier"]} for k, v in pids.items()
+            ]
```
max-moser, line 106: this basically just duplicates the information already available in `record['identifiers']`, massaged into a slightly more digestible format.
should we keep this around and/or throw `record['identifiers']` out, if the record is deleted?

invenio_rdm_records/records/models.py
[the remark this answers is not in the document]
max-moser: FTR: Alembic didn't detect any changes to the DB schema after switching to the `ChoiceType` with the `String(1)` implementation.
Not that this would've been expected, but still good to check and write down.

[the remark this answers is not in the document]
slint: Actually, the [column type detection is a bit flaky behavior with Alembic](https://alembic.sqlalchemy.org/en/latest/autogenerate.html#affecting-the-rendering-of-types-themselves), and sometimes you have to make the conversion "by-hand".

invenio_rdm_records/services/schemas/record.py
slint: I think I'm ok with the list you shared so :+1: for me. It also resembles an explicit "allowed/safe" list of fields, vs. hand-removing fields that might contain information that we don't want to disclose.

invenio_rdm_records/records/models.py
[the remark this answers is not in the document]
max-moser: So from what I understand there, the `repr(<column_type>)` would be relevant.
This is what I get in `invenio shell`:
```python
In [11]: repr(ChoiceType(RecordDeletionStatusEnum, impl=db.String(1)))
Out[11]: 'ChoiceType(length=1)'
```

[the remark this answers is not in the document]
max-moser: Update: `ChoiceType(length=1)` won't work because that's not a valid constructor call for `ChoiceType` (duh), but `String(1)` is fine.
However, testing the alembic script again did reveal an issue; namely the columns not being nullable but alembic didn't add default values.

I've just added a fix for that (by adding `server_default="P"` to the `add_column()`) to this PR.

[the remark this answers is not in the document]
slint: I think it's fine to hardcode the final DB column type in the alembic recipe, i.e. `db.String(1)`. 

I would remove the `server_default` though, since it sets the default on the column definition in the DB vs. the `default=...` we set in the Model classes field, which is Python/ORM-only. The difference is that with `server_default` if you would run an SQL INSERT statement via `psql` without passing `deletion_status` you would get the default value `"P"` in the row, which is actually not what we want: the application logic should be responsible for what values get set (so in case we want to update the default value in the future, we don't need a new alembic recipe).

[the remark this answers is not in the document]
slint: Also for the [`PersistentIdentifier.status` alembic recipe](https://github.com/inveniosoftware/invenio-pidstore/blob/d54359c185c92f87decc8b9ba951ed03790226f6/invenio_pidstore/alembic/999c62899c20_create_pidstore_tables.py#L32C57-L32C57) we're not specifying any kind of "default" either.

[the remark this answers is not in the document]
max-moser: hmm... yeah good point!
we're still going to need some sort of default value for the new column though, as the normal `default=...` won't update existing records, and leave them `null` (which contradicts the `not null` clause and fails the migration).
probably a simple `UPDATE` statement in the same transaction could do the trick.

invenio_rdm_records/services/schemas/record.py
max-moser: i guess the benefit of having the identifiers in the tombstone itself would be to have the `tombstone` information as self-contained as possible, so you can just send around that part of the whole record JSON and be fine with it.
but i also feel that this isn't exactly a compelling reason to have it around.

it's simple enough that we can re-add it at any point later, should we find that it would be actually beneficial to have.

invenio_rdm_records/records/models.py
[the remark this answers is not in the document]
max-moser: Rewrote the upgrade script, it's now a 3-step process that should work for both new instances as well as existing instances:
1. Create the new columns but make them nullable
2. Set a default value for all existing rows (s.t. they don't have null values) with a heuristic that tries to get the default value from the module
3. Set the new column to be *not nullable*

invenio_rdm_records/alembic/2186256e8d9b_add_deletion_status_to_rdmrecords.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # step 2: set default values for existing rows
+    default_value = _get_default_deletion_status()
+    op.execute(f"UPDATE rdm_records_metadata SET deletion_status = '{default_value}'")
+    op.execute(
+        f"UPDATE rdm_records_metadata_version SET deletion_status = '{default_value}'"
+    )
+
+    # step 3: make the original table not nullable
+    op.alter_column("rdm_records_metadata", "deletion_status", nullable=False)
```
max-moser, line 70: note: the `rdm_records_metadata_version.deletion_status` table was **not** set `nullable=False` by alembic even before i rewrote the upgrade script.

i'm not sure if it actually matters whether or not we have the column in the `[...]_version` table nullable or not; the critical part is the `rdm_records_metadata` table/column.

invenio_rdm_records/alembic/2186256e8d9b_add_deletion_status_to_rdmrecords.py
```diff
@@ -0,0 +1,76 @@
...
+            nullable=True,
+        ),
+    )
+
+    # step 2: set default values for existing rows
+    default_value = _get_default_deletion_status()
+    op.execute(f"UPDATE rdm_records_metadata SET deletion_status = '{default_value}'")
+    op.execute(
+        f"UPDATE rdm_records_metadata_version SET deletion_status = '{default_value}'"
+    )
```
slint, line 67: To avoid manually building SQL strings (not that we'll ever get SQL injection attacks in this part), you can check out [this example](https://github.com/inveniosoftware/invenio-sipstore/blob/c213d0738360f44eebd643f2e095a2ce53bd9dcc/invenio_sipstore/alembic/b31cad2f14c7_extend_sip_metadata_model.py#L79-L81).

invenio_rdm_records/alembic/2186256e8d9b_add_deletion_status_to_rdmrecords.py
slint: If you create the database with `invenio db create`, is `rdm_records_metadata_version.deletion_status` set to `nullable=False`? If yes, the alembic recipe should match that.

invenio_rdm_records/records/models.py
```diff
@@ -54,6 +56,13 @@ class RDMRecordMetadata(db.Model, RecordMetadataBase, ParentRecordMixin):
     media_bucket_id = db.Column(UUIDType, db.ForeignKey(Bucket.id))
     media_bucket = db.relationship(Bucket, foreign_keys=[media_bucket_id])
 
+    # The deletion status is stored in the model so that we can use it in SQL queries
+    deletion_status = db.Column(
+        ChoiceType(RecordDeletionStatusEnum, impl=db.String(1)),
+        nullable=False,
+        default=RecordDeletionStatusEnum.PUBLISHED.value,
```
slint, line 63:
```suggestion
        default=RecordDeletionStatusEnum.PUBLISHED,
```

question/nit: is it still needed now that we're using `ChoiceType`? I know it works probably, since the field is a bit more flexible with the accepted values, but just to confirm and advertise it a bit in case one wonders how to set the field.

slint (approved):
LGTM! Super minor comments/questions, but otherwise it's good to go.

invenio_rdm_records/alembic/2186256e8d9b_add_deletion_status_to_rdmrecords.py
max-moser: no, with just `invenio db create` it's not marked as `not null`:

```
my-site=# \d rdm_records_metadata_version
                    Table "public.rdm_records_metadata_version"
       Column       |            Type             | Collation | Nullable | Default
--------------------+-----------------------------+-----------+----------+---------
 created            | timestamp without time zone |           |          |
 updated            | timestamp without time zone |           |          |
 id                 | uuid                        |           | not null |
 json               | jsonb                       |           |          |
 version_id         | integer                     |           |          |
 index              | integer                     |           |          |
 bucket_id          | uuid                        |           |          |
 media_bucket_id    | uuid                        |           |          |
 deletion_status    | character varying(1)        |           |          |          <==
 parent_id          | uuid                        |           |          |
 transaction_id     | bigint                      |           | not null |
 end_transaction_id | bigint                      |           |          |
 operation_type     | smallint                    |           | not null |
Indexes:
    "pk_rdm_records_metadata_version" PRIMARY KEY, btree (id, transaction_id)
    "ix_rdm_records_metadata_version_end_transaction_id" btree (end_transaction_id)
    "ix_rdm_records_metadata_version_operation_type" btree (operation_type)
    "ix_rdm_records_metadata_version_transaction_id" btree (transaction_id)

```

i can't find any explicit table definition for it (and think it's automatically created by SQLAlchemy-Continuum?), so i can't tell exactly why it's not marked as `not null` while the "original" table is.

unfortunately i'm not familiar enough with the entire revision/version system to be able to foresee any potential problems here, but i guess it's fine.
some quick manual testing didn't show any issues when editing metadata for existing records and publishing the changes, or creating new versions of records.

invenio_rdm_records/records/models.py
max-moser: I've tried out storing the enum itself rather than the `.value`, and it does work nicely.
However, this will also cause the enum value rather than the string to be returned when accessing this model field, which is not expected in a number of places. :boom: 

Also, the alembic migration falls back to the string `"P"` if all other heuristics fail to fetch a default value.
So in order to stay consistent, I reverted back to setting the `PUBLISHED.value` rather than `PUBLISHED` here.

slint: Ouch, yeah that will definitely cause issues... I though by using the Enum, we're making our lives easier when doing comparisons/conversions across the codebase, i.e. `"P" == RecordDeletionStatusEnum.PUBLISHED`. 

What are the places that expect a string value coming back from the model? Is it serialization/indexing? It feels a bit dirty to rely on the "temporary" value that the model field is set to, i.e. if the following behaviour doesn't fail we're in trouble:

```python
record.deletion_status = "P"
# using "is" as in the exact value + type without implicit equality checks
assert record.deletion_status is "P"
assert record.deletion_status is not RecordDeletionStatusEnum.PUBLISHED

record.deletion_status = RecordDeletionStatusEnum.PUBLISHED
assert record.deletion_status is RecordDeletionStatusEnum.PUBLISHED
assert record.deletion_status is not "P"
```

I would expect something like:

```python
record.deletion_status = "P"
assert record.deletion_status == "P"
assert record.deletion_status == RecordDeletionStatusEnum.PUBLISHED
assert record.deletion_status is not "P"
assert record.deletion_status is RecordDeletionStatusEnum.PUBLISHED

record.deletion_status = RecordDeletionStatusEnum.PUBLISHED
assert record.deletion_status == RecordDeletionStatusEnum.PUBLISHED
assert record.deletion_status is not "P"
assert record.deletion_status is RecordDeletionStatusEnum.PUBLISHED
```

invenio_rdm_records/records/systemfields/tombstone.py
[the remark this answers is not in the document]
slint: Discussed with Nico IRL, and it's good to go as long as we're not mentioning anywhere in the docs explicitly that we're supporting Python v3.7 still.

invenio_rdm_records/records/models.py
max-moser: No, unfortunately it doesn't seem that comparisons between enums and strings are so easily possible:
```python
In [9]: RecordDeletionStatusEnum.DELETED.value
Out[9]: 'D'

In [10]: RecordDeletionStatusEnum.DELETED == "D"
Out[10]: False
```

However, checks similar to the following are possible (note that this will raise an exception if `"D"` isn't a valid enum value):
```python
In [14]: RecordDeletionStatusEnum("D") is RecordDeletionStatusEnum.DELETED
Out[14]: True
```

---

The enums are mostly used as a backbone in the implementation, to make sure no invalid values can be set in the model; they're used as the basis for normalization while _setting_ the record deletion status.
For _getting_ the deletion status, I usually provide several options:
* `record.deletion_status` gives back a `RecordDeletionStatus` object (calling `str()` on that will return the enum string value)
* `record.deletion_status.status` gives back the string value
* `record.deletion_status._status` gives back the `RecordDeletionStatusEnum` object

:heavy_plus_sign: I've just overridden the equality checks for the **system field**, s.t. it can handle comparisons with strings and enum values as expected.

---

:information_source:  **Update**
I've just experimented a bit, and it seems like the `ChoiceField` (in the **model**) accepts both strings as well as enum values.
When setting the field to a string (e.g. `"P"`), it will have the string value for the remainder of the session.
In a new session, the value will be normalized into an enum value:

```python
In [24]: r.model.deletion_status = "P"

In [25]: r.model.deletion_status
Out[25]: 'P'

In [26]: db.session.commit()

In [27]: r = rec_svc.record_cls.get_record(r.id)

In [28]: r.model.deletion_status
Out[28]: <RecordDeletionStatusEnum.PUBLISHED: 'P'>
```

Note that `db.session.commit()` will raise an exception (`StatementError`) if the value is set to something invalid.

Although it should be pretty rare for anybody to interact with the model's DB column directly, usually it should go through the system field.
And as mentioned above, that takes care of conversions and normalization s.t. the developers don't have to.

slint: Ok, that makes sense. Thanks a lot for digging into this, I learned something new as well today :)

In that case, since we're always going through the system field for everything (setting, serialization, etc.), we should be good. I don't think there's a need to explicitly check for all the weird scenarios I mentioned in https://github.com/inveniosoftware/invenio-rdm-records/pull/1386#discussion_r1289998046 (since there's no point in verifying the `ChoiceType`'s behavior).

slint (approved):
Fantastic, we can merge this :)

PR #1388, opened 2023-07-28
service: lock record files conditionally

kpsherva:
* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2259

## Breaking changes, to be released as major

### DIFF
--- a/invenio_rdm_records/services/permissions.py
+++ b/invenio_rdm_records/services/permissions.py
@@ -196,7 +196,7 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     can_remove_record = [CommunityCurators()]
 
     #
-    # Media files
+    # Media files - draft
     #
     can_draft_media_create_files = can_review
     can_draft_media_read_files = can_review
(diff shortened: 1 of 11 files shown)


invenio_rdm_records/services/permissions.py
```diff
@@ -214,6 +214,23 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
...
+    can_get_content_media_files = [
+        # note: even though this is closer to business logic than permissions,
+        # it was simpler and less coupling to implement this as permission check
+        IfFileIsLocal(then_=can_read, else_=[SystemProcess()])
+    ]
+    can_create_media_files = [Disable()]
+    can_set_content_media_files = [Disable()]
+    can_commit_media_files = [Disable()]
+    can_update_media_files = [Disable()]
+    can_delete_media_files = [Disable()]
```
zzacharo, line 233: @slint sanity check: if we disable `can_create_media_files` shouldn't be enough to say "users by default cannot create in any manner media files"?

slint: - Shouldn't it be `can_draft_media_create_files = [Disable()]` (and all other `can_draft_media_...` vars), since that's how users actually get to create the media files?
- The naming prefixes are a bit mixed actually, shouldn't it be `can_draft_media_files_create`? (note the last two words are flipped)

kpsherva, line 233:
```suggestion
    can_media_read_files = [
        IfRestricted("record", then_=can_view, else_=can_all),
        ResourceAccessToken("read"),
    ]
    can_media_get_content_files = [
        # note: even though this is closer to business logic than permissions,
        # it was simpler and less coupling to implement this as permission check
        IfFileIsLocal(then_=can_read, else_=[SystemProcess()])
    ]
    can_media_create__files = [Disable()]
    can_media_set_content_files = [Disable()]
    can_media_commit_files = [Disable()]
    can_media_update_files = [Disable()]
    can_media_delete_files = [Disable()]
```

we are missing

```
permission_action_prefix = "media_"
```
in 
https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/config.py#L477

thats why it was working with incorrect names (taking the permissions from the files below, not the newly added with incorrect naming)

Issue #1324, opened 2023-06-01
DOIs: improve error message


kpsherva:
>  "source" will show up if there is an issue with a specific metadata field, but not if there is a general error like incorrect credentials or a missing prefix. So I think the get is right, but the 'Missing field' text below is a bit confusing because you can get errors for reasons other than missing fields. 

One possible other improvement is the 403 error contains nothing useful in the `text`field beyond You are not authorized to access this resource. Maybe returning something like "403 Error: Check your credentials and prefix settings" would he more helpful for folks.

_Originally posted by @tmorrell in https://github.com/inveniosoftware/invenio-rdm-records/pull/1315#discussion_r1212035513_

TLGINO:
After quite a bit of debugging I was not able to produce any error on the client-side (regarding inputting valid and invalid DOIs).
The only issue I found was regarding invalid `DATACITE_PREFIX` configurations
Which in turn yielded this error to the front-end
> Oops, something went wrong! The identifier was not reserved. Please try again. If the problem persists, contact user support..

Which is misleading as it seems to imply the error comes from PIDs (especially if you look at the actual error variable used `RESERVE_PID_FAILED`)
As this is a configuration issue, and the server stack trace clearly shows the error, it is not necessary to inform the user of the 'root' issue, and so this imprecision is deemed ok.


![image](https://github.com/inveniosoftware/invenio-rdm-records/assets/59311729/a329cd99-a1ed-4a44-ac83-541fe92805d3)


@tmorrell does the problem still persist on your side? Thx!

tmorrell:
Ah, sorry for the confusion. My comment was specific to https://github.com/jrcastro2/invenio-rdm-records/blob/3b31fb26533e1ce2a74641c20a2bcd3aa27d056c/invenio_rdm_records/services/pids/providers/datacite.py#L117, which is for the logging that will show up in the console. I think the generic error for the UI is fine, since these are really technical errors that an admin needs to figure out.

I don't have a current master install, but here are some steps to reproduce what I'm talking about:

1. Set an incorrect password in `DATACITE_PASSWORD`
2. Try to register a DOI
3. Console shows error ```Error in  'Missing field' : 403.....```

That error is confusing because the error is because of a password problem, and has nothing to do with a missing field. The easiest fix is to redo the line to not say `Missing field`. A more optimal solution would be to specifically catch the 403 and suggest that it might be because of a password or prefix issue.

TLGINO:
[...]

PR #1390, opened 2023-07-31
pids: fix logging error when registering DOIs

ntarocco:


zzacharo (approved):

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -109,16 +110,21 @@ def __init__(
...
+        ex_txt = exception.args[0] or ""
+        try:
+            ex_json = json.loads(ex_txt)
+        except JSONDecodeError:
+            current_app.logger.error(f"No content error: {ex_txt}")
+            return
+
+        # the `errors` field is only available when a 4xx error happened (not 500)
+        for error in ex_json.get("errors", []):
+            field = error.get("source", "Missing field/error")
```
tmorrell, line 125: I'd switch the default text from "Missing field/error" to something like "DataCite service; you may want to check your password". I'm pretty sure that if you have a missing field you'll always get back a source, so it doesn't make sense to have missing field in the default. The most common error I've seen people run into is password issues, since DataCite just returns a plain 403. I'd like to put some indication in the logs that it's something folks should look for., but I'm not sure of the exact phrasing.

tmorrell:
Will close https://github.com/inveniosoftware/invenio-rdm-records/issues/1324

tmorrell (approved):
Looks good, with one change to the default logging

invenio_rdm_records/services/pids/providers/datacite.py
ntarocco: Thanks a lot for the review and for linking the issue: I should have checked if there was one.
We actually encountered this error because of a 500 Internal server error on Datacite APIs, not because of 403.
The error is caught earlier, thanks to different exceptions [here](https://github.com/inveniosoftware/datacite/blob/master/datacite/errors.py#L38): I will need some refactoring to have a better error msg.

ntarocco: @tmorrell I have refactored the func, have a look if now it is more clear. Tests should also clarify the behaviour.

tmorrell: Awesome-that's much more readable and should be clearer for folks. It also makes it easier if we have to add more log messages or help text in the future.

Issue #1387, opened 2023-07-27
requests: implement access request link expiration setting


anikachurilova:
Add an advanced options "Link expiration" setting

PR #1392, opened 2023-08-01
Implement access request link expiration setting

anikachurilova:
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1387

Note: only the last commit corresponds to the current task. Previous one will be squashed after merge of this https://github.com/inveniosoftware/invenio-rdm-records/pull/1385

### DIFF
--- a/invenio_rdm_records/services/schemas/parent/access.py
+++ b/invenio_rdm_records/services/schemas/parent/access.py
@@ -14,7 +14,7 @@
 
 from datetime import timezone
 
-from marshmallow import Schema, fields, validate
+from marshmallow import Schema, fields, pre_load, validate
 from marshmallow_utils.fields import (
     ISODateString,
     SanitizedHTML,
@@ -71,6 +71,20 @@ class AccessSettingsSchema(Schema):
     allow_guest_requests = fields.Boolean()
 
     accept_conditions_text = SanitizedHTML(allow_none=True)
+    secret_link_expiration = fields.Integer(
+        validate=validate.Range(max=365), allow_none=True
+    )
+
+    @pre_load
+    def translate_expiration_date(self, data, **kwargs):
+        """Translate secret_link_expiration from ui dropdown value."""
+        expiration_days = data["secret_link_expiration"]
+        if expiration_days == 0:
+            data["secret_link_expiration"] = None
+        else:
+            data["secret_link_expiration"] = expiration_days
+
+        return data
 
 
 class ParentAccessSchema(Schema, FieldPermissionsMixin):
--- a/invenio_rdm_records/records/jsonschemas/records/parent-v3.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/parent-v3.0.0.json
@@ -102,6 +102,10 @@
             "accept_conditions_text": {
               "description": "Custom description for access conditions to be displayed rather than the default text.",
               "type": ["string", "null"]
+            },
+            "secret_link_expiration": {
+              "description": "The default amount of days before secret link expires.",
+              "type": ["number", "null"]
             }
           }
         }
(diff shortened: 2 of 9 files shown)


invenio_rdm_records/services/access/service.py
```diff
@@ -749,3 +754,54 @@ def create_guest_access_request(self, identity, token, expand=False, uow=None):
...
+        )
+
+        # Update
+        settings_dict = data["access"]["settings"]
+        for key in settings_dict:
+            setattr(parent.access.settings, key, settings_dict[key])
+
+        # Commit
+        uow.register(RecordCommitOp(parent))
+        if record:
```
ntarocco, line 796: is `record` optional? Because I can see that it is checked in the permissions in line `777`.

invenio_rdm_records/services/schemas/parent/access.py
```diff
@@ -70,12 +70,18 @@ class AccessSettingsSchema(Schema):
     allow_user_requests = fields.Boolean()
     allow_guest_requests = fields.Boolean()
 
-    accept_conditions_text = SanitizedHTML()
+    accept_conditions_text = SanitizedHTML(allow_none=True)
+    secret_link_expiration = fields.Integer(allow_none=True)
```
ntarocco, line 74: Shall we put a `max limit`? E.g. 1 year? Or can I overflow the max integer? :D

invenio_rdm_records/services/access/service.py
```diff
@@ -749,3 +754,54 @@ def create_guest_access_request(self, identity, token, expand=False, uow=None):
...
+
+        expiration_days = self._translate_expiration_date(
+            data["access"]["settings"]["secret_link_expiration"]
+        )
+        data["access"]["settings"]["secret_link_expiration"] = expiration_days
+
+        # Validation
+        data, __ = self.schema_parent.load(
+            data, context=dict(identity=identity), raise_errors=True
+        )
```
max-moser, line 787: is it expected for `data` to have the structure of a full parent record (i.e. with all the [other fields](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/schemas/parent/__init__.py#L40-L48) there as well)?
given that a few lines further down it looks like everything except for the `settings` values are ignored anyway, i think this may be a bit overkill.
maybe we should only require the `access` structure?

invenio_rdm_records/services/schemas/parent/access.py
max-moser: my integer? :woozy_face:

invenio_rdm_records/services/access/service.py
zzacharo: there are some changes done in the other [PR](https://github.com/inveniosoftware/invenio-rdm-records/pull/1385/files#diff-f1ba3f9be59bb3905cbc2e8d700d43e2a22281ff63f9dd08b79b99e07ce26c45R772) @anikachurilova is working on so it might affect how this PR is looking now

invenio_rdm_records/services/access/service.py
anikachurilova: Got this fixed in the merged PR: https://github.com/inveniosoftware/invenio-rdm-records/pull/1385

invenio_rdm_records/services/access/service.py
```diff
@@ -754,6 +754,13 @@ def create_guest_access_request(self, identity, token, expand=False, uow=None):
             uow=uow,
         )
 
+    def _translate_expiration_date(self, value):
+        """Translate secret_link_expiration from ui dropdown value."""
```
ntarocco, line 758: We should not assume in the backend that there is a UI dropdown component in the frontend, this is REST API :)

invenio_rdm_records/services/schemas/parent/access.py
```diff
@@ -71,6 +71,9 @@ class AccessSettingsSchema(Schema):
     allow_guest_requests = fields.Boolean()
 
     accept_conditions_text = SanitizedHTML(allow_none=True)
+    secret_link_expiration = fields.Integer(
```
ntarocco, line 74: The conversion should happen here :) You can use a [pre/post_load](https://marshmallow.readthedocs.io/en/stable/extending.html#pre-processing-and-post-processing-methods) to achieve that.

ntarocco (approved):

PR #1399, opened 2023-08-07
moderation: add 'verified' field to records

### LINKED ISSUE #1398: [USER MODERATION] - 3rd deliverable -  sort records by 'verified'
`verified` information should be added to the record so we can filter or sort records during search based on the verified property. 

In this issue, we should:

- Add `verified` property to a record
- Add filtering / sorting to record's search (e.g. similar to what is done on [Zenodo](https://github.com/zenodo/zenodo/blob/407cab4248828da355b1fe76d9c1b56fc30b5efe/zenodo/modules/records/query.py#L47))
- Tests

---

**REQUIRES 2ND DELIVERABLE:**

- https://github.com/inveniosoftware/invenio-users-resources/issues/95

alejandromumo:
closes #1398 

NEEDS:

- https://github.com/inveniosoftware/invenio-records-resources/pull/481

### DIFF
--- a/invenio_rdm_records/records/mappings/os-v1/rdmrecords/drafts/draft-v6.0.0.json
+++ b/invenio_rdm_records/records/mappings/os-v1/rdmrecords/drafts/draft-v6.0.0.json
@@ -239,6 +239,9 @@
               }
             }
           },
+          "is_verified": {
+            "type": "boolean"
+          },
           "review": {
             "type": "object",
             "properties": {
--- a/invenio_rdm_records/records/systemfields/is_verified.py
+++ b/invenio_rdm_records/records/systemfields/is_verified.py
@@ -17,11 +17,8 @@ class IsVerifiedField(CalculatedField):
         super().__init__(key=key, use_cache=False)
 
     def calculate(self, record):
-        """Calculate the ``is_expired`` property of the request."""
-        owner = record.access.owner
+        """Calculate the ``is_verified`` property of the record."""
+        owner = record.access.owner.resolve()
         if not owner:
             return False
-        is_verified = (
-            owner.resolve().verified_at is not None
-        )  # TODO property is_verified in user
-        return is_verified
+        return owner.verified_at is not None
--- a/invenio_rdm_records/records/systemfields/is_verified.py
+++ b/invenio_rdm_records/records/systemfields/is_verified.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""Record 'verified' system field."""
+
+from invenio_records_resources.records.systemfields.calculated import CalculatedField
+
+
+class IsVerifiedField(CalculatedField):
+    """Systemfield for calculating whether or not the request is expired."""
+
+    def __init__(self, key=None):
+        """Constructor."""
+        super().__init__(key=key, use_cache=False)
+
+    def calculate(self, record):
+        """Calculate the ``is_expired`` property of the request."""
+        owner = record.access.owner
+        if not owner:
+            return False
+        is_verified = (
+            owner.resolve().verified_at is not None
+        )  # TODO property is_verified in user
+        return is_verified
--- a/invenio_rdm_records/services/config.py
+++ b/invenio_rdm_records/services/config.py
@@ -29,6 +29,7 @@ from invenio_drafts_resources.services.records.config import (
     is_draft,
     is_record,
 )
+from invenio_drafts_resources.services.records.search_params import AllVersionsParam
 from invenio_indexer.api import RecordIndexer
 from invenio_records_resources.services import ConditionalLink, FileServiceConfig
 from invenio_records_resources.services.base.config import (
@@ -46,6 +47,11 @@ from invenio_records_resources.services.records.config import (
 from invenio_records_resources.services.records.links import (
     RecordLink,
     pagination_links,
+)
+from invenio_records_resources.services.records.params import (
+    FacetsParam,
+    PaginationParam,
+    QueryStrParam,
 )
 from invenio_requests.services.requests import RequestItem, RequestList
 from invenio_requests.services.requests.config import RequestSearchOptions
@@ -75,6 +81,7 @@ from .schemas.parent.access import AccessSettingsSchema
 from .schemas.parent.access import Grant as GrantSchema
 from .schemas.parent.access import SecretLink as SecretLinkSchema
 from .schemas.record_communities import RecordCommunitiesSchema
+from .sort import VerifiedRecordsSortParam
 
 
 def is_draft_and_has_review(record, ctx):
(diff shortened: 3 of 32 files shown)


invenio_rdm_records/services/services.py
```diff
@@ -128,3 +128,19 @@ def oai_result_item(self, identity, oai_record_source):
             record,
             links_tpl=self.links_item_tpl,
         )
+
+    @unit_of_work()
+    def reindex_user_records(self, identity, id_, uow=None):
+        """Reindexes all the records from a given user."""
+        self.require_permission(identity, "manage")
```
alejandromumo, line 135: Not 100% sure about this permission.

The service exposes one "reindex" method in which permissions are set to `search`. 

However, for 'access' we usually require `manage` permissions

invenio_rdm_records/records/systemfields/is_verified.py
```diff
@@ -0,0 +1,27 @@
...
+        super().__init__(key=key, use_cache=False)
+
+    def calculate(self, record):
+        """Calculate the ``is_expired`` property of the request."""
+        owner = record.access.owner
+        if not owner:
+            return False
+        is_verified = (
+            owner.resolve().verified_at is not None
+        )  # TODO property is_verified in user
```
alejandromumo, line 26: WIP: Ideally, we want a computed property "is_verified" on the user

invenio_rdm_records/services/services.py
```diff
@@ -128,3 +128,26 @@ def oai_result_item(self, identity, oai_record_source):
             record,
             links_tpl=self.links_item_tpl,
         )
+
+    @unit_of_work()
+    def reindex_latest_first(self, identity, id_, extra_filter=None, uow=None):
```
alejandromumo, line 133: naming is open for suggestions.

question: should we have this in `drafts-resources` or leave it here? The idea is to have a way of re-indexing the latest version of a record first and siblings afterwards.

invenio_rdm_records/services/services.py
```diff
@@ -128,3 +128,26 @@ def oai_result_item(self, identity, oai_record_source):
...
+        records = self.scan(identity, params=params)
+
+        for record in records:
+            record, parent = self._get_record_and_parent_by_id(record["id"])
+            # Commit (and re-index) latest versions
+            uow.register(RecordCommitOp(parent))
+            if record:
+                uow.register(RecordCommitOp(record))
+
+            self._index_related_records(record, parent, uow=uow)
```
alejandromumo, line 151: This function might be a bottleneck since it will query DB directly and re-index one by one. 

Maybe we can use bulk indexing instead?

invenio_rdm_records/records/mappings/os-v1/rdmrecords/drafts/draft-v6.0.0.json
alejandromumo, line 1: question: I am not entirely sure about adding the mapping to drafts as well.

invenio_rdm_records/services/schemas/parent/__init__.py
```diff
@@ -46,6 +46,7 @@ class RDMParentSchema(ParentSchema, FieldPermissionsMixin):
         values=fields.Nested(PIDSchema),
         dump_only=True,
     )
+    is_verified = fields.Boolean(dump_only=True)
```
alejandromumo, line 49: question: do we want to add permissions to read this file?

e.g. in the users service we added:


```python
can_read_system_details = [AdminAction(user_management_action)]
```

invenio_rdm_records/records/mappings/os-v1/rdmrecords/drafts/draft-v6.0.0.json
ntarocco: The `is_verified` is a calculated system field, which will not be part of the record saved in the db. Or am I wrong? You might don't need to add it to the various db schemas. It is probably necessary for OpenSearch mappings.

invenio_rdm_records/services/config.py
```diff
@@ -126,6 +133,16 @@ class RDMSearchOptions(SearchOptions, SearchOptionsMixin):
         "access_status": facets.access_status,
     }
 
+    # TODO we could do SearchOptions.params_interpreters_cls
```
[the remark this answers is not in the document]
ntarocco: We have a similar problem with `service components`: if you change them via config, you might break functionalities.
At some point, we discussed that we should have a kind of "private" components, and to override them is not simply via a config var, but via a method that will explicitly list them somehow, a guard to warn you: Hey, do you know what you are doing? We didn't do it yet.

In components, we have a comment `# Order of components are important!`: you might want to add something similar, we another sentence.

ntarocco, line 136: I would remove the `TODO`.

invenio_rdm_records/services/schemas/parent/__init__.py
ntarocco: I don't see it has a sensitive field. What is the behaviour in Zenodo today?

tests/services/test_rdm_service.py
```diff
@@ -127,3 +128,42 @@ def test_embargo_lift_with_error(running_app, search_clear, minimal_record):
...
+    assert v_draft
+    v_record = service.publish(id_=v_draft.id, identity=v_user.identity)
+    assert v_record
+
+    res = service.search(nv_user.identity)
+    assert res.total == 2
+    hits = res.to_dict()["hits"]["hits"]
+
+    expected_order = [v_record.id, nv_record.id]
+    assert expected_order == [h["id"] for h in hits]
```
ntarocco, line 169: Nice test! However, it might be that the default sorting is latest first, and your test will pass anyway.
If you want to be 100% correct in your test, I would:
1. Make sure that the `unverified` record is the one published the last
2. Set `RDM_SEARCH_SORT_BY_VERIFIED = False`, the default
3. Search **sorting by latest** and assert that `unverified` is first
4. Set `RDM_SEARCH_SORT_BY_VERIFIED = True`
5. Search **sorting by latest** and assert that `verified`, even if older, is first

ntarocco (commented):
Nice! Good job!

tests/services/test_rdm_service.py
alejandromumo: Thanks for the suggestion!

invenio_rdm_records/records/mappings/os-v1/rdmrecords/drafts/draft-v6.0.0.json
alejandromumo: Indeed it won't be saved in the DB. 
However, it is currently being indexed also for drafts. Otherwise, other tests fail for drafts. 

I think it is not necessary to have the field in drafts, I need to check how to only add it for the records on indexing.

alejandromumo: I had a closer look. 

`is_verified` is a field on the `parent`, which is a common field for both records and drafts.

I don't see a problem with having `is_verified` on the draft since it won't be used (at least for now).

WDYT @ntarocco

invenio_rdm_records/services/schemas/parent/__init__.py
alejandromumo: Zenodo does not dump the field, therefore is not visible in API queries.

Ideally, we could protect the field and only make it visible to admins. However, it would "pollute" a bit of the record's namespace by introducing user-related permission.

Options:

1 - Do not dump the field at all (I don't see why not)
2 - Add a permission to dump it (e.g. `AdminAction(user_management_action)`)

WDYT?

ntarocco: I would personally leave it like this. Can you write this down to discuss it with Alex/Lars?

ntarocco (approved):

invenio_rdm_records/services/schemas/parent/__init__.py
alejandromumo: Created an issue https://github.com/inveniosoftware/invenio-rdm-records/issues/1411

And added it to the epic Tasks https://github.com/inveniosoftware/invenio-app-rdm/issues/2290

PR #1404, opened 2023-08-09
views: add signposting

fenekku:
See https://github.com/inveniosoftware/rfcs/pull/73

### DIFF
--- a/invenio_rdm_records/resources/serializers/signposting/schema.py
+++ b/invenio_rdm_records/resources/serializers/signposting/schema.py
@@ -1,0 +1,228 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2023 Northwestern University.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Signposting schemas."""
+
+import idutils
+from marshmallow import Schema, fields, missing
+
+from ...urls import download_url_for
+from ..utils import get_vocabulary_props
+
+
+class LandingPageSchema(Schema):
+    """Schema for serialization of typed links pertaining to the landing page.
+
+    Serialization input is a whole record dict projection.
+    """
+
+    anchor = fields.Method(serialize="serialize_anchor")
+    author = fields.Method(serialize="serialize_author")
+    cite_as = fields.Method(data_key="cite-as", serialize="serialize_cite_as")
+    describedby = fields.Method(serialize="serialize_describedby")
+    item = fields.Method(serialize="serialize_item")
+    license = fields.Method(serialize="serialize_license")
+    type = fields.Method(serialize="serialize_type")
+
+    def serialize_anchor(self, obj, **kwargs):
+        """Seralize to landing page URL."""
+        return obj["links"]["self_html"]
+
+    def serialize_author(self, obj, **kwargs):
+        """Serialize author(s).
+
+        For now, the first linkable identifier is taken.
+        """
+
+        def pick_linkable_id(identifiers):
+            for id_dict in identifiers:
+                url = idutils.to_url(
+                    id_dict["identifier"], id_dict["scheme"], url_scheme="https"
+                )
+                if url:
+                    return url
+                else:
+                    continue
+            return None
+
+        metadata = obj["metadata"]
+        result = [
+            {"href": pick_linkable_id(c["person_or_org"].get("identifiers", []))}
+            for c in metadata.get("creators", [])
+        ]
+        result = [r for r in result if r["href"]]
+        return result or missing
+
+    def serialize_cite_as(self, obj, **kwargs):
+        """Serialize cite-as."""
+        identifier = obj.get("pids", {}).get("doi", {}).get("identifier")
+        if not identifier:
+            return missing
+
+        url = idutils.to_url(identifier, "doi", "https")
+
+        return [{"href": url}] if url else missing
+
+    def serialize_describedby(self, obj, **kwargs):
+        """Serialize describedby."""
+        # Has to be placed here to prevent circular dependency.
+        from invenio_rdm_records.resources.config import record_serializers
+
+        result = [
+            {"href": obj["links"]["self"], "type": mimetype}
+            for mimetype in sorted(record_serializers)
+        ]
+
+        return result or missing
+
+    def serialize_item(self, obj, **kwargs):
+        """Serialize item."""
+        file_entries = obj.get("files", {}).get("entries", {})
+
+        result = [
+            {
+                "href": download_url_for(pid_value=obj["id"], filename=entry["key"]),
+                "type": entry["mimetype"],
+            }
+            for entry in file_entries.values()
+        ]
+
+        return result or missing
+
+    def serialize_license(self, obj, **kwargs):
+        """Serialize license.
+
+        Note that we provide an entry for each license (rather than just 1).
+        """
+        rights = obj["metadata"].get("rights", [])
+
+        def extract_link(right):
+            """Return link associated with right.
+
+            There are 2 types of right representations:
+             - custom
+             - controlled vocabulary
+
+            If no associated link returns None.
+            """
+            # Custom
+            if right.get("link"):
+                return right["link"]
+            # Controlled vocabulary
+            elif right.get("props"):
+                return right["props"].get("url")
+
+        result = [extract_link(right) for right in rights]
+        result = [{"href": link} for link in result if link]
+        return result or missing
+
+    def serialize_type(self, obj, **kwargs):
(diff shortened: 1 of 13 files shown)


fenekku:
failing tests have to do with seemingly unrelated code:

FAILED tests/requests/test_user_moderation_actions.py::test_user_moderation_approve
FAILED tests/services/test_rdm_service.py::test_search_sort_verified_enabled

Will check later if any actual connection.

invenio_rdm_records/resources/serializers/signposting/schema.py
```diff
@@ -0,0 +1,215 @@
...
+                try:
+                    url = idutils.to_url(
+                        id_dict["identifier"], id_dict["scheme"], url_scheme="https"
+                    )
+                    if url:
+                        return url
+                    else:
+                        continue
+                except Exception as e:
+                    print("pick linkable id exception")
```
lnielsen, line 53: Comment: print statements

invenio_rdm_records/resources/serializers/signposting/schema.py
```diff
@@ -0,0 +1,215 @@
...
+
+        result = [{"href": right["link"]} for right in rights if right.get("link")]
+
+        return result or missing
+
+    def serialize_type(self, obj, **kwargs):
+        """Serialize type."""
+        resource_type = obj["metadata"]["resource_type"]
+
+        rt_record = vocabulary_service.read(
```
lnielsen, line 119: I think this is going to hit the database, so it should be using some sort of cache to not get a performance hit. @ntarocco was looking into similar issues with the search results.

invenio_rdm_records/resources/urls.py
```diff
@@ -0,0 +1,57 @@
...
+This module contains minimal methods to generate URLs correctly without much
+engineering. Over time, it can be made more abstract, complex and powerful and even
+extracted into its own package to solve url generation across InvenioRDM once and for
+all.
+"""
+
+from flask import current_app
+
+
+def record_url_for(_app="ui", pid_value=""):
```
lnielsen, line 29: Comment: I understand that the current link generation doesn't work for this use case, but I'm concerned that introducing these two methods will muddy the picture of where link generation should happen.

Would a possible solution be to return a URITemplate in the ``links`` or ``linktpls`` which could be used to generate the links you need. I.e. these two methods would take as input a URItemplate which was generated by the service layer and simply render the string.

lnielsen (commented):
Thanks a lot for implementing all this. It's very relevant and timely for InvenioRDM so fully support this gets.

fenekku:
Working on it

invenio_rdm_records/resources/urls.py
fenekku: Hmm this is a longer discussion. The practical piece I got from the comment is to rely on the already generated links that are part of the record projection that is being serialized into `json+linkset` in the `schema.py`. I've adapted the code to rely on that e.g., https://github.com/inveniosoftware/invenio-rdm-records/pull/1404/files#diff-32235a765fef3b5ef2104531e468805d799d3d8b9db5b501d09295b4a365b4eaR33 :+1: .
The download link is not part of the provided serialized links, so I kept the use of `download_url_for` there. I also kept `record_url_for` in the header link generation, since we don't necessarily have access to those links there either.  

We can have a longer discussion about link generation in a dedicated ticket. Main desires for me are (to repeat some from the docs from urls.py): 
- avoid already potential conflicts between routes defined in `APP_RDM_ROUTES` and routes defined in `RDMRecordServiceConfig.links_item` (link generation is already muddied)
- be able to generate links anywhere with the same call pattern and different arguments i.e. like `url_for` but also from either UI or API application (in or out of request context). Maybe name it `inveniordm_url_for` 
- have well defined endpoint names to support that

fenekku (commented):
Changes completed. Re-check :)

invenio_rdm_records/resources/urls.py
```diff
@@ -0,0 +1,60 @@
...
+    url_path = current_app.config["APP_RDM_ROUTES"]["record_detail"].replace(
+        "<pid_value>", pid_value
+    )
+
+    return "/".join(p.strip("/") for p in [url_prefix, url_path])
+
+
+def download_url_for(pid_value="", filename=""):
+    """Return url for download route."""
+    url_prefix = current_app.config.get(f"SITE_UI_URL", "")
```
chokribr, line 50: Probablement le 'f' est superflu

fenekku:
Rebased + tests updated for new application/ld+json mimetype + take into account new files structure (fixes 

```
"href": download_url_for(pid_value=obj["id"], filename=entry["key"]),
TypeError: string indices must be integers
```

and other related errors.

invenio_rdm_records/resources/serializers/signposting/schema.py
```diff
@@ -0,0 +1,212 @@
...
+        result = [
+            {"href": pick_linkable_id(c["person_or_org"].get("identifiers", []))}
+            for c in metadata.get("creators", [])
+        ]
+        result = [r for r in result if r["href"]]
+        return result or missing
+
+    def serialize_cite_as(self, obj, **kwargs):
+        """Serialize cite-as."""
+        identifier = obj.get("pids", {}).get("doi", {}).get("identifier")
```
chokribr, line 62: Probably we need to move to the last version of the record (api/records/7ze6y-5fa68/versions/latest) in order  to get pids object since it depends on  the version of the record
 I wonder if  we can add other pids in addition to the  DOI such oai-pmh ?

fenekku: Ah, per the spec for the cite-as field:

> Provide a link that has the persistent identifier of the scholarly object (e.g. DOI expressed as an HTTP URI) as target. Note that dereferencing the target of a link with a cite-as relation type must lead to the landing page that provided that cite-as link. 

, the signposting is per scholarly object (i.e. version), so getting the current version's DOI as we do is right. Different versions are cited differently and that's good since we can refer to a specific version. 

The cardinality of cite-as is 1 and its link must follow the conditions above (persistent and redirects back to this page) so an oai-pmh link probably wouldn't do.

It would be good to bring up the notion of versions to the spec authors. FAIR profiling mentions the landing page, resource and metadata, but doesn't mention inter-version connections.

invenio_rdm_records/resources/serializers/signposting/schema.py
```diff
@@ -0,0 +1,212 @@
...
+        return result or missing
+
+    def serialize_license(self, obj, **kwargs):
+        """Serialize license.
+
+        Note that we provide an entry for each license (rather than just 1).
+        """
+        rights = obj["metadata"].get("rights", [])
+
+        result = [{"href": right["link"]} for right in rights if right.get("link")]
```
chokribr, line 103: I think that we need rights.values!

chokribr, line 103: probably you mean props.url?

invenio_rdm_records/resources/serializers/signposting/schema.py
chokribr: We have to check if the record didn't have a DOI identifier but others PIDs were  mentioned and to decide which identifier we will chose!

fenekku:
Just need to check the to_dict() representation for search before I put this back out of draft. (Although, the "application/linkset+json" serialization is not really meant to be search serializes anyway). By end of tomorrow my time, it should be resolved.

I currently suspect that most serialization tests are not using a proper to_dict() representation as input and therefore are potentially doing unnecessary work (but I may be off). That's why I included a static full_record_to_dict fixture and will check the search use case. I'd be happy to refactor the other serializations if that is the case. Thanks to @chokribr for bringing this to my attention when checking the functionality.

tests/resources/serializers/conftest.py
```diff
@@ -36,3 +36,573 @@ def enhanced_full_record(enhanced_full_record, parent_record):
...
     enhanced_full_record["parent"] = parent_record
     return enhanced_full_record
+
+
+# TODO: Use this representation in all serializer tests
+@pytest.fixture
+def full_record_to_dict():
+    """The to_dict() representation of a full record.
+
+    THIS is the representation that a serializer gets as input.
```
fenekku, line 46: The full_record / minimal_record in root conftest.py are (were) meant as record creation input from the REST API / service side of things, but they have been altered over time. Relying on them is more confusing/clunky than not, ergo the creation of clear to_dict() fixtures here.

invenio_rdm_records/resources/serializers/signposting/schema.py
fenekku: Resolved. See comments below :+1:

lnielsen (approved):

PR #1427, opened 2023-08-25
oai-pmh: read oai sets prefix from app config

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/202

### DIFF
--- a/invenio_rdm_records/oaiserver/services/services.py
+++ b/invenio_rdm_records/oaiserver/services/services.py
@@ -25,7 +25,6 @@ from sqlalchemy.sql import text
 
 from invenio_rdm_records.oaiserver.services.errors import (
     OAIPMHSetDoesNotExistError,
-    OAIPMHSetIDDoesNotExistError,
     OAIPMHSetNotEditable,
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/oaiserver/services/services.py
```diff
@@ -38,7 +37,7 @@ class OAIPMHServerService(Service):
     def __init__(self, config, extra_reserved_prefixes={}):
         """Init service with config."""
         super().__init__(config)
-        self.reserved_prefixes = config.reserved_prefixes.union(extra_reserved_prefixes)
+        self.extra_reserved_prefixes = extra_reserved_prefixes
```
rekt-hard, line 40: Having an object as a default value will use the same object for all instances of this class. Better to have `extra_reserved_prefixes` defined as `None` and then initializing like `self.extra_reserved_prefixes = extra_reserved_prefixes or {}`

rekt-hard (commented):
just a small comment on initializing an instance. Otherwise LGTM

zzacharo (approved):

PR #1428, opened 2023-08-25
Record deletion: service workflows

max-moser:
Requires https://github.com/inveniosoftware/invenio-drafts-resources/pull/246 in order to properly filter out deleted records from the user's dashboard.

### DIFF
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -33,6 +33,7 @@ from invenio_requests.resources.requests.config import RequestSearchRequestArgsS
 from ..services.errors import (
     AccessRequestExistsError,
     InvalidAccessRestrictions,
+    RecordDeletedException,
     ReviewExistsError,
     ReviewNotFoundError,
     ReviewStateError,
@@ -175,6 +176,17 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
             lambda e: HTTPJSONException(
                 code=400,
                 description=e.description,
+            )
+        ),
+        RecordDeletedException: create_error_handler(
+            lambda e: (
+                HTTPJSONException(code=404, description=_("Record not found"))
+                if not e.record.tombstone.is_visible
+                else HTTPJSONException(
+                    code=410,
+                    description=_("Record deleted"),
+                    tombstone=e.record.tombstone.dump(),
+                )
             )
         ),
     }
--- a/invenio_rdm_records/resources/serializers/ui/schema.py
+++ b/invenio_rdm_records/resources/serializers/ui/schema.py
@@ -3,7 +3,7 @@
 # Copyright (C) 2020 CERN.
 # Copyright (C) 2020 Northwestern University.
 # Copyright (C) 2021-2023 Graz University of Technology.
-# Copyright (C) 2022 TU Wien.
+# Copyright (C) 2022-2023 TU Wien.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
@@ -19,11 +19,12 @@ from invenio_communities.communities.resources.ui_schema import (
     _community_permission_check,
 )
 from invenio_i18n import get_locale
+from invenio_i18n import lazy_gettext as _
 from invenio_records_resources.services.custom_fields import CustomFieldsSchemaUI
 from invenio_vocabularies.contrib.awards.serializer import AwardL10NItemSchema
 from invenio_vocabularies.contrib.funders.serializer import FunderL10NItemSchema
 from invenio_vocabularies.resources import L10NString, VocabularyL10Schema
-from marshmallow import Schema, fields, missing, pre_dump
+from marshmallow import Schema, fields, missing, post_dump, pre_dump
 from marshmallow_utils.fields import FormatDate as FormatDate_
 from marshmallow_utils.fields import FormatEDTF as FormatEDTF_
 from marshmallow_utils.fields import SanitizedHTML, SanitizedUnicode, StrippedHTML
@@ -92,6 +93,22 @@ def record_version(obj):
         return f"v{obj['versions']['index']}"
 
     return field_data
+
+
+def mask_removed_by(obj):
+    """Mask information about who removed the record."""
+    return_value = _("Unknown")
+    removed_by = obj.get("removed_by", None)
+
+    if removed_by is not None:
+        user = removed_by.get("user", None)
+
+        if user == "system":
+            return_value = _("System (automatic)")
+        elif user is not None:
+            return_value = _("User")
+
+    return return_value
 
 
 class RelatedIdentifiersSchema(Schema):
@@ -220,6 +237,24 @@ def compute_publishing_information(obj, dummyctx):
     return result
 
 
+class TombstoneSchema(Schema):
+    """Schema for a record tombstone."""
+
+    removal_reason = fields.Nested(VocabularyL10Schema, attribute="removal_reason")
+
+    note = fields.String(attribute="note")
+
+    removed_by = fields.Function(mask_removed_by)
+
+    removal_date_l10n_medium = FormatEDTF(attribute="removal_date", format="medium")
+
+    removal_date_l10n_long = FormatEDTF(attribute="removal_date", format="long")
+
+    citation_text = fields.String(attribute="citation_text")
+
+    is_visible = fields.Boolean(attribute="is_visible")
+
+
 class UIRecordSchema(BaseObjectSchema):
     """Schema for dumping extra information for the UI."""
 
@@ -288,6 +323,8 @@ class UIRecordSchema(BaseObjectSchema):
         attribute="metadata.funding",
     )
 
+    tombstone = fields.Nested(TombstoneSchema, attribute="tombstone")
+
     @pre_dump
     def add_communities_permissions_and_roles(self, obj, **kwargs):
         """Inject current user's permission to community receiver."""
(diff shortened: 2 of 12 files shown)


invenio_rdm_records/services/permissions.py
```diff
@@ -240,12 +240,16 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     # - Records/files are updated/deleted via drafts so we don't support
     #   using below actions.
     can_update = [Disable()]
-    can_delete = [Disable()]
+    can_delete = [
+        SystemProcess()
```
max-moser, line 244: note: the service takes care of checking the preconditions of deletion workflow actions (i.e. if the record's current deletion status is wrong), and preventing the execution if they're not fulfilled by raising an exception.
i think, admins have permissions to do any actions that are not explicitly `Disable()`d

invenio_rdm_records/services/services.py
```diff
@@ -128,3 +145,289 @@ def oai_result_item(self, identity, oai_record_source):
...
+            params=None,
+            search_preference=None,
+            expand=False,
+            **kwargs,
+        )
+
+    #
+    # Base methods, extended with handling of deleted records
+    #
+    def read(self, identity, id_, expand=False, read_deleted=False):
```
max-moser, line 419: the `read_deleted` flag is new, this is how we could enable administrators to review "normal" landing pages for **deleted**/**marked** records.

invenio_rdm_records/services/services.py
```diff
@@ -128,3 +145,289 @@ def oai_result_item(self, identity, oai_record_source):
...
+    def restore_record(self, identity, id_, expand=False, uow=None):
+        """Restore a record that has been (soft) deleted."""
+        record = self.record_cls.pid.resolve(id_)
+        if record.deletion_status != RecordDeletionStatusEnum.DELETED:
+            raise DeletionStatusException(RecordDeletionStatusEnum.DELETED, record)
+
+        self.require_permission(identity, "delete", record=record)
+
+        record.deletion_status = RecordDeletionStatusEnum.PUBLISHED
+        record.tombstone = None
```
zzacharo, line 265: move to component?

invenio_rdm_records/services/services.py
zzacharo, line 419:
```suggestion
    def read(self, identity, id_, expand=False, with_deleted=False):
```
I think in other places we use the `with_deleted` convention e.g https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L370

invenio_rdm_records/services/services.py
```diff
@@ -128,3 +145,289 @@ def oai_result_item(self, identity, oai_record_source):
...
+
+        # TODO restore as metadata-only if there's no files left
```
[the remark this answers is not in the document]
max-moser: for the future, yes.
when we implement the hard-deletion (i.e. `cleanup` and `purge` workflows), then this needs to be taken into account.
it should be moved into a separate ticket or a comment on the upcoming issue, though.

invenio_rdm_records/services/services.py
```diff
@@ -128,3 +145,289 @@ def oai_result_item(self, identity, oai_record_source):
...
+        raise NotImplementedError()
+
+    #
+    # Search functions
+    #
+    def _add_deletion_filter(self, kwargs_dict, deletion_status, positive_filter=True):
+        """Add an 'extra_filter' for the record deletion status to the kwarg dict."""
+        # the "filter" clause has some benefits (like caching), but it can only filter
+        # for "good" values while sometimes it's necessary to exclude "bad" ones instead
+        if positive_filter:
```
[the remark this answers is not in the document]
max-moser: no, the difference here isn't related to anything being deleted or not.
instead, it's about filtering for the **presence** of a specific value (which i termed "positive" filter, hence the name) vs. filtering for the **absence** of a _set_ of values (which is required if the field can be completely absent as well, as in the case of `deletion_status` for drafts).

invenio_rdm_records/services/services.py
max-moser: update: as per comment, has been renamed to `with_deleted`

invenio_rdm_records/services/services.py
```diff
@@ -128,3 +145,289 @@ def oai_result_item(self, identity, oai_record_source):
...
+                "record": record,
+            },
+            raise_errors=True,
+        )
+
+        # If no `citation_text` has been supplied, create one
+        # TODO: move into component
+        if not data.get("citation_text", None):
+            serializer = CSLJSONSerializer()
+            style = get_style_location("APA")  # TODO from app config
```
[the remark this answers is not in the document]
max-moser: now makes use of the already existing `RDM_CITATION_STYLES_DEFAULT` config item

invenio_rdm_records/services/schemas/record.py
```diff
@@ -91,16 +91,12 @@ def default_nested(self, data):
...
-        is_deleted = data.get("deletion_status", {}).get("is_deleted", False)
+        is_deleted = (data.get("deletion_status") or {}).get("is_deleted", False)
+        tombstone_visible = (data.get("tombstone") or {}).get("is_visible", True)
 
-        if is_deleted:
-            keys_to_keep = ["tombstone", "deletion_status", "id", "pids", "links"]
-            for key in [k for k in data if k in keys_to_keep]:
-                data.pop(key, None)
-
-        else:
```
max-moser, line 103: this was a bit overzealous; the actual stripping of data now happens in the rendering (i.e. jinja template for UI, and resource [error handler] for the API)

invenio_rdm_records/resources/config.py
```diff
@@ -177,6 +178,17 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
...
             )
         ),
+        RecordDeletedException: create_error_handler(
+            lambda e: (
+                HTTPJSONException(code=404, description=_("Record not found"))
+                if not e.record.tombstone.is_visible
+                else HTTPJSONException(
+                    code=410,
+                    description=_("Record deleted"),
+                    tombstone=e.record.tombstone.dump(),
```
max-moser, line 188: if we want to keep other entries (like `links`, `identifier`, etc.) in the response, this is the place to put them for the REST API.

invenio_rdm_records/services/components/record_deletion.py
```diff
@@ -0,0 +1,75 @@
...
+class RecordDeletionComponent(ServiceComponent):
+    """Service component for record deletion."""
+
+    def delete_record(self, identity, data=None, record=None, **kwargs):
+        """Set the record's deletion status and tombstone information."""
+        # If no `citation_text` has been supplied, create one
+        if not data.get("citation_text", None):
+            locale = get_locale() or current_app.config.get(
+                "BABEL_DEFAULT_LOCALE", "en"
+            )
```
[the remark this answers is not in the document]
zzacharo: use this https://github.com/inveniosoftware/invenio-i18n/blob/master/invenio_i18n/ext.py#L190 ?

invenio_rdm_records/services/services.py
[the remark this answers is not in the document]
zzacharo: I think we need an extra descriptive docstring here to understand what that function does or what about splitting in 2 methods with a descriptive name?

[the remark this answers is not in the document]
max-moser: updated the docstring to be extra specific.
however, i'm a bit unsure if that's made matters better given that this is just an internal helper function

invenio_rdm_records/services/components/record_deletion.py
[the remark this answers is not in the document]
max-moser: thanks for the pointer, i've just used this instead.

however, maybe we should still just use the system's default locale instead, to be more consistent across the board (regardless of who deleted the record and what locale they use)?
that being said, probably it doesn't make a lot of difference in the real world anyway.

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -94,6 +94,22 @@ def record_version(obj):
...
+def mask_removed_by(obj):
+    """Mask information about who removed the record."""
+    return_value = "Unknown"
+    removed_by = obj.get("removed_by", None)
+
+    if removed_by is not None:
+        user = removed_by.get("user", None)
+
+        if user == "system":
+            return_value = "System (automatic)"
```
kpsherva, line 106: all the returned values are missing translations, unless I misunderstood this serializer and it is done somewhere else

invenio_rdm_records/services/permissions.py
```diff
@@ -167,7 +168,7 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     # Actions
     #
     # Allow to put a record in edit mode (create a draft from record)
-    can_edit = can_curate
+    can_edit = can_curate + [IfDeleted(then_=[Disable()], else_=[])]
```
zzacharo, line 171:
```suggestion
    can_edit = [IfDeleted(then_=[Disable()], else_=can_curate)]
```
no?

invenio_rdm_records/resources/serializers/ui/schema.py
zzacharo: to be extra sure, does flask-babel return the resolved value as a string-to-translate when we extract messages?

invenio_rdm_records/services/services.py
```diff
@@ -128,3 +141,282 @@ def oai_result_item(self, identity, oai_record_source):
...
         )
+
+    #
+    # Deletion workflows
+    #
+    @unit_of_work()
+    def delete_record(self, identity, id_, data, expand=False, uow=None):
+        """(Soft) delete a published record."""
+        record = self.record_cls.pid.resolve(id_)
+        if record.deletion_status.is_deleted:
```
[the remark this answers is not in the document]
max-moser: No, this shouldn't be the case as we're specifically resolving a `record` here.
In the generator, I went a bit more defensive because I think the `record` there could be an actual **draft** as well (and they don't have a `deletion_status`).

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -94,6 +95,22 @@ def record_version(obj):
...
+        elif user is not None:
+            return_value = "User"
+
+    return _(return_value)
```
kpsherva, line 111: I am afraid this won't work on extract messages step, since it is runtime evaluation, to be tested

zzacharo (approved):

PR #1430, opened 2023-08-28
Customize PID provider and record serializers, fix files serialization, add `x-bibtex` serializer, expose "owned_by", fix AJAX headers

slint:
* Closes https://github.com/zenodo/zenodo-rdm/issues/468

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositApiClient.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositApiClient.js
@@ -41,7 +41,7 @@ export class DepositApiClient {
       withCredentials: true,
       xsrfCookieName: "csrftoken",
       xsrfHeaderName: "X-CSRFToken",
-      headers: this.apiHeaders.json,
+      headers: this.apiHeaders["vnd+json"],
     };
     this.axiosWithConfig = axios.create(this.apiConfig);
     this.cancelToken = axios.CancelToken;
@@ -117,7 +117,6 @@ export class RDMDepositApiClient extends DepositApiClient {
     const payload = this.recordSerializer.serialize(draft);
     return this._createResponse(() =>
       this.axiosWithConfig.post(this.createDraftURL, payload, {
-        headers: this.apiHeaders["vnd+json"],
         params: { expand: 1 },
       })
     );
@@ -131,7 +130,6 @@ export class RDMDepositApiClient extends DepositApiClient {
   async readDraft(draftLinks) {
     return this._createResponse(() =>
       this.axiosWithConfig.get(draftLinks.self, {
-        headers: this.apiHeaders["vnd+json"],
         params: { expand: 1 },
       })
     );
@@ -146,7 +144,6 @@ export class RDMDepositApiClient extends DepositApiClient {
     const payload = this.recordSerializer.serialize(draft);
     return this._createResponse(() =>
       this.axiosWithConfig.put(draftLinks.self, payload, {
-        headers: this.apiHeaders["vnd+json"],
         params: { expand: 1 },
       })
     );
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -124,7 +124,10 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
 
     request_search_args = RDMSearchRequestArgsSchema
 
-    response_handlers = record_serializers
+    response_handlers = FromConfig(
+        "RDM_RECORDS_SERIALIZERS",
+        default=record_serializers,
+    )
 
     error_handlers = {
         DeserializerError: create_error_handler(
@@ -343,7 +346,10 @@ class RDMCommunityRecordsResourceConfig(RecordResourceConfig, ConfiguratorMixin)
     url_prefix = "/communities"
     routes = {"list": "/<pid_value>/records"}
 
-    response_handlers = record_serializers
+    response_handlers = FromConfig(
+        "RDM_RECORDS_SERIALIZERS",
+        default=record_serializers,
+    )
 
 
 class RDMRecordCommunitiesResourceConfig(CommunityResourceConfig, ConfiguratorMixin):
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -44,6 +44,7 @@ from .deserializers import ROCrateJSONDeserializer
 from .deserializers.errors import DeserializerError
 from .errors import HTTPJSONException, HTTPJSONValidationWithMessageAsListException
 from .serializers import (
+    BibtexSerializer,
     CSLJSONSerializer,
     DataCite43JSONSerializer,
     DataCite43XMLSerializer,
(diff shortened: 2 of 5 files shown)


invenio_rdm_records/services/config.py
```diff
@@ -250,8 +251,8 @@ class RDMRecordServiceConfig(RecordServiceConfig, ConfiguratorMixin):
     """RDM record draft service config."""
 
     # Record and draft classes
-    record_cls = RDMRecord
-    draft_cls = RDMDraft
+    record_cls = FromConfig("RDM_RECORD_CLS", default=RDMRecord)
+    draft_cls = FromConfig("RDM_DRAFT_CLS", default=RDMDraft)
```
slint, line 255: The usual argument about making "individual attributes" vs. "entire classes" config-based applies to this PR.

invenio_rdm_records/services/schemas/files.py
```diff
@@ -19,11 +19,12 @@
 class FileSchema(Schema):
     """File schema."""
 
-    checksum = fields.String()
+    id = fields.String(attribute="file.id")
```
slint, line 22: This is a tiny bit "internal", but could be helpful for caching and other purposes.

invenio_rdm_records/services/schemas/parent/access.py
```diff
@@ -80,9 +80,8 @@ class ParentAccessSchema(Schema, FieldPermissionsMixin):
 
     field_dump_permissions = {
         # omit fields from dumps except for users with 'manage' permissions
-        # allow only 'settings'
+        # allow only 'settings' and 'owned_by'
         "grants": "manage",
-        "owned_by": "manage",
```
slint, line 85: This is to expose the `owned_by.user.id` in the Zenodo Legacy REST API.

Besides that, another argument for exposing it could be that this doesn't really "leak" any private information. Someone can with some creative detective work figure out via a search query a user's records to find their user ID.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositApiClient.js
```diff
@@ -41,7 +41,7 @@ export class DepositApiClient {
       withCredentials: true,
       xsrfCookieName: "csrftoken",
       xsrfHeaderName: "X-CSRFToken",
-      headers: this.apiHeaders.json,
+      headers: this.apiHeaders["vnd+json"],
```
kpsherva, line 44: this change assumes vnd+json is defined

invenio_rdm_records/services/config.py
```diff
@@ -104,7 +104,7 @@ def is_record_and_has_parent_doi(record, ctx):
 def has_doi(record, ctx):
     """Determine if a record has a DOI."""
     pids = record.pids or {}
-    return "doi" in pids
+    return "doi" in pids and pids["doi"].get("identifier")
```
kpsherva, line 107: this change seems a bit not backwards compatible, even if it will be evaluated to true in normal comparison, if someone used `has_doi is True` or similar, it will explode or return unexpected behaviour
maybe we should consider using types at some point ...?

invenio_rdm_records/services/schemas/files.py
```diff
@@ -19,11 +19,12 @@
...
+    checksum = fields.String(attribute="file.checksum")
     ext = fields.String(attribute="file.ext")
     key = SanitizedUnicode()
     metadata = fields.Dict()
-    mimetype = fields.String(attribute="file.mimetype")
+    mimetype = fields.String(attribute="mimetype")
```
kpsherva, line 27: why not file.mimetype ? seems a bit inconsistent with the rest of the fields

invenio_rdm_records/services/schemas/parent/access.py
kpsherva: why does owners' user id need to be exposed in the api ?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositApiClient.js
kpsherva: just noticed it was like this before but below, still not sure what are the consequences

invenio_rdm_records/services/schemas/files.py
slint: The `mimetype` is actually coming from `ObjectVersion.mimetype`, not `ObjectVersion.file.mimetype`, so without this we don't get it at all.

invenio_rdm_records/services/schemas/parent/access.py
slint: > why does owners' user id need to be exposed in the API ?

Main reason is I need it at the Zenodo legacy REST API serializations, and if it's not dumped here, it's not available later on at the response serializer.

Besides that, I don't see an issue with exposing it here. User IDs can be discovered via search queries (one can do a binary search to find the user ID of a record a la `id:<record_id> AND parent.access.owned_by.user:(1 to 1000)`), so we're just obfuscating here, and not hiding.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositApiClient.js
slint: Doesn't the same apply for `.json`? Both are defined [in `BASE_HEADERS`](https://github.com/inveniosoftware/invenio-rdm-records/blob/76077d75756e946945e20efb96a0f6f7530766eb/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositApiClient.js#L11-L18).

invenio_rdm_records/resources/config.py
```diff
@@ -80,6 +81,7 @@ def csl_url_args_retriever():
         StringCitationSerializer(url_args_retriever=csl_url_args_retriever),
         headers={"content-type": "text/plain"},
     ),
+    "application/x-bibtex": ResponseHandler(BibtexSerializer()),
```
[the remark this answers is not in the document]
slint: Discussed with him IRL: we don't conflict because he's changing the serializer, which right now is only exposed via the record landing page's export formats. This change just adds one more place to expose the format (via content negotiation on the REST API).

invenio_rdm_records/services/schemas/files.py
kpsherva: maybe makes sense to leave a comment then, someone else will notice the inconsistency for sure. Maybe we are missing a test to cover the absence of this attribute...

slint: Actually, the previous value made much more sense... and my logic is not 100% correct based on what I see now in the dumper... I'm double-checking to see if it still works

slint: Ok, can confirm it worked with the previous value, because it uses [`RDMFileRecord.file`](https://github.com/inveniosoftware/invenio-records-resources/blob/a6d972fb0f28cebe7c5af1d768b8781a39460238/invenio_records_resources/records/api.py#L77-L80).

slint:
Included in #1443

PR #1431, opened 2023-08-29
Access requests: Move a filter condition into the db query

max-moser:
Because filtering is more effective the earlier on you can do it!


max-moser:
The changes for the requests DB query have been added in another PR, and the SQLAlchemy changes may need a bump of minimum SA to >=1.4 in Invenio-DB to be 100% safe.
closing for now

PR #1439, opened 2023-08-31
systemfield: add files field that restricts files count dumping

zzacharo:
requires https://github.com/inveniosoftware/invenio-records-resources/pull/492
requires https://github.com/inveniosoftware/invenio-records/pull/314
closes https://github.com/zenodo/rdm-project/issues/166

### DIFF
--- a/invenio_rdm_records/records/api.py
+++ b/invenio_rdm_records/records/api.py
@@ -43,6 +43,7 @@ from . import models
 from .dumpers import (
     EDTFDumperExt,
     EDTFListDumperExt,
+    FilesDumperExt,
     GrantTokensDumperExt,
     StatisticsDumperExt,
 )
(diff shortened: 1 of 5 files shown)


invenio_rdm_records/records/systemfields/files.py
```diff
@@ -0,0 +1,38 @@
...
+"""RDM Record Files field.
+
+The field is implemented here to be able to dump specific fields based on record
+access.
+"""
+
+from invenio_records_resources.records.systemfields.files.field import FilesField
+
+
+class RecordFilesField(FilesField):
```
slint, line 19: Would it make sense to implement this as a search dumper, that just removes the files? We only care about filtering what is indexed, not e.g. what is rendered for the item (since e.g. when the record owner is making a request they should be able to see the files)

PR #1450, opened 2023-09-07
User moderation: Delete a user's records when blocking them

max-moser:
Fixes https://github.com/inveniosoftware/invenio-app-rdm/issues/2379

Screenshot of a blocked user's deleted record:
image

### DIFF
--- a/invenio_rdm_records/requests/user_moderation/actions.py
+++ b/invenio_rdm_records/requests/user_moderation/actions.py
@@ -1,24 +1,101 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2023 CERN.
+# Copyright (C) 2023 TU Wien.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
 """RDM user moderation action."""
 
 from invenio_access.permissions import system_identity
+from invenio_pidstore.errors import PIDDoesNotExistError
+from invenio_vocabularies.proxies import current_service
 
-from invenio_rdm_records.proxies import current_rdm_records_service
+from ...proxies import current_rdm_records_service
+from ...records.systemfields.deletion_status import RecordDeletionStatusEnum
+
+
+def _get_records_for_user(user_id):
+    """Helper function for getting all the records of the user.
+
+    Note: This function performs DB queries yielding all records for a given
+    user (which is not hard-limited in the system) and performs service calls
+    on each of them. Thus, this function has the potential of being a very
+    heavy operation, and should not be called as part of the handling of an
+    HTTP request!
+    """
+    record_cls = current_rdm_records_service.record_cls
+    model_cls = record_cls.model_cls
+    parent_cls = record_cls.parent_record_cls
+    parent_model_cls = parent_cls.model_cls
+
+    # get all the parent records owned by the blocked user
+    parent_recs = [
+        parent_cls(m.data, model=m)
+        for m in parent_model_cls.query.filter(
+            parent_model_cls.json["access"]["owned_by"]["user"].as_string() == user_id
+        ).all()
+    ]
+
+    # get all child records of the chosen parent records
+    recs = [
+        record_cls(m.data, model=m)
+        for m in model_cls.query.filter(
+            model_cls.parent_id.in_([p.id for p in parent_recs])
+        ).all()
+    ]
+
+    return recs
 
 
 def on_block(user_id, uow=None, **kwargs):
-    """Removes records that belong to a user."""
-    pass
+    """Removes records that belong to a user.
+
+    Note: This function operates on all records of a user and thus has the potential
+    to be a very heavy operation! Thus it should not be called as part of the handling
+    of an HTTP request!
+    """
+    user_id = str(user_id)
+    tombstone_data = {"note": "User was blocked"}
+
+    # set the removal reason if the vocabulary item exists
+    try:
+        removal_reason_id = kwargs.get("removal_reason_id", "misconduct")
+        vocab = current_service.read(
+            identity=system_identity, id_=("removalreasons", removal_reason_id)
+        )
+        tombstone_data["removal_reason"] = {"id": vocab.id}
+    except PIDDoesNotExistError:
+        pass
+
+    # soft-delete all the published records of that user
+    for rec in _get_records_for_user(user_id):
+        if not rec.deletion_status.is_deleted:
+            current_rdm_records_service.delete_record(
+                system_identity,
+                rec.pid.pid_value,
+                tombstone_data,
+                uow=uow,
+            )
 
 
 def on_restore(user_id, uow=None, **kwargs):
-    """Restores records that belong to a user."""
-    pass
+    """Restores records that belong to a user.
+
+    Note: This function operates on all records of a user and thus has the potential
+    to be a very heavy operation! Thus it should not be called as part of the handling
+    of an HTTP request!
+    """
+    user_id = str(user_id)
+
+    # restore all the deleted records of that user
+    for rec in _get_records_for_user(user_id):
+        if rec.deletion_status == RecordDeletionStatusEnum.DELETED:
+            current_rdm_records_service.restore_record(
+                system_identity,
+                rec.pid.pid_value,
+                uow=uow,
+            )
 
 
 def on_approve(user_id, uow=None, **kwargs):
--- a/tests/requests/test_user_moderation_actions.py
+++ b/tests/requests/test_user_moderation_actions.py
@@ -1,6 +1,7 @@
 # # -*- coding: utf-8 -*-
 # #
 # # Copyright (C) 2023 CERN.
+# # Copyright (C) 2023 TU Wien.
 # #
 # # Invenio-RDM is free software; you can redistribute it and/or modify
 # # it under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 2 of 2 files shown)


invenio_rdm_records/requests/user_moderation/actions.py
```diff
@@ -8,12 +8,48 @@
...
+    parent_model_cls = parent_cls.model_cls
+    # TODO check if the removal_reason vocabulary entry exists
+    # TODO maybe set a different removal_reason? take it from the kwargs?
+    tombstone_data = {
+        "note": "User was blocked",
+        "removal_reason": {"id": "misconduct"},
+    }
+
+    # get all the parent records owned by the blocked user
+    parent_recs = [
```
kpsherva, line 29: these can be potentially costly queries to run, is this action done in celery task ?

invenio_rdm_records/requests/user_moderation/actions.py
```diff
@@ -8,12 +8,48 @@
...
+    recs = [
+        record_cls(m.data, model=m)
+        for m in model_cls.query.filter(
+            model_cls.parent_id.in_([p.id for p in parent_recs])
+        ).all()
+    ]
+
+    # soft-delete all the published records of that user
+    for rec in recs:
+        if not rec.deletion_status.is_deleted:
```
kpsherva, line 46: same as above

invenio_rdm_records/requests/user_moderation/actions.py
```diff
@@ -8,12 +8,48 @@
...
 from invenio_access.permissions import system_identity
 
-from invenio_rdm_records.proxies import current_rdm_records_service
+from ...proxies import current_rdm_records_service
 
 
 def on_block(user_id, uow=None, **kwargs):
     """Removes records that belong to a user."""
-    pass
+    user_id = str(user_id)
```
kpsherva, line 16: could you add handling "on_restore" action as well ? I think it is also already possible from the record side

invenio_rdm_records/services/errors.py
```diff
@@ -154,9 +153,9 @@ def __init__(self, record_id, community_id):
     def description(self):
         """Exception description."""
         return _(
-            "The record {record_id} in not included in the community {community_id}.".format(
-                record_id=self.record_id, community_id=self.community_id
-            )
+            "The record %(rec_id)s in not included in the community %(com_id)s.",
```
kpsherva, line 156: ping @zzacharo @ntarocco @slint just flagging this to have your opinion

I find `%` operator for string formatting very unreadable and less flexible than `.format`, as far as I know there is negligible to no difference in performance during the evaluation

invenio_rdm_records/services/errors.py
```diff
@@ -188,6 +187,9 @@ def __init__(self, request_id):
     def description(self):
         """Exception description."""
         if self.request_id:
-            return _(f"Identical access requests already exist: {self.request_id}")
+            return _(
+                "Identical access requests already exist: %(request_id)s",
```
kpsherva, line 191: if by any change request id is of numeric type, the formatting will raise an exception

invenio_rdm_records/requests/user_moderation/actions.py
max-moser: sure, i was about to add it already – but then i wasn't sure about _which_ of the user's deleted records to restore.
all of them?

because some records may have been deleted already before the user was blocked, and i guess we don't want to restore them...

invenio_rdm_records/services/errors.py
max-moser: i completely agree, the `%` format is not as nice as f-strings by any means!
however, from my understanding, they're required by `lazy_gettext()`...

invenio_rdm_records/services/errors.py
max-moser: it should just be cast into a string:
```
In [4]: "that's %(id)s example" % {"id": 1}
Out[4]: "that's 1 example"
```

invenio_rdm_records/services/errors.py
kpsherva: isn't lazy_gettext ok also with `.format(` ?
I remember f-strings not working with lazy_gettext but .format should be fine (to be verified - don't trust my memory!)

invenio_rdm_records/services/errors.py
kpsherva: doesn't it need to be `%(id)d` instead of `%(id)s` for the number to be cast to string ?

invenio_rdm_records/requests/user_moderation/actions.py
alejandromumo: Yes, in the user's service, we kick off a celery task and execute all the actions inside it.

We can leave a comment here (e.g. `this is meant to be used ...`).

invenio_rdm_records/requests/user_moderation/actions.py
kpsherva: thats a good question, I will raise it with Lars

lnielsen: I would restore everything for now. Later we need to perhaps allow an admin to select what to restore. For now I would do everything. Usually a blocked user is a misbehaving  and will normally not have any prior deleted records.

invenio_rdm_records/services/errors.py
max-moser: no – `%(id)d` will fail if given anything other than a number (decimal numbers will be truncated), while `%(id)s` will accept anything that can be converted to a string (`str(value)`, `value.__str__()`).
```python
In [47]: "%(x)s" % {"x": 1}
Out[47]: '1'

In [48]: "%(x)s" % {"x": "1"}
Out[48]: '1'

In [49]: "%(x)s" % {"x": "a"}
Out[49]: 'a'

In [50]: "%(x)d" % {"x": 1}
Out[50]: '1'

In [51]: "%(x)d" % {"x": "1"}
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[51], line 1
----> 1 "%(x)d" % {"x": "1"}

TypeError: %d format: a number is required, not str

In [52]: "%(x)d" % {"x": "a"}
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[52], line 1
----> 1 "%(x)d" % {"x": "a"}

TypeError: %d format: a number is required, not str

In [53]:
```

i guess the benefit of `%d` over `%s` is that you can control stuff like their sign relatively easily: https://pyformat.info/#number_sign

invenio_rdm_records/services/errors.py
lnielsen: Above is a correct analysis.

Correct usage:
```
lazy_gettext("String %(var)s", var="a var")
```

Incorrect usage:
```
lazy_gettext("String {var}".format(var="test")) # will try to translate "String "test" but only "String {var}" was extracted.
lazy_gettext("String {var}").format(var="test") # reduces to gettext and translates immediately.
```

Technically, ``gettext("String {var}").format(var="test")`` will yield a correct result, but you rarely know when you are working with lazy_gettext vs gettext, hence for consistency I would never use it.

lnielsen: Btw, https://inveniordm.docs.cern.ch/develop/howtos/i18n/ is wrong in this case and I see we're missing a section as well https://inveniordm.docs.cern.ch/develop/howtos/i18n/#lazy-vs-non-lazy

max-moser: @lnielsen is not only the fastest merger in the west, but also beat me to this :smile: 

---

I just verified the results in a local instance, having the following code:
`invenio_rdm_records/requests/user_moderation/actions.py`
```python
    with force_locale("de_DE"):
        print(get_locale(), _("User %(user_id)s was blocked", user_id=user_id))
```
`invenio_rdm_records/translations/de/LC_MESSAGES/messages.po` (same for `de_DE` instead of `de`, and similar (with an empty `msgstr` in `invenio_rdm_records/translations/messages.pot`)
```
#: invenio_rdm_records/requests/user_moderation/actions.py:51
#, python-format
msgid "User %(user_id)s was blocked"
msgstr "Benutzer %(user_id)s wurde blockiert"
```
`invenio.cfg`
```python
I18N_LANGUAGES = [
    ('de', _('German')),
    ('de_DE', _('German')),
]
```

This will print the expected values:
```
[2023-09-11 13:58:01,421: WARNING/ForkPoolWorker-17] de_DE
[2023-09-11 13:58:01,421: WARNING/ForkPoolWorker-17]
[2023-09-11 13:58:01,432: WARNING/ForkPoolWorker-17] Benutzer 5 wurde blockiert
```

---

If I simply replace the `%(user_id)s` with `{user_id}` (f-string / format syntax) like so (in the code as well as the translation files):
```python
    with force_locale("de_DE"):
        print(get_locale(), _("User {user_id} was blocked", user_id=user_id))
```

Then I get the following logs:
```
[2023-09-11 14:01:09,448: WARNING/ForkPoolWorker-17] de_DE
[2023-09-11 14:01:09,449: WARNING/ForkPoolWorker-17]
[2023-09-11 14:01:09,460: WARNING/ForkPoolWorker-17] Benutzer {user_id} wurde blockiert
```
Which is also expected, since `Flask-Babel` uses `%` formatting and the syntax is different for that.
The text is translated correctly, but the `user_id=user_id` value is simply discarded.

---

Calling the `.format()` "inside" the translation:
```python
    with force_locale("de_DE"):
        print(get_locale(), _("User {user_id} was blocked".format(user_id=user_id)))
```

Will break the translation:
```
[2023-09-11 14:02:39,452: WARNING/ForkPoolWorker-17] de_DE
[2023-09-11 14:02:39,452: WARNING/ForkPoolWorker-17]
[2023-09-11 14:02:39,464: WARNING/ForkPoolWorker-17] User 5 was blocked
```

---

Lastly, calling it "outside" will work like `gettext`, but waste a few CPU cycles (due to the construction and evaluation of the `LazyString`):
```python
    with force_locale("de_DE"):
        print(get_locale(), _("User {user_id} was blocked").format(user_id=user_id))
```
```
[2023-09-11 14:06:14,551: WARNING/ForkPoolWorker-17] de_DE
[2023-09-11 14:06:14,551: WARNING/ForkPoolWorker-17]
[2023-09-11 14:06:14,551: WARNING/ForkPoolWorker-17] Benutzer 5 wurde blockiert
```

invenio_rdm_records/requests/user_moderation/actions.py
max-moser: Updated the docstring to reflect this!

invenio_rdm_records/requests/user_moderation/actions.py
max-moser: Updated the docstring

zzacharo (approved):

PR #1480, opened 2023-09-18
resources: add administration and moderation actions

kpsherva:
* service: add search all method for admins
* service: add search param to filter by record deletion status
* permissions: add search all permission

* closes https://github.com/inveniosoftware/invenio-app-rdm/issues/2408

### DIFF
--- a/invenio_rdm_records/records/systemfields/is_verified.py
+++ b/invenio_rdm_records/records/systemfields/is_verified.py
@@ -10,7 +10,7 @@ from invenio_records_resources.records.systemfields.calculated import Calculated
 
 
 class IsVerifiedField(CalculatedField):
-    """Systemfield for calculating whether or not the request is expired."""
+    """System field for calculating whether the request is expired."""
 
     def __init__(self, key=None):
         """Constructor."""
--- a/invenio_rdm_records/resources/resources.py
+++ b/invenio_rdm_records/resources/resources.py
@@ -76,9 +76,88 @@ class RDMRecordResource(RecordResource):
                 p(routes["access-request-settings"]),
                 self.update_access_settings,
             ),
+            route("POST", p(routes["delete-record"]), self.delete_record),
+            route("POST", p(routes["restore-record"]), self.restore_record),
+            route("GET", p(routes["list-all"]), self.search_all),
+            route("POST", p(routes["set-record-quota"]), self.set_record_quota),
+            # TODO: move to users?
+            route("POST", routes["set-user-quota"], self.set_user_quota),
         ]
 
         return url_rules
+
+    @request_headers
+    @request_view_args
+    @request_data
+    def set_record_quota(self):
+        """Read the related review request."""
+        item = self.service.set_quota(
+            g.identity,
+            resource_requestctx.view_args["pid_value"],
+            quota_size=resource_requestctx.data["quota_size"],
+            max_file_size=resource_requestctx.data["max_file_size"],
+            notes=resource_requestctx.data["notes"],
+        )
+
+        return {}, 200
+
+    @request_headers
+    @request_view_args
+    @request_data
+    def set_user_quota(self):
+        """Read the related review request."""
+        item = self.service.set_user_quota(
+            g.identity,
+            id_=resource_requestctx.view_args["pid_value"],
+            quota_size=resource_requestctx.data["quota_size"],
+            max_file_size=resource_requestctx.data["max_file_size"],
+            notes=resource_requestctx.data["notes"],
+        )
+
+        return {}, 200
+
+    @request_extra_args
+    @request_search_args
+    @response_handler(many=True)
+    def search_all(self):
+        """Perform a search over the items."""
+        identity = g.identity
+        hits = self.service.search_all(
+            identity=identity,
+            params=resource_requestctx.args,
+            search_preference=search_preference(),
+            expand=resource_requestctx.args.get("expand", False),
+        )
+        return hits.to_dict(), 200
+
+    #
+    # Deletion workflows
+    #
+    @request_headers
+    @request_view_args
+    @request_data
+    def delete_record(self):
+        """Read the related review request."""
+        item = self.service.delete_record(
+            g.identity,
+            resource_requestctx.view_args["pid_value"],
+            resource_requestctx.data,
+        )
+
+        return item.to_dict(), 200
+
+    @request_headers
+    @request_view_args
+    @request_data
+    def restore_record(self):
+        """Read the related review request."""
+        item = self.service.restore_record(
+            g.identity,
+            resource_requestctx.view_args["pid_value"],
+            resource_requestctx.data,
+        )
+
+        return item.to_dict(), 200
 
     #
     # Review request
--- a/invenio_rdm_records/services/permissions.py
+++ b/invenio_rdm_records/services/permissions.py
@@ -8,7 +8,7 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 """Permissions for Invenio RDM Records."""
-
+from invenio_administration.generators import Administration
 from invenio_communities.generators import CommunityCurators
 from invenio_records_permissions.generators import (
     AnyUser,
@@ -94,6 +94,7 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     #
     # Allow searching of records
     can_search = can_all
+    can_search_all = [Administration(), SystemProcess()]
 
     # Allow reading metadata of a record
     can_read = [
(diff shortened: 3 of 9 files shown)


invenio_rdm_records/resources/resources.py
```diff
@@ -76,10 +76,89 @@ def p(route):
...
+    @request_headers
+    @request_view_args
+    @request_data
+    def set_user_quota(self):
+        """Read the related review request."""
+        item = self.service.set_user_quota(
+            g.identity,
+            id_=resource_requestctx.view_args["pid_value"],
+            quota_size=resource_requestctx.data["quota_size"],
+            max_file_size=resource_requestctx.data["max_file_size"],
```
kpsherva, line 113: this should go through marshmallow schema, but I didn't add it, I was not sure if it was done on purpose in the service this way

invenio_rdm_records/services/schemas/record.py
```diff
@@ -65,6 +65,7 @@ class Meta:
     revision = fields.Integer(dump_only=True)
     versions = NestedAttribute(VersionsSchema, dump_only=True)
     parent = NestedAttribute(RDMParentSchema)
+    is_published = fields.Boolean(dump_only=True)
```
kpsherva, line 68: adding it back, suggested to do by @max-moser

invenio_rdm_records/services/services.py
```diff
@@ -323,92 +323,97 @@ def purge_record(self, identity, id_, uow=None):
...
     ):
         """Search for drafts that have not been marked as deleted."""
         # because drafts don't have a 'deletion_status', a simple positive filter
         # won't work in cases where records and drafts are mixed...
         # instead, we exclude all the 'wrong' values here
-        status = RecordDeletionStatusEnum.PUBLISHED.value
-        kwargs = self._add_deletion_filter(kwargs, status, positive_filter=False)
+        # TODO this filter can be removed if we cleanup "orphaned" drafts after deletion
+        deletion_filter = dsl.query.Bool(
+            "must_not", **{"exists": {"field": "deletion_status"}}
```
kpsherva, line 406: Simplified filter from the removed `_add_deletion_filter`

invenio_rdm_records/resources/resources.py
```diff
@@ -76,10 +76,89 @@ def p(route):
                 p(routes["access-request-settings"]),
                 self.update_access_settings,
             ),
+            route("POST", p(routes["delete-record"]), self.delete_record),
```
ntarocco, line 79: Why `POST` and not `DELETE`?

invenio_rdm_records/resources/resources.py
```diff
@@ -76,10 +76,89 @@ def p(route):
...
+            route("POST", p(routes["restore-record"]), self.restore_record),
+            route("GET", p(routes["list-all"]), self.search_all),
+            route("POST", p(routes["set-record-quota"]), self.set_record_quota),
+            # TODO: move to users?
```
ntarocco, line 83: What does this mean? Let's don't add TODOs that will never be addressed :)
Otherwise, please move it under user if possible.

invenio_rdm_records/resources/resources.py
```diff
@@ -76,10 +76,89 @@ def p(route):
...
+            route("POST", routes["set-user-quota"], self.set_user_quota),
         ]
 
         return url_rules
 
+    @request_headers
+    @request_view_args
+    @request_data
+    def set_record_quota(self):
+        """Read the related review request."""
```
max-moser, line 93: Docstring

invenio_rdm_records/resources/resources.py
```diff
@@ -76,10 +76,89 @@ def p(route):
...
+            max_file_size=resource_requestctx.data["max_file_size"],
+            notes=resource_requestctx.data["notes"],
+        )
+
+        return {}, 200
+
+    @request_extra_args
+    @request_search_args
+    @response_handler(many=True)
+    def search_all(self):
```
ntarocco, line 122: what is the difference between this and the one in `records/drafts-resources`? Why duplicating the method with a new one `search_all`?

invenio_rdm_records/records/systemfields/is_verified.py
```diff
@@ -10,7 +10,7 @@
 
 
 class IsVerifiedField(CalculatedField):
-    """Systemfield for calculating whether or not the request is expired."""
+    """System field for calculating whether the request is expired."""
```
alejandromumo, line 13:
```suggestion
    """System field for calculating whether the record is verified."""
```

invenio_rdm_records/resources/resources.py
alejandromumo, line 83: Should this endpoint be moved to user's resource?

ntarocco (commented):
Added a few comments, but maybe some of the questions have been already answered in some earlier discussions.

invenio_rdm_records/services/schemas/record.py
max-moser: yes, having a look at my original [record deletion PR](https://github.com/inveniosoftware/invenio-rdm-records/pull/1386/files#diff-3fc71b0ed029b66a27f137d5f3367a0f066cbc0068aba9ca19048169d017dd10L68-R68) I'm relatively confident that this was an accidental rename on my side

invenio_rdm_records/resources/resources.py
alejandromumo: +1 on adding param validation

invenio_rdm_records/resources/resources.py
kpsherva: I am not sure why it was added to record service in the first place ping @zzacharo . I evaluated moving it to users-resources but users-resources module normally would not know about RDMUserQuota model so if I start moving it now, it would trigger an avalanche of changes, therefore if we decide to do that now, we should be sure where to place it

invenio_rdm_records/services/services.py
max-moser: from my understanding, it's expected that:
* drafts don't have the `deletion_status` field
* (proper) records do have this field

which entails that this query filter would remove all published records from the search results, and only drafts would be considered.
is that intended? (from the method's name I'd assume, yes :sweat_smile: )

of course, I might have missed something and the above assumptions might not even hold true in practice

invenio_rdm_records/services/services.py
```diff
@@ -323,92 +323,97 @@ def purge_record(self, identity, id_, uow=None):
     #
     # Search functions
     #
-    def _add_deletion_filter(self, kwargs_dict, deletion_status, positive_filter=True):
```
max-moser, line 326: probably for the best that this one goes...

invenio_rdm_records/services/permissions.py
```diff
@@ -239,7 +240,7 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     #
     # Record deletion workflows
     #
-    can_delete = [SystemProcess()]
+    can_delete = [Administration(), SystemProcess()]
```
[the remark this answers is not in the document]
alejandromumo: I agree. 
At some point, we discussed having granular action needs to allow different types of users to specialize in what they can manage. 

For example: a librarian can focus on managing records and does not have to worry about moderating users (just an example).

invenio_rdm_records/resources/resources.py
kpsherva: do you mean .search method ? (I assume yes because there is no .search_all)
the difference is in the fact that search_all returns all records, including deleted and marked for purge, which we don't want from the normal .search method, therefore the difference is in filters. 
I have a vague memory that we said deleted records should be indexed separately, but it was not implemented for some reason, maybe it didn't make the MVP cut. ping @max-moser @lnielsen for more information about the scope of this feature

invenio_rdm_records/resources/resources.py
zzacharo: I placed it in rdm-records as it was the fastest way to integrate the feature and I didn't want to go all the evaluation of moving it in users-resources (which I am not sure it even belongs to)... we can remove the TODO, I added more of a note in case someone was confused finding that code there at the moment.

zzacharo:
comments will be addressed in upcoming PR

invenio_rdm_records/services/services.py
kpsherva: I have tested the query it seems to have the same outcome as the previous one

invenio_rdm_records/resources/resources.py
kpsherva: interesting axios quirks made me think DELETE does not accept payload, but after verifying, it is fixed, thanks!

PR #1481, opened 2023-09-18
models: avoid flushing when getting records

ntarocco:


invenio_rdm_records/services/access/service.py
```diff
@@ -500,26 +501,27 @@ def _exists(self, identity, created_by, record_id, request_type):
...
 
-        requests = [
-            request
-            for request in (
-                api_cls(rm.data, model=rm)
-                for rm in model_cls.query.filter(
-                    model_cls.json["created_by"] == created_by,
-                    model_cls.json["topic"] == {"record": record_id},
-                    model_cls.json["type"].as_string() == request_type,
+        with db.session.no_autoflush:
```
ntarocco, line 504: Given that below we do `api_cls(rm.data, model=rm)`, this will trigger an autoflush (UPDATE query) in the db.
We should emulate the [get_records](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L379) behaviour instead.

max-moser: good catch, i'll have to put autoflushes on a mental note for the future!

max-moser (approved):
LGTM, it's just an extra "no autoflush" with indentation changes

zzacharo (approved):

PR #1484, opened 2023-09-19
Record moderation fixes

kpsherva:


invenio_rdm_records/resources/config.py
```diff
@@ -110,7 +110,7 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
     routes["restore-record"] = "/<pid_value>/restore"
     routes["set-record-quota"] = "/<pid_value>/quota"
     routes["set-user-quota"] = "/users/<pid_value>/quota"
-    routes["list-all"] = "/all"
+    routes["list-all"] = "/administration/records"
```
kpsherva, line 113: changed to dedicated endpoint, I will try to solve it in a better way in a separate PR

invenio_rdm_records/resources/resources.py
```diff
@@ -76,9 +76,9 @@ def p(route):
                 p(routes["access-request-settings"]),
                 self.update_access_settings,
             ),
-            route("POST", p(routes["delete-record"]), self.delete_record),
+            route("DELETE", p(routes["delete-record"]), self.delete_record),
```
kpsherva, line 79: thanks @ntarocco

invenio_rdm_records/services/schemas/record.py
```diff
@@ -66,7 +66,7 @@ class Meta:
     versions = NestedAttribute(VersionsSchema, dump_only=True)
     parent = NestedAttribute(RDMParentSchema)
     is_published = fields.Boolean(dump_only=True)
-    is_deleted = fields.Boolean(dump_only=True)
+    has_draft = fields.Boolean(dump_only=True)
```
kpsherva, line 69: for reviewers: I need this exposed, do you see any issues ?

is_deleted was not correct attribute, the deletion status is coming from system field

invenio_rdm_records/services/services.py
```diff
@@ -447,39 +452,45 @@ def set_quota(
...
             )
         else:
             # update record quota
-            self._update_quota(draft_quota, quota_size, max_file_size, notes)
+            self._update_quota(draft_quota, **data)
 
         db.session.add(draft_quota)
 
         # files_attr can be set to "media_files"
+        files_manager = getattr(draft, files_attr)
```
kpsherva, line 488: needed to make it work from the admin, otherwise the files were locked

Issue #1437, opened 2023-08-31
A11y: deposit form needs accessibility improvements


jennur:
Scanning the deposit form with axeDevTools gives 78 errors, and these need to be adressed.

<img width="520" alt="Screenshot 2023-08-31 at 15 25 11" src="https://github.com/inveniosoftware/invenio-rdm-records/assets/21052053/6b58a584-3983-45ae-bef7-d2519563e4c5">

Issue #1461, opened 2023-09-12
A11y: Form elements do not have associated labels


0einstein0:
### Analysed using **WAVE** [Web Accessibility Evaluation Tool]

- References Issue #1437 

-  Need to add respective attributes to the associated elements in the Deposit Form

![Screenshot 2023-09-12 at 16 06 05](https://github.com/inveniosoftware/invenio-rdm-records/assets/51617644/1aa56b03-5c46-4ac7-ac35-62d6f5b5684a)

![Screenshot 2023-09-12 at 15 36 11](https://github.com/inveniosoftware/invenio-rdm-records/assets/51617644/c14a2bdb-69f6-4656-b086-c311174af34f)

PR #1485, opened 2023-09-19
Fixed Accessibility of Deposit Form

0einstein0:
Partially Closes Issues #1437 and #1461 related to the Accessibility of Deposit Form

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/customFields/Thesis.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/customFields/Thesis.js
@@ -31,7 +31,7 @@ export class Thesis extends Component {
         <Grid padded>
           <Grid.Column width="16">
             <Input
-              fieldPath={`${fieldPath}.university`}
+              fieldPath={fieldPath}
               label={university.label}
               placeholder={university.placeholder}
             />
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/customFields/Thesis.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/customFields/Thesis.js
@@ -31,7 +31,7 @@ export class Thesis extends Component {
         <Grid padded>
           <Grid.Column width="16">
             <Input
-              fieldPath={fieldPath}
+              fieldPath={`${fieldPath}.university`}
               label={university.label}
               placeholder={university.placeholder}
             />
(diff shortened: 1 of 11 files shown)


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/customFields/Thesis.js
```diff
@@ -31,7 +31,7 @@ export class Thesis extends Component {
         <Grid padded>
           <Grid.Column width="16">
             <Input
-              fieldPath={fieldPath}
+              fieldPath={`${fieldPath}.university`}
```
kpsherva, line 34: ping @alejandromumo because I don't remember the outcome for thesis fields
shouldn't this path be set on the parent of this component ?

invenio_rdm_records/contrib/thesis/custom_fields.py
```diff
@@ -26,7 +26,7 @@
     "section": _("Thesis"),
     "fields": [
         {
-            "field": "thesis:university",
+            "field": "thesis:thesis",
```
kpsherva, line 29: ping @alejandromumo to verify, I am not sure what was the path in the end

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/customFields/Thesis.js
0einstein0: Screenshot 2023-09-19 at 16 29 42

invenio_rdm_records/contrib/thesis/custom_fields.py
alejandromumo: Indeed `thesis:university` was correct. It maps in many other places, we can't change it here I think:


https://github.com/search?q=org%3Ainveniosoftware%20%22thesis%3Auniversity%22&type=code
https://github.com/search?q=repo%3Azenodo%2Fzenodo-rdm%20%22thesis%3Auniversity%22&type=code

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/customFields/Thesis.js
alejandromumo: yeah I think this path is set on the parent. 

What was the issue there? I only see one occurrence of `thesis:thesis` in the codebase and it's on ZenodoRDM (it's likely wrong, had to investigate)

invenio_rdm_records/contrib/thesis/custom_fields.py
alejandromumo: I believe we can't change the custom field name now, even for migration would mean 💥 

However, if there's something wrong, I can help you out @0einstein0

PR #1489, opened 2023-09-20
services: filter deleted records in search by permissions

kpsherva:
* services: config admin parameter to allow seeing deleted records
            only on the administration panel
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1494

### DIFF
--- a/invenio_rdm_records/resources/args.py
+++ b/invenio_rdm_records/resources/args.py
@@ -19,3 +19,4 @@ class RDMSearchRequestArgsSchema(SearchRequestArgsSchema):
     style = fields.Str()
     locale = fields.Str()
     status = fields.Str()
+    is_admin = fields.Bool()
--- a/invenio_rdm_records/resources/args.py
+++ b/invenio_rdm_records/resources/args.py
@@ -19,4 +19,4 @@ class RDMSearchRequestArgsSchema(SearchRequestArgsSchema):
     style = fields.Str()
     locale = fields.Str()
     status = fields.Str()
-    is_admin = fields.Bool()
+    include_deleted = fields.Bool()
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -119,6 +119,7 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
     request_read_args = {
         "style": ma.fields.Str(),
         "locale": ma.fields.Str(),
+        "include_deleted": ma.fields.Bool(),
     }
 
     request_body_parsers = {
--- a/invenio_rdm_records/resources/config.py
+++ b/invenio_rdm_records/resources/config.py
@@ -110,7 +110,6 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
     routes["restore-record"] = "/<pid_value>/restore"
     routes["set-record-quota"] = "/<pid_value>/quota"
     routes["set-user-quota"] = "/users/<pid_value>/quota"
-    routes["list-all"] = "/administration/records"
 
     request_view_args = {
         "pid_value": ma.fields.Str(),
--- a/invenio_rdm_records/services/generators.py
+++ b/invenio_rdm_records/services/generators.py
@@ -8,6 +8,9 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 """Invenio-RDM-Records Permissions Generators."""
+import operator
+from functools import reduce
+from itertools import chain
 
 from flask_principal import UserNeed
 from invenio_communities.generators import CommunityRoleNeed, CommunityRoles
@@ -18,6 +21,7 @@ from invenio_search.engine import dsl
 
 from ..records import RDMDraft
 from ..records.systemfields.access.grants import Grant
+from ..records.systemfields.deletion_status import RecordDeletionStatusEnum
 from ..requests.access import AccessRequestTokenNeed
 from ..tokens.permissions import RATNeed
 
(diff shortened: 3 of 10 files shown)


invenio_rdm_records/services/search_params.py
```diff
@@ -24,3 +24,33 @@ def apply(self, identity, search, params):
...
+    def apply(self, identity, search, params):
+        """Evaluate the is_admin parameter on the search."""
+        value = params.pop("is_admin", None)
+        # Filter prevents from other users' drafts from displaying on Moderator's
+        # dashboard
+        if value is None:
+            search = search.filter(
+                "term", **{"parent.access.owned_by.user": current_user.id}
+            )
+        return search
```
jrcastro2, line 56: Comment: I personally find it a bit confusing to have a parameter `is_admin`, it feels like it will give superpowers if it's set to `true`. A part from that, I would personally go for more explicit parameters, dividing this 2 Interpreters in 2 different params maybe something like `deleted` and `all`.

zzacharo: What about renaming it to `with_deleted` to indicate that I want the deleted records or not and then return if you have the right permissions?

kpsherva: it will have a naming clash with with_deleted argument for details views, I would like to avoid that

invenio_rdm_records/resources/config.py
```diff
@@ -120,6 +119,7 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
     request_read_args = {
         "style": ma.fields.Str(),
         "locale": ma.fields.Str(),
+        "with_deleted": ma.fields.Bool()
```
ntarocco, line 122: Very minor:
```suggestion
        "deleted": ma.fields.Bool()
```
of `include-deleted`. The `with`, IMO, is more technical/for code.

invenio_rdm_records/services/generators.py
```diff
@@ -135,6 +139,58 @@ def _condition(self, record=None, **kwargs):
...
+        """Make a query for one set of generators."""
+        queries = [g.query_filter(**kwargs) for g in generators]
+        queries = [q for q in queries if q]
+        return reduce(operator.or_, queries) if queries else None
+
+    def query_filter(self, **kwargs):
+        """Filters for current identity."""
+        q_then = dsl.Q("match_all")
+        q_else = dsl.Q(
+            "term", **{"deletion_status": RecordDeletionStatusEnum.PUBLISHED.value}
```
ntarocco, line 179: Will all published records have this?

invenio_rdm_records/resources/args.py
```diff
@@ -19,3 +19,4 @@ class RDMSearchRequestArgsSchema(SearchRequestArgsSchema):
     style = fields.Str()
     locale = fields.Str()
     status = fields.Str()
+    is_admin = fields.Bool()
```
ntarocco, line 22: I don't understand this param, what does it mean?
Checking the code below, it looks like that it is a way to filter "only mine" -> current_user.
I would go for a name more in this direction.

invenio_rdm_records/services/search_params.py
```diff
@@ -24,3 +24,33 @@ def apply(self, identity, search, params):
...
             search = search.filter("term", **{"deletion_status": value})
         return search
+
+
+class AdminParam(ParamInterpreter):
+    """Evaluates the 'status' parameter."""
+
+    def apply(self, identity, search, params):
+        """Evaluate the status parameter on the search."""
+        value = params.pop("is_admin", None)
```
ntarocco, line 34: Is this correct?

invenio_rdm_records/services/services.py
```diff
@@ -429,7 +384,15 @@ def read(self, identity, id_, expand=False, with_deleted=False):
         record = self.record_cls.pid.resolve(id_)
         result = super().read(identity, id_, expand=expand)
 
-        if record.deletion_status.is_deleted and not with_deleted:
+        can_read_deleted = self.check_permission(
```
ntarocco, line 387: `with_deleted` name makes sense to me here.
I would check this permission only if `with_deleted = True`, WDTY?

invenio_rdm_records/services/generators.py
```diff
@@ -135,6 +139,58 @@ def _condition(self, record=None, **kwargs):
             return False
 
 
+class IfRecordDeleted(Generator):
```
zzacharo, line 142: @max-moser added https://github.com/inveniosoftware/invenio-communities/commit/31945c6954780fc32f56f1cd16c3163ca2bb8216#diff-2476adbb5df07ac52ab8b38741dacd9f144d1ee77dbcef244085581885d6c530R148 and it seems similar so we need probably to align if possible these 2 approaches

invenio_rdm_records/services/generators.py
kpsherva: yes, they do, this was added by Max

invenio_rdm_records/services/generators.py
kpsherva: I didn't use it intentionally, he created one also for records, see https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/permissions.py#L174
this was used for edit, and it has slightly different purpose, I didn't want to mix them. Maybe I should rename or Max'es, do you have any idea what would be a good naming ?

invenio_rdm_records/services/search_params.py
```diff
@@ -24,3 +24,33 @@ def apply(self, identity, search, params):
...
+            )
+        return search
+
+
+class AdminDraftsParam(ParamInterpreter):
+    """Evaluates the 'is_admin' parameter."""
+
+    def apply(self, identity, search, params):
+        """Evaluate the is_admin parameter on the search."""
+        value = params.pop("is_admin", None)
```
zzacharo, line 49: same here? should be changed no?

invenio_rdm_records/services/search_params.py
kpsherva: yes, there is no reason to pass the param down. It is also consistent with other query filters

invenio_rdm_records/services/services.py
```diff
@@ -424,13 +379,18 @@ def search_drafts(
...
 
-        if record.deletion_status.is_deleted and not with_deleted:
-            raise RecordDeletedException(record, result_item=result)
+        if not include_deleted and record.deletion_status.is_deleted:
+            can_read_deleted = self.check_permission(
+                identity, "read_deleted", record=record
+            )
+
+            if not can_read_deleted:
+                raise RecordDeletedException(record, result_item=result)
```
jrcastro2, line 393: Major: If I am not mistaken the permission should be checked when trying to read deleted records.

```suggestion
        if not include_deleted and record.deletion_status.is_deleted:
                raise RecordDeletedException(record, result_item=result)
        if include_deleted and record.deletion_status.is_deleted:
                can_read_deleted = self.check_permission(
                     identity, "read_deleted", record=record
                 )
            if not can_read_deleted:
                 raise RecordDeletedException(record, result_item=result)       
```

PR #1492, opened 2023-09-20
resources: add etag headers

zzacharo:
closes https://github.com/zenodo/rdm-project/issues/258

### DIFF
--- a/invenio_rdm_records/resources/resources.py
+++ b/invenio_rdm_records/resources/resources.py
@@ -28,6 +28,7 @@ from importlib_metadata import version
 from invenio_drafts_resources.resources import RecordResource
 from invenio_drafts_resources.resources.records.errors import RedirectException
 from invenio_records_resources.resources.errors import ErrorHandlersMixin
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/resources/resources.py
```diff
@@ -677,7 +678,7 @@ def create_url_rules(self):
 
     def _get_record_with_files(self):
         uuid = resource_requestctx.view_args["uuid"]
-        return self.service.read_record(uuid=uuid, identity=g.identity)
+        return self.service.read_record(uuid=uuid, identity=g.identity).to_dict()
```
lnielsen, line 681: Comment: Perhaps it's better to move this to the service method instead?

lnielsen (approved):

invenio_rdm_records/resources/resources.py
zzacharo: but we always return a ResultItem from a service and we convert `to_dict()` in the resource layer no?

PR #1499, opened 2023-09-26
communities: add resource and service handlers for setting default community

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/188

### DIFF
--- a/invenio_rdm_records/resources/resources.py
+++ b/invenio_rdm_records/resources/resources.py
@@ -296,6 +296,7 @@ class RDMRecordCommunitiesResource(ErrorHandlersMixin, Resource):
             route("POST", routes["list"], self.add),
             route("DELETE", routes["list"], self.remove),
             route("GET", routes["suggestions"], self.get_suggestions),
+            route("PUT", routes["list"], self.set_default),
         ]
         return url_rules
 
@@ -367,6 +368,18 @@ class RDMRecordCommunitiesResource(ErrorHandlersMixin, Resource):
             expand=resource_requestctx.args.get("expand", False),
         )
         return items.to_dict(), 200
+
+    @request_view_args
+    @request_data
+    def set_default(self):
+        """Set default community."""
+        item = self.service.set_default(
+            id_=resource_requestctx.view_args["pid_value"],
+            identity=g.identity,
+            data=resource_requestctx.data,
+        )
+
+        return item, 200
 
 
 class RDMRecordRequestsResource(ErrorHandlersMixin, Resource):
--- a/invenio_rdm_records/services/communities/service.py
+++ b/invenio_rdm_records/services/communities/service.py
@@ -51,6 +51,11 @@ class RecordCommunitiesService(Service, RecordIndexerMixin):
         return ServiceSchemaWrapper(self, schema=self.config.schema)
 
     @property
+    def communities_schema(self):
+        """Returns the communities schema instance."""
+        return ServiceSchemaWrapper(self, schema=self.config.communities_schema)
+
+    @property
     def record_cls(self):
         """Factory for creating a record class."""
         return self.config.record_cls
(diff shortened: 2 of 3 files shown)


invenio_rdm_records/resources/resources.py
```diff
@@ -368,6 +369,24 @@ def get_suggestions(self):
...
+            id_=resource_requestctx.view_args["pid_value"],
+            identity=g.identity,
+            data=resource_requestctx.data,
+        )
+
+        response = {}
+        if item:
+            response["result"] = item
+        if errors:
+            response["errors"] = errors
```
jrcastro2, line 386: Major: We usually manipulate the response and fill it with errors when we allow to pass multiple values that will be processed and some will success and others might fail. It doesn't seem the case here, we can only set one default community, therefore I would suggest to use error handlers to handle the error. You will need to create a error for [here](https://github.com/inveniosoftware/invenio-communities/pull/1033/files#diff-e062fa3bca790c1a96331a1093f0b6c354d7364099ad90915f96f72d9f046236R170) and then you add the handler for that error [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/config.py#L401).

invenio_rdm_records/services/communities/service.py
```diff
@@ -325,3 +325,26 @@ def search_suggested_communities(
...
+    @unit_of_work()
+    def set_default(self, identity, id_, data, uow):
+        """Set default community."""
+        record = self.record_cls.pid.resolve(id_)
+
+        self.require_permission(identity, "manage", record=record)
+
+        errors = []
+        try:
+            record.parent.communities.default = data["data"]
```
jrcastro2, line 338: Major: Usually when processing data, we do use marshmallow schemas to validate it before processing.

invenio_rdm_records/services/communities/service.py
```diff
@@ -325,3 +325,26 @@ def search_suggested_communities(
...
+        except AttributeError as ex:
+            errors.append({"message": str(ex)})
```
jrcastro2, line 340: Major: Connected to the comment on the resource layer, the errors we do only want to control if we are processing multiple operations, whichc doesn't seem the case here, I would just raise the exception and let the error_handler take care of it. Meaning that there is no need to control the errors here

invenio_rdm_records/services/config.py
```diff
@@ -179,6 +180,7 @@ class RDMRecordCommunitiesConfig(ServiceConfig, ConfiguratorMixin):
     )
 
     schema = RecordCommunitiesSchema
+    communities_schema = CommunitiesSchema
```
rekt-hard, line 183: Minor: `CommunitiesSchema` may also benefit from adding a unique and duplicate id check as in https://github.com/anikachurilova/invenio-rdm-records/blob/22fe2376431cfd84047c33a9b05dd6f5624d3cde/invenio_rdm_records/services/schemas/record_communities.py#L30-L55

invenio_rdm_records/services/communities/service.py
```diff
@@ -325,3 +330,27 @@ def search_suggested_communities(
...
+            data,
+            context={
+                "identity": identity,
+            },
+            raise_errors=True,
+        )
+
+        record = self.record_cls.pid.resolve(id_)
+        self.require_permission(identity, "manage", record=record)
+        record.parent.communities.default = valid_data["default"]
```
rekt-hard, line 347: minor: should we check if a community with the default id exists? Maybe even check if all passed ids exist?

rekt-hard (commented):
Minor input on value validation check for ids.

invenio_rdm_records/services/communities/service.py
rekt-hard: Sorry, just saw the other PR where the check happens :)

PR #1517, opened 2023-10-05
deposit form: report invalid value errors on each draft save

anikachurilova:
* closes https://github.com/zenodo/rdm-project/issues/244


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
```diff
@@ -271,6 +269,21 @@ const depositReducer = (state = {}, action) => {
         actionState: action.type,
         actionStateExtra: {},
       };
+    case DRAFT_FETCHED:
+    case DRAFT_SAVE_SUCCEEDED:
+      return {
+        ...state,
+        record: {
+          ...action.payload.data,
```
anikachurilova, line 277: Reason: on the save action populate record only with the new state. Fixes the case when previous state had fields with invalid values, but the new state doesn't have those fields at all

kpsherva (approved):

zzacharo (approved):

PR #1474, opened 2023-09-14
add schemaorg serializer 

mfenner:
### Description

First release of an SchemaorgJSONLDSerializer for InvenioRDM, This completes the work that was merged earlier today. 

### Checklist

Ticks in all boxes and 🟢 on all GitHub actions status checks are required to merge:

- [x] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [x] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [x] I've added relevant test cases.
- [x] I've added relevant documentation.
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/best-practices/i18n/) (for relevant code).
- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/develop/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/develop/best-practices/react/) guidelines (for relevant code).
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/develop/best-practices/accessibility/) guidelines (for relevant code).
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/develop/best-practices/ui/) guidelines (for relevant code).
- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/contribute/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).
- [ ] I've NOT included third-party code (copy/pasted source code or new dependencies).

**Third-party code**

I have included two Python libraries:

* [pydash](https://pypi.org/project/pydash/) for better parsing of nested dicts, a common activity when serializing metadata. Neither `obj["pids"]["doi"]["identifier"]` (doesn't handle missing keys) nor `obj.get("pids", {}).get("doi", {}).get("identifier", None)` (very verbose) are ideal. With pydash you can do `py_.get(obj, "pids.doi.identifier")`.
* [commonmeta-py](https://pypi.org/project/commonmeta-py/), a library for converting scholarly metadata between different formats that I wrote. I am re-using a number of utility methods that are useful for several serializers.

PR #1533, opened 2023-10-11
Add schemaorg serializer

alejandromumo:
closes https://github.com/zenodo/rdm-project/issues/223

This PR was mostly authored by @mfenner , the original [PR](https://github.com/inveniosoftware/invenio-rdm-records/pull/1474) was closed and I am re-opening it here with some minor fixes.

### DIFF
--- a/invenio_rdm_records/resources/serializers/schemaorg/schema.py
+++ b/invenio_rdm_records/resources/serializers/schemaorg/schema.py
@@ -197,8 +197,8 @@ class SchemaorgSchema(BaseSerializerSchema, CommonFieldsMixin):
 
     def get_modification_date(self, obj):
         """Get modification date."""
-        # TODO: implement with the help of test data
-        return missing
+        last_updated = obj.get("updated")
+        return last_updated or missing
 
     def get_language(self, obj):
         """Get language. Schemaorg expects either a string or language dict.
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,6 +52,8 @@ install_requires =
     pytz>=2020.4
     pyyaml>=5.4.0
     python-slugify>=8.0.1
+    # Upper pinning due to commonmeta-py
+    types-requests<2.31.0.7
 
 [options.extras_require]
 tests =
--- a/tests/resources/serializers/test_schemaorg_serializer.py
+++ b/tests/resources/serializers/test_schemaorg_serializer.py
@@ -119,6 +119,11 @@ def test_schemaorg_serializer_full_record(running_app, full_record):
     serializer = SchemaorgJSONLDSerializer()
     serialized_record = serializer.dump_obj(full_record)
 
+    assert serialized_record["dateModified"]
+
+    # Delete to facilitate the comparison with the expected data
+    del serialized_record["dateModified"]
+
     assert serialized_record == expected_data
 
 


setup.cfg
```diff
@@ -46,9 +47,13 @@ install_requires =
     invenio-stats>=4.0.0,<5.0.0
     invenio-vocabularies>=2.0.0,<3.0.0
     nameparser>=1.1.1
+    pycountry>=22.3.5
+    pydash>=6.0.0,<7.0.0
     pytz>=2020.4
     pyyaml>=5.4.0
     python-slugify>=8.0.1
+    # Upper pinning due to commonmeta-py
+    types-requests<2.31.0.7
```
alejandromumo, line 56: For reviewers:

It happens that `commonmeta-py` has a dependency on `types-requests` that is not upper pinned. Therefore, it ends up installing its latest which depends on `urllib3 >= 2` and that clashes with rdm dependencies (namely search dependencies). 

I pinned it here since it would be safer for RDM this way, however this might happen with other dependencies of `commonmeta-py` in the future. Please let me know if this is not the right way to do it

invenio_rdm_records/resources/serializers/schemaorg/schema.py
```diff
@@ -0,0 +1,343 @@
...
+    #
+    # Schema Methods
+    #
+
+    def get_name_type(self, obj):
+        """Get name type. Schemaorg expects either Person or Organisation.
+
+        Defaults to Person.
+        """
+        name_type = py_.get(obj, "person_or_org.type")
```
slint, line 63:
```suggestion
        from invenio_records.dictutils import dict_lookup
        name_type = dict_lookup(obj, "person_or_org.type")
```

minor: to avoid adding the `pydash` dependency.

setup.cfg
slint: It's right, and I think we don't depend on any of the `requests`/network-related features of `commonmeta-py`.

invenio_rdm_records/resources/serializers/schemaorg/schema.py
```diff
@@ -0,0 +1,343 @@
...
+
+        # if date is an interval, use the start date
+        if (type(parsed_date).__name__) == "Interval":
+            parsed_date = parsed_date.lower
+        return str(parsed_date)
+
+    def get_modification_date(self, obj):
+        """Get modification date."""
+        # TODO: implement with the help of test data
+        return missing
```
slint, line 201: minor: isn't this just `obj["updated"]`?

slint (approved):
Minor comments on possibly reducing dependencies, but otherwise looks good.

invenio_rdm_records/resources/serializers/schemaorg/schema.py
alejandromumo: `dict_lookup` raises if it doesn't exist. I fear we would have a lot of:

```python

try:
    x = dict_lookup(obj, "a.b.c.")
except KeyError:
    pass
```

It would be nice if we could pass a `default` and only raise if there is no default.

invenio_rdm_records/resources/serializers/schemaorg/schema.py
alejandromumo: you're right, thanks!

PR #1540, opened 2023-10-14
GitHub release fix

kpsherva:
closes https://github.com/zenodo/rdm-project/issues/339

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,10 @@
 Changes
 =======
 
+Version 4.36.1 (2023-10-14)
+
+- github: read releases by user identity permission
+
 Version 4.36.0 (2023-10-13)
 
 - service add version scan method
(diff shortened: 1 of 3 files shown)


alejandromumo (approved):

alejandromumo:
The user identity is computed from the owner of the repository, I am not sure whether this could have an impact on shared repos.

kpsherva:
> The user identity is computed from the owner of the repository, I am not sure whether this could have an impact on shared repos.
> 
> How is this relevant for the linked issue on sentry?

tries to read a deleted record and fails with 410
good point however, I will fix it another way

alejandromumo:
> > The user identity is computed from the owner of the repository, I am not sure whether this could have an impact on shared repos.
> > How is this relevant for the linked issue on sentry?
> 
> tries to read a deleted record and fails with 410 good point however, I will fix it another way

I think this is fine. If the record is deleted, that Openaire index task should fail and set a key in the cache. 
Afterwards, we have another task that reads from that cache and - if the record is deleted - sends a request to OpenAIRE.

alejandromumo (approved):

PR #1541, opened 2023-10-14
Resolve record

kpsherva:
* closes https://github.com/zenodo/rdm-project/issues/341

### DIFF
--- a/invenio_rdm_records/services/github/release.py
+++ b/invenio_rdm_records/services/github/release.py
@@ -33,8 +33,9 @@ class RDMGithubRelease(GitHubRelease):
 
     def resolve_record(self):
         """Resolves an RDM record from a release."""
-        if not self.release_object.record_id:
+        if (self.release_object is None) or (not self.release_object.record_id):
             return None
+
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/services/github/release.py
```diff
@@ -33,8 +33,9 @@ def metadata(self):
 
     def resolve_record(self):
         """Resolves an RDM record from a release."""
-        if not self.release_object.record_id:
+        if (self.release_object is None) or (not self.release_object.record_id):
```
alejandromumo, line 36: I don't this should happen. A release object is mandatory to have this API wrapper.

Checking the failed scenario, it seems that this user has a release `Processing` (from before the migration):

``` python
 'created': datetime.datetime(2023, 10, 12, 16, 20, 24, 90820),
  'status': <ReleaseStatus.PROCESSING: 'P'>,
```

This seems to be a data issue for me.

kpsherva: there was another ticket about the processing releases, could be this, to be checked if data issue, until then I will close the PR

PR #1568, opened 2023-10-20
serializers: fix bibtex requiring given_name field

anikachurilova:


invenio_rdm_records/resources/serializers/bibtex/schema.py
```diff
@@ -268,15 +269,16 @@ def _format_output_row(self, field, value):
...
         id = data["id"]
 
-        creator = data.get("creator", None)
-        if creator is None:
+        creators = original_data["metadata"].get("creators", [])
+        if not creators:
             return id
 
-        name = creator.get("name")
+        creator = creators[0].get("person_or_org", None)
```
alejandromumo, line 280:
```suggestion
        creator = creators[0].get("person_or_org", {})
```

Otherwise `creator.get()` will throw an exception.

invenio_rdm_records/resources/serializers/bibtex/schema.py
```diff
@@ -268,15 +269,16 @@ def _format_output_row(self, field, value):
...
+        name = creator.get("family_name", creator.get("name"))
```
alejandromumo, line 281:
```suggestion
        name = creator.get("family_name", creator["name"])
```

the default should be `creator["name"]`. If that blows up, then it's fine because the app does not know how to recover from that. At least the name should be there.

alejandromumo (approved):

PR #1571, opened 2023-10-23
github: added default repo creators.

alejandromumo:
closes https://github.com/zenodo/rdm-project/issues/434


invenio_rdm_records/services/github/release.py
```diff
@@ -32,6 +32,39 @@ def metadata(self):
         output = metadata.default_metadata
         output.update(metadata.extra_metadata)
         output.update(metadata.citation_metadata)
+
+        if not output.get("creators"):
```
alejandromumo, line 36: This "pollutes" the release API a bit, I would expect that all the metadata-related operations would be in the metadata class `RDMReleaseMetadata` (or `ZenodoReleaseMetadata`). 

However, that requires that we rework the code and we don't want to do it right now I suppose.

invenio_rdm_records/services/github/release.py
```diff
@@ -32,6 +32,39 @@ def metadata(self):
...
+                                "type": "personal",
+                                "family_name": "Unknown",
+                            },
+                        }
+                    ]
+                }
+            )
+        return output
+
+    def get_owner(self):
```
alejandromumo, line 58: Same as before, this should be contained inside the metadata class. This is about serializing the repo's owner information to RDM

slint (approved):
LGTM, but we should investigate also why `contributors` fail to compute in general.

PR #1579, opened 2023-10-25
schemaorg: serialize "creator"

slint:


invenio_rdm_records/resources/serializers/schemaorg/schema.py
```diff
@@ -135,8 +135,18 @@ class SchemaorgSchema(BaseSerializerSchema, CommonFieldsMixin):
     type_ = fields.Method("get_type", data_key="@type")
     identifier = fields.Method("get_identifiers")
     name = SanitizedUnicode(attribute="metadata.title")
+    creator = fields.List(
```
alejandromumo, line 138: minor: not sure whether this should be a list or just one object.

Based on their [docs](https://schema.org/creator) and schema, it seems that the creator is one person only:

```json
  {
      "@id": "schema:creator",
      "@type": "rdf:Property",
      "rdfs:comment": "The creator/author of this CreativeWork. This is the same as the Author property for CreativeWork.",
      "rdfs:label": "creator",
      "schema:domainIncludes": [
        {
          "@id": "schema:UserComments"
        },
        {
          "@id": "schema:CreativeWork"
        }
      ],
      "schema:rangeIncludes": [
        {
          "@id": "schema:Person"
        },
        {
          "@id": "schema:Organization"
        }
      ]
    },
```

They provide one [example](https://schema.org/CreativeWork) (example #4) where creator is also one object only:

```json
    {
      "@context": "https://schema.org",
      "@type": "TheaterEvent",
      "name": "Julius Caesar at Shakespeare's Globe",
      "location": {
        "@type": "PerformingArtsTheater",
        "name": "Shakespeare's Globe",
        "sameAs": "http://www.shakespearesglobe.com/",
        "address": "London, UK"
      },
      "offers": {
        "@type": "Offer",
        "url": "/examples/ticket/0012301230123"
      },
      "startDate": "2014-10-01T19:30",
      "workPerformed": {
        "@type": "CreativeWork",
        "name": "Julius Caesar",
        "sameAs": "http://en.wikipedia.org/wiki/Julius_Caesar_(play)",
        "sameAs": "http://worldcat.org/entity/work/id/1807288036",
         "creator": {
          "@type": "Person",
          "name": "William Shakespeare",
          "sameAs": "http://en.wikipedia.org/wiki/William_Shakespeare"
        }
      }
    }
```


However, having multiple creators validates in their [validator](https://validator.schema.org/) (I passed our test's expected data), even for multiple creators.

alejandromumo (commented):
Left a question about the field type

invenio_rdm_records/resources/serializers/schemaorg/schema.py
slint: Lists vs. single objects is the bane of JSON-LD :)

In reality, JSON-LD can also be represented as RDF XML... and in XML everything "could be" a list. See e.g. [Example 3 from the `Dataset` schema](https://schema.org/Dataset#eg-0478).

PR #1583, opened 2023-10-27
datacite: fix related identifiers serialization.

alejandromumo:
closes https://github.com/zenodo/rdm-project/issues/472

### DIFF
--- a/invenio_rdm_records/resources/serializers/datacite/schema.py
+++ b/invenio_rdm_records/resources/serializers/datacite/schema.py
@@ -27,6 +27,29 @@ from ....proxies import current_rdm_records_service
 from ...serializers.ui.schema import current_default_locale
 from ..utils import get_preferred_identifier, get_vocabulary_props
 
+RELATED_IDENTIFIER_SCHEMES = {
+    "ark",
+    "arxiv",
+    "bibcode",
+    "doi",
+    "ean13",
+    "eissn",
+    "handle",
+    "igsn",
+    "isbn",
+    "issn",
+    "istc",
+    "lissn",
+    "lsid1",
+    "pmid",
+    "purl",
+    "upc",
+    "url",
+    "urn",
+    "w3id",
+}
+"""Allowed related identifier schemes for DataCite. Vocabulary taken from DataCite 4.3 schema definition."""
+
 
 def get_scheme_datacite(scheme, config_name, default=None):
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/resources/serializers/datacite/schema.py
```diff
@@ -362,7 +385,8 @@ def get_related_identifiers(self, obj):
                 default=scheme,
             )
 
-            if id_scheme:
+            # Only serialize related identifiers with a valid scheme for DataCite.
+            if id_scheme and id_scheme.lower() in RELATED_IDENTIFIER_SCHEMES:
```
slint, line 389: minor: instead of that would it make sense to:

- remove the `"datacite": "Other"` value from the `RDM_RECORDS_IDENTIFIERS_SCHEMES` config (we only need "valid" DataCite values there)
- change it so that the `get_scheme_datacite` calls are a bit more strict? Right now I see `default=scheme` being passed in many places, which is wrong. We want to enforce that if there is not equivalent DataCite scheme for an identifier, it has to be skipped.

invenio_rdm_records/config.py
```diff
@@ -94,7 +94,6 @@ def always_valid(identifier):
     "url": {"label": _("URL"), "validator": idutils.is_url, "datacite": "URL"},
     "urn": {"label": _("URN"), "validator": idutils.is_urn, "datacite": "URN"},
     "w3id": {"label": _("W3ID"), "validator": always_valid, "datacite": "w3id"},
```
slint, line 96: Will it be that now all IDs will be identified as `w3id`, since it's the last entry with `always_valid`?

I'm not sure what the logic for scheme detecting is for related identifiers...

alejandromumo: I can't validate this right now. 
I had a quick look and it seems that this "fallback" does not happen, but I want to be more thorough with the analysis. I will do it in a separate [task](https://github.com/inveniosoftware/invenio-rdm-records/issues/1584)

invenio_rdm_records/resources/serializers/datacite/schema.py
alejandromumo: I added a [task](https://github.com/inveniosoftware/invenio-rdm-records/issues/1584) to tackle the invalid schemes.

We have more schemes that don't map to DataCite's related identifiers (e.g. crossreffunderid)

PR #1587, opened 2023-10-27
Remove old mappings

slint:


utnapischtim:
closed. there is not really a necessity to remove old mappings. if this should be really done, we can reopen it!

PR #1592, opened 2023-10-30
bibtex: added doi and url fields.

alejandromumo:
closes https://github.com/zenodo/rdm-project/issues/494

### DIFF
--- a/invenio_rdm_records/resources/serializers/bibtex/schema.py
+++ b/invenio_rdm_records/resources/serializers/bibtex/schema.py
@@ -40,6 +40,7 @@ class BibTexSchema(BaseSerializerSchema, CommonFieldsMixin):
     pages = fields.Method("get_pages")
     note = fields.Method("get_note")
     venue = fields.Method("get_venue")
+    url = fields.Method("get_url")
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/resources/serializers/bibtex/schema.py
```diff
@@ -227,9 +236,7 @@ def _fetch_fields_map(self, data):
             "doi": data.get("doi", None),
             "month": data.get("date_created", {}).get("month", None),
             "version": data.get("version", None),
-            "url": (lambda doi: None if doi is None else "https://doi.org/" + doi)(
-                data.get("doi", None)
-            ),
+            "url": data.get("url", None),
```
ntarocco, line 239: if the result is `None`, is the url field added with empty value or missing?

ntarocco (approved):

invenio_rdm_records/resources/serializers/bibtex/schema.py
alejandromumo: It is not added to the final string

PR #1593, opened 2023-10-31
jsonschema: remove unique items constrainsts

kpsherva:
* closes https://github.com/zenodo/rdm-project/issues/437

### DIFF
--- a/invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
@@ -387,8 +387,7 @@
   },
   "subjects": {
     "type": "array",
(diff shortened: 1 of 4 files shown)


ntarocco (approved):
Can we create an issue on Zenodo to:
1. fix the db to remove all the duplicates
2. revert this change, linking this PR

I am wondering if, again, a simple script to go through all records (even if many millions) and remove duplicates would have fixed the issue more easily

kpsherva:
> Can we create an issue on Zenodo to:
> 
> 1. fix the db to remove all the duplicates
> 2. revert this change, linking this PR
> 
> I am wondering if, again, a simple script to go through all records (even if many millions) and remove duplicates would have fixed the issue more easily

done: https://github.com/zenodo/zenodo-rdm/issues/643

PR #1596, opened 2023-11-02
schemaorg: added fields to schema to improve fair score.

alejandromumo:
closes https://github.com/zenodo/rdm-project/issues/507

Notes:

1 - there are fields specific to some work types (e.g. `Dataset`) that we haven't implemented yet. We had them in [legacy](https://github.com/zenodo/zenodo/blob/2dbd8ca0ddf9f9ee8b73a7d8e91e6157ce187fab/zenodo/modules/records/serializers/schemas/schemaorg.py#L279) (example for Dataset). 
2 - I did not implement `access_details` as the user requested in the ticket. I could not find the definition except [this](https://f-uji.net/vocab/metadata/property/access_level). I was trying to look at their [source code](https://github.com/pangaea-data-publisher/fuji/tree/master) to see how they parse it. On this one, I also found these [docs](https://www.fairsfair.eu/fairsfair-data-object-assessment-metrics-request-comments)

### DIFF
--- a/invenio_rdm_records/resources/serializers/schemaorg/schema.py
+++ b/invenio_rdm_records/resources/serializers/schemaorg/schema.py
@@ -19,11 +19,41 @@ from idutils import to_url
 from invenio_access.permissions import system_identity
 from invenio_records_resources.proxies import current_service_registry
 from marshmallow import Schema, ValidationError, fields, missing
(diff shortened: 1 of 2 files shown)


slint (commented):
LGTM, adding test cases would be nice, but let's just test on QA everything before deploying to prod.

PR #1606, opened 2023-11-14
schema: add dataset specific fields to jsonld 

alejandromumo:
closes https://github.com/zenodo/rdm-project/issues/525

Example:

image

### DIFF
--- a/invenio_rdm_records/resources/serializers/schemaorg/schema.py
+++ b/invenio_rdm_records/resources/serializers/schemaorg/schema.py
@@ -8,7 +8,7 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 """Schemaorg based Schema for Invenio RDM Records."""
-from typing import Optional
+from copy import deepcopy
 
 import pycountry
 from babel_edtf import parse_edtf
@@ -18,7 +18,7 @@ from flask_resources.serializers import BaseSerializerSchema
 from idutils import to_url
 from invenio_access.permissions import system_identity
 from invenio_records_resources.proxies import current_service_registry
-from marshmallow import Schema, ValidationError, fields, missing
+from marshmallow import Schema, ValidationError, fields, missing, post_dump, pre_dump
 from marshmallow_utils.fields import SanitizedHTML, SanitizedUnicode
 from pydash import py_
 
@@ -205,6 +205,10 @@ class SchemaorgSchema(BaseSerializerSchema, CommonFieldsMixin):
 
     url = fields.Method("get_url")
 
+    # Fields that are specific to certain resource types.
+    measurementTechnique = fields.Method("get_measurement_techniques")
+    distribution = fields.Method("get_distribution")
+
     def get_id(self, obj):
         """Get id. Use the DOI expressed as a URL."""
         doi = py_.get(obj, "pids.doi.identifier")
@@ -219,7 +223,8 @@ class SchemaorgSchema(BaseSerializerSchema, CommonFieldsMixin):
             ["props.schema.org"],
             py_.get(obj, "metadata.resource_type.id"),
         )
-        return props.get("schema.org", "CreativeWork")
+        ret = props.get("schema.org", "https://schema.org/CreativeWork")
+        return ret
 
     def get_size(self, obj):
         """Get size."""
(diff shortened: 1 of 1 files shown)


invenio_rdm_records/resources/serializers/schemaorg/schema.py
```diff
@@ -205,6 +206,20 @@ class SchemaorgSchema(BaseSerializerSchema, CommonFieldsMixin):
...
+
+    @pre_dump
+    def hide_closed_files(self, obj, many, **kwargs):
+        """Hide files if the record is not Open Access."""
+        ret = deepcopy(obj)
+        access = py_.get(ret, "access.files")
+        files = py_.get(ret, "files")
+        if access == "restricted" and files:
+            del ret["files"]
+        return ret
```
slint, line 221: minor/comment: aren't the files already removed by the service schema dump in case the record has restricted files?

invenio_rdm_records/resources/serializers/schemaorg/schema.py
```diff
@@ -442,6 +459,44 @@ def get_citation(self, obj):
         ids = _serialize_identifiers(relids)
         return ids or missing
 
+    def _is_dataset(self):
+        return self.context.get("type") == "https://schema.org/Dataset"
```
slint, line 463: major: since marshmallow field serialization order is not deterministic, it might be that `get_type` isn't called before `get_measurement_techniques` (or `get_distribution`), which means that the `"type"` won't be set in the context.

Maybe it makes sense to make `get_type` a cached property or similar. Then it wouldn't matter who accesses it first.

invenio_rdm_records/resources/serializers/schemaorg/schema.py
```diff
@@ -442,6 +459,44 @@ def get_citation(self, obj):
...
+        if not self._is_dataset():
+            return missing
+
+        files = py_.get(obj, "files.entries", {})
+        distribution_list = []
+        for f_name, f_object in files.items():
+            entry = {
+                "@type": "DataDownload",
+                "encodingFormat": f_object["mimetype"],
+                "contentUrl": obj["links"]["files"] + f"/{f_name}" + "/content",
```
slint, line 494:
```suggestion
                "contentUrl": obj["links"]["files"] + f"/{f_name}/content",
```

slint (commented):
One comment on computing the record type, but otherwise looks good. Maybe it's not possible at this level, but could we check the distribution field in the serializer tests?

Sign-off: @carlinmack @slint

invenio_rdm_records/resources/serializers/schemaorg/schema.py
jrcastro2: They are removed if the user has no permission to view the files, however if the user has permission to see the files they will not be removed by the service shema but still we don't want them to be shown in the jsonld (unless I am missing something)

invenio_rdm_records/resources/serializers/schemaorg/schema.py
jrcastro2: Good point, it turns out the expensive call is already being cached, therefore I don't think we need to cache the get_type method, see [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/resources/serializers/utils.py#L23) the call to read_all which is already cached. I just removed the context and use the `get_type` directly, please let me know if you have a better suggestion

invenio_rdm_records/resources/serializers/schemaorg/schema.py
slint: We want to keep it consistent throughout serializers no matter the end-use (like e.g. in the "regular" InvenioRDM JSON response), so it's ok to show the files to a user that has permission to see them.

jrcastro2: Ok, just to confirm in "old" Zenodo we were also hiding them, see here the [code](https://github.com/zenodo/zenodo/blob/2dbd8ca0ddf9f9ee8b73a7d8e91e6157ce187fab/zenodo/modules/records/serializers/schemas/schemaorg.py#L284-L289). Please confirm that now we don't need this anymore and I will remove it (sorry for being so annoying, just want to make sure we don't need to revert the change :sweat_smile: )

jrcastro2: Discussed IRL, we will remove it

PR #1616, opened 2023-11-28
metadata: auto-populate `publication_date` on new versions

slint:
- Closes https://github.com/zenodo/rdm-project/issues/588

### DIFF
--- a/invenio_rdm_records/services/components/metadata.py
+++ b/invenio_rdm_records/services/components/metadata.py
@@ -10,6 +10,7 @@
 """RDM service component for metadata."""
 
 from copy import copy
+from datetime import datetime
 
 from invenio_drafts_resources.services.records.components import ServiceComponent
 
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/services/components/metadata.py
```diff
@@ -19,6 +20,9 @@ class MetadataComponent(ServiceComponent):
 
     field = "metadata"
     new_version_skip_fields = ["publication_date", "version"]
+    new_version_generated_fields = {
```
alejandromumo, line 23: minor: why not removing `publication_date` from `new_version_skip_fields`? It would simply import the previous version's publication date and then the user could modify it.

slint: I considered that initially, but this is how we had it in legacy and is the reason we ended up with many records that just had the exact same `publication_date` across all versions.

alejandromumo (approved):
Tests are failing, otherwise LGTM :)

invenio_rdm_records/services/components/metadata.py
```diff
@@ -19,6 +20,9 @@ class MetadataComponent(ServiceComponent):
...
+        "publication_date": lambda draft, record: datetime.today().isoformat(),
```
Samk13, line 24: To avoid time zone discrepancies?
```suggestion
        "publication_date": lambda draft, record: datetime.today(timezone.utc).isoformat(),
```

lnielsen (approved):

PR #1628, opened 2023-12-06
schema: add validation for affiliations

jrcastro2:
* closes https://github.com/zenodo/rdm-project/issues/530

### DIFF
--- a/invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
@@ -13,8 +13,7 @@
   },
   "affiliations": {
     "type": "array",
-    "items": {"$ref": "#/affiliation"},
-    "uniqueItems": true
+    "items": {"$ref": "#/affiliation"}
   },
   "agent": {
     "description": "An agent (user, software process, community, ...).",
--- a/invenio_rdm_records/services/schemas/metadata.py
+++ b/invenio_rdm_records/services/schemas/metadata.py
@@ -139,12 +139,36 @@ class PersonOrOrganizationSchema(Schema):
         return data
 
 
+def validate_affiliations_data(data):
+    """Validate affiliations."""
+    affiliations = data.get("affiliations", [])
+    #  return early If there are no affiliations
+    if not affiliations:
+        return
+    # avoid nesting
+    seen_names = {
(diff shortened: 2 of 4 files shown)


invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json
```diff
@@ -13,8 +13,7 @@
   },
   "affiliations": {
     "type": "array",
-    "items": {"$ref": "#/affiliation"},
-    "uniqueItems": true
+    "items": {"$ref": "#/affiliation"}
```
jrcastro2, line 16: I am not sure we are following semantic versioning on this. It looks like not fully, see here history: https://github.com/inveniosoftware/invenio-rdm-records/commits/master/invenio_rdm_records/records/jsonschemas/records/definitions-v2.0.0.json

If you think I shold create a new one, let me know and I will do it.

slint (approved):

slint:
I think the failing test cases, are rely on the removed JSONSchema validation. We can remove the specific clauses for duplicates checking, since it's done in the marshmallow schema now.

invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -139,13 +139,32 @@ def update_names(self, data, **kwargs):
...
+    seen_names = set()
+    for affiliation in affiliations:
+        name = affiliation.get("name")
+        if not affiliation.get("id") and name:
+            if name in seen_names:
+                messages = [_("Duplicated affiliations.")]
+                raise ValidationError({"affiliations": messages})
+            else:
+                seen_names.add(name)
+
```
Samk13, line 154: How about:
```py
def validate_affiliations_data(data):
    """Validate affiliations."""
    affiliations = data.get("affiliations", [])
    #  return early If there are no affiliations
    if not affiliations:
        return
   # avoid nesting
    seen_names = {affiliation.get("name") for affiliation in affiliations if affiliation.get("name")}
    if len(seen_names) != len(affiliations):
       # provide more specific info
        messages = [_("Duplicated affiliations: ") + ", ".join(set(name for name in seen_names if name))]
        raise ValidationError({"affiliations": messages})
```
WDYT?

jrcastro2: RIght! Thank you for the suggestion!

Issue #1291, opened 2023-04-25
iiif: unskip and fix tests


kpsherva:
changes introduced in https://github.com/inveniosoftware/invenio-rdm-records/pull/1271
exposed an issue with dumping `@id` key, therefore the tests had to be skipped until an underlying issue is found (https://github.com/inveniosoftware/invenio-rdm-records/pull/1271/commits/2a53743496dcb7fb34ad33608f024e9c53ec6356)

dfdan:
I've taken a look at this - the test are failing "correctly", ie the manifests are no longer validating after #1271 was merged.

The serializer no longer recieves a file list with the links object included - these links are required by the manifest serializer to generate the @id for both the canvases and the images.

dfdan:
Just bumping...  had a quick word with @slint. Manifest is not valid because here:

https://github.com/inveniosoftware/invenio-rdm-records/blob/ca1be162117eb112d3ff190ee3c5ab4b54ae7645/invenio_rdm_records/resources/serializers/iiif/schema.py#L119C22-L119C22

(and elsewhere) `links` is no longer availablew after #1271 was merged. I'm not sure if we should get them back (will this clog up the index) or generate them here in the serialzer?

kpsherva:
from Lars:
> default_preview in /files but no in record
entries list vs dict
entries differences: bucket, checksum, created, file_id, status, storage class, updated, version_id, ext, links

top level links in files vs not in record
total_bytes in record not in files
so the two serializations should definitely be harmonized.......

probably what's needed is to:

Fix the service schema serialization of RDMFileRecordServiceConfig, RDMMediaFileRecordServiceConfig, RDMFileDraftServiceConfig and RDMMediaFileDraftServiceConfig......e.g. I don't think bucket, file_id, etc should be available there....also you might run into issues that when file metadata comes from the index you dont' have all the info available
Then make the service serialization us the file services link injection templates somehow......

PR #1629, opened 2023-12-06
serializers: iiif: add links to IIIF schema

yashlamba:
Fixes https://github.com/inveniosoftware/invenio-rdm-records/issues/1291

### DIFF
--- a/invenio_rdm_records/resources/serializers/iiif/schema.py
+++ b/invenio_rdm_records/resources/serializers/iiif/schema.py
@@ -218,12 +218,13 @@ class IIIFManifestV2Schema(Schema):
     @pre_dump
     def add_iiif_links(self, data, many):
         """Add IIIF links to each files entry."""
-        for file in data["files"]["entries"].values():
-            if "links" not in file:
-                links_tpl = current_rdm_records_service.files.file_links_item_tpl(
-                    data["id"]
-                )
-                obj = SimpleNamespace(**file)
-                links = links_tpl.expand(system_identity, obj)
-                file["links"] = links
+        if "entries" in data["files"]:
+            for file in data["files"]["entries"].values():
+                if "links" not in file:
+                    links_tpl = current_rdm_records_service.files.file_links_item_tpl(
+                        data["id"]
+                    )
+                    obj = SimpleNamespace(**file)
+                    links = links_tpl.expand(system_identity, obj)
+                    file["links"] = links
         return data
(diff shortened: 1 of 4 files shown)


yashlamba:
After approval, requires merge and pin of: https://github.com/inveniosoftware/invenio-records-resources/pull/551 to pass CI

invenio_rdm_records/services/iiif/service.py
```diff
@@ -14,6 +14,7 @@
 import importlib_metadata as metadata
 from flask_iiif.api import IIIFImageAPIWrapper
 from invenio_records_resources.services import Service
+from invenio_records_resources.services.errors import PermissionDeniedError
```
lnielsen, line 17: Comment: Not needed anymore

invenio_rdm_records/resources/serializers/iiif/schema.py
```diff
@@ -209,3 +214,17 @@ def sortcanvases(self, manifest, many, **kwargs):
...
+    def add_iiif_links(self, data, many):
+        """Add IIIF links to each files entry."""
+        if "entries" in data["files"]:
+            for file in data["files"]["entries"].values():
+                if "links" not in file:
+                    links_tpl = current_rdm_records_service.files.file_links_item_tpl(
+                        data["id"]
+                    )
+                    obj = SimpleNamespace(**file)
+                    links = links_tpl.expand(system_identity, obj)
```
lnielsen, line 228: Question: What is the impact of using the ``system_identity`` here? How does this behave with restricted files - i.e. is it potentially leaking information about restricted files to unauthorised users?

invenio_rdm_records/resources/serializers/iiif/schema.py
```diff
@@ -209,3 +214,17 @@ def sortcanvases(self, manifest, many, **kwargs):
...
         # manifest["sequences"][0]["canvases"].sort(key=lambda x: x["@id"])
         return manifest
+
+    @pre_dump
+    def add_iiif_links(self, data, many):
+        """Add IIIF links to each files entry."""
+        if "entries" in data["files"]:
+            for file in data["files"]["entries"].values():
+                if "links" not in file:
+                    links_tpl = current_rdm_records_service.files.file_links_item_tpl(
```
lnielsen, line 224: Comment: It does not seem right to use a service in here - I could be wrong, but I'd like to check first.

yashlamba: Why does is not seem right? Is it because we have service logic in the presentation layer?

lnielsen: It's because link generation is the responsibility of the service layer, and the service ought to pass all needed information to the serialise. Serializer's should focus only on transformation of service result into something a client understands.

slint:
Closed in favor of https://github.com/inveniosoftware/invenio-rdm-records/pull/1692

PR #1639, opened 2023-12-19
serializers: marcxml: Add leader to schema

yashlamba:
Closes part of https://github.com/zenodo/ops/issues/367

### DIFF
--- a/invenio_rdm_records/resources/serializers/marcxml/schema.py
+++ b/invenio_rdm_records/resources/serializers/marcxml/schema.py
@@ -9,6 +9,7 @@
 
 import bleach
 from dateutil.parser import parse
+from dojson.contrib.to_marc21.fields.bdleader import to_leader
 from flask import current_app, g
 from flask_resources.serializers import BaseSerializerSchema
 from invenio_access.permissions import system_identity
@@ -56,9 +57,9 @@ class MARCXMLSchema(BaseSerializerSchema, CommonFieldsMixin):
     host_information = fields.Method("get_host_information", data_key="773  ")
     leader = fields.Method("get_leader")
 
-    def get_leader(self, o):
+    def get_leader(self, obj):
         """Return the leader information."""
-        rt = o["metadata"]["resource_type"]["id"]
+        rt = obj["metadata"]["resource_type"]["id"]
         rec_types = {
             "image": "two-dimensional_nonprojectable_graphic",
             "video": "projected_medium",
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -54,6 +55,37 @@ class MARCXMLSchema(BaseSerializerSchema, CommonFieldsMixin):
...
+            "subfield_code_count": 2,
+            "base_address_of_data": "00000",
+            "encoding_level": "unknown",
+            "descriptive_cataloging_form": "unknown",
+            "multipart_resource_record_level": "not_specified_or_not_applicable",
+            "length_of_the_length_of_field_portion": 4,
+            "length_of_the_starting_character_position_portion": 5,
+            "length_of_the_implementation_defined_portion": 0,
+            "undefined": 0,
+        }
```
ptamarit, line 87: **remark**: This part is copied from: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/records/serializers/schemas/marc21.py#L124

invenio_rdm_records/resources/serializers/marcxml/schema.py
```diff
@@ -54,6 +55,37 @@ class MARCXMLSchema(BaseSerializerSchema, CommonFieldsMixin):
...
+        return to_leader(None, None, res)
```
ptamarit, line 88: **question**: Returning `res` directly was failing, so I am explicitly calling `to_leader` to get the conversion to the expected 24 characters-long string.
However, it feels wrong and I am guessing there's a way to automatically and implicitly get this method executed?

slint: I think I know why the final MARCXML encoding is not applied: In Invenio-OAIServer, I see a usage where we first call [`to_marc21.do(...)` on the record](https://github.com/inveniosoftware/invenio-oaiserver/blob/f2cfe79138b96864f393281fbbb818e08f1e9b89/invenio_oaiserver/utils.py#L69), and afterwards call `dumps_etree`... 

I think the way you have here is correct since there seems to be an overall transition overtime from using the `dojson.contrib.to_marc21` towards directly dumping the desired values using marshmallow.

tests/resources/serializers/test_marcxml_serializer.py
```diff
@@ -74,6 +74,7 @@ def test_marcxml_serializer_minimal_record(running_app, minimal_record, parent):
     expected_data = f"""
         <?xmlversion='1.0'encoding='utf-8'?>
         <record xmlns="http://www.loc.gov/MARC21/slim">
+            <leader>00000nam##2200000uu#4500</leader>
```
slint, line 77: This also loks correct, see e.g. the output from https://zenodo-dev.cern.ch/oai2d?verb=GetRecord&metadataPrefix=marc21&identifier=oai:zenodo.org:59917

slint (approved):
LGTM! I think we can merge.

PR #1658, opened 2024-02-08
datacite: updated schema.

alejandromumo:
closes https://github.com/zenodo/ops/issues/315

### DIFF
--- a/invenio_rdm_records/config.py
+++ b/invenio_rdm_records/config.py
@@ -541,3 +541,6 @@ RDM_RECORDS_MAX_FILES_COUNT = 100
 
 RDM_RECORDS_MAX_MEDIA_FILES_COUNT = 100
 """Max amount of media files allowed to upload in the deposit form."""
+
+RDM_DATACITE_FUNDER_IDENTIFIERS_PRIORITY = ("ror", "doi", "grid", "isni", "gnd")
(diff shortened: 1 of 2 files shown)


tmorrell:
I very much disagree with this change. CrossRef funder DOIs are being depreciated and replaced with ROR https://www.crossref.org/blog/open-funder-registry-to-transition-into-research-organization-registry-ror/. Favoring DOIs in the serialized metadata is going backwards, and will have to be undone when they are fully depreciated.

alejandromumo:
> I very much disagree with this change. CrossRef funder DOIs are being depreciated and replaced with ROR https://www.crossref.org/blog/open-funder-registry-to-transition-into-research-organization-registry-ror/. Favoring DOIs in the serialized metadata is going backwards, and will have to be undone when they are fully depreciated.

Thanks a lot for the input!

Perhaps this is something that should be discussed alongside @lnielsen and @slint

alejandromumo:
> I very much disagree with this change. CrossRef funder DOIs are being depreciated and replaced with ROR https://www.crossref.org/blog/open-funder-registry-to-transition-into-research-organization-registry-ror/. Favoring DOIs in the serialized metadata is going backwards, and will have to be undone when they are fully depreciated.

We kept the same order but added a configuration variable so it can be customized by instance

slint (approved):

PR #1680, opened 2024-02-29
services: add metrics param interpreter

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/747

### DIFF
--- a/invenio_rdm_records/services/search_params.py
+++ b/invenio_rdm_records/services/search_params.py
@@ -58,3 +58,17 @@ class MyDraftsParam(ParamInterpreter):
                 "term", **{"parent.access.owned_by.user": identity.id}
             )
         return search
+
+
+class MetricsParam(ParamInterpreter):
+    """Evaluates the metrics parameter."""
+
+    def apply(self, identity, search, params):
+        value = params.pop("metrics", {})
+        for key, metric_params in value.items():
+            name = metric_params.get("name", key)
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/services/search_params.py
```diff
@@ -58,3 +58,37 @@ def is_user_authenticated():
...
+            }
+        """
+        value = params.pop("metrics", {})
+        for key, metric_params in value.items():
+            name = metric_params.get("name", key)
+            _type = metric_params.get("type", None)
+            kwargs = metric_params.get("kwargs", {})
+            field_name = kwargs.get("field", None)
+            if name and _type and field_name:
+                search.aggs.metric(name, _type, field=field_name)
```
slint, line 93:
```suggestion
            kwargs = metric_params.get("kwargs", {})
            if name and _type:
                search.aggs.metric(name, _type, **kwargs)
```

minor: no need to be specific about the metric kwargs, since there are aggs that don't take `field` only, but also many different types of config (see e.g. [matrix stats](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-matrix-stats-aggregation.html), [top hits](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-top-hits-aggregation.html), etc.).

slint (approved):
LGTM, minor/nit about how to pass the metric param

PR #1687, opened 2024-03-08
alembic: fix wrongly assigned down revision

zzacharo:
The `down_revision` should be pointed to [4a15e8671f4d](https://github.com/zzacharo/invenio-rdm-records/blob/e14199ca860f7ec19747d04f02050b8ef61a999a/invenio_rdm_records/alembic/4a15e8671f4d_create_rdm_records_tables.py#L17) which creates the draft/record metadata/files. The existing value i.e `cfcb8cb78708` causes an error to run alembic upgrade `Requested revision ff860d48fb4b overlaps with other requested revisions cfcb8cb78708`.

### DIFF
--- a/invenio_rdm_records/alembic/ff860d48fb4b_create_media_files_table.py
+++ b/invenio_rdm_records/alembic/ff860d48fb4b_create_media_files_table.py
@@ -14,7 +14,7 @@ from sqlalchemy_utils import JSONType, UUIDType
 
 # revision identifiers, used by Alembic.
 revision = "ff860d48fb4b"
-down_revision = "cfcb8cb78708"
+down_revision = "4a15e8671f4d"
 branch_labels = ()
 depends_on = None
 


zzacharo:
ping also @max-moser as you were the last one changed the down revision there I think

jrcastro2:
Here the [stackoverflow](https://stackoverflow.com/a/46711032)  answer related to this issue.
I would check the `alembic_version` table to make sure that there is only 1 value (SELECT * FROM alembic_version;)
 And then run it normally (without any other change). The revision hierarchy of the tasks looks okay to me.

zzacharo:
Closing this. We investigated with @jrcastro2  that the problem happens with instances that are in v12 beta version. The solution to fix the `alembic_version` table and run the upgrade is to run `DELETE FROM alembic_version WHERE version_num = 'cfcb8cb78708';` as the latter was a head before and it was reassigned to a down revision now, see [commit](https://github.com/inveniosoftware/invenio-rdm-records/commit/d1f55f95bde32963357eba8127acd06b82ea7be9)

PR #1696, opened 2024-03-19
Add edition to imprint  fields and initial thesis fields

tmorrell:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/310

### Description

This PR adds new fields to the imprint section

- Edition is important to characterize books

This PR also does an initial expansion of thesis fields.

- The awarding department and type of thesis are both important to identify a specific thesis 

This is a breaking change for thesis, since the thesis field was structured differently from the other publication custom fields and wasn't set up to support additional subfields. This change makes the thesis field consistent with the others, but will require current v12 users of this field to do an upgrade. I'm not sure what's needed to support that or if folks will think it's worth it.

Screenshot 2024-03-19 at 9 46 27 PM

### Checklist

- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/) (for relevant code).

**Third-party code**

**Reminder**

### DIFF
--- a/invenio_rdm_records/contrib/imprint/custom_fields.py
+++ b/invenio_rdm_records/contrib/imprint/custom_fields.py
@@ -11,6 +11,7 @@ Implements the following fields:
 - imprint.pages
 - imprint.place
 - imprint.title
+- imprint.edition
 """
 
 from idutils import is_isbn
@@ -32,11 +33,12 @@ class ImprintCF(BaseCF):
                 "isbn": SanitizedUnicode(
                     validate=is_isbn,
                     error_messages={
-                        "validator_failed": [_("Please provide a valid ISBN.")]
+                        "validator_failed": _("Please provide a valid ISBN.")
                     },
                 ),
                 "pages": SanitizedUnicode(),
                 "place": SanitizedUnicode(),
+                "edition": SanitizedUnicode(),
             }
         )
 
@@ -53,6 +55,7 @@ class ImprintCF(BaseCF):
                 "isbn": {"type": "keyword"},
                 "pages": {"type": "keyword"},
                 "place": {"type": "keyword"},
+                "edition": {"type": "keyword"},
             },
         }
 
@@ -95,6 +98,11 @@ IMPRINT_CUSTOM_FIELDS_UI = {
                     "label": _("Pagination"),
                     "placeholder": _("e.g. 15-23 or 158"),
                     "description": "",
+                },
+                "edition": {
+                    "label": _("Edition"),
+                    "placeholder": _("e.g. 3"),
+                    "description": _("The edition of the book"),
                 },
                 "icon": "book",
             },
--- a/invenio_rdm_records/contrib/imprint/processors.py
+++ b/invenio_rdm_records/contrib/imprint/processors.py
@@ -27,6 +27,7 @@ class ImprintCSLDumper(DumperMixin):
         place = imprint_data.get("place")
         pages = imprint_data.get("pages")
         isbn = imprint_data.get("isbn")
+        edition = imprint_data.get("edition")
 
         if title:
             data["container_title"] = title
@@ -36,6 +37,8 @@ class ImprintCSLDumper(DumperMixin):
             data["publisher_place"] = place
         if isbn:
             data["ISBN"] = isbn
+        if edition:
+            data["edition"] = edition
 
         return data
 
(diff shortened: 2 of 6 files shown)


lnielsen:
@tmorrell Can you bring it up in the next telecon?  In principle I don't see issues with adding the fields. It might have some impact on the serializers to use the new fields (e.g. book edition in citation formats) but better let people agree on first if they want the fields.

invenio_rdm_records/contrib/imprint/custom_fields.py
```diff
@@ -53,13 +55,14 @@ def mapping(self):
...
                 "place": {"type": "keyword"},
+                "edition": {"type": "keyword"},
             },
         }
 
 
 IMPRINT_NAMESPACE = {
     # Imprint
-    "imprint": None,
+    "imprint": "",
```
kpsherva, line 65: was the change to mutable value done on purpose?

tmorrell: Good catch. Nope, just my copy of the code was older. I've fixed it.

ntarocco:
@tmorrell thank you for this change. As briefly mentioned during our last discussion, we will also need to have the following fields:
- `Date of submission to the university`
- `Thesis Defence Date`

Do you think these are fields that can be added to the default implementation, as many might need the same? Or they are too specific and we should go for new custom fields instead?

tmorrell:
We will definitely need `Thesis Defense Date` soon, but I'm not sure how best to implement it since I don't think there is custom field dates js widget. We basically need https://github.com/inveniosoftware/invenio-rdm-records/blob/12576835186dc837df7cdf20d905452eeab8291e/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/PublicationDateField/PublicationDateField.js, but I'm not sure how to get that working within custom fields.

So if you've got some suggestions how to implement it, I think they would be good additions to the default.

egabancho:
I am also looking into this. Should the dates be added to the `thesis:thesis` custom field, or would it be enough to use the current `dates` field and add the two options to the vocabulary on our instances?

I guess from a UX point of view, it might be better to have all the thesis information together ... curious to know what you think ☺️

ntarocco:
> I am also looking into this. Should the dates be added to the `thesis:thesis` custom field, or would it be enough to use the current `dates` field and add the two options to the vocabulary on our instances?
> 
> I guess from a UX point of view, it might be better to have all the thesis information together ... curious to know what you think ☺️

We will also work on this in the iteration at CERN, and we also have other fields to migrate (@egabancho you might have memories https://cds.cern.ch/collection/CERN%20Theses :)).
We did not yet reach that point. To me, they should be all together when they are specific to thesis.

As a general thought, we will have to decide if we extend the data model, or have dedicated custom fields, maybe as a "Thesis pack", similar to [Publishing/Journal fields](https://github.com/zenodo/zenodo-rdm/blob/master/site/zenodo_rdm/custom_fields/publishing.py#L15).
It might be a good idea to have these packs in `InvenioRDM` and optionally add them, WDTY?

tmorrell:
Yup, I'll propose a breakout for the next telecon to discuss the thesis fields. I'd lean toward having a specific thesis section....the only issue is if Zenodo/others are alright moving the v12 "thesis:university" into a dedicated thesis section. I think it's also be the first time RDM depreciates a metadata field.

tmorrell:
This was presented at the telecon, and there was support for depreciating thesis:university and replacing it with thesis:thesis.

One suggestion was to support multiple universities. Having this field be a VocabularyCF would make the field work better and meet this use case, pulling from the existing affiliations vocabulary. Having department and thesis type also be VocabularyCF might also be good, and I'll try to implement that in an improved PR.

Other metadata fields that were discussed were:
- Dates (Approval, Defense) - Similar to existing date field
- Advisor and Committee - Similar to existing contributor field

It's not clear whether these would be better in the thesis block or whether we should reuse the existing fields with new types. It's also not clear how we could make these values mandatory if we use the existing fields, but work on per-community validation could be useful.

Other fields such as major, minor, approved for public release, and retracted could be added by individual sites as custom fields.

There was general discussion of having a state field, but it wasn't clear whether this was specifically tied to theses or should be more general. There was a concern raised about not mixing system and bibliographic metadata fields.

invenio_rdm_records/contrib/imprint/custom_fields.py
```diff
@@ -96,6 +99,11 @@ def mapping(self):
                     "placeholder": _("e.g. 15-23 or 158"),
                     "description": "",
                 },
+                "edition": {
+                    "label": _("Edition"),
+                    "placeholder": "3",
```
jrcastro2, line 104:
```suggestion
                    "placeholder": _("e.g. 3"),
```

invenio_rdm_records/contrib/imprint/processors.py
jrcastro2, line 1: We think it would be nice if we would also add the field for the MarcXML serializer, in the `b` subfield, reference: https://www.loc.gov/marc/bibliographic/bd773.html

WDYT?

jrcastro2 (approved):
Peer reviewed with @0einstein0, LGTU!

invenio_rdm_records/contrib/imprint/custom_fields.py
```diff
@@ -96,6 +99,11 @@ def mapping(self):
...
+                    "description": "The edition of the book",
```
zzacharo, line 105:
```suggestion
                    "description": _("The edition of the book"),
```

invenio_rdm_records/contrib/imprint/processors.py
tmorrell: I've added it to `b` and re-organized some of the other fields to make more sense. Some of the metadata should probably go outside the 773 field, but that's for another PR.

@jrcastro2 could you take a quick look and then this should be good to merge.

tmorrell:
I've simplified this to just the edition field. We can tackle the thesis fields together https://github.com/inveniosoftware/invenio-rdm-records/issues/1978, hopefully after discussions at the meeting.

jrcastro2 (approved):
LGTM! 🚀

invenio_rdm_records/contrib/imprint/processors.py
jrcastro2: Looks good, thanks for this!

PR #1700, opened 2024-03-25
results: added support for drafts

alejandromumo:
closes https://github.com/zenodo/ops/issues/34

* Added support for drafts in the results list


invenio_rdm_records/services/results.py
```diff
@@ -69,3 +69,32 @@ def get_value_service(self, value):
     def pick(self, identity, resolved_rec):
         """Pick fields defined in the entity resolver."""
         return resolved_rec
+
+class RDMRecordList(RecordList):
```
alejandromumo, line 73: The issue was that drafts are searched through the records service, which in turn uses the default `RecordList` class to project the search results. 

The default implementation uses `self._service.record_cls` unconditionally, which in turn projects the result always to a record. This is not correct if the service shares the search for records and drafts. For instance, an unpublished draft always had the attribute `is_draft=False` because the attribute `is_draft` of a `Record` is always set to `False`.

invenio_rdm_records/services/results.py
```diff
@@ -69,3 +72,33 @@ def get_value_service(self, value):
...
+    """Record list with custom fields."""
+
+    @property
+    def hits(self):
+        """Iterator over the hits."""
+        for hit in self._results:
+            # Load dump
+            record_dict = hit.to_dict()
+            if record_dict["is_published"]:
```
slint, line 86: minor: wouldn't it be better to rely on the [`is_draft` property](https://github.com/inveniosoftware/invenio-drafts-resources/blob/116f25479db770312ee9672104d89a587c497cc1/invenio_drafts_resources/records/api.py#L63)? Though I think this should be a `ConstantField` systemfield so that it makes it to the indexed document.

alejandromumo: I think that's what caused the issue in the first place. the "search" method can be used for drafts or records, however, the method lives inside the record's service. So using `record.is_draft` always returns `False`.

utnapischtim:
@slint could we reactivate the discussion here please

invenio_rdm_records/services/results.py
slint: (@alejandromumo not for taking any action now, just commenting so that others might be able to move this forward)

I meant something like:

```suggestion
            if record_dict["is_draft"]:
```

You're right that accessing the `record_cls.is_draft` would always return `False`. But we should make sure also that whatever is indexed in the drafts index should have `is_draft: true` (I hope it does, otherwise it would be very confusing 😅)

utnapischtim: on my tests there exists no `is_draft` in `record_dict`

slint (approved):

PR #1721, opened 2024-04-09
vocabularies: update existing licenses

SotosTsepe:
### Description

- Change delimiter of CSV to comma
- Update license URLs:
Use base URLs `creativecommons.org` and `opensource.org` wherever possible with `spdx.dev` as fallback.

### Tests
Tested with invenio-app-rdm `v12.0.0b1`

### DIFF
--- a/invenio_rdm_records/fixtures/vocabularies.py
+++ b/invenio_rdm_records/fixtures/vocabularies.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2021-2022 CERN.
 # Copyright (C) 2021-2022 Northwestern University.
+# Copyright (C) 2024 TU Wien.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/fixtures/vocabularies.py
```diff
@@ -70,7 +71,7 @@ def map_row(self, header, row):
     def __iter__(self):
         """Iterate over records."""
         with open(self._data_file) as fp:
-            reader = csv.reader(fp, delimiter=";", quotechar='"')
+            reader = csv.reader(fp, delimiter=",", quotechar='"')
```
slint, line 74: minor: given that we're somewhat "breaking the API" for the CSV import, we could maybe use [the `csv.Sniffer`](https://docs.python.org/3/library/csv.html#csv.Sniffer), to detect the dialect (reading 1024 chars should be enough as done in the docs example).

slint (approved):
LGTM, minor improvement to keep backwards compat.

invenio_rdm_records/fixtures/vocabularies.py
SotosTsepe: Good point, I'll add this here :+1:

slint (approved):

PR #1724, opened 2024-04-13
fix restricted dois problem

utnapischtim:
- test: update test workflows
- fix: don't mint doi's on restricted

this PR will close https://github.com/inveniosoftware/product-rdm/issues/178

still open Problem:
- ~~the DOI is shown on the landing page, because the PID exists in the database~~
- ~~the DOI widget in the frontend has to be grayed out if the access is set to restricted~~
- is it really necessary to gray it out? it complicates the code. it has to be done independent of DOI but for all other pids too, to be generic in that way. The problem is, that the DOI could be already there (entry in pidstore_pid), in the InvenioRDM instance, because the record is set to restricted in the end of the drafting process. there would be a lot of work to do to make that consistent.

ATTENTION:
- to make it clear for the user, a message should be added to the landing page too. this has to be done on invenio-app-rdm

NOTE:
- with this solution it is easy to change from `restricted` to `public` and mint the DOI with this step. No further implementation or work from the user would be necessary

UI:
after grace period of a public record ended, the restriction button is disabled + the popup message
Screenshot 2024-05-21 at 09 54 57

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/MetadataAccess.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/MetadataAccess.js
@@ -10,17 +10,45 @@ import PropTypes from "prop-types";
 import { ProtectionButtons } from "./ProtectionButtons";
 import { i18next } from "@translations/invenio_rdm_records/i18next";
 import Overridable from "react-overridable";
+import { DateTime } from "luxon";
 
 export const MetadataAccess = (props) => {
-  const { recordAccess, communityAccess } = props;
+  const {
+    recordAccess,
+    communityAccess,
+    record,
+    recordRestrictionGracePeriod,
+    allowRecordRestriction,
+  } = props;
   const publicMetadata = recordAccess === "public";
   const publicCommunity = communityAccess === "public";
+
+  const isGracePeriodOver = () => {
+    const createdDate = DateTime.fromISO(record.created);
+    const endOfGracePeriod = createdDate.plus({ days: recordRestrictionGracePeriod });
+
+    return endOfGracePeriod < DateTime.now();
+  };
+
+  const canRestrictRecord = () => {
+    if (
+      !allowRecordRestriction &&
+      record?.created &&
+      record?.access?.record !== "restricted" &&
+      record.is_published
+    ) {
+      return !isGracePeriodOver();
+    }
+    return true;
+  };
 
   return (
     <Overridable id="ReactInvenioDeposit.MetadataAccess.layout" {...props}>
       <>
         {i18next.t("Full record")}
         <ProtectionButtons
+          canRestrictRecord={canRestrictRecord()}
+          record={record}
           active={publicMetadata && publicCommunity}
           disabled={!publicCommunity}
           fieldPath="access.record"
@@ -33,4 +61,7 @@ export const MetadataAccess = (props) => {
 MetadataAccess.propTypes = {
   recordAccess: PropTypes.string.isRequired,
   communityAccess: PropTypes.string.isRequired,
+  record: PropTypes.object.isRequired,
+  recordRestrictionGracePeriod: PropTypes.object.isRequired,
+  allowRecordRestriction: PropTypes.bool.isRequired,
 };
--- a/invenio_rdm_records/services/components/pids.py
+++ b/invenio_rdm_records/services/components/pids.py
@@ -12,6 +12,7 @@
 
 from copy import copy
 
+from flask import current_app
 from invenio_drafts_resources.services.records.components import ServiceComponent
 from invenio_drafts_resources.services.records.uow import ParentRecordCommitOp
 from invenio_records_resources.services.uow import TaskOp
@@ -75,7 +76,7 @@ class PIDsComponent(ServiceComponent):
         required_schemes = set(self.service.config.pids_required)
 
         # Validate the draft PIDs
-        self.service.pids.pid_manager.validate(draft_pids, record, raise_errors=True)
+        self.service.pids.pid_manager.validate(draft_pids, draft, raise_errors=True)
 
         # Detect which PIDs on a published record that has been changed.
         #
@@ -89,6 +90,8 @@ class PIDsComponent(ServiceComponent):
             if record_id != draft_id:
                 changed_pids[scheme] = record_pids[scheme]
 
+        self.service.pids.pid_manager.validate_restriction_level(draft)
+
         self.service.pids.pid_manager.discard_all(changed_pids)
 
         # Determine schemes which are required, but not yet created.
@@ -97,7 +100,11 @@ class PIDsComponent(ServiceComponent):
         pids = self.service.pids.pid_manager.create_all(
             draft,
             pids=draft_pids,
-            schemes=missing_required_schemes,
+            schemes=(
+                missing_required_schemes
+                if draft["access"]["record"] != "restricted"
+                else None
+            ),
         )
 
         # Reserve all created PIDs and store them on the record
--- a/invenio_rdm_records/services/pids/manager.py
+++ b/invenio_rdm_records/services/pids/manager.py
@@ -24,7 +24,7 @@ class PIDManager:
     def __init__(self, providers, required_schemes=None):
         """Constructor for RecordService."""
         self._providers = providers
-        self._required_schemes = required_schemes
+        self._required_schemes = required_schemes if required_schemes else []
 
     def _get_provider(self, scheme, provider_name=None):
         """Get a provider."""
@@ -35,6 +35,21 @@ class PIDManager:
             return providers[provider_name]
         except KeyError:
             raise ProviderNotSupportedError(provider_name, scheme)
+
+    def _get_providers(self, pids):
+        """Get all providers."""
+        schemes = set(pids.keys()) | set(self._required_schemes)
+        scheme_provider_names = [
+            # provider_name for an absent-but-required pid will be None which will
+            # in turn select the default provider below
+            (scheme, pids.get(scheme, {}).get("provider"))
+            for scheme in schemes
+        ]
+        provider_pid_dicts = [
+            (self._get_provider(scheme, provider_name), pids.get(scheme, {}))
+            for scheme, provider_name in scheme_provider_names
+        ]
+        return provider_pid_dicts
 
     def _validate_pids_schemes(self, pids):
         """Validate the pid schemes that are supported by the service.
@@ -96,17 +111,7 @@ class PIDManager:
         # Validate according to the schemes that the draft has and the schemes that
         # the draft would be given. _required_schemes are schemes that would be given
         # (if not already on the draft).
-        schemes = set(pids.keys()) | set(self._required_schemes)
-        scheme_provider_names = [
-            # provider_name for an absent-but-required pid will be None which will
-            # in turn select the default provider below
-            (scheme, pids.get(scheme, {}).get("provider"))
-            for scheme in schemes
-        ]
-        provider_pid_dicts = [
-            (self._get_provider(scheme, provider_name), pids.get(scheme, {}))
-            for scheme, provider_name in scheme_provider_names
-        ]
+        provider_pid_dicts = self._get_providers(pids)
 
         for provider, pid_dict in provider_pid_dicts:
             success, provider_errors = provider.validate(record=record, **pid_dict)
@@ -279,9 +284,26 @@ class PIDManager:
             try:
                 self.discard(
                     scheme,
-                    pid_attrs["identifier"],
-                    pid_attrs["provider"],
+                    pid_attrs.get("identifier"),
+                    pid_attrs.get("provider"),
                     soft_delete=soft_delete,
                 )
             except PIDDoesNotExistError:
                 pass  # might not have been saved to DB yet
+
+    def validate_restriction_level(self, record, **kwargs):
+        """Validates that the record has correct restriction levels to crate the PIDs."""
+        pids = record.get("pids", {})
+        provider_pid_dicts = self._get_providers(pids)
+
+        for provider, pid_dict in provider_pid_dicts:
+            provider.validate_restriction_level(
+                record, identifier=pid_dict.get("identifier")
+            )
+
+    def create_and_reserve(self, record, **kwargs):
+        """Create and reserve a PID for the given record, and update the record with the reserved PID."""
+        pids = record.get("pids", {})
+        provider_pid_dicts = self._get_providers(pids)
+        for provider, _ in provider_pid_dicts:
+            provider.create_and_reserve(record)
--- a/invenio_rdm_records/services/pids/providers/datacite.py
+++ b/invenio_rdm_records/services/pids/providers/datacite.py
@@ -2,7 +2,7 @@
 #
 # Copyright (C) 2021 CERN.
 # Copyright (C) 2023 Northwestern University.
-# Copyright (C) 2023 Graz University of Technology.
+# Copyright (C) 2023-2024 Graz University of Technology.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
@@ -26,6 +26,7 @@ from invenio_i18n import lazy_gettext as _
 from invenio_pidstore.models import PIDStatus
 
 from ....resources.serializers import DataCite43JSONSerializer
+from ....utils import ChainObject
 from .base import PIDProvider
 
 
@@ -153,6 +154,12 @@ class DataCitePIDProvider(PIDProvider):
         :param record: the record metadata for the DOI.
         :returns: `True` if is registered successfully.
         """
+        if isinstance(record, ChainObject):
+            if record._child["access"]["record"] == "restricted":
+                return False
+        elif record["access"]["record"] == "restricted":
+            return False
+
         local_success = super().register(pid)
         if not local_success:
             return False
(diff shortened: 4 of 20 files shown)


tmorrell (approved):
Great! This safely skips registering DOIs with DataCite on restricted records, and hides DOI metadata when users have changed the permissions. It also allows for DOIs to be registered when a user changes a restricted record to public or when an embargo expires (though maybe that should be tested just to be sure).

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -178,10 +185,19 @@ def update(self, pid, record, url=None, **kwargs):
...
+            if record._child["access"]["record"] == "restricted":
+                hide = True
+        elif record["access"]["record"] == "restricted":
+            hide = True
+
         try:
-            # Set metadata
-            doc = self.serializer.dump_obj(record)
-            self.client.api.update_doi(metadata=doc, doi=pid.pid_value, url=url)
+            if hide:
```
zzacharo, line 196: Shouldn't be able to update a doi even if it is is hidden? Also, how do I restore a hidden doi? Currently, if I change from restricted to public then the update method will update the metadata but not make the doi findable again, unless I miss something...

utnapischtim: yes, good catch. it should be that the hided should be put back to "findable" after a change from "restricted" to "public". i will have a look.

tmorrell: That's going to be a tricky one, since I don't think we want to run `show_doi` for every single public DOI update.

I can think of two solutions:
- I can add a `status` option to the `update_doi` function in the datacite library. Then I could pass along all the updates to DataCite in one call
- We can modify pidstore so it has a `hidden` state. Then we could check locally whether the state of the DOI needs to be changed.

Let me know what you think or if you want to chat.

utnapischtim: @zzacharo @tmorrell what do you think of the proposed solution to this problem? this does not make it necessary to keep track the state.

tmorrell: Yup, I think that should work too. Just make sure DataCite doesn't complain about having the event as "publish" for a published doi. Looks like the tests need to updated too

zzacharo: If DataCite indeed doesn't complain about the "publish" events that should be good. Otherwise, we need some kind of state, maybe the caller passes the old record value, to call the publish/show doi only on the change of permission.

utnapischtim: How i tested it: DataCite doesn't complain.

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -178,10 +185,20 @@ def update(self, pid, record, url=None, **kwargs):
...
+                self.client.api.hide_doi(doi=pid.pid_value)
+            else:
+                doc = self.serializer.dump_obj(record)
```
ntarocco, line 199: We reviewed this with @zzacharo and we think that when you change from restricted to public, here we are not calling `.register()`, so the local PID will still have the status `PIDStatus.NEW` or `PIDStatus.RESERVED`.
We are missing a test that will test the change from restricted to public and viceversa.

invenio_rdm_records/services/pids/providers/datacite.py
```diff
@@ -178,10 +185,20 @@ def update(self, pid, record, url=None, **kwargs):
...
+        elif record["access"]["record"] == "restricted":
+            hide = True
+
         try:
-            # Set metadata
-            doc = self.serializer.dump_obj(record)
-            self.client.api.update_doi(metadata=doc, doi=pid.pid_value, url=url)
+            if hide:
+                self.client.api.hide_doi(doi=pid.pid_value)
+            else:
```
ntarocco, line 198: Another way could be to delegate the state to DataCite: we could first read/get from DataCite, check if it is hidden or not, and therefore add the `event: publish` or not. We are in a celery task so it could work.
However, we need to take into account that DataCite might not respond, and what to do in that case.
At the same time, I think that it would make sense to have an `update` method that receive the old and the new version of the record, so the implementation could use the diff to make choices. It feels like a much bigger change.

utnapischtim: which concerns do you have with having always the `event: publish` there?

invenio_rdm_records/services/pids/providers/datacite.py
utnapischtim: the pid register method is already called, on the publish. that is independent of the way datacite is handling the restriction access setting.

tmorrell: Yes, that call won't happen until the record is public. The "doi" pid doesn't need to be registered until the DOI is actually sent to DataCite. So every time there are changes to the record it will try to register the DOI, but it will only get registered if the record is public. A test is probably a good idea though.

utnapischtim: > What about this [call](https://github.com/inveniosoftware/invenio-rdm-records/pull/1724/files#diff-804dbcc4c6b21d366befbb2b80d5f987c616878f680562143f8999bef3e7a0e8R163). That will not be called if the record is restricted. Isn't this suppose to register the "doi" PID?

what do you mean by "register" the doi PID? 

as Tom said, the doi should only be set to R in `pidstore_pid` if it has been registered at datacite. and that isn't the case if the record is set to `restricted`. this means that the call in the [register_or_update](https://github.com/inveniosoftware/invenio-rdm-records/blob/f1d5dc0da0c71d29fd21696b7d25af79dd5d14b3/invenio_rdm_records/services/pids/service.py#L210) task will register the pid DOI at datacite after the change from `restricted` to `public`

> We reviewed this with @zzacharo and we think that when you change from restricted to public, here we are not calling `.register()`, so the local PID will still have the status `PIDStatus.NEW` or `PIDStatus.RESERVED`. We are missing a test that will test the change from restricted to public and viceversa.

is [this](https://github.com/inveniosoftware/invenio-rdm-records/pull/1724/files#diff-4d69e1b21e25061eb03ca154d0f090414925e0529fede5b88a2b6044d8d9e913R88) test what you had in mind?

invenio_rdm_records/services/pids/providers/datacite.py
utnapischtim: in my opinion it would complicate the whole process if it has first to be checked on DataCite if the record is hidden. The state management will be more complex and we need at least one request to get the information. Yes one request is not that much but scaled on multiple InvenioRDM instances with multiple records DataCite has to handle a lot more requests.

invenio_rdm_records/services/pids/providers/datacite.py
tmorrell: Maybe add  `minimal_record["access"]["record"] = "public"` at the end and then `assert pid.status == PIDStatus.REGISTERED` ?

utnapischtim: like that: https://github.com/inveniosoftware/invenio-rdm-records/pull/1724/commits/fe042f84c78ecda19197a94dab05f17b7e99d981

tmorrell: Looks good to me

tmorrell: Just silly copyright conflict to fix https://github.com/inveniosoftware/invenio-rdm-records/pull/1724/conflicts

utnapischtim: so rebased and cleaned, lets see tomorrow for the rest :)

zzacharo: @utnapischtim thanks for the test, it seems good! I understand that the `PIDComponent` will register the doi PID, but I think before the datacite client when the method `client.register()` was called, it ensured that the local doi PID i.e PersitentIdentifier model was registered before publishing it to datacite. Now, that a "register" action can happen in the update and if the client is used directly e.g an admin is using the client programmatically then the doi will not be registered just by calling `client.update()` to a public record. I am thinking that we need to call 

```python
 local_success = super().register(pid)
 if not local_success:
     return False
```

in the case of the `event=publish`.

utnapischtim: i am not sure if i understand your scenario:

The `register` action in `update` can only happen if the `PersistentIdentifier` has been called in `dataciteprovicer.register` method. the register method will always be called first because of the check in the `register_or_update` method [see](https://github.com/inveniosoftware/invenio-rdm-records/blob/f1d5dc0da0c71d29fd21696b7d25af79dd5d14b3/invenio_rdm_records/services/pids/service.py#L210). since the DOI PID was not registered locally in the database if it was restricted the `pid.is_registered` will be false and therefore the `pid_manager.register` method will be called and therefore the `DatacitePIDProvider.register` method

do you mean that before of the `event:publish` the `DatacitePIDProvider.update` method would have failed because it the doi has not been registered at Datacite?

zzacharo: I mean using the `DatacitProvider` directly without going through the service :) It is not the usual/recommended way but as an admin you might do that and I would rather have an extra check to register the local PID doi before going to datacite. The normal flow via services is working correctly no problem with that!

utnapischtim: then yes you are right, we have to add a if to register the pid if it has not been registered at that point. which makes the whole code really messy

zzacharo: the `pid.register()` method already checks if a pid is registered otherwise do nothing.... Maybe I am overthinking it but I was just triggered by the fact that such a check was present in the register method..

utnapischtim: to be honest, in my point of view it is wrong to register the DOI pid in `pidstore_pid` within the provider class. the provider should only take care what's associated with the provider. so the `DataCitePIDProvider.register` should not know about the internal behavior and should only be called if it has been registered internally. But that is another discussion.

yes `pid.register` takes care of knowing if it has already bin registered but this is not called in `DataCitePIDProvider.update` as you pointed out. So if a record is set to `RESTRICTED` the DOI PID will not be set to R in `pidstore_pid`. If someone is using the `DataCitePIDProvider.update` method then directly the DOI will be registered at DataCite and will remain `K` in `pidstore_pid`. if the record is then set to `public` the code will mostly work, because the `public_doi` method of the datacite rest client does more or less the same as the update does, it uses the `event: publish` to register the doi.

the thing is, is it a problem if the doi is in state `K` in `pidstore_pid` for this scenario?

tmorrell: I think the /doi link won't work if the identifier is in state `K`. I guess we could add a check to pidstore and then register if needed, though it's a lot of extra calls to take care of someone who's not really following the normal flow.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/MetadataAccess.js
```diff
@@ -12,15 +12,41 @@ import { i18next } from "@translations/invenio_rdm_records/i18next";
...
+    recordRestrictionGracePeriod,
+    allowRecordRestriction,
+  } = props;
   const publicMetadata = recordAccess === "public";
   const publicCommunity = communityAccess === "public";
 
+  const checkGracePeriod = () => {
+    const created = record?.created;
+    if (!allowRecordRestriction && created) {
+      const createdDate = new Date(created);
```
kpsherva, line 28: if I am not mistaken for dates we use `luxon` library, also helps with readability of date calculations

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/MetadataAccess.js
```diff
@@ -12,15 +12,41 @@ import { i18next } from "@translations/invenio_rdm_records/i18next";
...
+      createdDate.setDate(createdDate.getDate() + recordRestrictionGracePeriod);
+
+      return createdDate.getTime() <= new Date().getTime();
+    }
+
+    return false;
+  };
+
+  const isButtonDisabled = () => {
+    return checkGracePeriod() && record?.access?.record !== "restricted";
```
kpsherva, line 40: variable already available, no need to recompare with restricted
```suggestion
    return checkGracePeriod() && publicMetadata;
```

kpsherva, line 40: this will avoid calling the grace period function when is it not needed. I moved those conditions from checkGracePeriod
```suggestion
  const isButtonDisabled = () => {
    if (!allowRecordRestriction && record && publicMetadata){
        return checkGracePeriod();
    }
```

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/MetadataAccess.js
```diff
@@ -12,15 +12,41 @@ import { i18next } from "@translations/invenio_rdm_records/i18next";
...
+    record,
+    recordRestrictionGracePeriod,
+    allowRecordRestriction,
+  } = props;
   const publicMetadata = recordAccess === "public";
   const publicCommunity = communityAccess === "public";
 
+  const checkGracePeriod = () => {
+    const created = record?.created;
+    if (!allowRecordRestriction && created) {
```
kpsherva, line 27: if https://github.com/inveniosoftware/invenio-rdm-records/pull/1724/files#r1601326255 is applied, this if can be removed

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/MetadataAccess.js
```diff
@@ -12,15 +12,41 @@ import { i18next } from "@translations/invenio_rdm_records/i18next";
...
+    communityAccess,
+    record,
+    recordRestrictionGracePeriod,
+    allowRecordRestriction,
+  } = props;
   const publicMetadata = recordAccess === "public";
   const publicCommunity = communityAccess === "public";
 
+  const checkGracePeriod = () => {
+    const created = record?.created;
```
kpsherva, line 26: if https://github.com/inveniosoftware/invenio-rdm-records/pull/1724/files#r1601326255 is applied, additional check if record exists is not needed

invenio_rdm_records/services/components/pids.py
```diff
@@ -89,6 +90,29 @@ def publish(self, identity, draft=None, record=None):
...
+                    if draft_doi.get("provider") == "external":
+                        continue
+                    if (
+                        draft_doi == {}
+                        or self.service.pids.pid_manager.read(
+                            "doi",
+                            draft_doi.get("identifier"),
+                            draft_doi.get("provider"),
+                        ).status
+                        == "N"
```
kpsherva, line 111:
```suggestion
                        == PIDStatus.NEW
```

invenio_rdm_records/services/components/pids.py
```diff
@@ -89,6 +90,29 @@ def publish(self, identity, draft=None, record=None):
             if record_id != draft_id:
                 changed_pids[scheme] = record_pids[scheme]
 
+        if current_app.config["DATACITE_ENABLED"]:
```
kpsherva, line 93: I remember vaguely a conversation about specilised pid providers, shouldn't this be implemented on the datacite provider instead?

invenio_rdm_records/services/components/pids.py
```diff
@@ -89,6 +90,29 @@ def publish(self, identity, draft=None, record=None):
...
+            if "doi" not in draft_schemes:
+                draft_schemes.add("doi")
+            if "doi" not in draft_pids:
+                draft_pids["doi"] = {}
+
+            for scheme in draft_schemes:
```
kpsherva, line 99: the amount of nested condititional statements makes it hard to untangle. Is it possible to simplify it?

invenio_rdm_records/services/pids/manager.py
```diff
@@ -181,7 +181,7 @@ def create_all(self, draft, pids=None, schemes=None):
             result[scheme] = self.create(
                 draft,
                 scheme,
-                pid_attrs["identifier"],
+                pid_attrs.get("identifier", None),
```
kpsherva, line 184: are we sure that direct key access wasn't there on purpose to fail early?
minor: adding None as get function argument is redundant since it is default

invenio_rdm_records/services/services.py
```diff
@@ -132,8 +136,37 @@ def lift_embargo(self, identity, _id, uow=None):
             "lift_embargo", identity, draft=draft, record=record, uow=uow
         )
 
-        # Commit and reindex record
-        uow.register(RecordCommitOp(record, indexer=self.indexer))
+        if current_app.config["DATACITE_ENABLED"] and "doi" not in record.pids:
```
kpsherva, line 139: shouldn't this be moved to a separate component dedicated to datacite?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/MetadataAccess.js
anikachurilova: `recordAccess` represents the current value of the button of the form. Here i am checking the access of the record itself

invenio_rdm_records/services/pids/manager.py
```diff
@@ -21,7 +21,7 @@
 class PIDManager:
     """RDM PIDs Manager."""
 
-    def __init__(self, providers, required_schemes=None):
+    def __init__(self, providers, required_schemes=[]):
```
kpsherva, line 24: is there any reason why we changed this to mutable argument? it is considered not to be a best practice

tests/conftest.py
```diff
@@ -1711,7 +1738,7 @@ def embargoed_record(running_app, minimal_record, superuser_identity):
 
     # Add embargo to record
     with mock.patch("arrow.utcnow") as mock_arrow:
-        minimal_record["access"]["files"] = "restricted"
+        minimal_record["access"]["record"] = "restricted"
```
kpsherva, line 1741: should we also recreate the test for files (can we have embargoed files)

PR #1727, opened 2024-04-16
iiif: added proxy to image server

alejandromumo:
* Added IIIFProxy interface and IIPServerProxy implementation
* Added proxy_if_enabled decorator to proxy requests to the image server if enabled

closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1728


invenio_rdm_records/resources/iiif.py
```diff
@@ -0,0 +1,124 @@
...
+        return False
+
+    def __call__(self):
+        """Proxy request to IIIF server if the endpoint is configured."""
+        if self.should_proxy():
+            return self.proxy_request()
+        return None
+
+
+class IIPServerProxy(IIIFProxy):
```
alejandromumo, line 78: For this implementation to be useful for broader use, we need to make `server_url` and `handle_url_rewrite` configurabl

invenio_rdm_records/resources/iiif.py
```diff
@@ -0,0 +1,124 @@
...
+        """List of routes that should be proxied."""
+        return ["iiif.image_api", "iiif.info"]
+
+    def proxy_request(self):
+        """Proxy request to IIIF server."""
+        url = self.handle_url_rewrite()
+        res = requests.request(
+            request.method, url, headers=request.headers, stream=True
+        )
+        return Response(
```
alejandromumo, line 98: We open a new connection to proxy the request and stream the result back in chunks of 10 KB. Do we want to test this number of at least make it configurable?

invenio_rdm_records/resources/iiif.py
```diff
@@ -0,0 +1,124 @@
...
+        For IIPServer, we need to match the folder structure where the images are stored.
+        I.e. /<recid>/<filename>
+        """
+        from urllib.parse import urljoin
+
+        uuid = resource_requestctx.view_args["uuid"]
+        recid = uuid.split(":")[1]
+        file_name = uuid.split(":")[-1]
+        path = request.path
+        path = path.replace("/iiif/", "").replace(uuid, f"{recid}/{file_name}")
```
alejandromumo, line 116: This assumes that the folder structure for the IIPServer storage is `<recid>/<file_name>`

ptamarit (approved):

invenio_rdm_records/resources/iiif.py
```diff
@@ -0,0 +1,379 @@
...
+        uuid = resource_requestctx.view_args["uuid"]
+        key = resource_requestctx.view_args["file_name"]
+        file_ = self.service.get_file(uuid=uuid, identity=g.identity, key=key)
+        return file_.to_dict(), 200
+
+    @cross_origin(origin="*", methods=["GET"])
+    @with_iiif_content_negotiation(IIIFInfoV2JSONSerializer)
+    @iiif_request_view_args
+    @response_handler()
+    @proxy_pass.__func__
```
alejandromumo, line 201: Out of curiosity, why is the `__func__` part needed?

slint: Because it's a `@staticmethod`, so it's "wrapped". Keeping it without the `@statimethod` looke more problematic/dangerous in case someone called this on an instance of the class.

Issue #1672, opened 2024-02-20
api: add `access/groups` endpoint


anikachurilova:
EPIC: https://github.com/CERNDocumentServer/cds-rdm/issues/109

Should be same as here https://github.com/inveniosoftware/invenio-rdm-records/issues/1671

PR #1729, opened 2024-04-16
grant access: implement groups grant record access

anikachurilova:
* add new endpoint for groups grant record access
* rename "role" to "group" to be consistent with groups in community memberships
* manage groups enabled feature flag by the permission
* closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1672


invenio_rdm_records/records/systemfields/access/grants.py
```diff
@@ -74,7 +74,7 @@ def subject(self):
             self._subject = self.resolve_subject()
 
             if self._subject is not None:
-                if self._subject_type == "role":
+                if self._subject_type == "group":
```
anikachurilova, line 77: Changes were made for consistency with community groups memebrs concept

invenio_rdm_records/services/access/service.py
```diff
@@ -385,6 +387,14 @@ def bulk_create_grants(self, identity, id_, data, expand=False, uow=None):
             raise GrantExistsError()
 
         for grant in grants:
+            # checks if groups are enabled in the instance
+            if grant["subject"]["type"] == RecordAccessService.group_subject_type:
```
anikachurilova, line 391: Groups functionality is depending on `USERS_RESOURCES_GROUPS_ENABLED` feature flag.
`can_groups_enabled` permission is managing it. Added checks for each service method to not allow users playing with groups through API if feature is disabled.

invenio_rdm_records/services/access/service.py
```diff
@@ -506,97 +533,25 @@ def read_all_grants(self, identity, id_, expand=False):
...
         # Fetching
         return self.grants_result_list(
             service=self,
             identity=identity,
-            results=user_grants,
-            expand=expand,
-        )
-
-    @unit_of_work()
-    def update_grant_by_subject(
```
anikachurilova, line 538: The victim of rebase of previous PRs (methods `update_grant_by_subject` and `read_all_grants_by_subject` were duplicated)

invenio_rdm_records/services/permissions.py
```diff
@@ -286,6 +287,14 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
     # correct permissions based on which the field will be exposed only to moderators
     can_moderate = [Disable()]
 
+    # Used to manage groups enabled feature flag
+    can_groups_enabled = [
+        AnyUser(),
```
anikachurilova, line 292: Any user in case if groups are enabled

invenio_rdm_records/records/systemfields/access/grants.py
```diff
@@ -130,7 +130,7 @@ def subject_id(self):
         if self._subject_id:
             return self._subject_id
 
-        if self.subject_type == "role":
+        if self.subject_type == "group":
```
kpsherva, line 133: minor: maybe the rename should be also discussed with zenodo team, not sure if there are consequences of changing the name here

invenio_rdm_records/resources/config.py
```diff
@@ -371,6 +371,13 @@ class RDMDraftMediaFilesResourceConfig(FileResourceConfig, ConfiguratorMixin):
     ),
 }
 
+group_access_error_handlers = {
+    **deepcopy(RecordResourceConfig.error_handlers),
+    LookupError: create_error_handler(
+        HTTPJSONException(code=404, description="No grant found by given group id.")
```
ntarocco, line 377: In general, are we making the error msg translatable?

invenio_rdm_records/services/access/service.py
```diff
@@ -385,6 +387,14 @@ def bulk_create_grants(self, identity, id_, data, expand=False, uow=None):
             raise GrantExistsError()
 
         for grant in grants:
+            # checks if groups are enabled in the instance
```
ntarocco, line 390: as we briefly chat, to improve performance, can we check it only the first time?
```
groups_enabled_checked = False

...

     if grant["subject"]["type"] == RecordAccessService.group_subject_type and not groups_enabled_checked:
         # check permissions once to improve performance
         self.require_permission(...
         groups_enabled_checked = True
```

invenio_rdm_records/services/access/service.py
```diff
@@ -506,97 +533,25 @@ def read_all_grants(self, identity, id_, expand=False):
...
-        # Permissions
-        self.require_permission(identity, "manage", record=record)
-
-        user_grants = []
-        for grant in parent.access.grants:
-            if grant.subject_type == subject_type:
-                user_grants.append(grant)
+        for grant in existing_grants:
+            # removes group grants if groups are not enabled in the instance
+            if grant.subject_type == this.group_subject_type:
```
ntarocco, line 540: same here about perf

invenio_rdm_records/services/access/service.py
```diff
@@ -993,6 +983,14 @@ def delete_grant_by_subject(
         self, identity, id_, subject_id, subject_type, uow=None
     ):
         """Delete an access grant for a record by subject."""
+        # checks if groups are enabled in the instance
+        if subject_type == RecordAccessService.group_subject_type:
+            self.require_permission(
```
ntarocco, line 988: Honestly, more I see this, more I would not do it checking permissions.
The `groups enabled` feature is a global config flag, that does not depend on the current `identity`.
A simple `config["...ENABLED"]` check will be computationally much faster.

invenio_rdm_records/services/access/service.py
```diff
@@ -882,6 +848,14 @@ def read_grant_by_subject(
         self, identity, id_, subject_id, subject_type, expand=False
     ):
         """Read a specific access grant of a record by subject."""
+        # checks if groups are enabled in the instance
+        if subject_type == RecordAccessService.group_subject_type:
```
ntarocco, line 852: I would put a decorator for these cases, so we remove "noise" from the code.

invenio_rdm_records/resources/config.py
anikachurilova: Yes, we should. For example this one is not only for the API, it could also be displayed in the UI. Thanks for noticing!

invenio_rdm_records/services/access/service.py
anikachurilova: In the current implementation, on the first occurrence of the group grant if groups are disabled, the application fails with `PermissionDenied` error and doesn't check the rest of the grants

invenio_rdm_records/services/access/service.py
```diff
@@ -506,97 +531,21 @@ def read_all_grants(self, identity, id_, expand=False):
...
-        for grant in parent.access.grants:
-            if grant.subject_type == subject_type:
-                user_grants.append(grant)
+        for grant in existing_grants:
+            # removes group grants if groups are not enabled in the instance
+            if (
+                not current_app.config.get("USERS_RESOURCES_GROUPS_ENABLED", False)
+                and grant.subject_type == RecordAccessService.group_subject_type
+            ):
+                existing_grants.remove(grant)
```
anikachurilova, line 542: Note, for `read_all` we don't fail with 403, instead we return only user grants, even if group grants are present.

invenio_rdm_records/services/access/service.py
kpsherva: shouldn't these if conditionals just be moved to `self.require_permission` instead? we would have less occurrences, and it would be checked any time permissions are checked (no need to remember to add it in case there is a new method in the service)

invenio_rdm_records/services/access/service.py
kpsherva: could you add the note in the code as a comment?

invenio_rdm_records/services/config.py
```diff
@@ -119,6 +119,11 @@ def archive_download_enabled(record, ctx):
     return current_app.config["RDM_ARCHIVE_DOWNLOAD_ENABLED"]
 
 
+def groups_enabled(record, ctx):
```
kpsherva, line 122:
```suggestion
def _groups_enabled(record, ctx):
```
we could make it internal for this file, otherwise there is a potential to mistake it for the decorator implemented below

invenio_rdm_records/services/access/service.py
kpsherva: on another note, just for me to understand: why do we need the conditional statements here if we have the permission generator GroupsEnabled?

anikachurilova: In the beginning, I did it via using permissions, but then chatted with Nico and decided that checking it with a simple `if config` would be much faster. Moreover, `require_permission` is checking the permission of the current **identity**, but whether groups are enabled or not does not depend on any identity, it depends only on the feature flag

ntarocco (approved):

PR #1734, opened 2024-04-22
Fix duplicate drafts on user dashboard after direct publish

slint:
- **tests: check user search results after direct publish to community**
- **review: fix draft indexing operations order**


ntarocco (approved):
Nice catch!

slint:
Just a note, this fix has the caveat that we'll index the draft and then delete it right after, which is a waste. 

What I had in mind though is that this is not the first place where we might be doing this (on purpose or accidentally) and we discussed that the right way to fix would be to have some sort of "roll-up" logic in the Unit of Work which e.g. knows the entire set of operations and might be smart enough to "skip index/update before delete", "avoid sending duplicate notification emails", etc.

Issue #1652, opened 2024-01-31
Parent PID configuration for non-DataCite setup


tmorrell:
Package version (if known): v12d

## Describe the bug
The current configuration of parent PIDS assumes that DataCite is enabled https://github.com/inveniosoftware/invenio-rdm-records/blob/a9eab71bf8b55efc90d2ab9268bbc45a4e2eab55/invenio_rdm_records/config.py#L377, but the default is that DataCite is not enabled https://github.com/inveniosoftware/invenio-rdm-records/blob/a9eab71bf8b55efc90d2ab9268bbc45a4e2eab55/invenio_rdm_records/config.py#L394. This will result in issues for a standard install.

- Should DataCite be a requirement?
- Do we need a config to turn off parent pids? (This would be useful for us)
- Can a different pid be used as the default?

utnapischtim:
Note: the parent_doi is used in [invenio_rdm_records/service/config.py](https://github.com/inveniosoftware/invenio-rdm-records/blob/bbdd2251ad275274a6282f47c075c51893c2b035/invenio_rdm_records/services/config.py#L362) i think this should be a ConditionalLink as in [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/bbdd2251ad275274a6282f47c075c51893c2b035/invenio_rdm_records/services/config.py#L324)

utnapischtim:
related to discussion around DOI's of restricted records: https://github.com/inveniosoftware/product-rdm/issues/178

utnapischtim:
related to https://github.com/inveniosoftware/product-rdm/issues/194

PR #1740, opened 2024-05-02
parent-doi: add configuration

utnapischtim:
should fix https://github.com/inveniosoftware/invenio-rdm-records/issues/1652

QUESTIONS:
- is this a valid approach?
- should the default be:
   - yes create parent dois
   - no parent doi should be created
- the PR implements it as DATACITE_ENABLE, where it has to be explicitly turned on
   the problem with that is, that it is not backwards compatible to the actual code, BUT it would be backwards compatible to v11
- i added a check to the migration script, to only create parent DOI's if datacite is enabled ([PR](https://github.com/inveniosoftware/invenio-app-rdm/blob/027a3ad02c0c6617a91a88bff72b2a539fa82556/invenio_app_rdm/upgrade_scripts/migrate_11_0_to_12_0.py#L54)). if the `PARENT_DOI_ENABLE` config will be merged this should be added to the migration script too

### DIFF
--- a/invenio_rdm_records/config.py
+++ b/invenio_rdm_records/config.py
@@ -2,7 +2,7 @@
 #
 # Copyright (C) 2019-2024 CERN.
 # Copyright (C) 2019 Northwestern University.
-# Copyright (C) 2021-2023 Graz University of Technology.
+# Copyright (C) 2021-2024 Graz University of Technology.
 # Copyright (C) 2023 TU Wien.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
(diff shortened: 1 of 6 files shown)


slint (approved):
I like the solution, and decoupling DataCite semantic/config from PID configuration :)

I have to admit I'm a bit confused by the "protocol" naming, since it reminds me of HTTP/transfer protocol. Since we've already did the bad thing and hardcoded special rules for the DOI in these classes, maybe it makes sense to just double-down and name the param/field `doi_enabled_cfg_key`.

I would later on deprecate this class and refactor to fully disassociate `DATACITE_ENABLED` from it, make the `"doi"` magic-string configurable, etc. and then either:

- make a major version bump, and break the API
- keep the old class for backwards compact and make a new one

PR #1746, opened 2024-05-10
services: fix permission for file edit.

alejandromumo:
closes https://github.com/zenodo/zenodo-rdm/issues/852


slint (approved):

slint:
@tmorrell I remember you mentioned using this function/config. The issue was that the Draft was being passed as the `record` parameter, and the Record wasn't passed at all, so we've added more explicit params and switched them around. Depending on what kind of attributes you were checking this might not affect you at all, but just a heads up.

tmorrell:
Thanks @slint, I've updated our test instance. You might want to update https://github.com/inveniosoftware/docs-invenio-rdm/blob/dev/docs/customize/files-and-versioning.md as well.

PR #1758, opened 2024-07-02
fix: abort on record deletion exception

utnapischtim:
* this closes
  https://github.com/inveniosoftware/invenio-app-rdm/issues/2472

* the RecordDeletedException should be catched and hided from the user

NOTE: this has to be backported to v10.7.X to be included into v12

### DIFF
--- a/invenio_rdm_records/resources/iiif.py
+++ b/invenio_rdm_records/resources/iiif.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2024-2024 CERN.
 # Copyright (C) 2022 Universität Hamburg.
+# Copyright (C) 2024 Graz University of Technology.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
@@ -43,6 +44,7 @@ from invenio_records_resources.services.base.config import ConfiguratorMixin, Fr
 from PIL.Image import DecompressionBombError
 from werkzeug.utils import cached_property, secure_filename
 
+from ..services.errors import RecordDeletedException
 from .serializers import (
     IIIFCanvasV2JSONSerializer,
     IIIFInfoV2JSONSerializer,
@@ -95,6 +97,14 @@ class IIIFResourceConfig(ResourceConfig, ConfiguratorMixin):
         DecompressionBombError: create_error_handler(
             lambda e: HTTPJSONException(
                 code=403, description=_("Image size limit exceeded")
+            )
+        ),
+        RecordDeletedException: create_error_handler(
+            lambda e: HTTPJSONException(
+                code=410,
+                description=_(
+                    "The record associated with this file has been deleted. See deletion notice."
+                ),
             )
         ),
     }


invenio_rdm_records/resources/iiif.py
```diff
@@ -48,6 +50,24 @@
...
+    the method was called for a deleted record. The exception should be hidden
+    from the user.
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except RecordDeletedException:
+            abort(404)
```
utnapischtim, line 66: discussion: should we use another error code here?

jrcastro2: IMHO we should be consistent and display the same code (410) for all the record related endpoints:
Screenshot from 2024-07-10 09-06-06

WDYT?

utnapischtim: i was not aware of this error code. is it enough to do `abort(410)` or do i have to implement something that the return dict will look like the json in the screenshot?

jrcastro2: Hmm, I think this is a bit more complex, the best/easiest solution I can think of is to simply add the error_handler that catches that error (RecordDeletedException) in the iiif resource layer returning a 410 with a message stating that the record is not available anymore, here: https://github.com/inveniosoftware/invenio-rdm-records/blob/6f028b22c5c2ea91a3df9162f9e385f3b5b5439c/invenio_rdm_records/resources/iiif.py#L94-L100

By doing this we are not displayig the full information about why the record was deleted but this is stored in the record data, maybe it wouldn't be too difficult to pass that data, but to me it would be enoughto have the 410 and a message saying the record was deleted. WDYT?

utnapischtim: something like https://github.com/inveniosoftware/invenio-rdm-records/pull/1758/commits/3c655e7f357c8910e743cbb52f846138de013c82 ?

invenio_rdm_records/resources/iiif.py
```diff
@@ -97,6 +99,11 @@ class IIIFResourceConfig(ResourceConfig, ConfiguratorMixin):
                 code=403, description=_("Image size limit exceeded")
             )
         ),
+        RecordDeletedException: create_error_handler(
+            lambda e: HTTPJSONException(
+                code=410, description=_("Record not accessible.")
```
jrcastro2, line 104: Minor: Since the record was deleted and anyone can access that information I would explictly say it in the message by either stating that it's gone or deleted.
```suggestion
                code=410, description=_("Record is gone.")
```

When stating not accessible it might be not fully clear and people trying to access it - without an extensive knowledge about status codes -  might think it's a permission issue.

jrcastro2 (approved):
LGTM! 
Just left a minor comment about the error message.

invenio_rdm_records/resources/iiif.py
utnapischtim: the problem i see is that with `Record is gone` someone also can ask `why is it gone?` there will always be a `why` question. My intention was to not reveal the real reason why it is not accessible. i had a discussion some months ago about if it is ok to reveal that a record has been deleted. the conclusion was that it should not be done, but unfortunately i forgot the justification.

kpsherva: How about:
"The record associated with this file was deleted. See deletion notice" <---link to tombstone of the record

The link would be there to explain deletion reasons, since questions about this might be raised by someone, as Christoph pointed out. Not sure how difficult would be to implement the tombstone link, maybe to be done as an improvement later on, if it is not straightforward.

utnapischtim: i would prefer "The record associated with this file has been deleted. See deletion notice." I am not sure about the "See deletion notice", because without a link it is not that helpful, but it helps more than no indication where to find relevant information. 

my first impression is that getting a link there will be a little bit complicated. i will create an issue for that.

PR #1761, opened 2024-07-05
iiif: schema: only return images within size limit in manifest

yashlamba:
IIIF Manifest should only serve images within minimum size constraints.

### DIFF
--- a/invenio_rdm_records/resources/serializers/iiif/schema.py
+++ b/invenio_rdm_records/resources/serializers/iiif/schema.py
@@ -132,10 +132,18 @@ class ListIIIFFilesAttribute(fields.List):
 
     def get_value(self, obj, *args, **kwargs):
         """Return the value for a given key from an object attribute."""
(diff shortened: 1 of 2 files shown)


0einstein0 (approved):

0einstein0 (approved):

invenio_rdm_records/resources/serializers/iiif/schema.py
```diff
@@ -132,10 +132,18 @@ class ListIIIFFilesAttribute(fields.List):
 
     def get_value(self, obj, *args, **kwargs):
         """Return the value for a given key from an object attribute."""
+        iiif_config = current_app.config.get("IIIF_TILES_CONVERTER_PARAMS")
+        valid_metadata = (
```
alejandromumo, line 136: nit: I'd prefer this as a separate function for readability. But just personal preference, nothing wrong with it per se.

alejandromumo (approved):

alejandromumo:
Note that tests are failing, but they also do in master

PR #1763, opened 2024-07-08
tests: update the uploader of reviews

kpsherva:
* fixes the identity for open review policy (community owner gets auto accept) tests should be running with regular users


zzacharo (approved):

alejandromumo (approved):
Thanks a lot for the fix!

PR #1770, opened 2024-07-12
codemeta: added identifier to schema

alejandromumo:
closes https://github.com/inveniosoftware/invenio-swh/issues/33

### DIFF
--- a/invenio_rdm_records/resources/serializers/codemeta/schema.py
+++ b/invenio_rdm_records/resources/serializers/codemeta/schema.py
@@ -20,6 +20,7 @@ class CodemetaSchema(SchemaorgSchema):
     CodeMeta schema derives from Schema.org
     """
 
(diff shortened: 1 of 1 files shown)


invenio_rdm_records/resources/serializers/codemeta/schema.py
```diff
@@ -20,6 +20,7 @@ class CodemetaSchema(SchemaorgSchema):
     CodeMeta schema derives from Schema.org
     """
 
+    identifier = fields.Method("get_id")
```
alejandromumo, line 23: This is defined by `SchemaorgSchema`, however for Codemeta it must be added as in the `identifier` field.

See [schema](https://raw.githubusercontent.com/codemeta/codemeta/2.0/codemeta.jsonld) for reference

slint (approved):

PR #1774, opened 2024-07-17
records: update index settings for accent search

sakshamarora1:
closes: https://github.com/inveniosoftware/invenio-app-rdm/issues/2746
closes: https://github.com/inveniosoftware/invenio-app-rdm/issues/2766

docs for reference:
https://opensearch.org/docs/latest/analyzers/#built-in-analyzers
https://opensearch.org/docs/latest/analyzers/token-filters/index/

`strip_special_chars` is a custom char_filter which uses pattern_replace (https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-charfilters.html)


carlinmack:
I've only checked os-v2 but it seems like there's an extra level of indentation on the whole file after the first line

carlinmack:
It also seems like the entire `mappings.properties.parent.properties.communities.properties.entries` path has been deleted, around 200 lines, is this on purpose?

invenio_rdm_records/records/jsonschemas/records/record-v7.0.0.json
```diff
@@ -0,0 +1,460 @@
+{
```
jrcastro2, line 1: Why do we need this one? it looks like it's the same content as v6, unless I am missing something, see here comparaison: https://www.diffchecker.com/FZPE8bDa/

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
```diff
@@ -0,0 +1,1480 @@
...
+      "metadata.subjects.subject",
+      "metadata.version"
+    ],
+    "analysis": {
+      "char_filter": {
+        "strip_special_chars": {
+          "type": "pattern_replace",
+          "pattern": "[\\p{Punct}\\p{S}]",
+          "replacement": ""
+        }
```
jrcastro2, line 1461: minor: I see that we have also removed special chars, by doing so we are affecting equations, let's say we have the following input in the title "a/b = c*d - e", with this changes if I want to search for it I would need to write "ab  cd  e", which doesn't seem very intuitive for the user. I would leave this one out from RDM, unless we can match both cases, WDYT?

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
```diff
@@ -0,0 +1,1480 @@
...
+            "type": "text",
+            "fields": {
+              "keyword": {
+                "type": "keyword",
+                "ignore_above": 256,
+                "normalizer": "custom_normalizer"
+              }
+            },
+            "analyzer": "custom_analyzer",
+            "search_analyzer": "custom_analyzer"
```
jrcastro2, line 1111: Comment: Should this also be applied to other fields such as additional titles, ... ?

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
sakshamarora1: Added more fields!

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
sakshamarora1: Added more fields!

invenio_rdm_records/records/jsonschemas/records/record-v7.0.0.json
sakshamarora1: Yes but do we not need to add the same schema for v7 if we are adding v7 mappings? is it different from ILS?

jrcastro2 (approved):
LGTM! 🚀 

Maybe in the future we want to apply these to more fields but it's a good start!

invenio_rdm_records/records/mappings/os-v1/rdmrecords/records/record-v7.0.0.json
```diff
@@ -0,0 +1,1491 @@
...
+              }
+            }
+          }
+        }
+      }
+    ],
+    "dynamic": "strict",
+    "date_detection": false,
+    "numeric_detection": false,
+    "properties": {
```
kpsherva, line 76: for which fields did you apply the new analysers ?

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
kpsherva: was it addressed? could someone confirm?

invenio_rdm_records/records/mappings/os-v1/rdmrecords/records/record-v7.0.0.json
sakshamarora1: It's the titles, additional titles, description, and name sub fields of creators, contributors and person_or_org
Here is the diff for the record index:
```diff
@@ -613,7 +613,9 @@
                     "type": "keyword"
                   },
                   "name": {
-                    "type": "text"
+                    "type": "text",
+                    "analyzer": "accent_analyzer",
+                    "search_analyzer": "accent_analyzer"
                   },
                   "identifiers": {
                     "properties": {
@@ -682,7 +684,9 @@
                     "type": "keyword"
                   },
                   "name": {
-                    "type": "text"
+                    "type": "text",
+                    "analyzer": "accent_analyzer",
+                    "search_analyzer": "accent_analyzer"
                   },
                   "identifiers": {
                     "properties": {
@@ -715,7 +719,9 @@
                     }
                   },
                   "name": {
-                    "type": "text"
+                    "type": "text",
+                    "analyzer": "accent_analyzer",
+                    "search_analyzer": "accent_analyzer"
                   },
                   "type": {
                     "type": "keyword"
@@ -768,7 +774,9 @@
             }
           },
           "description": {
-            "type": "text"
+            "type": "text",
+            "analyzer": "accent_analyzer",
+            "search_analyzer": "accent_analyzer"
           },
           "additional_descriptions": {
             "properties": {
@@ -1100,10 +1108,15 @@
           },
           "title": {
             "type": "text",
+            "analyzer": "accent_analyzer",
+            "search_analyzer": "accent_analyzer",
             "fields": {
               "keyword": {
                 "type": "keyword",
                 "ignore_above": 256
+              },
+              "original": {
+                "type": "text"
               }
             }
           },
@@ -1125,7 +1138,14 @@
                 }
               },
               "title": {
-                "type": "text"
+                "type": "text",
+                "analyzer": "accent_analyzer",
+                "search_analyzer": "accent_analyzer",
+                "fields": {
+                  "original": {
+                    "type": "text"
+                  }
+                }
               },
               "type": {
                 "type": "object",
```

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
sakshamarora1: Sorry I missed replying
Yes, this has been addressed by adding an `original` title sub field which doesn't use any analyzers.
So we are not matching both cases.

tests/resources/test_resources_search.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_advance_record_search(
+    client,
+    headers,
+    record_community,
+    minimal_record,
+    uploader,
+):
+    """Test search record's communities."""
```
anikachurilova, line 20: Could you please also add a test with punctuation? For example, searching by just `title` and getting no results, then trying to search by `title.original`?

zzacharo (approved):

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
```diff
@@ -0,0 +1,1491 @@
+{
+  "settings": {
+    "index.query.default_field": [
```
ntarocco, line 3: While reviewing this, me and @kpsherva were wondering if we are all ok with restricting the common search (without specifying the field(s)) to these fields. A quick discussion would be nice.

In any case, it looks like that the `text` fields have been added here, which makes sense. However these are missing:
```
dates.description <--- probably not needed
additional_descriptions.description
references.reference
additional_titles.title
```

Given that `pids` is not added, if we search for a DOI, will the record be found?

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
```diff
@@ -0,0 +1,1491 @@
...
+      "id",
+      "metadata.title",
+      "metadata.title.original",
```
ntarocco, line 6: the `title.original` is not part of the mapping, but looks like coming from the `dynamic` section, correct?

tests/resources/test_resources_search.py
ntarocco: Looks like done :)

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
sakshamarora1: It is part of the mapping
[See line 1164](https://github.com/inveniosoftware/invenio-rdm-records/pull/1774/files#diff-40a24708a1fed90c5895e1320701bb212c6230ad4bf6d9846a67e973cc5e9b23R1164)

PR #1781, opened 2024-07-25
services: added record extension registry

alejandromumo:
This PR introduces a new mechanism to extend the configured record class.

The idea is that some modules might want to add a single field (or method) to the record class, without polluting the core's RDM namespace. 

The interface of these "extensions" is a callback that returns a valid Record class. 

The extension point right now uses entry points since they can be loaded before all the extensions are initialised.

One example of a registration could be the following:

```python
class SWHExtension(object):
    swhid = SystemField("swhid")


def add_swh_extension(app, record_cls):
    # generate a new class with the field and return it
    if current_swh_ext.is_enabled(app) and current_swh_ext.is_configured(app):
        return type(record_cls.__name__, (record_cls, SWHExtension), {})
    return record_cls
```

There are still unanswered questions to move forward:

- The field can't be stored in DB or ES, if both mappings are set to `strict`
- Dumpers are not extensible at the moment
- Serializer's schemas are not extensible from config, instead, they are extensible on object initialisation. Therefore, we can't extend it from other modules than `invenio-rdm-records` (without subclassing the serializer entirely)

### DIFF
--- a/invenio_rdm_records/ext.py
+++ b/invenio_rdm_records/ext.py
@@ -10,6 +10,8 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 
 """DataCite-based data model for Invenio."""
+from importlib_metadata import entry_points
+import pkg_resources
 from flask import Blueprint
 from flask_iiif import IIIF
 from flask_principal import identity_loaded
@@ -95,6 +97,7 @@ class InvenioRDMRecords(object):
     def init_app(self, app):
         """Flask application initialization."""
         self.init_config(app)
+        self.init_record_service_registry(app)
         self.init_services(app)
         self.init_resource(app)
         app.extensions["invenio-rdm-records"] = self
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/ext.py
```diff
@@ -279,6 +282,24 @@ def fix_datacite_configs(self, app):
...
                 app.config[config_item] = str(app.config[config_item])
 
+    def init_record_service_registry(self, app):
+        # TODO load entry points
+        self.record_service_registry = {}
+        self._register_entry_point(
+            self.record_service_registry,
+            "invenio_rdm_records.services.record_service_registry",
+        )
+        # Discussed w/Alex
```
alejandromumo, line 292: TODO: to be removed

invenio_rdm_records/services/config.py
```diff
@@ -277,8 +278,20 @@ class RDMFileRecordServiceConfig(FileServiceConfig, ConfiguratorMixin):
...
+    def record_cls(self):
+        """Record class."""
+        cfg_cls = self._app.config.get("RDM_RECORD_CLS", RDMRecord)
+        exts = getattr(
+            self._app.extensions["invenio-rdm-records"], "record_service_registry", {}
+        )
+        new_cls = cfg_cls
+        for name, _ext in exts.items():
+            tmp = _ext(self._app, new_cls)
+            assert type(tmp) == type(cfg_cls), f"Invalid record type. Expected: {type(cfg_cls)}, got: {type(tmp)}"
```
alejandromumo, line 291: This is the place where the interface is being enforced

PR #1783, opened 2024-07-29
User moderation reliability fixes

slint:
- Closes https://github.com/zenodo/zenodo-rdm/issues/939
- **user_moderation: improve DB queries and use Celery tasks**

### DIFF
--- a/invenio_rdm_records/requests/user_moderation/actions.py
+++ b/invenio_rdm_records/requests/user_moderation/actions.py
@@ -8,11 +8,12 @@
 """RDM user moderation action."""
 
 from invenio_access.permissions import system_identity
+from invenio_db import db
 from invenio_pidstore.errors import PIDDoesNotExistError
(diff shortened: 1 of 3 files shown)


carlinmack:
very minor but I'd prefer the chore things to be in a separate PR which could be very easily merged, and this PR would be easier to review :)

invenio_rdm_records/requests/user_moderation/actions.py
carlinmack, line 1: minor: perhaps the docstrings need updating?

slint: I thought the same, but the operation is still slow because of the "scan" database query. It's somewhat quicker, because the work for each record is distributed to Celery tasks, but still not in the range where you can use it in an HTTP request.

carlinmack: good to know!

PR #1808, opened 2024-09-03
(UI) Publish buttons: refactored isDisabled function

alejandromumo:
**Issue**

On Zenodo, we have seen multiple inclusion requests being created for records that have "Pending" files. When the receiver tries to accept the request - and publish the record - it fails and does not show any useful information for the user. 

**How to replicate**

1. Create a draft
2. Fill in the required metadata
3. Select a community
4. Add a file but don't let it finish (e.g. use network throttling or simply reload the page)
5. Validate that the button "Submit" is enabled

**Changes**

* Refactored isDisabled function in PublishButton and SubmitReviewButton components to check if all files are finished uploading before enabling the button.
* Passed filesState instead of just the number of files to the function.
* Added a check for non-finished files.
* Added early returns to avoid unnecessary checks on possible large objects / arrays.


jrcastro2 (approved):
LGTM! 🚀

alejandromumo:
Created an issue for the backend: https://github.com/inveniosoftware/invenio-rdm-records/issues/1809

PR #1810, opened 2024-09-04
user_moderation: use search for faster actions

slint:
* Use search results to determine the user's list of records.
* Use a TaskOp and Unit of Work to avoid sending Celery tasks immediately.
* Add a cleanup task that will perform a more thorough check using the
  DB to lookup the user's records.

### DIFF
--- a/invenio_rdm_records/requests/user_moderation/actions.py
+++ b/invenio_rdm_records/requests/user_moderation/actions.py
@@ -8,38 +8,18 @@
 """RDM user moderation action."""
 
 from invenio_access.permissions import system_identity
-from invenio_db import db
 from invenio_pidstore.errors import PIDDoesNotExistError
+from invenio_records_resources.services.uow import TaskOp
 from invenio_vocabularies.proxies import current_service
 
 from ...proxies import current_rdm_records_service
-from .tasks import delete_record, restore_record
-
-
-def _get_records_for_user(user_id):
-    """Helper function for getting all the records of the user.
-
-    Note: This function performs DB queries yielding all records for a given
-    user (which is not hard-limited in the system) and performs service calls
-    on each of them. Thus, this function has the potential of being a very
-    heavy operation, and should not be called as part of the handling of an
-    HTTP request!
-    """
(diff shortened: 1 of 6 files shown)


invenio_rdm_records/requests/user_moderation/actions.py
```diff
@@ -63,8 +43,18 @@ def on_block(user_id, uow=None, **kwargs):
...
+    for recid in get_user_records(user_id):
+        uow.register(TaskOp(delete_record, recid=recid, tombstone_data=tombstone_data))
+
+    # Send cleanup task to make sure all records are deleted
+    uow.register(
+        TaskOp.for_async_apply(
+            user_block_cleanup,
+            kwargs=dict(user_id=user_id, tombstone_data=tombstone_data),
+            # wait for 10 minutes before starting the cleanup
+            countdown=10 * 60,
```
slint, line 55: This is a bit of a "magic number" here. The idea is that we first send off the individual tasks for each record but also send off a delayed task that will perform a more "thorough" check by querying the DB.

invenio_rdm_records/requests/user_moderation/utils.py
```diff
@@ -0,0 +1,48 @@
...
+
+"""RDM user moderation utilities."""
+
+from invenio_db import db
+from invenio_search.api import RecordsSearchV2
+
+from ...proxies import current_rdm_records_service
+
+
+def get_user_records(user_id, from_db=False, status=None):
```
kpsherva, line 17: shouldn't this be defined as a service method?

slint: After looking into this, it looks like it will require multiple changes to the underlying [`RecordService.search()`](https://github.com/inveniosoftware/invenio-records-resources/blob/e297181296eef56bdfb0d1486c3e570fa741a0aa/invenio_records_resources/services/records/service.py#L241), to allow for controlling the `source` field (i.e. what fields are included in the search results).

I would wait for another use-case where we need to do more flexible querying, before introducing a new set of parameters, since the change now would take a considerable effort.

PR #1813, opened 2024-09-16
Services: Communities: Add feature to require community based on config

sakshamarora1:
closes: https://github.com/CERNDocumentServer/cds-rdm/issues/88

Screenshot 2024-09-16 at 17 24 43

Screenshot 2024-09-16 at 17 25 04


invenio_rdm_records/services/communities/service.py
```diff
@@ -205,6 +207,16 @@ def _remove(self, identity, community_id, record):
         if community_id not in record.parent.communities.ids:
             raise RecordCommunityMissing(record.id, community_id)
 
+        # If config is True and there is only 1 communities left to remove
+        # Then, check for permissions to remove last community
+        if (
```
kpsherva, line 212: check the comment about readability in drafts-resources

tests/resources/test_resources_communities.py
```diff
@@ -889,3 +890,179 @@ def test_add_record_to_restricted_community_submission_open_member(
     assert not response.json.get("errors")
     processed = response.json["processed"]
     assert len(processed) == 1
+
+# Assure Records community exists tests
+# -------------------------------------
+from invenio_records_resources.services.errors import CommunityNotSelectedError, CannotRemoveCommunityError
```
kpsherva, line 896: please move imports on the top of the file

kpsherva:
UI comment: could you make sure the background on the error message is uniform? otherwise it looks like a bug

invenio_rdm_records/services/communities/service.py
```diff
@@ -205,10 +203,22 @@ def _remove(self, identity, community_id, record):
         if community_id not in record.parent.communities.ids:
             raise RecordCommunityMissing(record.id, community_id)
 
-        # check permission here, per community: curator cannot remove another community
-        self.require_permission(
+        # If config is true and there is only 1 communities left to remove
+        is_community_required = current_app.config["RDM_RECORD_ALWAYS_IN_COMMUNITY"]
+        is_last_community = len(record.parent.communities.ids) == 1
+        # Then, check for permissions to remove last community
+        can_remove_last_community = self.check_permission(
```
kpsherva, line 210: the name of this variable is misleading, because you don't use is really only for checking last community permission, but in general, if the user can remove community
```suggestion
        can_remove_community = self.check_permission(
```

invenio_rdm_records/services/communities/service.py
```diff
@@ -205,10 +203,22 @@ def _remove(self, identity, community_id, record):
...
             identity, "remove_community", record=record, community_id=community_id
         )
+        if (
+            is_community_required
+            and is_last_community
+            and not can_remove_last_community
```
kpsherva, line 216:
```suggestion
            and not can_remove_community
```

invenio_rdm_records/services/communities/service.py
```diff
@@ -205,10 +203,22 @@ def _remove(self, identity, community_id, record):
         if community_id not in record.parent.communities.ids:
             raise RecordCommunityMissing(record.id, community_id)
 
-        # check permission here, per community: curator cannot remove another community
-        self.require_permission(
+        # If config is true and there is only 1 communities left to remove
+        is_community_required = current_app.config["RDM_RECORD_ALWAYS_IN_COMMUNITY"]
+        is_last_community = len(record.parent.communities.ids) == 1
```
kpsherva, line 208: could you remind me why we are rechecking these conditions here since it is already covered by the permission generator?

invenio_rdm_records/services/services.py
```diff
@@ -404,6 +406,29 @@ def purge_record(self, identity, id_, uow=None):
...
+        """Publish a draft.
+
+        Check for permissions to publish a draft and then call invenio_drafts_resourcs.services.records.services.publish()
+        """
+        # Get the draft
+        draft = self.draft_cls.pid.resolve(id_, registered_only=False)
+
+        # If config is true and there are no communities selected
+        is_community_required = current_app.config["RDM_RECORD_ALWAYS_IN_COMMUNITY"]
+        is_community_missing = len(draft.parent.communities.ids) == 0
```
kpsherva, line 420: same question as in remove method, could you remind me why we check these conditions both in the generator and there?

invenio_rdm_records/services/communities/service.py
sakshamarora1: The check for removing a community is performed by permissions, here we are changing the permission check based on the config. So if the uploader has permission to remove the community, he won't be able to remove just the last community since a record should always have a community if the config is set to true.

invenio_rdm_records/services/services.py
```diff
@@ -404,6 +406,29 @@ def purge_record(self, identity, id_, uow=None):
...
+        # Then, check for permissions to upload without community
+        can_publish_draft = self.check_permission(identity, "publish", record=draft)
+
+        if is_community_required and is_community_missing and not can_publish_draft:
+            raise CommunityNotSelectedError()
+        elif not can_publish_draft:
```
kpsherva, line 426: is it needed, since the parent method will raise if permissions are not met?

invenio_rdm_records/services/generators.py
```diff
@@ -414,3 +414,14 @@ def needs(self, request=None, **kwargs):
...
+
+class IfOneCommunity(ConditionalGenerator):
+    """Conditional generator for records always in communities case."""
+
+    def _condition(self, record=None, **kwargs):
+        """Check if the record is associated with zero or one community."""
+        if record is None:
+            return True
+        rec_communities = record.parent.communities.ids
+        return len(rec_communities) == 1
```
kpsherva, line 427: record can have more than one community when it is republished

sakshamarora1:
Looks like there was some issue with my assets watch command, the UI looks fine after I created a new instance in a new virtual env.
Screenshot 2024-09-26 at 11 22 05

Also, reopened https://github.com/inveniosoftware/invenio-app-rdm/pull/2846, as it was erroring out when you try to remove a community that has already been removed from another tab/window.

invenio_rdm_records/services/communities/service.py
ntarocco, line 210: This is checking if the user can remove the community, not really the last one:
```suggestion
        can_remove_community = self.check_permission(
```

invenio_rdm_records/services/communities/service.py
```diff
@@ -205,10 +203,22 @@ def _remove(self, identity, community_id, record):
...
             identity, "remove_community", record=record, community_id=community_id
         )
+        if (
+            is_community_required
+            and is_last_community
+            and not can_remove_last_community
+        ):
+            raise CannotRemoveCommunityError()
+        # check permission here, per community: curator cannot remove another community
+        elif not can_remove_last_community:
```
ntarocco, line 220: There is something telling me that we could simplify the `if/else` logic here?

invenio_rdm_records/services/communities/service.py
ntarocco, line 216: Is this `not ... ` branch correct?

```
if not can_remove:
  fail
elif is_community_required and is_last_community:
  fail
```

invenio_rdm_records/services/permissions.py
```diff
@@ -209,11 +221,23 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
...
         CommunityCurators(),
         SystemProcess(),
     ]
+    can_remove_community = [
+        IfConfig(
+            "RDM_RECORD_ALWAYS_IN_COMMUNITY",
+            then_=[
+                IfOneCommunity(
+                    then_=[Administration(), SystemProcess()],
+                    else_=can_remove_community_,
```
ntarocco, line 235: we know that using permissions like this is dangerous (ping @zzacharo), even if we don't have better solutions at the moment.

invenio_rdm_records/config.py
```diff
@@ -131,6 +131,12 @@ def always_valid(identifier):
 RDM_ALLOW_RESTRICTED_RECORDS = True
 """Allow users to set restricted/private records."""
 
+#
+# Record communities
+#
+RDM_RECORD_ALWAYS_IN_COMMUNITY = False
```
ntarocco, line 137: Any idea if we can find a better name to this config var?

sakshamarora1: How about `RDM_COMMUNITY_REQUIRED` or `RDM_REQUIRE_COMMUNITY`?

invenio_rdm_records/services/permissions.py
```diff
@@ -199,7 +201,17 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
         ),
     ]
     # Allow publishing a new record or changes to an existing record.
-    can_publish = can_review
+    can_publish = [
```
ntarocco, line 204: Let's do the same as described above: we define a new permission `can_publish_` where, when the flag `RDM_RECORD_ALWAYS_IN_COMMUNITY` is enabled, the configured roles can still publish, and we remove the `if` from here.

invenio_rdm_records/services/errors.py
```diff
@@ -206,3 +206,15 @@ class RecordSubmissionClosedCommunityError(PermissionDenied):
     """Record submission policy forbids non-members from submitting records to community."""
 
     description = "Submission to this community is only allowed to community members."
+
+
+class CommunityNotSelectedError(Exception):
+    """Error thrown when a record is being created/updated with less than 1 community."""
+
+    description = "Cannot publish without selecting a community."
```
ntarocco, line 214: Please make all strings translatable:
```suggestion
    description = _("Cannot publish without selecting a community.")
```

invenio_rdm_records/services/errors.py
```diff
@@ -206,3 +206,15 @@ class RecordSubmissionClosedCommunityError(PermissionDenied):
...
+
+
+class CannotRemoveCommunityError(Exception):
+    """Error thrown when the last community is being removed from the record."""
+
+    description = "Cannot remove. A record should be part of atleast 1 community."
```
ntarocco, line 220:
```suggestion
    description = "Cannot remove. A record should be part of at least 1 community."
```

invenio_rdm_records/config.py
kpsherva: maybe to be more explicit `RDM_COMMUNITY_REQUIRED_TO_PUBLISH`?

invenio_rdm_records/services/generators.py
```diff
@@ -414,3 +414,14 @@ def needs(self, request=None, **kwargs):
             return [AccessRequestTokenNeed(request["payload"]["token"])]
 
         return []
+
+
+class IfOneCommunity(ConditionalGenerator):
+    """Conditional generator for records always in communities case."""
+
+    def _condition(self, record=None, **kwargs):
+        """Check if the record is associated with zero or one community."""
```
Samk13, line 423:
```suggestion
        """Check if the record is associated with at least one community."""
```

invenio_rdm_records/services/generators.py
```diff
@@ -414,3 +414,25 @@ def needs(self, request=None, **kwargs):
...
+            return True
+        rec_communities = record.parent.communities.ids
+        return len(rec_communities) == 1
+
+
+class IfAtLeastOneCommunity(ConditionalGenerator):
+    """Conditional generator for records always in communities case."""
+
+    def _condition(self, record=None, **kwargs):
+        """Check if the record is associated with zero or one community."""
```
Samk13, line 434:
```suggestion
        """Check if the record is associated with at least one community."""
```

invenio_rdm_records/services/generators.py
```diff
@@ -414,3 +414,25 @@ def needs(self, request=None, **kwargs):
...
+        if record is None:
+            return True
+        rec_communities = record.parent.communities.ids
+        return len(rec_communities) > 0
```
Samk13, line 438:
```suggestion
        return bool(record and record.parent.communities.ids)
```

invenio_rdm_records/services/generators.py
```diff
@@ -414,3 +414,25 @@ def needs(self, request=None, **kwargs):
...
+
+class IfOneCommunity(ConditionalGenerator):
+    """Conditional generator for records always in communities case."""
+
+    def _condition(self, record=None, **kwargs):
+        """Check if the record is associated with zero or one community."""
+        if record is None:
+            return True
+        rec_communities = record.parent.communities.ids
+        return len(rec_communities) == 1
```
Samk13, line 427:
```suggestion
        return bool(record and len(record.parent.communities.ids) == 1)
```

invenio_rdm_records/resources/config.py
```diff
@@ -255,6 +257,12 @@ class RDMRecordResourceConfig(RecordResourceConfig, ConfiguratorMixin):
                 description=e.description,
             )
         ),
+        CommunityRequiredError: create_error_handler(
+            HTTPJSONException(
+                code=400,
+                description="Cannot publish without selecting a community.",
```
zzacharo, line 263: missing translations

invenio_rdm_records/services/communities/service.py
```diff
@@ -205,10 +203,20 @@ def _remove(self, identity, community_id, record):
...
+            self.require_permission(
+                identity, "remove_community", record=record, community_id=community_id
+            )
+            # By default, admin/superuser has permission to do everything, so PermissionDeniedError won't be raised for admin in any case
+        except PermissionDeniedError as exc:
+            # If user doesn't have permission, then check if community is required
+            if current_app.config["RDM_COMMUNITY_REQUIRED_TO_PUBLISH"]:
+                # If community is required for a record and if it is the last community to remove, raise error
+                if len(record.parent.communities.ids) == 1:
+                    raise CannotRemoveCommunityError()
```
kpsherva, line 216: for readability
```suggestion
            # if permission is denied, decide which error to raise based on config
            community_required = current_app.config["RDM_COMMUNITY_REQUIRED_TO_PUBLISH"]
            community_missing = len(record.parent.communities.ids) <= 1
            if community_required && community_missing:
                raise CannotRemoveCommunityError()
```

also, I wouldn't check `==1` because you might have zero communities in edge cases. fe.
when someone with permissions has published before the record, then user without permission goes to publish again, this if should also catch this case

invenio_rdm_records/services/generators.py
```diff
@@ -414,3 +414,19 @@ def needs(self, request=None, **kwargs):
...
 
         return []
+
+
+class IfOneCommunity(ConditionalGenerator):
+    """Conditional generator for records always in communities case."""
+
+    def _condition(self, record=None, **kwargs):
+        """Check if the record is associated with one community."""
+        return bool(record and len(record.parent.communities.ids) == 1)
```
kpsherva, line 424: we talked about ==1, if you remember, it will not work in edge cases

invenio_rdm_records/services/services.py
```diff
@@ -404,6 +406,28 @@ def purge_record(self, identity, id_, uow=None):
...
+        """Publish a draft.
+
+        Check for permissions to publish a draft and then call invenio_drafts_resourcs.services.records.services.publish()
+        """
+        try:
+            draft = self.draft_cls.pid.resolve(id_, registered_only=False)
+            self.require_permission(identity, "publish", record=draft)
+            # By default, admin/superuser has permission to do everything, so PermissionDeniedError won't be raised for admin in any case
+        except PermissionDeniedError as exc:
+            # If user doesn't have permission, then check if community is required
```
kpsherva, line 420:
```suggestion
            # If user doesn't have permission to publish, determine which error to display, based on config
```

kpsherva (approved):

ntarocco (approved):

Samk13 (approved):

PR #1824, opened 2024-09-25
Add collections

alejandromumo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1820


invenio_rdm_records/collections/models.py
```diff
@@ -0,0 +1,136 @@
...
+    - slug
+    """
+
+    __tablename__ = "collections"
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    slug = db.Column(db.String(255), unique=False, nullable=False, index=True)
+    path = db.Column(db.Text, nullable=False, index=True)
+    tree_id = db.Column(db.Integer, db.ForeignKey("collection_tree.id"), nullable=False)
+    logo = db.Column(db.Text, nullable=True)
```
alejandromumo, line 83: For communities, logos are defined in a separate table. Are we following that model? If so, we need to change DB tables here

invenio_rdm_records/collections/models.py
```diff
@@ -0,0 +1,136 @@
...
+    title = db.Column(db.String(255), nullable=False)
+    query_text = db.Column("query", db.Text, nullable=False)
+    order = db.Column(db.Integer, nullable=True)
+    # TODO index depth
+    depth = db.Column(
+        db.Integer, db.Computed("array_length(string_to_array(path, ','), 1) - 2")
```
alejandromumo, line 89: `depth` is computed server side. This funcion is PostgreSQL specific, however it can easily be moved to client-side if needed.

invenio_rdm_records/collections/models.py
```diff
@@ -0,0 +1,136 @@
...
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    slug = db.Column(db.String(255), unique=False, nullable=False, index=True)
+    path = db.Column(db.Text, nullable=False, index=True)
+    tree_id = db.Column(db.Integer, db.ForeignKey("collection_tree.id"), nullable=False)
+    logo = db.Column(db.Text, nullable=True)
+    title = db.Column(db.String(255), nullable=False)
+    query_text = db.Column("query", db.Text, nullable=False)
+    order = db.Column(db.Integer, nullable=True)
+    # TODO index depth
```
alejandromumo, line 87: Initially we discussed having `depth` as an non-key column of the `path` index (see [PostgreSQL indices](https://www.postgresql.org/docs/current/sql-createindex.html) `INCLUDE`) . 

I removed it in this version, but can be added afterwards if needed (we could also use a composite index `(path, depth)`. )

invenio_rdm_records/collections/api.py
```diff
@@ -0,0 +1,348 @@
...
+        return self.model.logo
+
+    @property
+    def query(self):
+        """Get the collection query."""
+        q = ""
+        for _a in self.ancestors:
+            q += f"({_a.model.query_text}) AND "
+        q += f"({self.model.query_text})"
+        return q
```
alejandromumo, line 159: use dsl queries and join them 

consider luqum validation

invenio_rdm_records/collections/api.py
```diff
@@ -0,0 +1,348 @@
...
+        """Get the collection ancestors."""
+        if not self.model:
+            return None
+
+        cps = self.path.split(",")
+        ret = []
+        for cid in cps:
+            if not cid:
+                continue
+            cl = Collection.get(cid)
```
alejandromumo, line 172: we could have a `get_many` and do `SELECT * WHERE id in ()`

invenio_rdm_records/collections/api.py
```diff
@@ -0,0 +1,348 @@
...
+            ret.append(cl)
+        return list(sorted(ret, key=lambda x: (x.path, x.order)))
+
+    @cached_property
+    def sub_collections(self):
+        """Fetch all the descendants."""
+        return self.get_subcollections()
+
+    @cached_property
+    def direct_subcollections(self):
```
alejandromumo, line 182: naming: use `children`

invenio_rdm_records/collections/api.py
```diff
@@ -0,0 +1,348 @@
...
+            .filter(
+                self.model_cls.path == f"{self.path}{self.id},",
+                self.model_cls.tree_id == self.ctree_id,
+            )
+            .order_by(self.model_cls.path, self.model_cls.order)
+        )
+        ret = db.session.execute(stmt).scalars().all()
+        return [type(self)(r) for r in ret]
+
+    def get_subcollections(self, max_depth=3):
```
alejandromumo, line 204: this makes sense to have in `CollectionTree` API

invenio_rdm_records/collections/api.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def get_direct_subcollections(self):
+        """Get the collection first level (direct) children.
+
+        More preformant query to retrieve descendants, executes an exact match query.
+        """
+        if not self.model:
+            return None
+        stmt = (
+            select(self.model_cls)
```
alejandromumo, line 194: move `select` and the like to ORM

invenio_rdm_records/collections/api.py
```diff
@@ -0,0 +1,348 @@
...
+            "slug": collection.slug,
+            "depth": collection.depth,
+            "logo": collection.logo,
+            "order": collection.order,
+            "id": collection.id,
+            "query": collection.query,
+        }
+        return res
+
+    def to_dict(self) -> dict:
```
alejandromumo, line 237: missing max_depth

invenio_rdm_records/collections/service.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class CollectionsService:
+    """Collections service."""
+
+    collection_cls = Collection
+
+    def create(self, identity, community_id, tree_slug, slug, title, query, **kwargs):
+        """Create a new collection."""
+        current_communities.service.require_permission(
```
alejandromumo, line 71: if it has a REST API, permissions could be added directly to the service so they can e.g. be overwritten by Zenodo to `System`

invenio_rdm_records/collections/service.py
```diff
@@ -0,0 +1,118 @@
...
+
+        ctree = CollectionTree.get_by_slug(tree_slug, community_id)
+        if not ctree:
+            raise ValueError(f"Collection tree {tree_slug} not found.")
+
+        collection = self.collection_cls.resolve(slug, ctree.id, use_slug=True)
+        if not collection:
+            raise ValueError(f"Collection {slug} not found.")
+
+        return CollectionItem(collection)
```
alejandromumo, line 108: collapse into just one read that is smart enough, same as API get

consider `depth` as well

invenio_rdm_records/collections/models.py
alejandromumo: Created https://github.com/inveniosoftware/invenio-rdm-records/issues/1827

invenio_rdm_records/collections/models.py
```diff
@@ -0,0 +1,144 @@
...
+    path = db.Column(db.Text, nullable=False, index=True)
+    tree_id = db.Column(
+        db.Integer, db.ForeignKey("collections_collection_tree.id"), nullable=False
+    )
+    title = db.Column(db.String(255), nullable=False)
+    search_query = db.Column("query", db.Text, nullable=False)
+    order = db.Column(db.Integer, nullable=True)
+    # TODO index depth
+    depth = db.Column(
+        db.Integer, db.Computed("array_length(string_to_array(path, ','), 1) - 2")
```
anikachurilova, line 100: Will it fail if i'm using MySQL instead of Postgres? If yes, then this must be changed

invenio_rdm_records/collections/models.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @classmethod
+    def get(cls, id_):
+        """Get a collection by ID."""
+        return cls.query.get(id_)
+
+    @classmethod
+    def get_by_slug(cls, slug, tree_id):
+        """Get a collection by slug."""
+        return cls.query.filter(cls.slug == slug, cls.tree_id == tree_id).one_or_none()
```
anikachurilova, line 144: Do we return None, or should it fail when it's not found?

invenio_rdm_records/collections/models.py
```diff
@@ -0,0 +1,144 @@
...
+    def get(cls, id_):
+        """Get a collection tree by ID."""
+        return cls.query.get(id_)
+
+    @classmethod
+    def get_by_slug(cls, slug, community_id):
+        """Get a collection tree by slug."""
+        return cls.query.filter(
+            cls.slug == slug, cls.community_id == community_id
+        ).one_or_none()
```
anikachurilova, line 67: Same question here about `one_or_none`

invenio_rdm_records/collections/api.py
```diff
@@ -0,0 +1,346 @@
...
+
+from invenio_db import db
+from sqlalchemy import select
+from werkzeug.utils import cached_property
+
+from .models import Collection as CollectionModel
+from .models import CollectionTree as CollectionTreeModel
+
+
+class Collection:
```
anikachurilova, line 19: Are we sure that this api layer is really needed, when the data model is not a classic record with a json schema? Can't the models be directly used in the service layer (as done in `invenio-jobs` for example)?
This class looks like a wrapper of the model, and the various methods can be moved to the model class.

invenio_rdm_records/collections/service.py
```diff
@@ -0,0 +1,118 @@
...
+
+from invenio_communities.proxies import current_communities
+
+from .api import Collection, CollectionTree
+
+
+class CollectionItem:
+    """Collection item."""
+
+    def __init__(self, collection: Collection):
```
anikachurilova, line 17: Was using type hints discussed? We don't usually use them in Invenio

invenio_rdm_records/collections/service.py
```diff
@@ -0,0 +1,118 @@
...
+        """Instantiate a Collection object.
+
+        Optionally pass a community to cache its information in the collection's instance.
```
anikachurilova, line 20: This looks copy-pasted from somewhere

invenio_rdm_records/collections/service.py
anikachurilova, line 71: Maybe we should introduce a collection specific permissions instead of using the communities' ones?

`can_update` default value is `CommunityOwners`, is this correct?

invenio_rdm_records/collections/service.py
```diff
@@ -0,0 +1,118 @@
...
+        current_communities.service.require_permission(
+            identity, "update", community_id=community_id
+        )
+        ctree = CollectionTree.get_by_slug(tree_slug, community_id)
+        if not ctree:
+            raise ValueError(f"Collection tree {tree_slug} not found.")
+        collection = self.collection_cls.create(
+            slug=slug, title=title, query=query, ctree=ctree, **kwargs
+        )
+        return CollectionItem(collection)
```
anikachurilova, line 80: `create` and `add` methods are missing the commit operations (for example `uow.register(ModelCommitOp(...))`)

tests/collections/test_collections_api.py
```diff
@@ -0,0 +1,72 @@
...
+# Invenio-RDM is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""Test suite for the collections programmatic API."""
+
+from invenio_rdm_records.collections.api import Collection, CollectionTree
+
+
+def test_create(running_app, db, community, community_owner):
+    """Test collection creation via API."""
+    tree = CollectionTree.create(
```
ntarocco, line 14: I suggest changing the test to read from the DB when asserting the creation of the collection [tree]. Given that the db commit is missing, they should fail.

invenio_rdm_records/collections/models.py
alejandromumo: Yes, it would fail, I will move it to the client side in this PR.

invenio_rdm_records/collections/models.py
alejandromumo: I see the point of failing fast, however, I prefer to return `None` for separation of concerns. Whether the collection does not exist is an exception, I think it should be up to the business layer to decide. From a data point of view, it tried to fetch the collection but nothing was found. 

I agree that we need to throw an exception at some point to have the presentation layer(s) acting accordingly (and that's currently missing in this PR). I will add that part in the next iteration since I refactored part of the public API / service layers from Alex's comments.

invenio_rdm_records/collections/models.py
alejandromumo: replied above

invenio_rdm_records/collections/api.py
alejandromumo: On one hand, I agree this class feels like a wrapper to the model, and it could be richer (e.g. I and Alex discussed storing the relation between collections in memory and then they could be "navigated" in memory). 

However, some ideas that made me implement a public API for collections:

1. Abstract some concepts about how we store collections in DB (materialized path). In my opinion, for the developer, things like `path` should be abstracted and encapsulated under a "nice looking" API.
2. Related to the one above, adding some public-facing utilities to improve the usability of collections, e.g. `collection.ancestors`, `collection.children`, create a collection based on a parent, resolving a collection by `id` or `slug, community_id` pair. 
3. `dump` / `to_dict` methods for serialization.

invenio_rdm_records/collections/service.py
alejandromumo: I copied some code from my prototype, thanks for spotting it 😄 

Will be removed in this PR

invenio_rdm_records/collections/service.py
alejandromumo: This code was from my prototype, and then I changed the signature and forgot about the docs.  

Thanks for spotting it!

invenio_rdm_records/collections/service.py
alejandromumo: You are right, conceptually, communities are not required to have a collection (think about a collection on the repository level). However, the current state of this service is to require a community (this is our use case right now). 

I agree that the collections' permission policy at some point has to be specific for collections, but it still needs more thought and the service to be adapted to it. 

For a first launch, the permission `can_update` from communities is good since we don't want any user of a community creating collections for a community

invenio_rdm_records/collections/service.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def read(self, identity, id_):
+        """Get a collection by ID or slug."""
```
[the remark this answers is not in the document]
alejandromumo: This is going to be refactored in a second iteration when we add collections to the browse page. The idea is that `service.read()` is flexible enough that it can be done by ID or slug

invenio_rdm_records/collections/service.py
alejandromumo: Will be added now :)

invenio_rdm_records/collections/models.py
ntarocco: Maybe `None` works in this case, but you will see soon its limitations. With a None return value there are 2 main problems (and maybe more):
1. every time you call the func, you need to check if the return value is None or not. And the subsequent code will have to handle this case. So, as in your PR, you end up having `if` everywhere. Not very elegant.
2. even if less applicable in this func, in general you can't differentiate different type of unhappy path.

An exception is much more elegant because the caller will have many more options on how to handle the unhappy path.
I don't really understand what are concerns are better separate with returning None.

No problem with solving this later on, in subsequent PRs.

invenio_rdm_records/collections/api.py
ntarocco: My 2 cents

I understand your reasoning. At the same time, with this approach, we will end up having a wrapper classes that have most of the attributes and methods duplicated from the encapsulated model. And then we add a couple of extra methods.
I would not know what is the separation of concerns or abstraction, and if I have to add a new feature (a new query), I would not know if I have to put in the API or in the Model. What is the criteria that tells me where to put it?

For the `path`, maybe, I don't see it for the moment in the class.
For your second point, it can be done in the model, I don't see the difference.

A model is already an API class. What has been done for records is actually very different and in that case very much needed: the model is loaded into a `dict` object, with many other extra methods.

If we start creating wrappers everywhere, we will end up having more complex code, and less DRY. As a comparison, it looks a little bit like the props drilling problem in React.

As an alternative, if you really feel like that we need to separate things, then in this case I would go with inheritance instead of composition.

To recap: I think that our final objective is to have the simplest possible code. If the majority of the APIs exposed by a model are intended to be used, I would use the model directly or create inheritance.
Composition and wrapping make sense to me when you want to hide the majority of the features provided by the wrapped object.

alejandromumo: Those are excellent points, I will keep them in mind in the next PR. 

Thanks for the detailed explanation 🚀

PR #1826, opened 2024-09-27
dependencies: bump flask-iiif

fenekku:
Did we forget to bump this dependency or was there a larger reason for
keeping it clamped so low? None of the flask-iiif improvements from the past year
have been adopted by InvenioRDM it means.

I will backport this to v12 maint branch once merged as it prevents our instance
from previewing specific medical images.


utnapischtim:
no idea why this hasn't been done already, but i have it on my bucket list for flask>=3.0.0 so hopefully next week i would have merged this :)

Samk13 (approved):

PR #1835, opened 2024-10-01
js remove warnings

utnapischtim:
- fix: propTypes warning
- fix: propTypes warning

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/ProtectionButtons.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/ProtectionButtons.js
@@ -1,7 +1,7 @@
 // This file is part of Invenio-RDM-Records
 // Copyright (C) 2020-2023 CERN.
 // Copyright (C) 2020-2022 Northwestern University.
-// Copyright (C) 2021 Graz University of Technology.
+// Copyright (C) 2021-2024 Graz University of Technology.
 //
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/ProtectionButtons.js
```diff
@@ -110,6 +110,6 @@ export class ProtectionButtons extends Component {
 }
 
 ProtectionButtons.propTypes = {
-  canRestrictRecord: PropTypes.bool.isRequired,
+  canRestrictRecord: PropTypes.bool,
```
Samk13, line 113: It Looks like it's still missing `defaultProps` declaration for the `canRestrictRecord` prop?

utnapischtim: didn't saw that on the browser. will check the next time the tests too. beside that:
what should be the default value for it?

Samk13: The default value appears to be true as per the reference [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/a4fbec8dbde537244a58d905495fb6919029abaa/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/MetadataAccess.js#L33C1-L44C1). Feel free to double-check!

kpsherva (approved):

Issue #1852, opened 2024-10-21
collections: update number of records regularly


alejandromumo:
This task aims to implement a mechanism to regularly update the number of records in all the collections.

PR #1853, opened 2024-10-21
collections: added task to compute num of records

alejandromumo:
* added task to compute number of records for all the collections
* added "collection_id" parameter to record search
* added service methods to read collections (many, all)
* added tests

closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1852

### DIFF
--- a/invenio_rdm_records/collections/api.py
+++ b/invenio_rdm_records/collections/api.py
@@ -91,8 +91,21 @@ class Collection:
     @classmethod
     def resolve_many(cls, ids_=None, depth=2):
         """Resolve many collections by ID."""
-        _ids = ids_ or []
-        return [cls(c, depth) for c in cls.model_cls.read_many(_ids)]
+        if not ids_:
+            return []
+        return [cls(c, depth) for c in cls.model_cls.read_many(ids_)]
+
+    @classmethod
+    def resolve_all(cls, depth=2):
+        """Resolve all collections."""
+        return [cls(c, depth) for c in cls.model_cls.read_all()]
+
+    def update(self, **kwargs):
+        """Update the collection."""
+        if "search_query" in kwargs:
+            Collection.validate_query(kwargs["search_query"])
+        self.model.update(**kwargs)
+        return self
 
     def add(self, slug, title, query, order=None, depth=2):
         """Add a subcollection to the collection."""
--- a/invenio_rdm_records/collections/models.py
+++ b/invenio_rdm_records/collections/models.py
@@ -156,9 +156,30 @@ class Collection(db.Model, Timestamp):
         return cls.query.filter(cls.slug == slug, cls.tree_id == tree_id).one_or_none()
 
     @classmethod
-    def read_many(cls, ids):
+    def read_many(cls, ids_):
         """Get many collections by ID."""
-        return cls.query.filter(cls.id.in_(ids)).order_by(cls.path, cls.order)
+        return cls.query.filter(cls.id.in_(ids_)).order_by(cls.path, cls.order)
+
+    @classmethod
+    def read_all(cls):
+        """Get all collections."""
+        return cls.query.order_by(cls.path, cls.order)
+
+    def update(
+        self, /, slug=None, title=None, search_query=None, order=None, num_records=None
+    ):
+        """Update a collection."""
+        if slug is not None:
+            self.slug = slug
+        if title is not None:
+            self.title = title
+        if search_query is not None:
+            self.search_query = search_query
+        if order is not None:
+            self.order = order
+        if num_records is not None:
+            self.num_records = num_records
+        return self
 
     @classmethod
     def get_children(cls, model):
--- a/invenio_rdm_records/collections/schema.py
+++ b/invenio_rdm_records/collections/schema.py
@@ -14,7 +14,8 @@ class CollectionSchema(Schema):
 
     slug = fields.Str()
     title = fields.Str()
-    depth = fields.Int()
+    depth = fields.Int(dump_only=True)
     order = fields.Int()
-    id = fields.Int()
+    id = fields.Int(dump_only=True)
     num_records = fields.Int()
+    search_query = fields.Str(load_only=True)
--- a/invenio_rdm_records/collections/service.py
+++ b/invenio_rdm_records/collections/service.py
@@ -9,13 +9,19 @@
 import os
 
 from flask import current_app, url_for
+from invenio_records_resources.services import ServiceSchemaWrapper
 from invenio_records_resources.services.base import Service
 from invenio_records_resources.services.uow import ModelCommitOp, unit_of_work
+
+from invenio_rdm_records.proxies import (
+    current_community_records_service,
+    current_rdm_records_service,
+)
 
 from .api import Collection, CollectionTree
 from .errors import LogoNotFoundError
 from .links import CollectionLinkstemplate
-from .results import CollectionItem, CollectionTreeList
+from .results import CollectionItem, CollectionList, CollectionTreeList
 
 
 class CollectionsService(Service):
@@ -30,7 +36,7 @@ class CollectionsService(Service):
     @property
     def collection_schema(self):
         """Get the collection schema."""
-        return self.config.schema()
+        return ServiceSchemaWrapper(self, schema=self.config.schema)
 
     @property
     def links_item_tpl(self):
@@ -121,6 +127,31 @@ class CollectionsService(Service):
             identity, new_collection, self.collection_schema, self.links_item_tpl
         )
 
+    @unit_of_work()
+    def update(self, identity, collection_or_id, data=None, uow=None):
+        """Update a collection."""
+        if isinstance(collection_or_id, int):
+            collection = self.collection_cls.resolve(id_=collection_or_id)
+        else:
+            collection = collection_or_id
+        self.require_permission(
+            identity, "update", community_id=collection.community.id
+        )
+
+        data = data or {}
+
+        valid_data, errors = self.collection_schema.load(
+            data, context={"identity": identity}, raise_errors=True
+        )
+
+        res = collection.update(**valid_data)
+
+        uow.register(ModelCommitOp(res.model))
+
+        return CollectionItem(
+            identity, collection, self.collection_schema, self.links_item_tpl
+        )
+
     def read_logo(self, identity, slug):
         """Read a collection logo.
 
@@ -131,3 +162,41 @@ class CollectionsService(Service):
         if _exists:
             return url_for("static", filename=logo_path)
         raise LogoNotFoundError()
+
+    def read_many(self, identity, ids_=None, depth=2):
+        """Get many collections."""
+        self.require_permission(identity, "read")
+
+        ids_ = ids_ or []
+        res = self.collection_cls.resolve_many(ids_, depth=depth)
+        return CollectionList(
+            identity, res, self.collection_schema, None, self.links_item_tpl
+        )
+
+    def read_all(self, identity, depth=2):
+        """Get all collections."""
+        self.require_permission(identity, "read")
+        res = self.collection_cls.resolve_all(depth=depth)
+        return CollectionList(
+            identity, res, self.collection_schema, None, self.links_item_tpl
+        )
+
+    def search_records(self, identity, collection_or_id, params=None):
+        """Search records in a collection."""
+        params = params or {}
+
+        if isinstance(collection_or_id, int):
+            collection = self.collection_cls.resolve(id_=collection_or_id)
+        else:
+            collection = collection_or_id
+
+        params.update({"collection_id": collection.id})
+        if collection.community:
+            res = current_community_records_service.search(
+                identity,
+                community_id=collection.community.id,
+                params=params,
+            )
+        else:
+            raise NotImplementedError("Search for collections without community not supported.")
+        return res
--- a/invenio_rdm_records/collections/tasks.py
+++ b/invenio_rdm_records/collections/tasks.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2024 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+"""Collections celery tasks."""
+
+from celery import shared_task
+from flask import current_app
+from invenio_access.permissions import system_identity
+
+from invenio_rdm_records.proxies import (
+    current_rdm_records,
+)
+
+
+@shared_task(ignore_result=True)
+def update_collections_size():
+    """Calculate and update the size of all the collections."""
+    collections_service = current_rdm_records.collections_service
+    res = collections_service.read_all(system_identity, depth=0)
+    for citem in res:
+        try:
+            collection = citem._collection
+            res = collections_service.search_records(system_identity, collection)
+            collections_service.update(
+                system_identity, collection, data={"num_records": res.total}
+            )
+        except Exception as e:
+            current_app.logger.exception(str(e))
--- a/invenio_rdm_records/collections/results.py
+++ b/invenio_rdm_records/collections/results.py
@@ -66,7 +66,7 @@ class CollectionItem(ServiceItemResult):
         res = {
             "root": self._collection.id,
             self._collection.id: {
-                **self._schema.dump(self._collection),
+                **self._schema.dump(self._collection, context={"identity": self._identity}),
                 "children": list(),
                 "links": self._links_tpl.expand(self._identity, self._collection),
             },
@@ -76,7 +76,7 @@ class CollectionItem(ServiceItemResult):
             if _c.id not in res:
                 # Add the subcollection to the dictionary
                 res[_c.id] = {
-                    **self._schema.dump(_c),
+                    **self._schema.dump(_c, context={"identity": self._identity}),
                     "children": list(),
                     "links": self._links_tpl.expand(self._identity, _c),
                 }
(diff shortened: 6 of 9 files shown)


invenio_rdm_records/collections/models.py
```diff
@@ -156,9 +156,30 @@ def get_by_slug(cls, slug, tree_id):
...
-        return cls.query.filter(cls.id.in_(ids)).order_by(cls.path, cls.order)
+        return cls.query.filter(cls.id.in_(ids_)).order_by(cls.path, cls.order)
+
+    @classmethod
+    def read_all(cls):
+        """Get all collections."""
+        return cls.query.order_by(cls.path, cls.order)
+
+    def update(
+        self, /, slug=None, title=None, search_query=None, order=None, num_records=None
```
alejandromumo, line 169: My idea with this is to make the update very explicit, to avoid things being updated without the developer's knowledge. 

Moreover, I don't think some fields should be updated (e.g. `id`, `depth` or `path` since those are computed).

invenio_rdm_records/collections/api.py
```diff
@@ -91,8 +91,21 @@ def resolve(cls, *, id_=None, slug=None, ctree_id=None, depth=2):
     @classmethod
     def resolve_many(cls, ids_=None, depth=2):
         """Resolve many collections by ID."""
-        _ids = ids_ or []
-        return [cls(c, depth) for c in cls.model_cls.read_many(_ids)]
+        if not ids_:
+            return []
```
ntarocco, line 95: Very minor comment: why checking this year, and not delegating it to the caller?
The caller could avoid calling the method if no `ids` given.
Otherwise, it feels like that we should check the values of all params, including `depth`.

invenio_rdm_records/collections/api.py
```diff
@@ -91,8 +91,21 @@ def resolve(cls, *, id_=None, slug=None, ctree_id=None, depth=2):
...
+        return [cls(c, depth) for c in cls.model_cls.read_many(ids_)]
+
+    @classmethod
+    def resolve_all(cls, depth=2):
```
ntarocco, line 99: Any reason why it is called `resolve_all` and/or `read_all` between service, api and models?

invenio_rdm_records/collections/service.py
```diff
@@ -131,3 +162,43 @@ def read_logo(self, identity, slug):
...
+    def search_records(self, identity, collection_or_id, params=None):
+        """Search records in a collection."""
+        params = params or {}
+
+        if isinstance(collection_or_id, int):
+            collection = self.collection_cls.resolve(id_=collection_or_id)
+        else:
+            collection = collection_or_id
+
+        params.update({"collection_id": collection.id})
```
ntarocco, line 193: I am puzzled by fetching the collection query inside the community service [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/community_records/service.py#L71).
It feels like a specific case, and if in the future when need to add another `extra_filter` there, we will copy the pattern and add an extra if/else there.
Why not fetching here the collection query, and injecting it in the `search` as an extra filter?
It also feels like we are fetching multiple times the collection object around.

invenio_rdm_records/collections/service.py
```diff
@@ -131,3 +162,43 @@ def read_logo(self, identity, slug):
         if _exists:
             return url_for("static", filename=logo_path)
         raise LogoNotFoundError()
+
+    def read_many(self, identity, ids_=None, depth=2):
```
ntarocco, line 166: I think it would be better to remove the default. It does not make much sense to have a `read_many` method without passing any id.
```suggestion
    def read_many(self, identity, ids_, depth=2):
```

invenio_rdm_records/collections/models.py
ntarocco, line 169: What is the `/` param? Did you mean `_`?

ntarocco: What about (pseudocde):
```python
    def update(self, **kwargs):
        """A very nice docstring to explain what key/values are expected."""
        allowed_keys = ["slug", ...]
        for key, value in kwargs.items():
          if key in allowed_keys:
            setattr(self, key, value)
```

invenio_rdm_records/collections/models.py
```diff
@@ -156,9 +156,30 @@ def get_by_slug(cls, slug, tree_id):
...
+            self.slug = slug
+        if title is not None:
+            self.title = title
+        if search_query is not None:
+            self.search_query = search_query
+        if order is not None:
+            self.order = order
+        if num_records is not None:
+            self.num_records = num_records
+        return self
```
ntarocco, line 182: The object has been already mutated, I think that it is not very useful to return `self`.

invenio_rdm_records/collections/service.py
```diff
@@ -121,6 +127,31 @@ def add(self, identity, collection, slug, title, query, uow=None, **kwargs):
...
+            identity, "update", community_id=collection.community.id
+        )
+
+        data = data or {}
+
+        valid_data, errors = self.collection_schema.load(
+            data, context={"identity": identity}, raise_errors=True
+        )
+
+        res = collection.update(**valid_data)
```
ntarocco, line 147: The schema definition is already "protecting" what can be updated. Do we need the extra check inside the method to allow only certain attr to be updated?

invenio_rdm_records/collections/models.py
```diff
@@ -156,9 +156,30 @@ def get_by_slug(cls, slug, tree_id):
...
     @classmethod
-    def read_many(cls, ids):
+    def read_many(cls, ids_):
         """Get many collections by ID."""
-        return cls.query.filter(cls.id.in_(ids)).order_by(cls.path, cls.order)
+        return cls.query.filter(cls.id.in_(ids_)).order_by(cls.path, cls.order)
+
+    @classmethod
+    def read_all(cls):
+        """Get all collections."""
```
ntarocco, line 165: Do you mind adding some docstring to make it explicit that it guarantees a certain order of results?

ntarocco (commented):
Reviewed with @ptamarit

invenio_rdm_records/collections/schema.py
```diff
@@ -14,7 +14,8 @@ class CollectionSchema(Schema):
 
     slug = fields.Str()
     title = fields.Str()
-    depth = fields.Int()
+    depth = fields.Int(dump_only=True)
```
kpsherva, line 17: won't this prevent us from creating collections with desired depth via service in the future ?

invenio_rdm_records/collections/schema.py
```diff
@@ -14,7 +14,8 @@ class CollectionSchema(Schema):
...
     order = fields.Int()
-    id = fields.Int()
+    id = fields.Int(dump_only=True)
     num_records = fields.Int()
+    search_query = fields.Str(load_only=True)
```
kpsherva, line 21: search query will be stripped out when dumping, don't we want it visible in the UI in the future? 
Not sure, what was the reason to instroduce this change?

invenio_rdm_records/collections/service.py
```diff
@@ -131,3 +162,43 @@ def read_logo(self, identity, slug):
...
+        """Get many collections."""
+        self.require_permission(identity, "read")
+
+        ids_ = ids_ or []
+        res = self.collection_cls.resolve_many(ids_, depth=depth)
+        return CollectionList(
+            identity, res, self.collection_schema, None, self.links_item_tpl
+        )
+
+    def read_all(self, identity, depth=2):
```
kpsherva, line 176: shouldn't the name be `search` to align with the usual service method naming convention?

invenio_rdm_records/collections/service.py
```diff
@@ -131,3 +162,43 @@ def read_logo(self, identity, slug):
...
+        """Get all collections."""
+        self.require_permission(identity, "read")
+        res = self.collection_cls.resolve_all(depth=depth)
+        return CollectionList(
+            identity, res, self.collection_schema, None, self.links_item_tpl
+        )
+
+    def search_records(self, identity, collection_or_id, params=None):
```
kpsherva, line 184: nit: just for clarity of usage
```suggestion
    def search_collection_records(self, identity, collection_or_id, params=None):
```

invenio_rdm_records/collections/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+
+@shared_task(ignore_result=True)
+def update_collections_size():
+    """Calculate and update the size of all the collections."""
+    collections_service = current_rdm_records.collections_service
+    res = collections_service.read_all(system_identity, depth=0)
+    for citem in res:
+        try:
+            collection = citem._collection
+            res = collections_service.search_records(system_identity, collection)
```
kpsherva, line 24: does it make sense to take advantage of the OS count query instead?

invenio_rdm_records/collections/service.py
```diff
@@ -121,6 +127,31 @@ def add(self, identity, collection, slug, title, query, uow=None, **kwargs):
             identity, new_collection, self.collection_schema, self.links_item_tpl
         )
 
+    @unit_of_work()
+    def update(self, identity, collection_or_id, data=None, uow=None):
+        """Update a collection."""
+        if isinstance(collection_or_id, int):
+            collection = self.collection_cls.resolve(id_=collection_or_id)
+        else:
+            collection = collection_or_id
```
kpsherva, line 136: just for me to understand: in what circumstances do we pass collection, not the id? I think most of the services only accept id, not the obj

invenio_rdm_records/collections/api.py
alejandromumo: Yes, this was done explicitly to separate both.  I expect that  'read` and `resolve` to return very different things (service result item X API object ). 

But re-reading the code we have in other places, `resolve` seems to be used more for resolvers (e.g., entity resolvers), systemfields, and `pid` (where the service method is also named `resolve`). Since we're not in that situation here, I think it can be modified to `read*` all over.

invenio_rdm_records/collections/service.py
alejandromumo: I think this is a really good point and I had the same feeling that it could go just in one place. Plus, collections are not bound to communities at its core so the search should be possible without going through the community records service.

For this change, however, we need a new endpoint inside collections to expose the collection service search method

invenio_rdm_records/collections/models.py
alejandromumo: I modified the method now based on your suggestions above. 

Anyway, I wanted to use **keyword-only** [parameters](https://docs.python.org/3/whatsnew/3.8.html#positional-only-parameters) (`*`) and not `positional-only`. The [idea](https://github.com/inveniosoftware/invenio-rdm-records/pull/1853#discussion_r1808909774) was that if someone uses `collection.update` they need to be explicit of what they want to update, to avoid unexpected errors coming from e.g. swapping the order unintentionally or drilling objects.

Since I now refactored the `update` code to have a `set` of allowed parameters, it accomplishes the same and this was removed.

invenio_rdm_records/collections/service.py
alejandromumo: I prefer the model to be protected because it “knows” which fields are safe to modify. For instance, depth is calculated based on the path during creation, and the path isn’t updated since we don’t rearrange the tree. Currently, these fields aren’t in the schema, but there’s a risk that they could be added in the future, inadvertently introducing bugs. To me, I am not a big fan of relying on Marshmallow because the schema feels “far” from the model and it’s also used in other ways beyond validating what can and cannot be updated.

But I think it's more of a personal fear so I will remove it.

invenio_rdm_records/collections/schema.py
alejandromumo: No, the `depth` of a collection tells you how deep it is in the tree and is a `read-only` field. It is computed based on the `path` when you create a collection. The `path` is the field that matters here since collections are based on the [materialized path pattern](https://www.mongodb.com/docs/manual/tutorial/model-tree-structures-with-materialized-paths/). 

Initially, we added `depth` as a field to improve the performance of read queries.

The service has two methods to create collections:

- `create`: creates a new "root" collection (e.g. no parent) inside a `CollectionTree`. The path is always "empty" since it's a root (`path=','`).  I added a docstring to be more explicit about it. 
- `add`: adds a collection as a child of another one. The path will be computed based on the parent.

invenio_rdm_records/collections/service.py
alejandromumo: **Naming it search**

 I don't agree, I would expect `search` to search in the search engine. Plus `read` (and `read_many`, `read_all`) are also widely used in other services. 

Moreover, `collections_service.search` could also be ambiguous. Does it mean search collections or search records inside the collections? In my opinion the latter makes more sense, so we added `collections_service.search_records` to be explicit about it.

---

**Difference between `read_many` and `read_all`**

Read many is a specific query `SELECT  ... FROM ... WHERE  id in (1,2,3)` while read all means `SELECT ... FROM`. So I split it because:

1 - I want it to be very explicit they do different things 
2- To avoid `ids_=[]` having any implicit meaning. e.g. would it mean fetch all the collections OR fetch 0 collections?

---

**Underling `resolve` methods**

These were renamed to be consistent with the service methods.

invenio_rdm_records/collections/tasks.py
alejandromumo: I am not sure whether it's faster or not (I assume it should be, but to be tested), but is it supported by the other services methods? Searching for records inside a collection ultimately means using the `CommunityRecords` or `RecordsService` search.

invenio_rdm_records/collections/service.py
alejandromumo: This was added to avoid resolving collections twice. I.e. this is useful when chaining service method calls. For the celery task implemented here, I need to read all the collections and then search each one of them. 

The flow of having only an `id` and resolving the entity inside works well when in a request context and when all the logic is contained within one service method.

An alternative would be having a service method that does all of that (e.g. `search_all_collections`), but it's not as granular and reusable as I would expect from a service method.

invenio_rdm_records/collections/service.py
alejandromumo: We discussed IRL and we were on the same page: the collection's records search is better fitted in the collection service. Therefore, I added a new commit with the required changes. 

In order not to block the development, both commits of this PR are working so we can split them if needed.

jrcastro2 (approved):
Peer reviewed with @0einstein0 LGTU! 🚀 
We just left a just a minor question

invenio_rdm_records/collections/services/service.py
```diff
@@ -131,3 +165,47 @@ def read_logo(self, identity, slug):
...
+
+    def read_all(self, identity, depth=2):
+        """Get all collections."""
+        self.require_permission(identity, "read")
+        res = self.collection_cls.read_all(depth=depth)
+        return CollectionList(
+            identity, res, self.collection_schema, None, self.links_item_tpl
+        )
+
+    def search_collection_records(self, identity, collection_or_id, params=None):
```
ntarocco, line 192:
```suggestion
    def search_records(self, identity, collection_or_id, params=None):
```

ntarocco (approved):
Reviewed with ❤️ by Nico and Carlin.

We think that you could merge, and eventually improve/fix things in a subsequent PR.

ntarocco:
Before merging, please fix the tests.

invenio_rdm_records/collections/api.py
```diff
@@ -89,10 +89,21 @@ def resolve(cls, *, id_=None, slug=None, ctree_id=None, depth=2):
...
+
+    @classmethod
+    def read_all(cls, depth=2):
+        """Read all collections."""
+        return [cls(c, depth) for c in cls.model_cls.read_all()]
+
+    def update(self, **kwargs):
+        """Update the collection."""
+        if "search_query" in kwargs:
+            Collection.validate_query(kwargs["search_query"])
```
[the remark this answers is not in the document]
alejandromumo: created an issue: https://github.com/inveniosoftware/invenio-rdm-records/issues/1867

invenio_rdm_records/collections/services/service.py
```diff
@@ -121,6 +130,31 @@ def add(self, identity, collection, slug, title, query, uow=None, **kwargs):
             identity, new_collection, self.collection_schema, self.links_item_tpl
         )
 
+    @unit_of_work()
+    def update(self, identity, collection_or_id, data=None, uow=None):
+        """Update a collection."""
+        if isinstance(collection_or_id, int):
+            collection = self.collection_cls.read(id_=collection_or_id)
```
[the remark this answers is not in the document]
alejandromumo: Created an issue: https://github.com/inveniosoftware/invenio-rdm-records/issues/1865

invenio_rdm_records/collections/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+
+@shared_task(ignore_result=True)
+def update_collections_size():
+    """Calculate and update the size of all the collections."""
+    collections_service = current_rdm_records.collections_service
+    res = collections_service.read_all(system_identity, depth=0)
+    for citem in res:
+        try:
+            collection = citem._collection
+            res = collections_service.search_collection_records(
```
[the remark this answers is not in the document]
alejandromumo: Created an issue: https://github.com/inveniosoftware/invenio-rdm-records/issues/1866

invenio_rdm_records/collections/services/service.py
alejandromumo: That was the original name, I modified it after reviews :(

invenio_rdm_records/collections/services/service.py
```diff
@@ -131,3 +165,47 @@ def read_logo(self, identity, slug):
         if _exists:
             return url_for("static", filename=logo_path)
         raise LogoNotFoundError()
+
+    def read_many(self, identity, ids_, depth=2):
```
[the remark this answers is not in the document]
alejandromumo: Good point. 

IMO it's useful to have it alongside `read_all` to make a clear distinction since the underlying queries are distinct and have performance differences. 

Therefore I prefer it stay there from early on.

PR #1858, opened 2024-10-25
metadata: add internal notes

kpsherva:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/227
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/240

### DIFF
--- a/invenio_rdm_records/config.py
+++ b/invenio_rdm_records/config.py
@@ -13,11 +13,12 @@
 from datetime import timedelta
 
 import idutils
-from invenio_administration.permissions import administration_permission
+from invenio_access.permissions import system_permission
 from invenio_i18n import lazy_gettext as _
 from invenio_records_resources.services.records.queryparser import QueryParser
 from invenio_records_resources.services.records.queryparser.transformer import (
     RestrictedTerm,
+    RestrictedTermValue,
     SearchFieldTransformer,
 )
 
@@ -271,12 +272,12 @@ RDM_SEARCH = {
     ],
     "query_parser_cls": QueryParser.factory(
         mapping={
-            "internal_notes.note": RestrictedTerm(administration_permission),
-            "internal_notes.id": RestrictedTerm(administration_permission),
-            "internal_notes.added_by": RestrictedTerm(administration_permission),
-            "internal_notes.timestamp": RestrictedTerm(administration_permission),
-            "_exists_": RestrictedTerm(
-                administration_permission, word=word_internal_notes
+            "internal_notes.note": RestrictedTerm(system_permission),
+            "internal_notes.id": RestrictedTerm(system_permission),
+            "internal_notes.added_by": RestrictedTerm(system_permission),
+            "internal_notes.timestamp": RestrictedTerm(system_permission),
+            "_exists_": RestrictedTermValue(
+                system_permission, word=word_internal_notes
             ),
         },
         tree_transformer_cls=SearchFieldTransformer,
--- a/invenio_rdm_records/config.py
+++ b/invenio_rdm_records/config.py
@@ -13,7 +13,13 @@
 from datetime import timedelta
 
 import idutils
+from invenio_administration.permissions import administration_permission
 from invenio_i18n import lazy_gettext as _
+from invenio_records_resources.services.records.queryparser import QueryParser
+from invenio_records_resources.services.records.queryparser.transformer import (
+    RestrictedTerm,
+    SearchFieldTransformer,
+)
 
 from . import tokens
 from .resources.serializers import DataCite43JSONSerializer
@@ -21,6 +27,7 @@ from .services import facets
 from .services.config import lock_edit_published_files
 from .services.permissions import RDMRecordPermissionPolicy
 from .services.pids import providers
+from .services.queryparser import word_internal_notes
 
 # Invenio-RDM-Records
 # ===================
@@ -251,6 +258,7 @@ RDM_SORT_OPTIONS = {
 
 """
 
+
 RDM_SEARCH = {
     "facets": ["access_status", "file_type", "resource_type"],
     "sort": [
@@ -261,6 +269,18 @@ RDM_SEARCH = {
         "mostviewed",
         "mostdownloaded",
     ],
+    "query_parser_cls": QueryParser.factory(
+        mapping={
+            "internal_notes.note": RestrictedTerm(administration_permission),
+            "internal_notes.id": RestrictedTerm(administration_permission),
+            "internal_notes.added_by": RestrictedTerm(administration_permission),
+            "internal_notes.timestamp": RestrictedTerm(administration_permission),
+            "_exists_": RestrictedTerm(
+                administration_permission, word=word_internal_notes
+            ),
+        },
+        tree_transformer_cls=SearchFieldTransformer,
+    ),
 }
 """Record search configuration.
 
--- a/invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
@@ -341,6 +341,22 @@
               }
             }
           }
+        },
+        "_internal_notes": {
+            "properties": {
+              "note": {
+                "type": "string"
+              },
+              "timestamp": {
+                "type": "string",
+                "description": "ISO8601 formatted timestamp in UTC.",
+                "format": "date-time"
+              },
+              "user": {
+                "type": "string",
+                "description": "User id of the person who created the note"
+              }
+            }
         }
       }
     },
--- a/invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
+++ b/invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
@@ -341,22 +341,6 @@
               }
             }
           }
-        },
-        "_internal_notes": {
-            "properties": {
-              "note": {
-                "type": "string"
-              },
-              "timestamp": {
-                "type": "string",
-                "description": "ISO8601 formatted timestamp in UTC.",
-                "format": "date-time"
-              },
-              "user": {
-                "type": "string",
-                "description": "User id of the person who created the note"
-              }
-            }
         }
       }
     },
@@ -392,6 +376,24 @@
         "is_visible": {
           "type": "boolean",
           "description": "Whether or not the tombstone page is publicly visible."
+        }
+      }
+    },
+    "internal_notes": {
+      "properties": {
+        "id": {
+          "type": "string"
+        },
+        "note": {
+          "type": "string"
+        },
+        "timestamp": {
+          "type": "string",
+          "description": "ISO8601 formatted timestamp in UTC.",
+          "format": "date-time"
+        },
+        "added_by": {
+          "$ref": "local://records/definitions-v2.0.0.json#/agent"
         }
       }
     },
--- a/invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
+++ b/invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
@@ -45,8 +45,13 @@
         "accent_analyzer": {
           "tokenizer": "standard",
           "type": "custom",
-          "char_filter": ["strip_special_chars"],
-          "filter": ["lowercase", "asciifolding"]
+          "char_filter": [
+            "strip_special_chars"
+          ],
+          "filter": [
+            "lowercase",
+            "asciifolding"
+          ]
         }
       }
     }
@@ -702,19 +707,6 @@
           "_default_preview": {
             "type": "object",
             "enabled": false
-          },
-          "_internal_notes": {
-            "properties": {
-              "note": {
-                "type": "text"
-              },
-              "timestamp": {
-                "type": "date"
-              },
-              "user": {
-                "type": "keyword"
-              }
-            }
           },
           "contact": {
             "type": "keyword"
@@ -1593,6 +1585,28 @@
           },
           "is_visible": {
             "type": "boolean"
+          }
+        }
+      },
+      "internal_notes": {
+        "type": "object",
+        "properties": {
+          "id": {
+            "type": "keyword"
+          },
+          "note": {
+            "type": "text"
+          },
+          "timestamp": {
+            "type": "date"
+          },
+          "added_by": {
+            "type": "object",
+            "properties": {
+              "user": {
+                "type": "keyword"
+              }
+            }
           }
         }
       }
--- a/invenio_rdm_records/services/components/internal_notes.py
+++ b/invenio_rdm_records/services/components/internal_notes.py
@@ -9,6 +9,7 @@
 import uuid
 from copy import copy, deepcopy
 from datetime import datetime, timezone
+
 from invenio_drafts_resources.services.records.components import ServiceComponent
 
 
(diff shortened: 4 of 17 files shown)


invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
```diff
@@ -341,6 +341,22 @@
               }
             }
           }
+        },
+        "_internal_notes": {
```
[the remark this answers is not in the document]
kpsherva: because of https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/mappings/v7/rdmrecords/records/record-v6.0.0.json#L654
it will be still discussed, that's why it is a draft PR, too early to review since it is not the final form of the code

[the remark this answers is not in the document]
utnapischtim: thanks :)

invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -351,6 +353,13 @@ class FeatureSchema(Schema):
     features = fields.List(fields.Nested(LocationSchema))
 
 
+class InternalNoteSchema(Schema):
+    """Schema for internal notes."""
+    timestamp = TZDateTime(timezone=timezone.utc, format="iso", dump_only=True)
+    user = fields.Nested(Agent, dump_only=True)
+    note = SanitizedUnicode()
```
lnielsen, line 360: html

invenio_rdm_records/services/config.py
```diff
@@ -231,6 +239,11 @@ class RDMSearchOptions(SearchOptions, SearchOptionsMixin):
         MetricsParam,
     ]
 
+    query_parser_cls = QueryParser.factory(
```
kpsherva, line 242: for reviewers: I find this very confusing to have to define the query_parser_cls both in `RDM_SEARCH` config variable and here. Does anyone know the reason for this config being spread in many places?

invenio_rdm_records/services/schemas/record.py
```diff
@@ -36,6 +44,20 @@ def validate_scheme(scheme):
         raise ValidationError(_("Invalid persistent identifier scheme."))
 
 
+class InternalNoteSchema(Schema):
+    """Schema for internal notes."""
+
+    id = SanitizedUnicode(load_default=lambda: RecordIdProviderV2.generate_id())
```
kpsherva, line 50: I have doubts about using the RecordIDProvider in the schema of the record, maybe someone has an idea for something more elegant
I need to identify the notes within the record, otherwise there is no way of keeping track

invenio_rdm_records/config.py
```diff
@@ -261,6 +269,18 @@ def always_valid(identifier):
         "mostviewed",
         "mostdownloaded",
     ],
+    "query_parser_cls": QueryParser.factory(
+        mapping={
+            "internal_notes.note": RestrictedTerm(administration_permission),
```
kpsherva, line 274: I would gladly align on using the correct permission here, I used admin for self explaining PR

invenio_rdm_records/services/components/internal_notes.py
```diff
@@ -0,0 +1,64 @@
...
+        record.update({"internal_notes": notes})
+
+    def update_draft(self, identity, data=None, record=None, **kwargs):
+        """Inject parsed metadata to the record."""
+        notes_updated = data.get(self.field, [])
+        notes = deepcopy(record.get(self.field, []))
+        existing_ids = set([note["id"] for note in notes])
+        new_ids = set([note["id"] for note in notes_updated])
+        to_delete = existing_ids - new_ids
+        if sorted(existing_ids) != sorted(new_ids):
```
kpsherva, line 40: there is a bit of manual checks since the field was designed as a list, each new or removed element need to be checked by its unique id

ntarocco: With the discussion we had IRL, the `id` was not needed. Why adding it now?
If not needed, it might be confusing.

invenio_rdm_records/services/components/internal_notes.py
```diff
@@ -0,0 +1,64 @@
...
+class InternalNotesComponent(MetadataComponent):
+    """Service component for custom fields."""
+
+    field = "internal_notes"
+    new_version_skip_fields = []
+
+    def create(self, identity, data=None, record=None, **kwargs):
+        """Inject note to the record."""
+        notes = data.get(self.field, [])
+        for note in notes:
```
ntarocco, line 28: Given that it is a new record, do we need to use the `setdefault`?
```python
 record["internal_notes"] = [{
    "added_by": {"user": identity.id}),
    "timestamp": datetime.now(timezone.utc).isoformat(),
    "note": note["note"],
    } for note in data.data.get(self.field, [])
] 
```

invenio_rdm_records/config.py
```diff
@@ -261,6 +269,18 @@ def always_valid(identifier):
...
+            "internal_notes.id": RestrictedTerm(administration_permission),
+            "internal_notes.added_by": RestrictedTerm(administration_permission),
+            "internal_notes.timestamp": RestrictedTerm(administration_permission),
+            "_exists_": RestrictedTerm(
```
ntarocco, line 278: this requires some explanations :)

invenio_rdm_records/services/queryparser.py
```diff
@@ -0,0 +1,16 @@
...
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Query Parser module for InvenioRdmRecords."""
+from luqum.tree import Word
+
+
+def word_internal_notes(node):
+    """Quote DOIs."""
```
ntarocco, line 13: Wrong docstring?

invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -40,6 +39,8 @@
 from marshmallow_utils.schemas import GeometryObjectSchema, IdentifierSchema
 from werkzeug.local import LocalProxy
 
+from invenio_rdm_records.services.schemas.parent.access import Agent
```
ntarocco, line 42: Unused?

invenio_rdm_records/services/schemas/record.py
ntarocco: See my comment above

tests/services/components/test_internal_notes_component.py
```diff
@@ -0,0 +1,105 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# Invenio-RDM-Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Tests for the service Metadata component."""
```
ntarocco, line 8: I don't know if here or in [invenio-records-resources](https://github.com/inveniosoftware/invenio-records-resources/pull/599/files), but we need to see tests for searching text that is in the `internal_notes` field, without specifying the field name in the search query.

invenio_rdm_records/config.py
kpsherva: agreed with @ntarocco to settle on super-user permission for now

invenio_rdm_records/services/components/internal_notes.py
kpsherva: we had another discussion which we concluded that id is needed to identify the note, the position in the list is not enough (f.e. when one note is deleted and a new added in the same request)

ntarocco: What we discussed is that if you mark the deletion sending an empty element, you can rely on the list position:
```
[
    {},   # means delete
    { new one or updated }
]

File: invenio_rdm_records/config.py (hunk 8 above)
Comment (kpsherva) [reply invenio_rdm_records/config.py:278]: I am not sure if you added the comment before we talked or after :) we can rediscuss if it was after!

File: tests/services/components/test_internal_notes_component.py (hunk 11 above)
Comment (kpsherva) [reply tests/services/components/test_internal_notes_component.py:8]: done, in https://github.com/inveniosoftware/invenio-rdm-records/pull/1858/files#diff-fa2dc1cd0a4a4556f57462d4044d5a57208aec2455d7a237e84686641a38518dR951

File: invenio_rdm_records/services/components/internal_notes.py (hunk 6 above)
Comment (kpsherva) [reply invenio_rdm_records/services/components/internal_notes.py:40]: `id` is needed also for identifying which note is new, which already exists, to propagate the timestamp and user on the existing ones (when updating drafts)

Comment (ntarocco) [reply invenio_rdm_records/services/components/internal_notes.py:42]: > `id` is needed also for identifying which note is new, which already exists, to propagate the timestamp and user on the existing ones (when updating drafts)

Not really, as the client will have to send the timestamp and creator, and can be compared with the existing.
It all boils down to what contract you want to have for this API.
The `id` is really necessary only if we need to support re-ordering of the list in conjunction with a insert/update/delete operation.

I am ok with an id, if you can ensure that a previous id cannot be re-used.

File: invenio_rdm_records/services/components/internal_notes.py [hunk 12]
```diff
@@ -0,0 +1,66 @@
...
+
+    def update_draft(self, identity, data=None, record=None, **kwargs):
+        """Inject parsed metadata to the record."""
+        notes_updated = data.get(self.field, [])
+        notes = deepcopy(record.get(self.field, []))
+        existing_ids = set([note["id"] for note in notes])
+        new_ids = set([note["id"] for note in notes_updated])
+        to_delete = existing_ids - new_ids
+
+        if sorted(existing_ids) != sorted(new_ids):
```
zzacharo, line 42: why sorted here?

invenio_rdm_records/services/components/internal_notes.py
```diff
@@ -0,0 +1,66 @@
...
+            for note in notes_updated:
+                if note["id"] in existing_ids:
```
zzacharo, line 44: I understand that the new list is of:
```
new_notes = [{id: 'existing1', note: 'blabla'}, {note: 'new note'}, ....]
```

if the above is correct then we cannot assume that `note['id']` is valid for the new notes.

tests/services/components/test_internal_notes_component.py
```diff
@@ -0,0 +1,105 @@
...
+    # check if internal notes have correct attributes
+    component.update_draft(identity_simple, data=new_data, record=draft)
+    assert "internal_notes" in draft
+    notes = draft["internal_notes"]
+    assert "id" in notes[0]
+    assert "timestamp" in notes[0]
+    assert "added_by" in notes[0]
+
+    # check if new internal note is added properly
+    new_data["internal_notes"] = notes + [{"note": "def", "id": "frgt-09df"}]
```
zzacharo, line 39: Here I think the test is wrong because it assumes that a new note will have an `id` unless I am mistaken

invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
```diff
@@ -379,6 +379,24 @@
         }
       }
     },
+    "internal_notes": {
```
ntarocco, line 382: why changing v6?

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
```diff
@@ -1595,6 +1587,28 @@
             "type": "boolean"
           }
         }
+      },
+      "internal_notes": {
```
ntarocco, line 1591: is it outside the `metadata` object in the end? it is a bit hard to see if from the diff here

invenio_rdm_records/services/components/internal_notes.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def create(self, identity, data=None, record=None, **kwargs):
+        """Inject note to the record."""
+        notes = data.get(self.field, [])
+        for note in notes:
+            note.update(
+                {
+                    "added_by": {"user": identity.id},
+                    "timestamp": datetime.now(timezone.utc).isoformat(),
+                    "id": str(uuid.uuid4()),
```
ntarocco, line 28: do we need such a large UUID here? 16 chars?
wouldn't it a 4 chars id enough, or similar?

```python
def _get_node_id(...):
    """Generate a random short id."""
    # import secrets
    return secrets.token_hex(3)
```

invenio_rdm_records/services/components/internal_notes.py
```diff
@@ -0,0 +1,68 @@
...
+                }
+            )
+        record.update({"internal_notes": notes})
+
+    def update_draft(self, identity, data=None, record=None, **kwargs):
+        """Inject parsed metadata to the record."""
+        notes_to_update = data.get(self.field, [])
```
zzacharo, line 35:
```suggestion
        updated_notes = deepcopy(data.get(self.field, []))
```
Becase data is passed around I would probably deepcopy it.

invenio_rdm_records/services/components/internal_notes.py
```diff
@@ -0,0 +1,68 @@
...
+        notes = deepcopy(record.get(self.field, []))
```
zzacharo, line 36:
```suggestion
        existing_notes = deepcopy(record.get(self.field, []))
```

invenio_rdm_records/services/components/internal_notes.py
```diff
@@ -0,0 +1,68 @@
...
+        ids_to_check = [note["id"] for note in notes]
+
+        # reverse list to keep proper iterator reference after remove
+        for note in reversed(notes_to_update):
+            note.setdefault("id", str(uuid.uuid4()))
+            if note["id"] in ids_to_check:
+                # exclude existing IDs from incoming data
+                # because we don't modify existing notes
+                notes_to_update.remove(note)
+                ids_to_check.remove(note["id"])
```
zzacharo, line 46: Maybe a note here that we remove the ids so we identify the in the end the ones need to be deleted?

zzacharo (approved):
Some minor comments, you can merge afterwards!

invenio_rdm_records/services/components/internal_notes.py
```diff
@@ -0,0 +1,68 @@
...
+                continue
```
ntarocco, line 47: are we simply skipping previously existing notes here? 

I think that it is not very clear the contract of the API. If I understand correctly, we only accept:
1. insert of new notes: the client will only send the notes text (no `id` sent or other fields)
2. deletion
We don't accept updates. In any case, the client should send ALL existing notes.

I feel like it is a bit too complicated the code here, and not easy to capture the constraints here. I suggest checking if there is an easier way to implement it, and add docstring.

invenio_rdm_records/services/config.py
```diff
@@ -64,6 +65,13 @@
     PaginationParam,
     QueryStrParam,
 )
+from invenio_records_resources.services.records.queryparser import (
+    QueryParser,
+    SearchFieldTransformer,
+)
+from invenio_records_resources.services.records.queryparser.transformer import (
+    RestrictedTerm,
```
ntarocco, line 73: Unused imports?

invenio_rdm_records/services/permissions.py
```diff
@@ -61,6 +61,9 @@ class RDMRecordPermissionPolicy(RecordPermissionPolicy):
         "object-read": "read_files",
     }
 
+    # permission meant for global curators of the instance
+    # (for now applies to internal notes field only
+    can_manage_internal = [Administration()]
```
ntarocco, line 66: I believe we should not use `Administration`, there is no global curator, and it is not meant for that. The comment is misleading.
This was discussed in IRL. Why changing it back?

invenio_rdm_records/services/components/internal_notes.py
kpsherva: lets discuss this together with Zach, he suggested the uuid for consistency. I used initially the cool pid.

invenio_rdm_records/services/components/internal_notes.py
kpsherva: I didn't go with "updated_notes" since past tense is not true until we actually reach the record.commit and it didn't sit well semantially. I hope you don't mind if I leave the name as is.

invenio_rdm_records/records/mappings/os-v2/rdmrecords/records/record-v7.0.0.json
kpsherva: yes, it is outside

invenio_rdm_records/records/jsonschemas/records/record-v6.0.0.json
slint: The JSONSchema (and datamodel) is still on v6, though the mapping was bumped because of the incompatible change in the indexing analyzers.

We can have the two being in different versions (although it's a bit confusing). In the end it doesn't affect indexing, since in the Record API class these can be separately specified.

ntarocco: Ah sorry! I confused it with the mapping there. Thanks!

invenio_rdm_records/services/components/internal_notes.py
kpsherva: added docstring as discussed

ntarocco (approved):

PR #1860, opened 2024-10-29
assets: allow multiple DndProviders

egabancho:
* If the context is not set to window we can end up in a situation where multiple backends are instantiated resulting in an error. More info https://github.com/react-dnd/react-dnd/issues/3257

### Description

### Checklist

- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect](https://github.com/orgs/inveniosoftware/teams/architects/members).

**Frontend**

**Reminder**

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
@@ -71,7 +71,7 @@ class CreatibutorsFieldForm extends Component {
       error || (creatibutorsList === formikInitialValues && initialError);
 
     return (
-      <DndProvider backend={HTML5Backend}>
+      <DndProvider backend={HTML5Backend} context={window}>
         <Form.Field
           required={schema === "creators"}
           className={creatibutorsError ? "error" : ""}
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/License/LicenseField.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/License/LicenseField.js
@@ -74,7 +74,7 @@ class LicenseFieldForm extends Component {
     const uiRights = getIn(values, uiFieldPath, []);
 
     return (
-      <DndProvider backend={HTML5Backend}>
+      <DndProvider backend={HTML5Backend} context={window}>
         <Form.Field required={required}>
           <FieldLabel htmlFor={fieldPath} icon={labelIcon} label={label} />
           <List>


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
```diff
@@ -71,7 +71,7 @@ class CreatibutorsFieldForm extends Component {
       error || (creatibutorsList === formikInitialValues && initialError);
 
     return (
-      <DndProvider backend={HTML5Backend}>
+      <DndProvider backend={HTML5Backend} context={window}>
```
zzacharo, line 74: Hey, just for me to understand: did you encounter the issue? I am asking because I see [here](https://github.com/react-dnd/react-dnd/blob/7c88c37489a53b5ac98699c46a506a8e085f1c03/packages/react-dnd/src/core/DndProvider.tsx#L92C51-L92C57) that if no context passed then it should use by default window.

egabancho: Yes, we are re-arranging the deposit form, still using invenio-rdm-records' components directly. As soon as we add two components that use the `DndProvier` (Creatibutors and Licenses, for example) it raises the error.
We are building the form dynamically upon resource type selection, that might be the reason why we don't see the error on the "regular: deposit form.

I looked at the code before making the PR and I don't fully understand why it fails, you are right that if not set it should `window` already ... 🤷‍♂️

egabancho: You got it, https://gitlab.com/egabancho/invenio-ubiquity/-/tree/custom-deposit-form/site/up/assets/semantic-ui/js/up

zzacharo: I checked briefly but I cannot detect something wrong without really digging into the internals of the library... I think it is worth trying to understand it before we assign `window` there and be afraid to remove it afterwards :)

egabancho: You're definitely right. I believe I jumped the gun on this one. Something is going on when I try to change the deposit form (and the value of `context` is indeed `window` already), but I'm not sure what.
I'm going to close this one, but I will report back what I find in case some poor soul has to go through the same process I am going through right now :wink:

Thanks @zzacharo!

PR #1868, opened 2024-11-04
semantic-ui: CommunitySelectionModal: Remove unused Footer component

sakshamarora1:
Fixes: https://github.com/inveniosoftware/invenio-communities/issues/1209


zzacharo (approved):

0einstein0 (approved):
Reviwed with @sakshamarora1 👀

zzacharo (commented):
There are more places of this being hardcoded, see for example [here](https://github.com/inveniosoftware/invenio-app-rdm/blob/35165ab5a0509dd7b7e83bad0af8e7c28b240297/invenio_app_rdm/users_ui/templates/semantic-ui/invenio_app_rdm/users/header.html#L33)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/components/CommunitySelectionModal/CommunitySelectionFooter.js
```diff
@@ -17,7 +17,7 @@ export const CommunitySelectionFooter = () => {
           <p>
             <Trans>
               Did not find a community that fits you? Upload without a community or{" "}
-              <a href="/communities/new">create your own.</a>
+              <a href="/communities-new">create your own.</a>
```
zzacharo, line 20: also all these places should be changed to utilize the config rather the hardcoded values

sakshamarora1: It's a javascript file, I'm not sure if we are already injecting config?

PR #1872, opened 2024-11-08
community submission: remove notification about request auto-accept

anikachurilova:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/203

### DIFF
--- a/invenio_rdm_records/requests/community_submission.py
+++ b/invenio_rdm_records/requests/community_submission.py
@@ -8,6 +8,7 @@
 
 """Community submission request."""
 
+from invenio_access.permissions import system_identity
 from invenio_drafts_resources.services.records.uow import ParentRecordCommitOp
 from invenio_i18n import lazy_gettext as _
 from invenio_notifications.services.uow import NotificationOp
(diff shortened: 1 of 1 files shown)


invenio_rdm_records/requests/community_submission.py
```diff
@@ -81,13 +82,16 @@ def execute(self, identity, uow):
         # Publish the record
         # TODO: Ensure that the accepting user has permissions to publish.
         service.publish(identity, draft.pid.pid_value, uow=uow)
-        uow.register(
-            NotificationOp(
-                CommunityInclusionAcceptNotificationBuilder.build(
-                    identity=identity, request=self.request
+
+        # don't send notification about request auto-accept
+        if identity != system_identity:
```
zzacharo, line 87: discussion: I am thinking if it makes sense to check the `community.settings.curation_policy` instead so we are more specific.

anikachurilova: If I understand correctly, then you probably mean `review_policy` (correct me if I'm wrong). But in this case we would also have to check that identity is a member of the community, because for non-members, even if `review_policy` is open - they still have to go through the review process.

zzacharo: `Review policy` yes sorry. What I meant is basically to check that if the review policy is open then we will automatically accept so only in that case skip the notification. I am thinking of action centric approach versus identity centric.

zzacharo: hmm I see...thanks for the explanations! Maybe then is better to keep this implementation for now until we have an actual case.

ntarocco: As discussed in IRL, let's merge so it will stop failing, but then we will fix it in a different way. More explanations in a different way.

ntarocco (approved):

PR #1873, opened 2024-11-08
Identifiers validation

kpsherva:
DEPENDS ON https://github.com/inveniosoftware/marshmallow-utils/pull/92/files

### DIFF
--- a/invenio_rdm_records/services/schemas/metadata.py
+++ b/invenio_rdm_records/services/schemas/metadata.py
@@ -33,6 +33,7 @@ from marshmallow_utils.fields import (
     EDTFDateString,
     EDTFDateTimeString,
     IdentifierSet,
+    IdentifierValueSet,
     SanitizedHTML,
     SanitizedUnicode,
 )
@@ -369,7 +370,7 @@ class MetadataSchema(Schema):
     dates = fields.List(fields.Nested(DateSchema))
     languages = fields.List(fields.Nested(VocabularySchema))
     # alternate identifiers
-    identifiers = IdentifierSet(
+    identifiers = IdentifierValueSet(
         fields.Nested(
             partial(IdentifierSchema, allowed_schemes=record_identifiers_schemes)
         )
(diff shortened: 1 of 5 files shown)


invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -33,6 +33,7 @@
     EDTFDateString,
     EDTFDateTimeString,
     IdentifierSet,
+    IdentifierValueSet,
```
ntarocco, line 36: You can remove the unused `IdentifierSet` above

kpsherva: it is used in authors identifiers. There is makes sense to keep the deduplication by schema (as we said yesterday)

tmorrell:
I'm not opposed to this change, but more curious why related identifiers wouldn't work for this use case. All our uses of this field benefited from the uniqueness requirement (record identifiers from legacy system, sample identifiers, external DOIs that are on multiple records), and removing it makes it a bit more likely that people misuse the field.

zzacharo:
> I'm not opposed to this change, but more curious why related identifiers wouldn't work for this use case. All our uses of this field benefited from the uniqueness requirement (record identifiers from legacy system, sample identifiers, external DOIs that are on multiple records), and removing it makes it a bit more likely that people misuse the field.

Hey @tmorrell thanks for your input! I guess technically that is feasible but conceptually if I understand correctly the difference between alternate identifiers and related identifies is that first is describing equally the resource that is being registered. We have cases where people have obtained for example multiple DOIs for their paper and they would like to insert all of them. I understand your concern about the field being misused but I am not sure that the way to control it is by eliminating some realistic cases such as the multiple DOIs for the same resource. Would you think that this should be configurable?

ntarocco:
@zzacharo Tom has a good point: multiple DOIs could be covered by the related identifiers field. However, @tmorrell another use case that we need to cover is having multiple "report numbers" (internal IDs) for a given publication, created internally at CERN, see for example [here](https://cds.cern.ch/record/2916297).
In general, we don't see any drawback of relaxing the constraint and allowing multiple IDs with the same scheme, to better represent real-life cases.

PR #1898, opened 2024-12-05
datacite: fix funding serialization for optional award fields

slint:
* Makes sure that we handle missing values for optional award fields
  like "title" and "number".


tmorrell (approved):
Yup, I forgot that folks could just provide a title now.

PR #1926, opened 2025-01-13
fix: use uow not self.uow

utnapischtim:
* fix: use uow not self.uow

### DIFF
--- a/invenio_rdm_records/services/communities/moderation.py
+++ b/invenio_rdm_records/services/communities/moderation.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2024 CERN.
+# Copyright (C) 2025 Graz University of Technology.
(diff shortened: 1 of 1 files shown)


slint (approved):
Fine with the change, not sure though what in the end should be the convention we follow for passing the UoW to components, i.e. rely on class attribute and `self.uow` vs. explicit parameter.

I somewhat like the class attribute since it makes the method signatures smaller/simpler and more focused on the action.

utnapischtim:
the problem is, that `self.uow` is not set in this class. the new base class doesn't handle the case to set `self.uow`.

slint:
> the problem is, that `self.uow` is not set in this class. the new base class doesn't handle the case to set `self.uow`.

Thanks, I didn't check the context and thought this was a component class 😅 

In that case, I guess it's necessary so this is good to merge :)

PR #1927, opened 2025-01-14
pids: improve UI for optional DOI

zzacharo:
closes https://github.com/CERNDocumentServer/cds-rdm/issues/303


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/components/UnmanagedIdentifierCmp.js
```diff
@@ -0,0 +1,80 @@
...
+ * the unmanaged PID.
+ */
+export class UnmanagedIdentifierCmp extends Component {
+  constructor(props) {
+    super(props);
+
+    const { identifier } = props;
+
+    this.state = {
+      localIdentifier: identifier,
```
jrcastro2, line 24: Question: I might be missing something, but I don't see added value on adding the usage of `localIdentifier`, couldn't we rely on the prop `identifier` alone?

jrcastro2 (approved):
Some of the copyright headers are not updated to 2025, besides that the PR looks good to me, some logic is a bit too complex to understand in a first look but I don't have better proposals.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/components/UnmanagedIdentifierCmp.js
zzacharo: I don't think the component would be updated i.e. there will not be a state update to force a rerender if we rely only on the `props.identifier`.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/PIDFieldCmp.js
```diff
@@ -0,0 +1,70 @@
...
+    super(props);
+
+    this.validatePropValues();
+  }
+
+  validatePropValues = () => {
+    const { canBeManaged, canBeUnmanaged, fieldPath } = this.props;
+
+    if (!canBeManaged && !canBeUnmanaged) {
+      throw Error(`${fieldPath} must be managed, unmanaged or both.`);
```
ntarocco, line 28: Minor: translations here?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/OptionalPIDField.js
```diff
@@ -0,0 +1,243 @@
...
+      required,
+      unmanagedHelpText,
+      pidType,
+      field,
+      record,
+      optionalDOItransitions,
+      setNoINeedDOI,
+    } = this.props;
+
+    let { doiDefaultSelection } = this.props;
```
ntarocco, line 89: Trivial: you can use `const` instead, and below
```suggestion
    const { doiDefaultSelection } = this.props;
```

invenio_rdm_records/services/components/pids.py
```diff
@@ -22,68 +22,40 @@
 from ..pids.tasks import register_or_update_pid
 
 
-class PIDsComponent(ServiceComponent):
-    """Service component for PIDs."""
+def _get_optional_doi_transitions(record):
+    """Reusable method to validate optional DOI."""
+    RDM_OPTIONAL_DOI_TRANSITIONS = current_app.config.get(
```
ntarocco, line 27: Classic comment from me... do we want to use `.get(` here? Or fail if not provided?
If it is OK to be optional, the code should work fine if I set the config to `{}`. Otherwise, let's use `[<key>]` instead :)

ntarocco (commented):
👏🏼

invenio_rdm_records/services/components/pids.py
zzacharo: you are right, I changed that

PR #1932, opened 2025-01-22
optional-doi: fix new upload disabled states

zzacharo:
On new upload, unreserving a local DOI was not re-enabling all options. This PR refactors the relevant code to fix that.

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/components/OptionalDOIoptions.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/components/OptionalDOIoptions.js
@@ -8,13 +8,14 @@ import { i18next } from "@translations/invenio_rdm_records/i18next";
 
 import PropTypes from "prop-types";
 import React, { Component } from "react";
+import { connect } from "react-redux";
 import { Form, Radio, Popup } from "semantic-ui-react";
-import _isEmpty from "lodash/isEmpty";
+import { DepositStatus } from "../../../../state/reducers/deposit";
 
 /**
  * Manage radio buttons choices between managed (i.e. datacite), unmanaged (i.e. external) and no need for a PID.
  */
-export class OptionalDOIoptions extends Component {
+class OptionalDOIoptionsCmp extends Component {
   handleChange = (e, { value }) => {
     const { onManagedUnmanagedChange } = this.props;
     const isManagedSelected = value === "managed";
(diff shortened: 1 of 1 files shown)


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/components/OptionalDOIoptions.js
```diff
@@ -23,24 +23,36 @@ export class OptionalDOIoptions extends Component {
...
+    const {
+      isManagedSelected,
+      isNoNeedSelected,
+      pidLabel,
+      optionalDOItransitions,
+      record,
+    } = this.props;
+
+    const doi = record?.pids?.doi?.identifier || "";
+    const isNewUpload = record.status === "draft";
```
jrcastro2, line 38: Minor: We could user [this](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js#L35) to avoid hardcoding the state.

Also is `isNewUpload` a correct name?  just find it a bit confusing 😅 Does this mean that after the first version is published if we create a new version it doesn't has the draft status?

jrcastro2 (approved):

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/components/OptionalDOIoptions.js
zzacharo: no actually it has `status=='new_version_draft'`. Thanks for the suggestion, I will integrate it now!

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/components/OptionalDOIoptions.js
```diff
@@ -23,24 +24,39 @@ export class OptionalDOIoptions extends Component {
...
+      record,
+    } = this.props;
+
+    const doi = record?.pids?.doi?.identifier || "";
+    const alreadyPublished = [
+      DepositStatus.PUBLISHED,
+      DepositStatus.NEW_VERSION_DRAFT,
+    ].includes(record.status);
+
+    const hasDoi = doi !== "";
```
kpsherva, line 44: minor: wouldn't isEmpty be safer?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/components/OptionalDOIoptions.js
```diff
@@ -23,24 +24,39 @@ export class OptionalDOIoptions extends Component {
...
-        !optionalDOItransitions.allowed_providers.includes("not_needed"));
+    const {
+      isManagedSelected,
+      isNoNeedSelected,
+      pidLabel,
+      optionalDOItransitions,
+      record,
+    } = this.props;
+
+    const doi = record?.pids?.doi?.identifier || "";
```
kpsherva, line 38: minor: see my comment below(can be replacing empty string) why the empty string is needed in the first place?

zzacharo: The empty string is the initial state when the form is rendered. This was carried over from the existing implementation and I didn't refactor it. If you don't think it is a major blocker I would avoid refactoring more now...

PR #1944, opened 2025-02-04
schema.org: add uploadDate for VideoObject serialization [+]

fenekku:
This is the remnant of a dead-end investigation in providing more VideoObject structured
data per https://developers.google.com/search/docs/appearance/structured-data/video
since we were getting 'chided' by Google search console. Unfortunately, the
reality is that a record page is not a video watch page (it may contain multiple
videos + focus of page is on wider metadata) and a generic instance doesn't
have video thumbnails and perhaps doesn't want to expose video content file too
much. All of this resulted in only implementing the uploadDate field :shrug: .


tmorrell (approved):
Nice small improvement, and a good model if folks want to add other item-type specific serializations in the future.

PR #1947, opened 2025-02-08
cff: add default "message" field

slint:
* Addresses https://github.com/zenodo/zenodo-rdm/issues/960


tmorrell (approved):
Good change, and should pass tests once rebased.

PR #1955, opened 2025-02-14
WIP packaging: migrate everything to pyproject.toml

slint:


utnapischtim:
the following link would be an example how to create a publishing configuration which can be used to for `setup.cfg` and `pyproject.toml` files, which would help in the transition phase

https://github.com/tu-graz-library/.github/blob/main/.github/workflows/pypi-publish.yml

utnapischtim:
https://github.com/tu-graz-library/invenio-edugain would be an example for a `pyproject.toml` only package

Issue #1948, opened 2025-02-12
Sharing a draft/record doesn't give access to corresponding record requests


zzacharo:
Package version (if known): <version>

## Describe the bug
When a user is sharing their draft/record using the **Share** functionality, although they can act on the record/draft e.g edit a draft they cannot see act on the subsequent request. 

## Expected behavior

If a user is granted access to see/edit a record then it is reasonable to also be able to see/act on any associated request.

## Proposed Solution

To address the original issue, several changes have been implemented across different architecture layers, summarized below:

#### Requests
* The can_read permission now considers the topic of a request. In the case of a record, this is expanded to generate grant_tokens for users with can_preview access, allowing them to access the associated request.

#### Record

* Whenever the shared access settings of a record are updated, any corresponding review request is reindexed to update its grant_tokens. This ensures that all entities granted access to the record via sharing also gain access to the associated request.

### Upload Form

* The ability to manage community selection and record submission has changed. Now, only users with manage access can perform these actions. Users with edit access can modify metadata and publish a record only if it is not associated with a community, submitted to one, or preselected on the upload form. These permissions are correctly enforced for individuals, groups, and links with whom the record has been shared.

### Landing Page
 
* The community inclusion widget is now visible only to users with manage access to the record.

### Share of a record or draft

The wording for each permission has been changed to depict the fact that users can have access to the associated request.

![Image](https://github.com/user-attachments/assets/8c05c0fc-a5d9-45c2-8aaf-cfb5b263dde1)

#### Uploads
* A new “shared_with_me” filter has been introduced to distinguish between uploads you can access because you are the owner and those you can access due to granted preview, edit, or manage permissions.
* Since the backend does not provide granular details about a user’s specific access level (e.g., preview or edit), the below change was introduced:
* The Edit button has been **removed**, and only the “View” button is displayed, following these rules:
* If the upload is published, the “View” button redirects the user to the published record. Depending on their access level, they can edit the record from there to access the draft.
* If the upload is a draft, the “View” button redirects the user to the deposit page of the record. Users with preview access are automatically redirected to the draft’s preview.

#### Requests
* A new “shared_with_me” filter has been introduced to differentiate between requests that you have access to because you are the creator or recipient and those you can view or comment on due to the request’s topic being a record to which you have preview, edit, or manage access.

##### Current implementation
![Image](https://github.com/user-attachments/assets/20d6e512-1bf6-4e1c-8a7b-4a0ffed69f03)
[...]

PR #1956, opened 2025-02-15
resources: add param to filter shared with my uploads

zzacharo:
closes https://github.com/inveniosoftware/invenio-rdm-records/issues/1948

### DIFF
--- a/invenio_rdm_records/services/access/service.py
+++ b/invenio_rdm_records/services/access/service.py
@@ -22,7 +22,7 @@ from invenio_i18n import lazy_gettext as _
 from invenio_notifications.services.uow import NotificationOp
 from invenio_records_resources.services.errors import PermissionDeniedError
 from invenio_records_resources.services.records.schema import ServiceSchemaWrapper
-from invenio_records_resources.services.uow import unit_of_work
+from invenio_records_resources.services.uow import unit_of_work, RecordCommitOp
 from invenio_requests.proxies import current_requests_service
 from invenio_search.engine import dsl
 from invenio_users_resources.proxies import current_user_resources
@@ -103,6 +103,16 @@ class RecordAccessService(RecordService):
         return [
             GrantSubjectExpandableField("subject"),
         ]
+
+    #
+    # Update parent request on access changes
+    #
+
+    def _update_parent_request(self, parent, uow):
+        """Update the parent record request."""
+        if parent.review:
+            request = parent.review.get_object()
+            uow.register(RecordCommitOp(request, indexer=self.indexer))
 
     #
     # Secret links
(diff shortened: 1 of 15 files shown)


invenio_rdm_records/services/search_params.py
```diff
@@ -55,9 +60,14 @@ def is_user_authenticated():
...
 
         if value is None and is_user_authenticated():
-            search = search.filter(
-                "term", **{"parent.access.owned_by.user": identity.id}
-            )
+            my_uploads = dsl.Q("term", **{"parent.access.owned_by.user": identity.id})
+
+            if params.get("shared_with_me") is True:
+                # Shared with me
+                return search.filter(~my_uploads)
```
ptamarit, line 67: Is a record not owned by me necessarily shared with me?
Can there be other use cases, maybe related to communities, where a record was not explicitly shared with me, but where I am also not the owner?

invenio_rdm_records/services/config.py
```diff
@@ -97,7 +97,7 @@
 from .schemas.tombstone import TombstoneSchema
 from .search_params import (
     MetricsParam,
-    MyDraftsParam,
+    MyOrSharedWIthMeDraftsParam,
```
jrcastro2, line 100:
```suggestion
    MyOrSharedWithMeDraftsParam,
```

invenio_rdm_records/services/search_params.py
jrcastro2: Basically does this mean that on a community inclusion, will it appear as "shared with me", when it's not the case right?

jrcastro2 (commented):
Isort failing, peer-reviewed with @sakshamarora1

invenio_rdm_records/services/search_params.py
zzacharo: Exactly, you will see all your community related requests there. In principle, the shared with me is at the moment an umbrella to see all the records that yourself didn't create.

zzacharo:
Regarding https://github.com/inveniosoftware/invenio-rdm-records/pull/1956#discussion_r1965627243 : Yes if you have access you will see it in the shared view. So, a community inclusion to a community you can curate gives you access to the record.

invenio_rdm_records/services/access/service.py
```diff
@@ -104,6 +104,16 @@ def expandable_fields(self):
             GrantSubjectExpandableField("subject"),
         ]
 
+    #
+    # Update parent request on access changes
+    #
+
+    def _update_parent_request(self, parent, uow):
```
kpsherva, line 111: just for my information to understand how this works:  why is it needed?

zzacharo: It is needed so when a user changes the access in a record via the share button, the record/request is shown to the user that the record was shared with under their dashboard.

zzacharo:
> Regarding [#1956 (comment)](https://github.com/inveniosoftware/invenio-rdm-records/pull/1956#discussion_r1965627243) : Yes if you have access you will see it in the shared view. So, a community inclusion to a community you can curate gives you access to the record.

that is now changed to not include community related uploads/requests

PR #1962, opened 2025-02-20
file_links: prepare for changes in invenio-records-resources [+]

fenekku:
Invenio-records-resources will not provide default `file_links_{item,list}`
values anymore, because with the usage of `invenio_url_for` those default values cannot
be generic. Links need a specific endpoint which the default config
cannot provide (doesn't provide a blueprint for starters). So to prevent
breakage when invenio-records-resources is released and invenio-rdm-records
uses it (see usage change in this PR) before its own code has been changed, we need to make the pre-emptive change in this PR.

### DIFF
--- a/invenio_rdm_records/services/config.py
+++ b/invenio_rdm_records/services/config.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2020-2024 CERN.
-# Copyright (C) 2020-2021 Northwestern University.
+# Copyright (C) 2020-2025 Northwestern University.
 # Copyright (C)      2021 TU Wien.
 # Copyright (C) 2021-2023 Graz University of Technology.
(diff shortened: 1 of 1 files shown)


max-moser (approved):
i just did a quick check via `invenio shell`, and as expected the values inserted here are the default values (so there's no effective change):
```python
In [1]: from invenio_rdm_records.services.config import FileServiceConfig

In [2]: FileServiceConfig.file_links_list
Out[2]: {'self': <invenio_records_resources.services.records.links.RecordLink at 0x7479a9143830>}

In [3]: FileServiceConfig.file_links_item
Out[3]: 
{'self': <invenio_records_resources.services.files.links.FileLink at 0x7479a8ff7560>,
 'content': <invenio_records_resources.services.files.links.FileLink at 0x7479a563f4d0>}
```

PR #1971, opened 2025-03-07
form: revamp errors/warning in feedback 

ptamarit:
* Fixes inveniosoftware/react-invenio-forms#277

> - Updates FormFeedback messages and structure to show relevant sections and count, with directed links to them
> - Doesn't stop publishing in case of severity checks
> - Updates files upload button to have error class

Screenshot 2025-03-13 at 09 52 16

Screenshot 2025-03-13 at 16 23 27

Screenshot 2025-03-13 at 16 29 07

PR #1975, opened 2025-03-14
277 metadata check warnings fields visibililty acccess files

ptamarit:
Fixes inveniosoftware/react-invenio-forms#277


ptamarit:
Converting to draft for now.
The one extra commit in this PR compared to #1971 is `form: added feedback label in visibility access files`.

PR #1997, opened 2025-03-31
fix tombstone on commit

utnapischtim:
- fix: tombstone gone after record.commit
- fix: deletion_status gone after record.commit

### DIFF
--- a/invenio_rdm_records/records/systemfields/deletion_status.py
+++ b/invenio_rdm_records/records/systemfields/deletion_status.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2023 TU Wien.
+# Copyright (C) 2025 Graz University of Technology.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
@@ -89,6 +90,10 @@ class RecordDeletionStatusField(SystemField):
         if record is None:
             return self  # returns the field itself.
 
+        return self.get_obj(record, owner)
+
+    def get_obj(self, record, owner=None):
+        """Get obj."""
         status = self._get_cache(record) or RecordDeletionStatus(
             record.model.deletion_status
         )
@@ -103,7 +108,7 @@ class RecordDeletionStatusField(SystemField):
 
     def pre_commit(self, record):
         """Dump the deletion status to the record before committing."""
-        status = self._get_cache(record) or RecordDeletionStatus(None)
+        status = self.get_obj(record)
         record.model.deletion_status = record.get("deletion_status", status._status)
 
     def pre_dump(self, record, data, **kwargs):
--- a/invenio_rdm_records/records/systemfields/tombstone.py
+++ b/invenio_rdm_records/records/systemfields/tombstone.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2023 TU Wien.
+# Copyright (C) 2025 Graz University of Technology.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
@@ -156,6 +157,10 @@ class TombstoneField(SystemField):
         if record is None:
             return self  # returns the field itself.
 
+        return self.get_obj(record, owner)
+
+    def get_obj(self, record, owner=None):
+        """Get obj."""
         tombstone = self._get_cache(record)
         if tombstone is not None:
             return tombstone
@@ -167,6 +172,10 @@ class TombstoneField(SystemField):
 
     def __set__(self, record, value):
         """Set a record's tombstone entry."""
+        self.set_obj(record, value)
+
+    def set_obj(self, record, value):
+        """Set obj."""
         if value is None:
             tombstone = None
         elif isinstance(value, dict):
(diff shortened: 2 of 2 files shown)


invenio_rdm_records/records/systemfields/deletion_status.py
```diff
@@ -103,7 +108,7 @@ def __set__(self, record, value):
 
     def pre_commit(self, record):
         """Dump the deletion status to the record before committing."""
-        status = self._get_cache(record) or RecordDeletionStatus(None)
```
max-moser, line 106: sounds like the bug was the `RecordDeletionStatus(None)` when the record wasn't cached (yet)?

max-moser (approved):
LGTM

invenio_rdm_records/records/systemfields/deletion_status.py
utnapischtim: i think so, since in the now `get_obj` method the `RecordDeletionStatus` will get the value from the model. https://github.com/inveniosoftware/invenio-rdm-records/blob/6da09ba2291a3cae84ab964bc7dde3b74cf72df1/invenio_rdm_records/records/systemfields/deletion_status.py#L98

max-moser: which makes me wonder...
i think there are some cases when there _is_ no active model backing the record, e.g. when the source for the record in question is a search result (i.e. no DB query is involved) [1]
maybe, in such cases (when the model is `None`) we should just take what the `record` dictionary says here?

[1] i've had such cases with requests recently, in my effort to create "reminder notifications" for rdm-curation reviewers.
accessing some (model-backed) systemfields would fail because the model for the `Record` object is currently `None` (and would require a DB query to set properly)

utnapischtim: since the tests are not green yet maybe i see this special case in the tests and have to change the code anyway. to be honest, i would like to move this code into the `get_obj` function, otherwise it would recreate some code from there to avoid to call that function

invenio_rdm_records/records/systemfields/tombstone.py
```diff
@@ -148,6 +149,22 @@ def __repr__(self):
...
+
+    def is_marked_to_be_removed(self, instance):
+        """Is marked to be removed."""
+        if not hasattr(instance, "_obj_cache"):
+            instance._obj_cache = {}
+
+        try:
+            return instance._obj_cache["remove"]
+        except KeyError:
+            return False
```
max-moser, line 166:
```suggestion
       return instance._obj_cache.get("remove", False)
```

invenio_rdm_records/records/systemfields/tombstone.py
```diff
@@ -148,6 +149,22 @@ def __repr__(self):
 class TombstoneField(SystemField):
     """System field for accessing a record's deletion status."""
```
max-moser, line 150:
```suggestion
    """System field for accessing a record's deletion status.
    
    Note: To remove a record's tombstone, use ``del record.tombstone``.
    This will take care of properly de-initializing the information.
    """
```

invenio_rdm_records/services/components/record_deletion.py
```diff
@@ -57,7 +58,7 @@ def restore_record(self, identity, data=None, record=None, **kwargs):
         record.deletion_status = RecordDeletionStatusEnum.PUBLISHED
 
         # Remove the tombstone information
-        record.tombstone = None
+        del record.tombstone
```
max-moser, line 61: i've suggested an addition to the tombstone field's documentation, just because the `del` keyword isn't something you see every day anymore

utnapischtim: personally i would have liked to use some sort of `record.tombstone.remove()` call, but there is some black magic i don't understand. `tombstone` is of type `Tombstone` and not `TombstoneField` so adding `def remove` to `TombstoneField` doesn't work, but adding `def __delete__` to `TombstoneField` and using `del record.tombstone` works.

slint (approved):

invenio_rdm_records/records/systemfields/tombstone.py
```diff
@@ -178,10 +206,16 @@ def __set__(self, record, value):
 
         self._set_cache(record, tombstone)
 
+    def __delete__(self, record):
+        """Remove tombstone."""
+        self.mark_to_be_removed(record)
```
slint, line 211: minor/question: shouldn't doing `record.tombstone = None` also call `self.mark_to_be_removed(record)`?

invenio_rdm_records/services/components/record_deletion.py
max-moser: that might be referring to how invenio uses the way python descriptors (which is what systemfields are) work.
you can get the systemfield (i.e. descriptor) itself by accessing the class (e.g. `RDMRecord.tombstone` or `type(record).tombstone`).
if you access the systemfield (i.e. descriptor) via an object of the class, you'll get the tombstone rather than the tombstone field.

you can distinguish between access via object vs. class by virtue of some argument being set; in the case of the tombstone field, the check is here: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/records/systemfields/tombstone.py#L157

invenio_rdm_records/records/systemfields/tombstone.py
utnapischtim: yes it should, done

max-moser:
i assume this should be backported to v12?

PR #1999, opened 2025-04-01
feedback-ui: Add specific ID to disconnected feedback form

edivalentinitu:

### Description

### Checklist

- [ ] I'm aware of the [code of conduct](https://inveniordm.docs.cern.ch/contribute/code-of-conduct/).
- [ ] I've created [logical separate commits](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commits) and followed the [commit message format](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#commit-message).
- [ ] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

**Reminder**

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/errors/FormFeedback.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/errors/FormFeedback.js
@@ -214,11 +214,19 @@ class DisconnectedFormFeedback extends Component {
     // if no field is specified on the backend, then the message is on the `_schema` field
     const backendErrorMessage = errors._schema;
 
-    // Retrieve the corresponding icon and type
-    const { icon, type } = feedbackConfig[feedback] || {};
+    // Retrieve the corresponding icon and type if the feedback is a valid key,
+    // else fallback to warning.
+    const { icon, type } = feedbackConfig[feedback] || feedbackConfig["warning"];
 
+    // generate dynamic id for easy getElementById calls.
+    // see feedbackConfig for possible values.
     return (
-      <Message visible {...{ [type]: true }} className="flashed top attached">
+      <Message
+        visible
+        {...{ [type]: true }}
+        className="flashed top attached"
+        id={type + "-feedback-div"}
+      >
         <Grid container>
           <Grid.Column width={15} textAlign="left">
             <strong>


tmorrell (approved):
Seems like a nice improvement!

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/errors/FormFeedback.js
```diff
@@ -217,7 +217,7 @@ class DisconnectedFormFeedback extends Component {
     const { icon, type } = feedbackConfig[feedback] || {};
 
     return (
-      <Message visible {...{ [type]: true }} className="flashed top attached">
+      <Message visible {...{ [type]: true }} className="flashed top attached" id={type + "-feedback-div"}>
```
max-moser, line 220: could we have a comment around what an expected value for `type` would be, and thus an expected resulting value for the ID?

fenekku (approved):

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/errors/FormFeedback.js
utnapischtim, line 220: please check what type can be and add a guard to make sure that it doesn't break at runtime

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/errors/FormFeedback.js
```diff
@@ -216,8 +216,20 @@ class DisconnectedFormFeedback extends Component {
     // Retrieve the corresponding icon and type
     const { icon, type } = feedbackConfig[feedback] || {};
 
+    let finalType = type;
```
kpsherva, line 219: Hello! thank you for the contribution! I have couple of questions to better understand the goal.

Could you explain a little bit what is the motivation behind checking types here? 
Normally we have dynamic typing in our code, with a check of prop types for component interfaces, so it is not totally clear why would we check the type here.

I also understand that partially this code is supposed to help with the testing suite which detects specific selectors. I am worried however that we are mixing a test utility with our code base. I would hesitate a lot to take this approach, since it will increase the code complexity - and I believe tests should not make the code base more complex. After a time adding test utilities in between functional code will increase our maintenance cost, and we won't be able to distinguish what is a testing utility and what is a feature. 
I believe testing code should remain isolated from the application code.
Just to clarify - the above concern is not related to adding html attributes such as `id=`, since those are normally an integral part of the frontend. 
For the id's syntax I think we should have a naming strategy to make it clear what is their purpose (could it be a subject of a telecon breakout room?).

Is there any alternative solution to this type check?

edivalentinitu: hi, thanks for the feedback. dropped the type check and replaced it with a null check

kpsherva: Thank you! it looks better in terms of the code integration, how about taking it further and fixing the original issue?
The `null` you are checking for is originating from here https://github.com/inveniosoftware/invenio-rdm-records/pull/1999/files#diff-03d8aee41762cc84385040d5241b661f1f043b0c93cba256e75f1ae2d5e60698R189, meaning the feedback variable will end up as undefined, since we are probably receiving some action which is not listed in https://github.com/inveniosoftware/invenio-rdm-records/pull/1999/files#diff-03d8aee41762cc84385040d5241b661f1f043b0c93cba256e75f1ae2d5e60698R33

meaning that we should probably also report this unforseen behaviour or action to the user somehow, agreed @ptamarit @0einstein0 ?

I would try to do this:

```javascript
if (feedback === undefined){
const {icon, type} = feedbackConfig["warning"];
```

simply because it solves the issue of undefined values both for the feature and for your test (two birds with one stone :))

an alternative would be also adding a "default" action in https://github.com/inveniosoftware/invenio-rdm-records/pull/1999/files#diff-03d8aee41762cc84385040d5241b661f1f043b0c93cba256e75f1ae2d5e60698R33 to which we fall back in case we receive unknown action type, but maybe @ptamarit or @0einstein0 have more thoughts on this option

edivalentinitu: implemented the recommended fallback fix

kpsherva (approved):

PR #2007, opened 2025-04-08
localizing dates based on application selected locale

Ducica:

### Description

Currently, the dates in AccessRights component, are not being localized based on the language selected in the app and we would like to have it consistent i.e. if I select Czech language, I would like the dates to also be localized to that language. This is a minor modification, that uses locale set in i18next. Please let us know if it would be OK for you. Thanks!

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/AccessMessage.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/AccessMessage.js
@@ -27,7 +27,9 @@ export const AccessMessage = ({ access, metadataOnly, accessCommunity }) => {
   const fullEmbargo = !recordPublic && embargoActive;
 
   const fmtDate = access.embargo?.until
-    ? DateTime.fromISO(access.embargo.until).toLocaleString(DateTime.DATE_FULL)
+    ? DateTime.fromISO(access.embargo.until)
+        .setLocale(i18next.language)
+        .toLocaleString(DateTime.DATE_FULL)
     : "???";
 
   if (fullyPublic) {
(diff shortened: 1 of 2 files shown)


utnapischtim:
@Ducica could you please add a before and after screenshot of that particular change. that makes it easier to review.

Ducica:
@utnapischtim Sure, I will make sure to do that initially in the future. We use in app language selector to switch between languages and the way code is currently, it does not take into account the language selected in the app.

Before

image
image


After

image
image


Please ignore the rest of the text around it, because I think this is translated somewhere in transifex, but it will not be available until RDM13 (as I understood). We are here just talking about the date locaization, which was reported to us by user of one of our apps.

tmorrell:
We're pretty busy with the v13 release, but I've put it on the board so hopefully someone else can add a review.

Ducica:
@wgresshoff Hello thanks for the message, let me please double check again and I will get back to you.

Ducica:
@wgresshoff  Sorry about that, we don't use full record restriction so I missed that.

I can implement the fix, however, I think there is also an issue with translations.json 

image

image

Which can be seen on Zenodo too

image

If you agree, I will correct both issues. Could you please tell me if I can modify translation.json files directly or you somehow generate them from somewhere?

It should be enough to remove this i18nkey and use the same principle as in other trans components where keys/values are the same.
image
image

utnapischtim:
@Ducica @wgresshoff about the wrong message for fully restricted records. it is simply the wrong message added: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/AccessMessage.js#L70

@Ducica lets fix that in another PR

Ducica:
@utnapischtim Thanks for the message. I would say it is just wrong content of translations.json file for english and czech for the key "access-message". Just to be clear, here the issue is fullEmbargo, meaning when you fully restrict and then try to add embargo as well. 

https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/AccessField/components/AccessMessage.js#L53

For me it is OK to merge, but If I understood your previous comment, this would need to be fixed on transifex and then pulled into the library

PR #2014, opened 2025-04-11
metadata: add copyrights

kpsherva:
* closes https://github.com/CERNDocumentServer/cds-rdm/issues/381
* proposal of the solution to add copyrights field, discussed on the 8th of April telecon
* I tried to apply what we came up with in the breakout room and:
    *  I explored the possibility of adding the field alongside license field. Unfortunately the current implementation of license field is a bit too rigid to handle this, license field would need to be refactored (unfortunately this is not feasible in the scope of this PR)
    * putting the field inside the modal requires some thoughts and design on the UX of representing the current value when updating the record
    * explored adding the value to datacite serializer, it seems we don't have a field to accommodate it at the moment https://datacite-metadata-schema.readthedocs.io/en/4.5/properties/rights/

### DIFF
--- a/invenio_rdm_records/services/schemas/metadata.py
+++ b/invenio_rdm_records/services/schemas/metadata.py
@@ -385,6 +385,7 @@ class MetadataSchema(Schema):
     )
     version = SanitizedUnicode()
     rights = fields.List(fields.Nested(RightsSchema))
+    copyrights = SanitizedHTML(validate=validate.Length(min=3))
     description = SanitizedHTML(validate=validate.Length(min=3))
     additional_descriptions = fields.List(fields.Nested(DescriptionSchema))
     locations = fields.Nested(FeatureSchema)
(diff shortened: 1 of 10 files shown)


invenio_rdm_records/services/schemas/metadata.py
```diff
@@ -385,6 +385,7 @@ class MetadataSchema(Schema):
     )
     version = SanitizedUnicode()
     rights = fields.List(fields.Nested(RightsSchema))
+    copyrights = SanitizedHTML(validate=validate.Length(min=3))
```
tmorrell, line 388: "copyright" seems like a better label to me. My main argument is that since it's a single text field (not a list) the singular makes more sense. This is probably a bit of a US vs EU thing too (I've never used "copyrights" as a term). 

I'll also raise a minor point about the length restriction. Could it be one character in case someone just wanted to use the copyright symbol?

tmorrell (commented):
For the DataCite serialization, you can add it to rightsList with structure:

```
{
      "rights": "Put text from the copyright field here",
      "rightsUri": "http://rightsstatements.org/vocab/InC/1.0/"
 }
```

invenio_rdm_records/services/schemas/metadata.py
kpsherva: done!

PR #2015, opened 2025-04-14
Meetings

kpsherva:


tmorrell (approved):
The meeting identifier change makes sense, with just some minor suggestions. This will also need a migration script for v13. If it could be just run on records that have the conference website field (and not every record), that would be preferable.

Issue #1978, opened 2025-03-18
New Thesis Fields Discussion


tmorrell:
Summary issue for discussions around upgrades to the thesis fields from telecon and https://github.com/inveniosoftware/invenio-rdm-records/pull/1696

Initial proposal:

Fields to upgrade to a vocabulary:

- University
- Department
- Thesis type

Possible new fields in the thesis section:

- Dates (Re-use date field widget)
- Advisor (Re-use creator field widget)
- Committee (Re-use creator field widget)
- Major (Vocabulary)
- Minor (Vocabulary)

tmorrell:
Final proposal:

Vocabulary: University (pull from affiliations). Also accept free text.
Type: vocabulary

New fields:
Defense date
Submission date
(Multi-add date and date type would also work, but would be trickier to implement)

There are some issues with a nested custom field that has vocabularies inside them.
Date field should be validation should be able to be added instead of text

egabancho:
@tmorrell, can you remind me if we discussed doing something similar to `codemeta` custom fields and why we discarded the idea?

I am talking about defining the fields following `thesis:university` or `thesis:dates` rather than `thesis:thesis`, and having a complex object inside.

tmorrell:
We're following the pattern used in imprint, journal, and meeting which is the repeated label. I'm not arguing that's the best structure...just if we decided to change it it would require deprecating a lot of existing fields and would be a lot of work.

I wasn't part of the decision when it was made originally, but I think the idea was the namespace for the custom fields could be used for site-specific fields. So sites could use "caltech:thesis.division" for fields that were specific to them. Then the default fields get the "thesis:thesis" label.

PR #2020, opened 2025-04-17
thesis: add defense and submission date

kpsherva:
* partially addresses https://github.com/inveniosoftware/invenio-rdm-records/issues/1978

### DIFF
--- a/invenio_rdm_records/contrib/thesis/custom_fields.py
+++ b/invenio_rdm_records/contrib/thesis/custom_fields.py
@@ -16,7 +16,7 @@ Implements the following fields:
 from invenio_i18n import lazy_gettext as _
 from invenio_records_resources.services.custom_fields import BaseCF
 from marshmallow import fields
-from marshmallow_utils.fields import SanitizedUnicode
+from marshmallow_utils.fields import SanitizedUnicode, EDTFLevel2DateString
 
 
 class ThesisCF(BaseCF):
@@ -30,6 +30,8 @@ class ThesisCF(BaseCF):
                 "university": SanitizedUnicode(),
                 "department": SanitizedUnicode(),
                 "type": SanitizedUnicode(),
+                "date_submitted": EDTFLevel2DateString(),
+                "date_defended": EDTFLevel2DateString(),
             }
         )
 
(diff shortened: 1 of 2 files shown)


tmorrell:
I think we'll need to get this added to the UI as well.  I forgot to make a PR for the UI for department and type, which was just a label change: https://github.com/caltechlibrary/caltechauthors/blob/7dd9a7b5b19362a659dbc434a2930a4253cde43a/templates/semantic-ui/invenio_app_rdm/records/details/side_bar/details.html#L29. 

The dates will probably need a separate serialized form as individual fields, or we could refactor it differently.

tmorrell (approved):
Looks good once the etdf changes are merged and tests pass.

invenio_rdm_records/contrib/thesis/custom_fields.py
```diff
@@ -42,6 +44,8 @@ def mapping(self):
                 "university": {"type": "keyword"},
                 "department": {"type": "keyword"},
                 "type": {"type": "keyword"},
+                "date_submitted": {"type": "keyword"},
```
egabancho, line 47: I might be wrong here, but I believe we agreed during the workshop session that we would have a `dates` field to which we will initially add only `submitted` and `defended`, somewhat artificially using the interface.

kpsherva: > I think the conclusion was that it would be too difficult to port the date functionality into custom fields. But if it's possible that would be a better option.

yes that was the conclusion, I followed the last comment added in the issue by Tom. We can always improve in the future versions - but at this moment I couldn't unfortunately dig deeper on this

slint: Another point of the discussion in the workshop, was that we could have `dates` (with the limited types), and just handle the complexity in the UI by just showing two fields for input that serialize to a list. This would leave open the door for other date types, without having to go through adding more mapping fields to the custom field.

The only clear benefit I see with having them separate is that you can directly do a search query on either type of date (which wouldn't be possible if it was a list of objects, because of the usual ElasticSearch array-of-object flatenning issues).

> Disclaimer: I have zero stakes on the table for adding a thesis field, so take all this with a grain of salt 😅

PR #2036, opened 2025-05-06
schemas: set default value for notes in quota increases

max-moser:
On the administration panel, the `notes` field is marked as optional:
image

However, if this value is omitted, this may result in the `data` not having a `notes` field at all, which crashes the service layer here: https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/services/services.py#L694

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,10 @@
 Changes
 =======
 
+Version <next>
+
+- schemas: provide default value for quota increase notes
+
 Version v18.8.0 (released 2025-05-06)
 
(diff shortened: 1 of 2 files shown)


Samk13 (approved):
LGTM! 
Thanks for that, don't forget to add your copyrights "if you want to 😄"

tmorrell (approved):
Great! It's a small enough change that it doesn't need copyrights unless you want to.

max-moser:
yeah, this would barely qualify as creative work :sweat_smile:

PR #2044, opened 2025-05-13
fix(ui): record access not changing

utnapischtim:
* if a restricted community is removed from the deposit the record
  should change back to public


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
```diff
@@ -333,6 +334,15 @@ const depositReducer = (state = {}, action) => {
           },
         };
       }
+      if (!isRecordPublic && !isCommunityRestricted) {
```
utnapischtim, line 337: NOTE:
this PR fixes the problem by setting the record access setting back to public/public. it could be that it would be better to unblock the visibility setting. that should be discussed somehow!

kpsherva:
hello! overall looks good, I would additionally check before merging if the deposit behaves correctly on ssave/publish and other actions (if the restrictions are correct) and if it is possible to re-select the restricted community

PR #2047, opened 2025-05-15
serializers/datacite: configurable dumping of access right

### LINKED ISSUE #2046: info:eu-repo-Access-Terms vocabulary in DataCite export
Package version (if known): unknown

In the DataCite export invenio used to expose the access restrictions according to the OpenAIRE recommendations:
https://guidelines.openaire.eu/en/latest/data/field_rights.html#rightsuri-ma

But sometime ~2023/2024 this was no longer availible.
Example form resources tagged with openAccess from DataCites API (Zenodo):
https://api.datacite.org/dois?query=rightsList.rightsUri:%22info:eu-repo/semantics/openAccess%22&client-id=cern.zenodo

Image

This information is still available in oai_dc but it would be good to keep it in DataCite as well because it makes it alot simpler to determine what resources have emargo/restrictions etc by using only DataCite metadata.

It seems to still be some code related to handling datacite-dcat for this but i did not find when/why this was removed from the DataCite export:
https://github.com/inveniosoftware/invenio-rdm-records/blob/809923a05a0a5281a90e044d613413e5832d9628/invenio_rdm_records/resources/serializers/dcat/datacite-to-dcat-ap.xsl#L879-L886 

## Steps to Reproduce
1. Go to any resource in a current invenio instance
2. Click on export metadata in DataCite

## Expected behavior

access conditions with the license:
```xml
<rightsList>
  <rights rightsURI=”info:eu-repo/semantics/openAccess” />
  <rights rightsURI=”http://creativecommons.org/licenses/by/4.0/”>
    Creative Commons Attribution 4.0 International
  </rights>
</rightsList>
```

slint:
* Closes #2046.
* Adds a new `RDM_DATACITE_DUMP_ACCESS_RIGHTS` config variable to
  control if the access right level is included in the DataCite
  serialization. The value is based on the vocabulary documented at
  <https://wiki.surfnet.nl/spaces/standards/pages/11055603/info-eu-repo#infoeurepo-AccessRights>

### DIFF
--- a/invenio_rdm_records/config.py
+++ b/invenio_rdm_records/config.py
@@ -628,8 +628,13 @@ RDM_FILES_DEFAULT_MAX_FILE_SIZE = None
 RDM_DATACITE_FUNDER_IDENTIFIERS_PRIORITY = ("ror", "doi", "grid", "isni", "gnd")
 """Priority of funder identifiers types to be used for DataCite serialization."""
 
-RDM_DATACITE_DUMP_ACCESS_RIGHTS = False
-"""Flag to show whether DataCite access rights level should be dumped or not."""
+RDM_DATACITE_DUMP_OPENAIRE_ACCESS_RIGHTS = False
+"""Flag to control dumping DataCite OpenAIRE access rights.
+
+See https://guidelines.openaire.eu/en/latest/data/field_rights.html for further 
+information on how the OpenAIRE Guidelines expect access rights to be exposed 
+via the DataCite schema.
+"""
 
 RDM_IIIF_MANIFEST_FORMATS = [
     "gif",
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/config.py
```diff
@@ -628,6 +628,9 @@ def lock_edit_published_files(service, identity, record=None, draft=None):
 RDM_DATACITE_FUNDER_IDENTIFIERS_PRIORITY = ("ror", "doi", "grid", "isni", "gnd")
 """Priority of funder identifiers types to be used for DataCite serialization."""
 
+RDM_DATACITE_DUMP_ACCESS_RIGHTS = False
```
tmorrell, line 631: I wonder if we want to default this to True. I'm not clear on the history of adding and removing this from the serializer...but it generally seems like a good thing to include.

tmorrell (approved):
Looks like a good improvement with tests.

invenio_rdm_records/config.py
tmorrell: I wonder if we want to use a different config variable then? Maybe RDM_DATACITE_DUMP_OPENAIRE_ACCESS_RIGHTS?

I agree the rest can be put in an issue and then we can pick one of those options

PR #2099, opened 2025-07-10
PublishButton: Separate PublishModal and add checkbox easily via params

sakshamarora1:
fixes: https://github.com/CERNDocumentServer/cds-rdm/issues/137


ntarocco:
This looks very good to me, great work!
The only extra check I would implement is that we must ensure that `id` and `fieldPath` are unique and not duplicated. Can we do that easily?
Please have a look to the lint issues.

PR #2111, opened 2025-07-22
fix: raise validation errors for publish without reserving DOI

sakshamarora1:
fixes: https://github.com/CERNDocumentServer/cds-rdm/issues/523

- When a published record with default field as `No, I need one` is re-published, the state is not set for the validation check on the optional DOI field.
- On Submit for review button, there was no validation for DOI reservation for optional DOIs.

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/controls/PublishButton/PublishButton.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/controls/PublishButton/PublishButton.js
@@ -35,38 +35,30 @@ class PublishButtonComponent extends Component {
     const { setSubmitContext } = this.context;
     const {
       formik,
-      raiseDOINeededButNotReserved,
       isDOIRequired,
       noINeedDOI,
+      doiReservationCheck,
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/components/OptionalDOIoptions.js
```diff
@@ -26,6 +26,13 @@ class OptionalDOIoptionsCmp extends Component {
   _render = (cmp, shouldWrapPopup, message) =>
     shouldWrapPopup && message ? <Popup content={message} trigger={cmp} /> : cmp;
 
+  componentDidMount() {
+    const { isManagedSelected, isNoNeedSelected, onManagedUnmanagedChange } =
+      this.props;
+    // When the component is mounted we need to call the callback to sync the state
+    onManagedUnmanagedChange(isManagedSelected, isNoNeedSelected);
```
zzacharo, line 33: hmmm I don't understand this.... here we call a property passed down from the parent component.... it feels a bit hackish...can you ellaborate on what is the problem please?

PR #2117, opened 2025-07-26
remove context usage

utnapischtim:
- refactor(schema): remove usage of object_key

NOTE: please ignore the unpin commit. it is not part of this PR, it makes only my life little bit easier

check about the `is_parent` PR
and fix the `max_community` setting, it is also using the context


utnapischtim:
closed in favor of #2170 

if the problems should be handled independently, please update this PR

PR #2123, opened 2025-08-06
services: create RDMCommunityRecordSearchOptions

carlinmack:

### Description

Currently the community records search uses the same search options as the record search, which means that they both currently de-rank records from unverified users. This creates confusion and frustration in community administrators as they review and accept records into their communities from unverified users, and then cannot find it. Either they think that the inclusion request failed or that the search is broken for their community. To fix this, instance administrators have to verify the community members and the owners of records in that community

As records are manually reviewed before they are accepted into communities, there is no reason to apply the de-ranking in communities.

### Checklist

- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/develop/howtos/i18n/).
    * If you have added [third-party code (copy/pasted or new dependencies)](https://inveniordm.docs.cern.ch/develop/best-practices/commits/#third-party-codedependencies), please reach out to an [architect on Discord](https://discord.gg/8qatqBC).

**Frontend**

**Reminder**

### DIFF
--- a/invenio_rdm_records/services/config.py
+++ b/invenio_rdm_records/services/config.py
@@ -210,6 +210,26 @@ class RDMSearchOptions(SearchOptions, SearchOptionsMixin):
     ]
 
 
+class RDMCommunityRecordSearchOptions(SearchOptions, SearchOptionsMixin):
+    """Search options for community record search."""
+
+    facets = {
+        "resource_type": facets.resource_type,
+        "languages": facets.language,
+        "access_status": facets.access_status,
+    }
+
+    # without VerifiedRecordsSortParam as communities review records they accept
+    params_interpreters_cls = [
+        AllVersionsParam.factory("versions.is_latest"),
+        QueryStrParam,
+        PaginationParam,
+        FacetsParam,
+        StatusParam,
+        PublishedRecordsParam,
+    ] + SearchOptions.params_interpreters_cls
+
+
 class RDMSearchDraftsOptions(SearchDraftsOptions, SearchOptionsMixin):
     """Search options for drafts search."""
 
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/services/config.py
```diff
@@ -827,8 +847,8 @@ class RDMCommunityRecordsConfig(BaseRecordServiceConfig, ConfiguratorMixin):
         "RDM_SEARCH",
         "RDM_SORT_OPTIONS",
         "RDM_FACETS",
-        search_option_cls=RDMSearchOptions,
-        search_option_cls_key="RDM_SEARCH_OPTIONS_CLS",
+        search_option_cls=RDMCommunityRecordSearchOptions,
+        search_option_cls_key="RDM_COMMUNITY_RECORD_SEARCH_OPTIONS_CLS",
```
slint, line 851: comment: I was about to say to add this to the top-level config.py of the module, but I see the previous config isn't there either... So LGTM :)

invenio_rdm_records/services/config.py
```diff
@@ -210,6 +210,26 @@ class RDMSearchOptions(SearchOptions, SearchOptionsMixin):
...
+
+    # without VerifiedRecordsSortParam as communities review records they accept
+    params_interpreters_cls = [
+        AllVersionsParam.factory("versions.is_latest"),
+        QueryStrParam,
+        PaginationParam,
+        FacetsParam,
+        StatusParam,
+        PublishedRecordsParam,
+    ] + SearchOptions.params_interpreters_cls
```
slint, line 230:
```suggestion
    ]
```

This would already include many of the other parameters already specified, and I don't want to know what happens when we repeat these classes 😅 

I think now that we've identified where the issue was, I would actually go another solution, which is to keep the explicit `RDMCommunityRecordSearchOptions`, but modify `VerifiedRecordsSortParam` so that it checks for `self.config.verified_sorting_enabled`, which we can then set like:

```python
class RDMCommunityRecordSearchOptions(...):

   verified_sorting_enabled = False
```

Thus, we keep the original records search config via inheritance and don't have to duplicate new facets or parameters.

slint (commented):
I propose a slightly different implementation which will save us from some eventual copy-pasting everytime we add more params/facets in the records search config

PR #2125, opened 2025-08-06
ci: switch to uv for Python tests

slint:
> [!IMPORTANT]
> This is not meant to be merged, but just as a precautionary check to see if the `uv` CI workflow works without issues. Once done, we'll actually merge [the workflows `uv` branch](https://github.com/inveniosoftware/workflows/tree/uv) to `master` and thus propagate the change to all repos

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,7 +31,7 @@ on:
 
 jobs:
   Python:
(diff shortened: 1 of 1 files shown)


slint:
Closed since this passes tests, so it's a good sign that we can enable `uv` globally

PR #2133, opened 2025-08-21
deposit-form: add fallback message if checksum is not yet available

max-moser:
This is relevant for the multipart file upload with local file storage, where the checksum gets calculated in a background task after finalizing the upload.

See also: https://github.com/inveniosoftware/invenio-records-resources/pull/650


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/FileUploader/FileUploaderArea.js
```diff
@@ -120,7 +120,7 @@ const FileTableRow = ({
             </a>
           )}
           <br />
-          {file.checksum && (
+          {(file.checksum && (
```
max-moser, line 123: this extra pair of brackets seems to be `prettier` than without

slint (approved):

PR #2141, opened 2025-08-28
optional-doi: fixes

zzacharo:


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositRecordSerializer.js
```diff
@@ -408,7 +412,12 @@ export class RDMDepositRecordSerializer extends DepositRecordSerializer {
     ]);
 
     // Save pids so they are not removed when an empty value is passed
-    let savedPIDsFieldValue = originalRecord.pids || {};
+    const savedNoDOIPIDS = this.originalPublishedRecord.pids
```
zzacharo, line 415: save any already added published PID that is not part of the form. From the form we reset the pids field so we can e.g "unreserve" an externally managed DOI (it is just a metadata update).

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/components/OptionalDOIoptions.js
```diff
@@ -42,13 +42,18 @@ class OptionalDOIoptionsCmp extends Component {
...
-      : hasDoi && isManagedSelected;
-
-    const isNoNeedDisabled = alreadyPublished
-      ? !optionalDOItransitions.allowed_providers.includes("not_needed")
-      : hasDoi && isManagedSelected;
+    const doiManaged = hasDoi && isManagedSelected;
+
+    function isDisabled(provider) {
+      return (
+        doiManaged ||
```
zzacharo, line 49: disable the options "Yes, I have" and "No, I need one" in the UI if there is locally managed DOI reserved. We do that so that the user is forced to unreserve first the managed DOI.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/Identifiers/PIDField/OptionalPIDField.js
```diff
@@ -207,7 +210,8 @@ class OptionalPIDFieldCmp extends Component {
 
         {canBeManaged && _isManagedSelected && (
           <ManagedIdentifierCmp
-            disabled={hasDoi && isEditingPublishedRecord}
+            // You cannot change the managed DOI option if you have already published with a local DOI
+            disabled={hasPublishedLocalDOI}
```
zzacharo, line 214: We disable the unreserve button for locally managed DOIs only if there is a published version with one. In all the other cases, unreserve is there if the option is enabled.

PR #2165, opened 2025-09-24
fix(record-deletion): add safety net permission

slint:
* Adds a `can_request_deletion` permission for the
  `RDMRecordService.request_deletion` service call.
* This is meant to be a "safety net" to narrow down the set of
  actors that can manage the record. The configured deletion policy will
  actually check the specifics of whether a record is within policy for
  immediate deletion or a deletion request.


ntarocco:
As discussed in the [related PR](https://github.com/inveniosoftware/invenio-rdm-records/pull/2136), the policy should cover all use cases. We can close this one.

PR #2203, opened 2025-11-18
backport component service and co

fenekku:
- **community-records: Adds service component calls to RecordCommunitiesService**
- **fix(services): Make 'notes' parameter optional in _update_quota**
- **release: v19.6.0**


tmorrell (approved):
Good to go!

PR #2228, opened 2025-12-18
`SanitizedHTML`: Take allowed HTML tags & attributes from the application configuration

max-moser:
Previously, the set of allowed HTML tags and attributes was actually taken from a hard-coded list in [`marshmallow_utils.html`](https://github.com/inveniosoftware/marshmallow-utils/blob/master/marshmallow_utils/html.py#L24) because they weren't specified as arguments to the constructor.

This means that the configuration wasn't taken into account, and it easily possible to change the list of allowed tags and attributes, e.g. to allow the `` tag.

The latter is not part of the hard-coded list, causing the image feature in the TinyMCE editor to not work properly.


zzacharo (approved):
LGTM! That would be the easiest to propagate the change across the RDM.... I like the fallback too.

PR #2234, opened 2026-01-08
fix(iiif): validate expected shape of identifiers

max-moser:
Some IIIF API endpoints expect passed identifiers to have a specific shape (either two or three separator colons).
Currently, this is not validated beyond a simple `id.split(":")` which causes a 500 internal server error if the identifier doesn't match the expected shape.

This PR introduces a new exception type that indicates the case that an identifier did not match the expected shape.

### DIFF
--- a/invenio_rdm_records/services/iiif/service.py
+++ b/invenio_rdm_records/services/iiif/service.py
@@ -11,6 +11,8 @@ import io
 from flask_iiif.api import IIIFImageAPIWrapper
 from invenio_records_resources.services import Service
 
+from ..errors import IdentifierShapeException
+
 try:
     metadata.distribution("wand")
     from wand.image import Image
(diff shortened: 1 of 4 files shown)


invenio_rdm_records/services/iiif/service.py
```diff
@@ -52,16 +54,24 @@ def _iiif_uuid(self, uuid):
 
         We assume the uuid is build as ``<record|draft>:<pid_value>``.
         """
-        type_, id_ = uuid.split(":", 1)
```
max-moser: remove the `maxsplit` argument to have a more accurate reading of the number of separator colons in the full string

this could be an issue if `pid_value` was allowed to have a colon inside it, which is not the case for RECIDs out of the box
if they do, the `_iiif_image_uuid()` check might be even worse off

max-moser:
lol, i forgot about this PR and investigated a bit about IIIF error notifications yesterday and today...
after an hour or two i came up with a fix that looked almost exactly like this PR again (except for naming).

tmorrell:
Re-kicking off CI

max-moser:
Rebased to get the fixes in

max-moser:
sibling PR: https://github.com/inveniosoftware/invenio-rdm-records/pull/2380

tmorrell (approved):
A clear improvement in the validation with tests. Looks good to me.

fenekku:
This is part of invenio-rdm-records 33.x but invenio-app-rdm v14 currently only takes invenio-rdm-records 32.x . invenio-app-rdm should bump its invenio-rdm-records to 34.x actually.

tmorrell:
@fenekku This has been backported to 32.x, which is what is going into RDM v14. https://github.com/inveniosoftware/invenio-rdm-records/commits/maint-32.x/

fenekku:
@tmorrell ah I see now thank you!

PR #2235, opened 2026-01-16
V23.2.0

zzacharo:


palkerecsenyi (approved):
Thanks, I'm not sure how we missed this! It looks good to me, I also found the same issue

PR #2253, opened 2026-02-14
fix: role.name to role.id

utnapischtim:
needs https://github.com/inveniosoftware/flask-security-fork/pull/90, so tests will not be green without the flask-security-fork release.

i am unsure if we should solve it that way


utnapischtim:
superseded by #2254

PR #2290, opened 2026-03-27
resource_types: fix OpenAIRE resource type for computational notebook

ptamarit:
* Problem reported by the OpenAIRE team:
    > OpenAIRE relies on the resourceType code to apply the [dnet:publication_resource vocabulary](https://api.openaire.eu/vocabularies/dnet:publication_resource.json), that defines 0040 as Other software type.
    > 
    > OpenAIRE codes have fixed-length prefixes (e.g., 0029 for Software, 0040 for Other software type).
    > 
    > If InvenioRDM sends 40 instead of 0040, OpenAIRE's vocab resolver doesn't match it, resulting in unresolved instance types on OpenAIRE Explore.
    > OpenAIRE confirmed that if they manually change the value from "40" to "0040" and reindex, the mapping works correctly.
* The problem is only for this specific resource type, all other resource types have a proper format.
* The fixture's `openaire_resourceType` prop needs to be manually updated:
    ```sql
    from invenio_vocabularies.proxies import current_service
    from invenio_db import db

    record = current_service.read(system_identity, id_ = ("resourcetypes", "software-computationalnotebook"))._record
    record["props"]["openaire_resourceType"] = "0040"

    record.commit()
    db.session.commit()
    current_service.indexer.index(record)
    ```
    I don't think we need to invalidate the [cache for the resource types](https://github.com/inveniosoftware/invenio-vocabularies/blob/c5dc8a5b72b341776070e72bfdb43e08a8c88aaf/invenio_vocabularies/services/service.py#L126) in the form, nor reindex records, since this prop is only used for direct indexation.
* I also fixed tests in a separate commit to be more consistent.
* Other repositories with badly formatted `openaire_resourceType` in test fixtures:
    * inveniosoftware/invenio-collections#121
    * inveniosoftware/invenio-app-rdm#3401
    * inveniosoftware/invenio-swh#48
    * zenodo/zenodo-rdm#1330


tmorrell (approved):
Sensible change

PR #2294, opened 2026-03-31
feat(RecordCommunitiesService): Add get_record_requests service method

### LINKED ISSUE #568: "View comments" button under the community is not visible on other record versions (landing page)
Currently, if a request is accepted for a record version, the "View Comments" button under the accepted community only shows up for that specific version. It is stored in the "topic" sub-field of the request's payload.

There should be a way to view the request comments from the landing page of all versions.

This will require to have the parent_id on the requests index, to be able to show the button on all other versions.

### Solution proposal (to be discussed and finalized)
The link can redirect to the dashboard with the results containing all the related requests on the parent id.
This way if there is scenario where a community inclusion request of version was not accepted but the version 2 was accepted, the link will redirect to the requests dashboard containing all requests for those records in that community to be able to refer to the older request for reference.

#### Steps
1. Add parent id to the request index and update index (all new requests will contain parent_id)
2. Re-index the requests (where parent_id doesn't exist, all requests before we updated the index)
3. Update the link in the UI at the end (this way we avoid any searches on the parent till we have it indexed on all)

## Additional context
Another thing to consider is that on the request dashboard currently there is no way to view all open/closed/owned/shared request in a single list view. We can have a way to be able to view.

sakshamarora1:
closes: https://github.com/inveniosoftware/invenio-requests/issues/568
requires: https://github.com/inveniosoftware/invenio-communities/pull/1361

  * Moved from invenio_app_rdm/records_ui/views/records.py
  * Since the function checks for permissions and also access the API layer, it should be a service method following the existing methods in this service class also accessing the communities API

### DIFF
--- a/invenio_rdm_records/services/communities/service.py
+++ b/invenio_rdm_records/services/communities/service.py
@@ -10,6 +10,7 @@
 """RDM Record Communities Service."""
 
 from flask import current_app
+from flask_principal import AnonymousIdentity
 from invenio_access.permissions import system_identity
 from invenio_communities.proxies import current_communities
 from invenio_drafts_resources.services.records.uow import ParentRecordCommitOp
@@ -34,7 +35,7 @@ from sqlalchemy.orm.exc import NoResultFound
 
 from ...notifications.builders import CommunityInclusionSubmittedNotificationBuilder
 from ...proxies import current_rdm_records, current_rdm_records_service
-from ...requests import CommunityInclusion
+from ...requests import CommunityInclusion, CommunitySubmission
 from ..errors import (
     CannotRemoveCommunityError,
     CommunityAlreadyExists,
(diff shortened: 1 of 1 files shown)


invenio_rdm_records/services/communities/service.py
```diff
@@ -493,3 +494,55 @@ def bulk_add(self, identity, community_id, record_ids, set_default=False, uow=No
...
+                ],
+            ),
+            params={"sort": "oldest"},
+        )
+        record_requests = {
+            request["receiver"]["community"]: request["id"]
+            for request in record_requests
+        }
+
+        community_requests.update(record_requests)
```
OliverGeneser, line 546: Doesn't this potentially overwrite requests? If a community have an accepted request from parent.communities.get_requests() and a pending request?

sakshamarora1: Then the pending request is shown. Only in case there is no request for the record as a topic, it will show the accepted request

mesemus:
If the index for record_id is required, it needs to be defined at the sqlalchemy level, not just at the alembic level. The reason is that if user calls `invenio db init create` (when initializing a new repository), this call does not invoke alembic at all and then the index would not get created.

The fix is to add:

```python
from sqlalchemy import Index

# existing class
class RDMParentCommunity(db.Model, CommunityRelationMixin):
    """Relationship between parent record and communities."""

    __tablename__ = "rdm_parents_community"
    __record_model__ = RDMParentMetadata

# newly added line
Index("ix_rdm_parents_community_record_id", RDMParentCommunity.record_id)
```

After this addition the test_alembic.py will start working.

sakshamarora1:
@mesemus Yes, I've added it directly in the mixin class in `invenio-communities`
https://github.com/inveniosoftware/invenio-communities/pull/1361/changes#diff-f879da97b230ab47bad06c7db3eb34568ff92b73820dc120670afead2f577f4cR49

Once the module is released the tests should pass
Indeed, moving it here would contain it in this PR, WDYT? @slint

mesemus:
> @mesemus Yes, I've added it directly in the mixin class in `invenio-communities` https://github.com/inveniosoftware/invenio-communities/pull/1361/changes#diff-f879da97b230ab47bad06c7db3eb34568ff92b73820dc120670afead2f577f4cR49
> 
> Once the module is released the tests should pass Indeed, moving it here would contain it in this PR, WDYT? @slint

Correct, that is ok, no need to move it - I've overlooked that it depends on the other PR :)

kpsherva:
> > @mesemus Yes, I've added it directly in the mixin class in `invenio-communities` https://github.com/inveniosoftware/invenio-communities/pull/1361/changes#diff-f879da97b230ab47bad06c7db3eb34568ff92b73820dc120670afead2f577f4cR49
> > Once the module is released the tests should pass Indeed, moving it here would contain it in this PR, WDYT? @slint
> 
> Correct, that is ok, no need to move it - I've overlooked that it depends on the other PR :)

@mesemus, please check my comment https://github.com/inveniosoftware/invenio-communities/pull/1361/changes#r3386855622

PR #2303, opened 2026-04-07
feat(auditlog): Add audit logs for file create/delete

### LINKED ISSUE #681: Audit logs: Files and grants
- [ ] add grants audit logs to store information on access changes
- [ ] add files audit logs to track file changes (updates before grace period applied, updates by system user, might be handy to include the changes of files when done by link holder or other person who holds an edit right)
- [ ] remember to add parent id to draft context in the existing draft logs

sakshamarora1:
closes: https://github.com/CERNDocumentServer/cds-rdm/issues/681

For files, from what I've observed users can only delete and upload a new file. There aren't any changes done via the service layer on the same file (eg. metadata).

If there is, and maybe I have missed it, please let me know and we can add tracking for that as well.

### DIFF
--- a/invenio_rdm_records/auditlog/actions.py
+++ b/invenio_rdm_records/auditlog/actions.py
@@ -18,7 +18,7 @@ from invenio_rdm_records.services.schemas.parent.access import (
     Grant,
 )
 
-from .context import LogChangesContext, ResourceDataContext
+from .context import FileContext, LogChangesContext, ResourceDataContext
 
 
 class ParentBaseAuditLog(BaseAuditLog):
@@ -127,4 +127,33 @@ class RDMDraftAccessSettingsAuditLog(RDMRecordAccessSettingsAuditLog):
     id = "draft.access_settings_update"
     message_template = _(
         "User {user_id} updated access settings of {resource_type} {resource_id} via draft."
+    )
+
+
+class FileCreateAuditLog(BaseAuditLog):
+    """Audit log for file create."""
+
+    resource_type = "draft"
+
+    context = BaseAuditLog.context + [
+        FileContext(),
+    ]
+
+    id = "file.create"
+    message_template = _(
+        "User {user_id} created file {file_key} of {resource_type} {resource_id}."
+    )
+
+    metadata_schema = {
+        **BaseAuditLog.metadata_schema,
+        "file_key": ma.fields.String(required=True),
+    }
+
+
+class FileDeleteAuditLog(FileCreateAuditLog):
+    """Audit log for file delete."""
+
+    id = "file.delete"
+    message_template = _(
+        "User {user_id} deleted file {file_key} of {resource_type} {resource_id}."
     )
--- a/invenio_rdm_records/auditlog/context.py
+++ b/invenio_rdm_records/auditlog/context.py
@@ -38,3 +38,12 @@ class ResourceDataContext(object):
             "type": "draft" if isinstance(resource, RDMDraft) else "record",
         }
         dict_set(data, "metadata.triggered_by", triggered_by)
+
+
+class FileContext(object):
+    """Payload generator for setting file data."""
+
+    def __call__(self, data, **kwargs):
+        """Update data with file data."""
+        file_key = kwargs.get("file_key", None)
+        dict_set(data, "metadata.file_key", file_key)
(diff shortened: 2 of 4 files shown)


invenio_rdm_records/services/files/service.py
```diff
@@ -32,3 +38,23 @@ def _get_record(self, id_, identity, action, file_key=None):
...
 
         return record
+
+    @unit_of_work()
+    def commit_file(self, identity, id_, file_key, uow=None):
+        """Commit a file upload."""
+        result = super().commit_file(identity, id_, file_key, uow=uow)
+
+        uow.register(
+            AuditLogOp(FileCreateAuditLog.build(identity, id_, file_key=file_key))
```
OliverGeneser, line 48: Why did we choose to use the drafts id for file creation logs and not the parent.pid.pid_value as we do In the other logs?

invenio_rdm_records/auditlog/context.py
```diff
@@ -38,3 +38,12 @@ def __call__(self, data, **kwargs):
...
         }
         dict_set(data, "metadata.triggered_by", triggered_by)
+
+
+class FileContext(object):
+    """Payload generator for setting file data."""
+
+    def __call__(self, data, **kwargs):
+        """Update data with file data."""
+        file_key = kwargs.get("file_key", None)
```
OliverGeneser, line 48: nit; the marshmallow schema is required, but here it can possibly be None

invenio_rdm_records/auditlog/actions.py
```diff
@@ -128,3 +128,32 @@ class RDMDraftAccessSettingsAuditLog(RDMRecordAccessSettingsAuditLog):
     message_template = _(
         "User {user_id} updated access settings of {resource_type} {resource_id} via draft."
     )
+
+
+class FileCreateAuditLog(BaseAuditLog):
+    """Audit log for file create."""
+
+    resource_type = "draft"
```
OliverGeneser, line 136: Is this log only for the initial file creation? What if a record is modified and files are added?

invenio_rdm_records/services/files/service.py
sakshamarora1: For the share access logs, the metadata of the parent is updated. Here, the metadata of the files attached to the record are updated. And through the UI, we can only create or delete a file.

invenio_rdm_records/auditlog/actions.py
sakshamarora1: Then it's a new file and a new audit log entry, through the FileService `commit_file` method

0einstein0: so for a published record edit, file changes are logged independently of the publish action and only if the file is successfully uploaded? Because we have some cases where people upload very large files and they get stuck in pending state sometimes as they failed upload, so they have to be removed and uploaded again.

0einstein0 (approved):

invenio_rdm_records/auditlog/actions.py
sakshamarora1: Yes, independently of the publish and only if it passes through this service method successfully.

slint (approved):

slint:
Sorry, I should have left it in the review, but maybe a more general question on audit logs: is it possible for instances to enable only specific types of audit logs, or is it expected that we log everything, and the (soon to come) retention period will be responsible for cleaning-up? E.g. for Zenodo we would disable these entirely, purely from a volume standpoint.

sakshamarora1:
@slint It is possible to disable certain actions via this config flag.

https://github.com/inveniosoftware/invenio-audit-logs/blob/d4a06057df76423546e8adb28cddc8becbb3584e/invenio_audit_logs/config.py#L48-L56

Screenshot 2026-06-18 at 11 58 08

slint:
@sakshamarora1 @kpsherva @zzacharo I realised in the shower, I'm not a big fan of having a deny/disabled-list instead of an explicit allow-list. The issue is that as we add new audit log actions, instances have to update their config and if they miss out or forget might accidentally get an unwanted audit log event sneak-in.

If you don't mind, I'll put my clanker to add an allow-list version of the config, like e.g. `AUDIT_LOGS_ENABLED_ACTIONS = None`, where `None` (the default) means all actions are allowed.

slint:
FYI, here's the PR for having an allowlist config: https://github.com/inveniosoftware/invenio-audit-logs/pull/29

PR #2307, opened 2026-04-09
Overridable added to tabs in community header search app

Ducica:

### Description
A few fixes related to community header. I've split it into logically related commits, so that you can more easily reject or accept some. The fixes are:

1. ability to pass search app config to community header as a prop (imo this was probably simply forgotten when component was originally written).
2. Modal close button not working (ditto).
3. The third change is the largest. We need a good way to override those tabs, because somewhere we could have different number of tabs (1 or 3). The use case for us, is that we have some repositories where it is much stricter who can submit to which community, so we require some flexibility here. I've decided to make the overrides passed to search app via prop, and contained to the component, instead of pasting overrideStore.getAll to the overridable provider, as I thought this would make it cleaner and contained to the component itself. Also in some repositories, we could have 5 or more forms, and using global override store is not a good idea for such a use case. 

Please let us know what you think. 

Thanks,

Dusan S

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/components/CommunitySelectionModal/CommunitySelectionSearch.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/components/CommunitySelectionModal/CommunitySelectionSearch.js
@@ -7,7 +7,7 @@
 
 import { i18next } from "@translations/invenio_rdm_records/i18next";
 import React, { Component } from "react";
-import { OverridableContext, parametrize } from "react-overridable";
+import Overridable, { OverridableContext, parametrize } from "react-overridable";
 import {
   EmptyResults,
   Error,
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/components/CommunitySelectionModal/CommunitySelectionSearch.js
```diff
@@ -75,49 +77,57 @@ export class CommunitySelectionSearch extends Component {
...
             <Modal.Content as={Grid} className="m-0 pb-0 centered">
               {myCommunitiesEnabled && (
-                <Grid.Column
-                  mobile={16}
-                  tablet={8}
-                  computer={8}
-                  textAlign="left"
-                  floated="left"
-                  className="pt-0 pl-0"
+                <Overridable
```
mirekys: Container should not be conditional (rendered only when mycommunities enabled)

Ducica: You're right my bad

kpsherva (approved):

PR #2309, opened 2026-04-09
fix: fixed confusing feedback message when preview fails 

Ducica:

### Description
When using preview functionality, the draft first gets saved. If save returns some validation errors, the action in redux store ends up as DRAFT_HAS_VALIDATION_ERRORS, which in turns causes the message in form feedback to be Record saved with validation feedback in. This is very confusing for people, i.e. they clicked on preview and message has nothing to do with preview. The PR fixes it in a way consistent with DRAFT_PUBLISH_FAILED_WITH_VALIDATION_ERRORS for example.


ntarocco (approved):
Thank you!

PR #2310, opened 2026-04-09
fix: add policy_id to request metadata

### LINKED ISSUE #2313: fix: add policy_id to file modification request metadata
Add payload with policy_id to request metadata for File Modification, like it is done for Immediate Deletion.

mairasalazar:
:heart: Thank you for your review!

### Description

Adds payload with policy_id to request metadata for File Modification, like it is done for Immediate Deletion.

Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/2313

### Checklist

- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/community/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/community/code/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/community/code/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/community/code/best-practices/ui/) guidelines.

**Reminder**

### DIFF
--- a/invenio_rdm_records/requests/record_deletion.py
+++ b/invenio_rdm_records/requests/record_deletion.py
@@ -87,7 +87,7 @@ class AcceptAction(actions.AcceptAction):
 
         # For immediate deletions, we track the deletion policy ID in the tombstone
         policy_id = self.request.get("payload", {}).get("policy_id")
-        if policy_id:
+        if policy_id and policy_id not in ("request-deletion-v1",):
             # NOTE: Deletion policies are not a vocabulary, but we follow the same
             # format for consistency and potential extensions.
             tombstone_data["deletion_policy"] = {"id": policy_id}
--- a/invenio_rdm_records/requests/record_deletion.py
+++ b/invenio_rdm_records/requests/record_deletion.py
@@ -87,7 +87,7 @@ class AcceptAction(actions.AcceptAction):
 
         # For immediate deletions, we track the deletion policy ID in the tombstone
         policy_id = self.request.get("payload", {}).get("policy_id")
-        if policy_id and policy_id not in ("request-deletion-v1",):
+        if policy_id:
             # NOTE: Deletion policies are not a vocabulary, but we follow the same
             # format for consistency and potential extensions.
             tombstone_data["deletion_policy"] = {"id": policy_id}
--- a/invenio_rdm_records/services/services.py
+++ b/invenio_rdm_records/services/services.py
@@ -281,8 +281,8 @@ class RDMRecordService(RecordService):
             raise PermissionDeniedError()
 
         # Keep track of the immediate deletion policy ID
-        if deletion_policy := (immediate_deletion.policy or request_deletion.policy):
-            data["policy_id"] = deletion_policy.get("id")
+        if immediate_deletion.policy:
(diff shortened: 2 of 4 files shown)


invenio_rdm_records/services/services.py
```diff
@@ -324,6 +323,9 @@ def file_modification(self, identity, id_, data=None, uow=None, **kwargs):
         if disabled or forbidden:  # bail early
             raise PermissionDeniedError()
 
+        if immediate_file_mod.policy:
+            data["policy_id"] = immediate_file_mod.policy.get("id")
+
```
carlinmack:
```suggestion
        data["policy_id"] = immediate_file_mod.policy.get("id")
```

As we are `.allowed`, there must be a policy so we don't have to check for it first. Also we require there to be a policy so we have to set `policy_id` :)

carlinmack: Once you change it, I will also try this locally before approving

carlinmack (approved):
works as expected :)

Image

invenio_rdm_records/requests/record_deletion.py
```diff
@@ -87,7 +87,7 @@ def execute(self, identity, uow, **kwargs):
 
         # For immediate deletions, we track the deletion policy ID in the tombstone
         policy_id = self.request.get("payload", {}).get("policy_id")
-        if policy_id:
+        if policy_id and policy_id not in ("request-deletion-v1",):
```
mairasalazar: @carlinmack Added this because, if adding the policy_id to requested deletions, it would show removed_by the user who created the request (not the one that approved it). Perhaps is better to revert this and only have the policy_id for immediate deletions if this can cause some compatibility issues for instances with different policies 🤔

carlinmack: ah interesting thanks for figuring this out! maybe best to add a comment about the request deletions

carlinmack: also I think the check for `not in` is incorrect as the user might have custom request deletion polices. Maybe instead just don't use `request_deletion.policy`

invenio_rdm_records/services/services.py
```diff
@@ -324,6 +324,8 @@ def file_modification(self, identity, id_, data=None, uow=None, **kwargs):
         if disabled or forbidden:  # bail early
             raise PermissionDeniedError()
 
+        data["policy_id"] = immediate_file_mod.policy.get("id")
```
ntarocco: @mairasalazar what happens if the policy id is `None`/missing? can it be?
In that case, should we omit the `policy_id` field?

In general, I would limit as much as possible the usage of `.get()`, as it can lead to unexpected values. It is preferable to control if a value should be added or not, or how it should be handled, based on its existance.

mairasalazar: Will be more attentive with that, thank you! In this case, policy will have an id; otherwise an error would have been raised before in `request_policies.py`, so I just changed it to policy['id']

PR #2314, opened 2026-04-13
feat: add new RDMSubCommunity request class

OliverGeneser:

### DIFF
--- a/invenio_rdm_records/requests/__init__.py
+++ b/invenio_rdm_records/requests/__init__.py
@@ -13,6 +13,7 @@ from .community_submission import CommunitySubmission
 from .file_modification import FileModification
 from .quota_increase import QuotaIncrease
 from .record_deletion import RecordDeletion
+from .subcommunities import RDMSubCommunityInvitationRequest, RDMSubCommunityRequest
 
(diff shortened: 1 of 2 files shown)


slint (approved):

slint:
In theory this is still the "wrong place" since the endpoint is in app-rdm... Does it make sense to just move the new class there?

palkerecsenyi:
@slint while `app-rdm` might be a more technically correct place (since this request is not related to records), all other RDM request types are declared in this repo. We don't have a precedent for registering request types in app-rdm (I think?) so I feel like it would be a little clunky to create a new directory and everything just for this one override. Might be easier to just leave it in here even if it's not the "correct" place. What do you think?

slint:
Alright, I trust you zoomers for now, but we have to find a better solution for all requests... I remember there was an issue somewhere on handling requests definition, registration, and extensions a bit better...

palkerecsenyi:
Yes I also vaguely remember this issue but I can't seem to find it anymore...

palkerecsenyi:
@OliverGeneser is there a corresponding PR in app-rdm to override the config var?

Issue #2315, opened 2026-04-13
Flaky test due to new resource types


Ducica:
Package version 28.2.0

## Describe the bug
Here https://github.com/inveniosoftware/invenio-rdm-records/blob/50934b84a213baa0b94c4a0fc8c4fd4be6783a93/invenio_rdm_records/fixtures/demo.py#L185 there is an issue, because it is using a random resource type available in the vocabulary. 

Here, we have added some entries that contain tags "linkable" only https://github.com/inveniosoftware/invenio-rdm-records/blob/50934b84a213baa0b94c4a0fc8c4fd4be6783a93/invenio_rdm_records/fixtures/data/vocabularies/resource_types.yaml#L1075-L1096

Here https://github.com/inveniosoftware/invenio-rdm-records/blob/50934b84a213baa0b94c4a0fc8c4fd4be6783a93/invenio_rdm_records/records/api.py#L184

it is checking for depositable and if it is not the test fails 

```
app = <Flask 'invenio'>, location = pytest-location

  db = <SQLAlchemy ***localhost:5432/invenio>

  search_clear = <OpenSearch([{'host': 'localhost', 'port': 9200}])>

  vocabularies = None, users = [<User 7>, <User 8>]

  

      def test_create_fake_demo_invitation_requests(

          app, location, db, search_clear, vocabularies, users

      ):

          """Assert that demo invitation requests creation works without failing."""

          first_user_id = users[0].id

      

  >       create_demo_record(first_user_id, create_fake_record(), publish=True)

  

  tests/fixtures/test_cli.py:97: 

  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

  .venv/lib/python3.14/site-packages/celery/local.py:182: in __call__

      return self._get_current_object()(*a, **kw)

             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  .venv/lib/python3.14/site-packages/celery/app/task.py:411: in __call__

      return self.run(*args, **kwargs)

             ^^^^^^^^^^^^^^^^^^^^^^^^^

  invenio_rdm_records/fixtures/tasks.py:76: in create_demo_record

      draft = service.create(data=data, identity=identity)

              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[...]

PR #2317, opened 2026-04-14
fix: using only depositable resource type for fake data

Ducica:
* in case fake data ends up with resource type that is not
* depositable, relations check will fail and therefore the
* test using the fake data

### Description

Coming from here https://github.com/inveniosoftware/invenio-rdm-records/issues/2315#issuecomment-4238709248


tmorrell (approved):
Thanks!

PR #2320, opened 2026-04-15
fix(vcs): avoid failing on unrecognised license; include warning messages in notifications

### LINKED ISSUE #2: Make `invenio-github` support other VCS providers
This issue is part of https://github.com/inveniosoftware/product-rdm/issues/226 to add a GitLab integration. In order to reduce code duplication, the approach will be to adapt `invenio-github` and turn it into a "generic" module that supports any Version Control System (VCS) as long as it provides the necessary APIs and functionality. Implementations for specific VCSes like GitHub and GitLab will be provided in new contrib files.

# Stage 1

Aiming to complete a fully functional, production-ready, well-documented MVP. We will regard it as complete when:

- it provides an end-user experience equivalent to the current GitHub integration, preferably as similar as possible.
  - This has some scalability issues but for now we will avoid changing too much
- a clear migration script and guide are available and have been thoroughly tested with existing Zenodo data
- unit tests have been updated
- as many bugs fixed as possible

Work is split between several PRs to make reviewing easier. These will be merged into this repo's (`invenio-vcs`) `master` branch. We will only create a release once all the functionality is ready. Before the first release, `master` may contain incomplete unrunnable code. **For a snapshot of the latest runnable state of the VCS integration, please refer to [my fork's `master`](https://github.com/palkerecsenyi/invenio-vcs-new/tree/master)**.

- #1 
- #3 
- #4 
- #5 
- #6 
- #7 
- #8
- #10
- https://github.com/inveniosoftware/invenio-rdm-records/pull/2283
- https://github.com/inveniosoftware/invenio-rdm-records/pull/2284
- https://github.com/inveniosoftware/invenio-app-rdm/pull/3162

## Todo for stage 1

- [x] GitLab contrib. This is a priority as it's needed to test a lot of the other features (e.g. auth). It's very difficult to test e.g. OAuth without it.
- [x] OAuth user ID correlation
  - i.e. if the VCS provider uses the same OAuth server to authenticate the user as the Invenio instance, we should check the user IDs to make sure they match. This is useful for CDS-RDM where users will be able to link CERN GitLab, which uses the same CERN SSO.
  - We could express this through a more versatile hook function that returns whether/not we should accept the authenticated user.
  - **Update**: This can be done relatively easily by configuring a custom `info_serializer` handler in `invenio.cfg`. See the example for CDS: https://github.com/CERNDocumentServer/cds-rdm/pull/554
- [x] Sync VCS repositories straight into the `vcs_repositories` table instead of the OAuth remote user `extra_data` field. 
  - This will make querying a lot easier so we can paginate/search on the repository list page, which is currently very slow for users on e.g. GitLab instances where they have membership of thousands of repos due to group membership.
- [x] Check duplication for organisational/team repos if multiple people activate them
  - What happens if a user is deleted? How can we transfer the repos?
- [x] Repo name should not be unique individually. It is unique as a tuple of (provider_id,provider,name)
- [x] UI bug with menu not being able to differentiate between multiple dynamically-registered entries
    - For example: image
- [x] Unit tests
- [x] Documentation
- [x] Migration script and guide
- [x] Careful testing of DB migration for existing GitHub repos/releases
- [x] Some UI pages have not been adapted and continue to throw errors
- [x] JSONB extra_data in oauthclient
  - See https://github.com/inveniosoftware/invenio-oauthclient/pull/360
- [x] ~Correct handling of dependency in InvenioRDM~
  - We are keeping invenio-vcs as a mandatory dependency for now
- [x] Check permissions
- [x] Notifications on failed/successful archive
  - See https://github.com/inveniosoftware/product-rdm/issues/226#issuecomment-3197378210
  - If the release fails, we should still save the draft. We will send a notification to the user so they can edit the draft until the errors are resolved.
  - Requires implementing a number of classes for https://github.com/inveniosoftware/invenio-notifications
- [x] Check sync algorithm for race conditions and performance issues.
  - Copy over changes from https://github.com/inveniosoftware/invenio-github/pull/197
- [x] The license keys returned from GitLab are not compliant SPDX IDs so this can sometimes cause errors when publishing records that have license IDs that don't comply with InvenioRDM's list of SPDX IDs.
  - https://github.com/inveniosoftware/invenio-rdm-records/pull/2320

# Stage 2

The following features will **only be implemented in future PRs** once Stage 1 has been fully completed and merged:

- ~~Refresh token support~~ Done
    - In the existing GitHub impl we use access tokens which are [_non-expiring_ by default](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/refreshing-user-access-tokens?apiVersion=2022-11-28#about-user-access-tokens-that-expire). This is a security issue in case of a database leak and is recommended against by [RFC 6749](https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2).
    - A PR exists (https://github.com/inveniosoftware/invenio-oauthclient/pull/328) but needs some more work (last commit May 2024)
- Support for private repositories
- ~~"Link-only" OAuth without an option to "sign in with" a remote~~ Done
  - https://github.com/inveniosoftware/invenio-oauthclient/issues/365
- React + API-based UI for pagination/search of repos, using OpenSearch
  - See https://github.com/inveniosoftware/product-rdm/discussions/235#discussioncomment-14257358 for details\
- ~Selecting community to directly publish the repo to (especially on community-mandatory instances): https://github.com/inveniosoftware/product-rdm/discussions/235#discussioncomment-14484598~ Done
- Propagate permissions so users who have access to a repo also have access to records created from releases
  - https://github.com/inveniosoftware/product-rdm/discussions/235#discussioncomment-14497837
- Correct handling of orphaned repos
  - Right now when a user disconnects their VCS accounts, we disable the hook on all repos they have access to. Obviously this is the wrong behaviour (e.g. what if there are other users still left connected that have access to the repo) so we need to implement something more logical.
- Allow customising the relation type on records created from repos on InvenioRDM.
  - Currently, we add the repo as a Related Identifier with a relation type of `is supplement to` which seems a little unusual. We should consider whether this is a reasonable default to have, and probably allow for more easily customising it via a config variable.
- Badge redirect doesn't work if user is not signed in
- When using the [config override](https://github.com/palkerecsenyi/invenio-vcs/blob/8c795d8b7817b32a96d1cef8102fccf2ad19f07e/invenio_vcs/providers.py#L73) the OAuth remote app config is not updated as this is initialised before the override function is called. The remote app config is copied not referenced, so when we update our config it doesn't get passed down. This makes it hard to e.g. change the `base_url` without editing `invenio.cfg`.
- ~Licenses are not being synced for GitLab, which makes all records be CC licensed by default. This is slightly tricky to fix due to GitLab API limitations.~ Done
  - https://github.com/palkerecsenyi/invenio-vcs/blob/a057ba82e3c6d669d4781b7bbd2264506cebc314/invenio_vcs/contrib/gitlab.py#L278-L280
  - https://gitlab.com/gitlab-org/gitlab/-/issues/580429

# Migration considerations
The current plan for the GitHub to VCS migration is as follows:

- We merge all the code into the `invenio-github` `feature/vcs` branch. Once it's ready, we move that branch to a new repository, `invenio-vcs`. This is also published as the `invenio-vcs` PyPI package.
- We drop support for `invenio-github` and remove the mandatory dependency in `invenio-rdm-records`. However, we continue to allow it as an optional dependency (on the instance level), so users do not have migrate immediately before the official RDM v14 release.
  - To allow for this, we keep the old GitHub bindings in `invenio-rdm-records`. The [`GITHUB_RELEASE_CLASS`](https://github.com/inveniosoftware/invenio-app-rdm/blob/e18f3fcbab367e59c6859ae1ed4a6bee0a54d603/invenio_app_rdm/config.py#L1484) config var (now renamed to `VCS_RELEASE_CLASS`) can be used to set the old bindings.
- `invenio-vcs` will be an optional dependency at the instance level. To activate it, it needs to be installed manually, which the documentation will explain how to do. Both `invenio-github` and `invenio-vcs` will have checks to ensure they cannot be installed simultaneously.
- `invenio-vcs` will check if the `github_repositories`/`github_releases` tables (which exist on all InvenioRDM instances) are empty. If not, it will migrate the data into the new tables. Otherwise, it will simply initialise the new tables as empty.
- We will need to ensure `invenio-github` is compatible with the latest `invenio-rdm-records` (e.g. OAuthClient versions need to be the same).

palkerecsenyi:
Added a system of "warnings" that can be included in repository release notifications. To keep things simple, these are not stored in the DB for the release (unlike full errors), they are just included in the notification message. Right now they are only used for missing contributors and missing licenses, but more can be added later (e.g. if we improve the repository metadata extraction to support more features).

### DIFF
--- a/invenio_rdm_records/notifications/vcs.py
+++ b/invenio_rdm_records/notifications/vcs.py
@@ -64,10 +64,12 @@ class RepositoryReleaseSuccessNotificationBuilder(RepositoryReleaseNotificationB
         generic_repository: GenericRepository,
         generic_release: GenericRelease,
         record,
+        warnings: list[str],
     ):
         """Build the notification."""
         notification = super().build(provider, generic_repository, generic_release)
         notification.context["record"] = EntityResolverRegistry.reference_entity(record)
+        notification.context["warnings"] = warnings
         return notification
 
     context = [EntityResolve(key="record")]
@@ -150,6 +152,7 @@ class RepositoryReleaseCommunitySubmittedNotificationBuilder(
         generic_release: GenericRelease,
         request,
         community,
+        warnings: list[str],
     ):
         """Build the notification."""
         notification = super().build(provider, generic_repository, generic_release)
@@ -159,6 +162,7 @@ class RepositoryReleaseCommunitySubmittedNotificationBuilder(
         notification.context["community"] = EntityResolverRegistry.reference_entity(
             community
         )
+        notification.context["warnings"] = warnings
         return notification
 
     context = [EntityResolve(key="request"), EntityResolve(key="community")]
--- a/invenio_rdm_records/services/vcs/metadata.py
+++ b/invenio_rdm_records/services/vcs/metadata.py
@@ -85,15 +85,10 @@ class RDMReleaseMetadata(object):
             title=self.title,
             resource_type={"id": "software"},
             creators=self.contributors,
-            publisher=current_app.config.get("APP_RDM_DEPOSIT_FORM_DEFAULTS").get(
+            publisher=current_app.config.get("APP_RDM_DEPOSIT_FORM_DEFAULTS", {}).get(
                 "publisher", "CERN"
             ),
         )
-
-    @property
-    def repo_license(self):
-        """Get license from repository, if any."""
-        return self.rdm_release.generic_repo.license_spdx
 
     @property
     def contributors(self):
--- a/invenio_rdm_records/services/vcs/release.py
+++ b/invenio_rdm_records/services/vcs/release.py
@@ -7,6 +7,8 @@
 # it under the terms of the MIT License; see LICENSE file for more details.
 """VCS release API implementation."""
 
+from __future__ import annotations
+
 from flask import current_app
 from invenio_access.permissions import authenticated_user, system_identity
 from invenio_access.utils import get_identity
@@ -14,10 +16,14 @@ from invenio_db import db
 from invenio_drafts_resources.resources.records.errors import DraftNotCreatedError
 from invenio_i18n import lazy_gettext as _
 from invenio_notifications.services.uow import NotificationOp
+from invenio_pidstore.errors import PIDDoesNotExistError
 from invenio_records_resources.services.uow import UnitOfWork
 from invenio_vcs.api import VCSRelease
 from invenio_vcs.errors import CustomVCSReleaseNoRetryError
-from invenio_vcs.models import ReleaseStatus
+from invenio_vcs.models import Release, ReleaseStatus
+from invenio_vcs.providers import RepositoryServiceProvider
+from invenio_vocabularies.proxies import current_service as current_vocabularies_service
+from sqlalchemy.exc import NoResultFound
 
 from invenio_rdm_records.notifications.vcs import (
     RepositoryReleaseCommunityRequiredNotificationBuilder,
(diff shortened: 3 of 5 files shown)


zzacharo (approved):
LGTM, my only idea to consider is either instead of throwing a warning for the license if not found, we select a default one.

PR #2324, opened 2026-04-16
feat(reviews): add config for record version review policy

### LINKED ISSUE #2305: Update config structure for record review requests
Right now, the config for record reviews is [quite minimal](https://github.com/inveniosoftware/invenio-rdm-records/blob/66df87d36e893432ee3cbe6eff29a9ab7c0bbbf8/invenio_rdm_records/config.py#L145-L153). We should update this structure in a way that is as backward-compatible as possible, while adding support for **configuring how new versions trigger review requests**.

See https://github.com/inveniosoftware/product-rdm/discussions/241 for a more precise description of the intended functionality.

---

For this issue, we should implement the necessary changes to the config on the backend (without the rest of the functionality) in a backward-compatible way (preferably).

**Importantly, the existing default behaviour should be unchanged.** Without changes to configuration, the current behaviour (the first bullet point) should apply.

palkerecsenyi:
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/2305

**I am happy to make changes to the structure of this config or the general approach to the task of allowing per-record-version reviews. We should try to make changes now before we "lock in" the structure of the config.**

See the [discussion](https://github.com/inveniosoftware/product-rdm/discussions/241) for more details.

---

* Added a new class to define a policy for whether a new version draft of an existing record should trigger a community submission request.
* The default behaviour is to never trigger a request for new versions (which is the current behaviour).
* The implementation of this policy is in https://github.com/inveniosoftware/invenio-rdm-records/pull/2325
* The configured policy can be changed easily using a new config variable by inheriting from the class and overriding the method.

### DIFF
--- a/invenio_rdm_records/config.py
+++ b/invenio_rdm_records/config.py
@@ -26,6 +26,7 @@ from invenio_records_resources.services.records.queryparser.transformer import (
 import invenio_rdm_records.services.communities.moderation as communities_moderation
 from invenio_rdm_records.services.components.pids import validate_optional_doi
 from invenio_rdm_records.services.components.verified import UserModerationHandler
+from invenio_rdm_records.services.review.policy import RecordVersionReviewPolicy
 
 from . import tokens
 from .requests.community_inclusion import CommunityInclusion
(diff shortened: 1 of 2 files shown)


palkerecsenyi:
Already included and merged in #2325

PR #2325, opened 2026-04-16
per-record-version requests (model + service updates)

### LINKED ISSUE #2312: Update data model to support reviews per record version
Based on the discussed [supported features](https://github.com/inveniosoftware/product-rdm/discussions/241) and [config](https://github.com/inveniosoftware/invenio-rdm-records/issues/2305), update the RDM data model to support having a submission review request per each record version.

Currently, we store the review that corresponds to a record in the [`review` field of the `RDMParent` record](https://github.com/inveniosoftware/invenio-rdm-records/blob/afc9603d04bf4cd4fceddae5d984e156ccbf3873/invenio_rdm_records/records/api.py#L92). We need to discuss the best thing to do here. We should keep backwards compatibility and minimise the complexity of a migration, so **removing this existing `review` field is probably not a possibility**. In any case, the migration should be as simple as possible and the current behaviour (reviews only for the first version) should be retained as the default option unless the instance/community configures it otherwise.

We will also need to modify the [`rdm_parents_community` table](https://github.com/inveniosoftware/invenio-requests/issues/568#issuecomment-3893302643) which currently only allows for one request per record / community combo.

Maybe we should add a `review` field to `RDMDraft` (and correspondingly also to `RDMRecord`)? This is up for discussion.

- [ ] Implement the data model changes
- [ ] Document the needed deployment steps, i.e. any Alembic migrations / OpenSearch template updates

### LINKED ISSUE #2323: Update review service to support reviews per record version as configured
Based on the configuration in https://github.com/inveniosoftware/invenio-rdm-records/issues/2305 and the data model changes in https://github.com/inveniosoftware/invenio-rdm-records/issues/2312, we should modify the existing `ReviewService` in this repository to support creating a review request for new record versions beyond the first one.

If possible, we should aim to keep `CommunitySubmission` as the request type; we should avoid creating a new request type for non-first-version submission requests, as hopefully the existing one can be modified in a way that allows this.

All new functionality must be backward-compatible and must not require database migrations beyond the ones already done in #2312.

palkerecsenyi:
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/2312
Closes https://github.com/inveniosoftware/invenio-rdm-records/issues/2323

Also includes commit from https://github.com/inveniosoftware/invenio-rdm-records/pull/2324 (unmerged right now).

---

- Updated the data models to include a `review` field on the record. The existing `review` field on the parent record is retained, and a new method `get_own_or_parent_review` on the record is used (nearly) everywhere the review was referenced. This method prefers to return the record's review, but returns the parent's review if it doesn't exist. This way we retain compatibility with existing records.
- Updated all methods that reference the review to ensure compatibility with the parent/record review.
- Added the `new_version` method to the review service component, so that a draft community submission request is always added to a new version draft when the policy requires it. This review can be submitted via the frontend.
- Updated the frontend deposit form components so that a "submit review" button is shown for a new-version draft instead of a "publish" button where the policy requires a review.
  - If the draft is a new version and the draft has a review on it, we assume that the "submit review" button should be shown. If no review is present, the "publish" button continues to be shown.
- Added unit tests

---

**I have _not_ tested the migration** from having the `review` on the parent to on the record. I have also not tested support for existing parent's with `review`s with the new code.

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
@@ -148,11 +148,7 @@ function getSelectedCommunityMetadata(record, selectedCommunity) {
  * @param {object} selectedCommunity: the selected community, `null` to deselect.
  * @returns a new state for the deposit form
  */
-export function computeDepositState(
-  record,
-  permissions,
-  selectedCommunity = undefined
-) {
+export function computeDepositState(record, selectedCommunity = undefined) {
   const depositStatusAllowsReviewDeletion = hasStatus(
     record,
     DepositStatus.allowsReviewDeletionStates
@@ -290,7 +286,6 @@ const depositReducer = (state = {}, action) => {
         },
         editorState: computeDepositState(
           action.payload.data,
-          state.permissions,
           state.editorState.selectedCommunity
         ),
         errors: {},
@@ -307,7 +302,6 @@ const depositReducer = (state = {}, action) => {
         },
         editorState: computeDepositState(
           action.payload.data,
-          state.permissions,
           state.editorState.selectedCommunity
         ),
         errors: {},
@@ -326,7 +320,6 @@ const depositReducer = (state = {}, action) => {
         },
         editorState: computeDepositState(
           action.payload.data,
-          state.permissions,
           state.editorState.selectedCommunity
         ),
         errors: { ...action.payload.errors },
@@ -366,11 +359,7 @@ const depositReducer = (state = {}, action) => {
       return {
         ...state,
         record: recordCopy,
-        editorState: computeDepositState(
-          recordCopy,
-          state.permissions,
-          action.payload.community
-        ),
+        editorState: computeDepositState(recordCopy, action.payload.community),
       };
     }
     case SET_DOI_NEEDED: {
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
@@ -36,6 +36,7 @@ import {
 export class DepositStatus {
   static DRAFT = "draft";
   static NEW_VERSION_DRAFT = "new_version_draft";
+  static NEW_VERSION_DRAFT_WITH_REVIEW = "new_version_draft_with_review";
   static DRAFT_WITH_REVIEW = "draft_with_review";
   static IN_REVIEW = "in_review";
   static DECLINED = "declined";
@@ -58,7 +59,6 @@ export class DepositStatus {
   static disallowsSubmitForReviewStates = [
     DepositStatus.PUBLISHED,
     DepositStatus.IN_REVIEW,
-    DepositStatus.NEW_VERSION_DRAFT,
   ];
 }
 
@@ -66,11 +66,20 @@ function hasStatus(record, statuses = []) {
   return statuses.includes(record.status);
 }
 
+// Returns true if the current record version is published, or if this is not the first version record
+// within its parent.
+// This essentially corresponds to whether the parent record has been published.
+function isParentAlreadyPublished(record) {
+  return hasStatus(record, [DepositStatus.PUBLISHED]) || record.versions?.index !== 1;
+}
+
 function getSelectedCommunityMetadata(record, selectedCommunity) {
   switch (selectedCommunity) {
     case undefined: {
       // when `undefined`, retrieve the community from the record, if previously selected
-      const reviewCommunityId = record.parent?.review?.receiver?.community;
+      const reviewCommunityId =
+        record.parent?.review?.receiver?.community ??
+        record.review?.receiver?.community;
       const defaultCommunityId = record.parent?.communities?.default;
       const hasCommunity = reviewCommunityId || defaultCommunityId;
       if (!hasCommunity) {
@@ -78,13 +87,8 @@ function getSelectedCommunityMetadata(record, selectedCommunity) {
         return undefined;
       }
 
-      const alreadyPublished = hasStatus(record, [
-        DepositStatus.PUBLISHED,
-        DepositStatus.NEW_VERSION_DRAFT,
-      ]);
-
       // record should be expanded
-      return alreadyPublished
+      return isParentAlreadyPublished(record)
         ? record.expanded.parent.communities.default
         : record.expanded.parent.review.receiver;
     }
@@ -112,10 +116,15 @@ function getSelectedCommunityMetadata(record, selectedCommunity) {
  *     - user has deselected a community
  *     - the draft status is one of `DepositStatus.allowsReviewDeletionStates`
  * - `actions.communityStateMustBeChecked`: true if one of the `shouldUpdateReview` or `shouldDeleteReview` is true
- * - `ui.showSubmitForReviewButton`: true if all of the following are true:
- *     - user has selected a community
- *     - the associated review for the selected community is not declined/expired
- *     - the record is not published
+ * - `ui.showSubmitForReviewButton`:
+ *     - true if all of the following are true:
+ *         - user has selected a community
+ *         - the associated review for the selected community is not declined/expired
+ *         - the record is not published
+ *     - OR true if all of the following are true:
+ *         - parent record belongs to a community
+ *         - record is a new version draft
+ *         - user does not have permission to skip community review for a new record version
  * - `ui.showDirectPublishButton`: true if all of the following are true:
  *     - user has selected a community
  *     - user can direct publish to a community
@@ -139,7 +148,11 @@ function getSelectedCommunityMetadata(record, selectedCommunity) {
  * @param {object} selectedCommunity: the selected community, `null` to deselect.
  * @returns a new state for the deposit form
  */
-export function computeDepositState(record, selectedCommunity = undefined) {
+export function computeDepositState(
+  record,
+  permissions,
+  selectedCommunity = undefined
+) {
   const depositStatusAllowsReviewDeletion = hasStatus(
     record,
     DepositStatus.allowsReviewDeletionStates
@@ -157,7 +170,10 @@ export function computeDepositState(record, selectedCommunity = undefined) {
   const _selectedCommunity = getSelectedCommunityMetadata(record, selectedCommunity);
   const communityIsSelected = !_isEmpty(_selectedCommunity);
 
-  const draftReview = record?.parent?.review;
+  const parentAlreadyPublished = isParentAlreadyPublished(record);
+  // If the parent is already published, the review can only be on the record itself (for new-version reviews).
+  // Otherwise, the review must be on the parent.
+  const draftReview = parentAlreadyPublished ? record?.review : record?.parent?.review;
 
   // check if the selected community has a request created
   const isReviewForSelectedCommunityCreated =
@@ -181,25 +197,28 @@ export function computeDepositState(record, selectedCommunity = undefined) {
   const _showDirectPublishButton =
     communityIsSelected &&
     !hasStatus(record, [DepositStatus.PUBLISHED, DepositStatus.NEW_VERSION_DRAFT]) &&
-    _selectedCommunity.ui.permissions.can_include_directly;
+    _selectedCommunity.ui?.permissions.can_include_directly;
 
   // show submit for review button conditions extracted to be reused
   const _showSubmitReviewButton =
     communityIsSelected &&
-    !isReviewForSelectedCommunityDeclinedOrExpired &&
+    (!isReviewForSelectedCommunityDeclinedOrExpired || parentAlreadyPublished) &&
     !hasStatus(record, [DepositStatus.PUBLISHED, DepositStatus.NEW_VERSION_DRAFT]);
 
   // show community selection button conditions extracted to be reused
-  const _showCommunitySelectionButton = !hasStatus(record, [
-    DepositStatus.PUBLISHED,
-    DepositStatus.NEW_VERSION_DRAFT,
-  ]);
+  const _showCommunitySelectionButton =
+    !hasStatus(record, [
+      DepositStatus.PUBLISHED,
+      DepositStatus.NEW_VERSION_DRAFT,
+      DepositStatus.NEW_VERSION_DRAFT_WITH_REVIEW,
+    ]) && !(isReviewForSelectedCommunityDeclinedOrExpired && parentAlreadyPublished);
 
   const shouldUpdateReview =
     communityIsSelected &&
     depositStatusAllowsReviewUpdate &&
     !isReviewForSelectedCommunityCreated &&
-    !isReviewForSelectedCommunityDeclinedOrExpired;
+    !isReviewForSelectedCommunityDeclinedOrExpired &&
+    !parentAlreadyPublished;
 
   const shouldDeleteReview = !communityIsSelected && depositStatusAllowsReviewDeletion;
 
@@ -215,8 +234,14 @@ export function computeDepositState(record, selectedCommunity = undefined) {
       showSubmitForReviewButton: _showSubmitReviewButton,
       showDirectPublishButton: _showDirectPublishButton,
       disableSubmitForReviewButton:
-        _showSubmitReviewButton && depositStatusDisallowsSubmitForReview,
-      showChangeCommunityButton: isReviewForSelectedCommunityDeclinedOrExpired,
+        _showSubmitReviewButton &&
+        (depositStatusDisallowsSubmitForReview ||
+          isReviewForSelectedCommunityDeclinedOrExpired),
+      // Can only change the community if the parent record isn't already published.
+      showChangeCommunityButton:
+        isReviewForSelectedCommunityDeclinedOrExpired && record.versions.index === 1,
+      requireDeletionForDeclinedDraft:
+        isReviewForSelectedCommunityDeclinedOrExpired && record.versions.index !== 1,
       showCommunitySelectionButton: _showCommunitySelectionButton,
       showCommunityHeader: !isRecordPublishedWithoutOrUnresolvedCommunity,
       disableCommunitySelectionButton: _disableCommunitySelectionButton,
@@ -265,6 +290,7 @@ const depositReducer = (state = {}, action) => {
         },
         editorState: computeDepositState(
           action.payload.data,
+          state.permissions,
           state.editorState.selectedCommunity
         ),
         errors: {},
@@ -281,6 +307,7 @@ const depositReducer = (state = {}, action) => {
         },
         editorState: computeDepositState(
           action.payload.data,
+          state.permissions,
           state.editorState.selectedCommunity
         ),
         errors: {},
@@ -299,6 +326,7 @@ const depositReducer = (state = {}, action) => {
         },
         editorState: computeDepositState(
           action.payload.data,
+          state.permissions,
           state.editorState.selectedCommunity
         ),
         errors: { ...action.payload.errors },
@@ -338,7 +366,11 @@ const depositReducer = (state = {}, action) => {
       return {
         ...state,
         record: recordCopy,
-        editorState: computeDepositState(recordCopy, action.payload.community),
+        editorState: computeDepositState(
+          recordCopy,
+          state.permissions,
+          action.payload.community
+        ),
       };
     }
     case SET_DOI_NEEDED: {
--- a/invenio_rdm_records/records/api.py
+++ b/invenio_rdm_records/records/api.py
@@ -93,6 +93,12 @@ class RDMParent(ParentRecordBase):
         Request,
         keys=["type", "receiver", "status"],
     )
+    """
+    Deprecated: We have now moved to storing the review on the draft record, in order to support reviews per-record-version.
+    This field is retained for backward-compatibility.
+
+    The field will remain for existing records but will not be set for new ones.
+    """
 
     communities = CommunitiesField(models.RDMParentCommunity)
 
@@ -284,6 +290,22 @@ class CommonFieldsMixin:
 
     #: Custom fields system field.
     custom_fields = DictField(clear_none=True, create_if_missing=True)
+
+    review = RelatedRecord(
+        Request,
+        keys=["type", "receiver", "status"],
+    )
+
+    def get_review(self):
+        """
+        We used to store the review relation on the parent record, and we still do for existing/old records.
+
+        New records store it on the draft directly.
+        This method returns the review on the draft, or on the parent if it doesn't exist.
+
+        If the review does not exist on the record or the parent, None is returned.
+        """
+        return self.review if self.review is not None else self.parent.review
 
 
 #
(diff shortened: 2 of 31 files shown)


invenio_rdm_records/records/api.py
```diff
@@ -93,6 +93,12 @@ class RDMParent(ParentRecordBase):
         Request,
         keys=["type", "receiver", "status"],
     )
+    """
+    Deprecated: We have now moved to storing the review on the draft record, in order to support reviews per-record-version.
+    This field is retained for backward-compatibility.
+
+    The field will remain for existing records but will not be set for new ones.
+    """
```
palkerecsenyi, line 100: Remove message: for newly submitted first-version records we want to keep the review in the parent always. the record review is only for new versions if the policy requires it.

invenio_rdm_records/services/review/policy.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# Copyright (C) 2026 CERN.
+#
+# Invenio RDM Records is free software; you can redistribute it and/or modify
+# it under the terms of the MIT License; see LICENSE file for more details.
+
+"""Policy configuration to define which versions of records require community review."""
+
+
+class RecordVersionReviewPolicy:
```
palkerecsenyi: `NewRecordVersionReviewPolicy`

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
```diff
@@ -85,7 +86,7 @@ function getSelectedCommunityMetadata(record, selectedCommunity) {
       // record should be expanded
       return alreadyPublished
         ? record.expanded.parent.communities.default
-        : record.expanded.parent.review.receiver;
+        : record.review?.receiver ?? record.expanded.parent.review?.receiver;
```
palkerecsenyi: Keep this as was - if already published, the review has to be on the record. If not published then the review has to be on the parent.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
```diff
@@ -138,7 +144,11 @@ function getSelectedCommunityMetadata(record, selectedCommunity) {
  * @param {object} selectedCommunity: the selected community, `null` to deselect.
  * @returns a new state for the deposit form
  */
-export function computeDepositState(record, selectedCommunity = undefined) {
+export function computeDepositState(
+  record,
+  permissions,
```
palkerecsenyi: This is removed in next commit (double check)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
```diff
@@ -156,17 +166,17 @@ export function computeDepositState(record, selectedCommunity = undefined) {
   const _selectedCommunity = getSelectedCommunityMetadata(record, selectedCommunity);
   const communityIsSelected = !_isEmpty(_selectedCommunity);
 
-  const draftReview = record?.parent?.review;
+  const draftReview = record?.expanded?.review ?? record?.expanded?.parent?.review;
```
palkerecsenyi: Also make conditional on status

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
```diff
@@ -156,17 +166,17 @@ export function computeDepositState(record, selectedCommunity = undefined) {
...
 
   // check if the selected community has a request created
   const isReviewForSelectedCommunityCreated =
     hasStatus(record, [DepositStatus.DRAFT_WITH_REVIEW]) &&
-    draftReview?.receiver?.community === _selectedCommunity?.id;
+    draftReview?.receiver?.id === _selectedCommunity?.id;
```
palkerecsenyi: Double check why this is needed

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
palkerecsenyi: Check if we defo need `expanded` here

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
```diff
@@ -182,11 +192,17 @@ export function computeDepositState(record, selectedCommunity = undefined) {
     !hasStatus(record, [DepositStatus.PUBLISHED, DepositStatus.NEW_VERSION_DRAFT]) &&
     _selectedCommunity.ui.permissions.can_include_directly;
 
+  const isRecordNewVersionDraftWithReview =
+    communityIsSelected &&
+    hasStatus(record, [DepositStatus.NEW_VERSION_DRAFT]) &&
+    draftReview?.receiver?.id === _selectedCommunity?.id;
```
palkerecsenyi: Double check, maybe we can use the record parent default community directly instead of the selected one.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
```diff
@@ -182,11 +192,17 @@ export function computeDepositState(record, selectedCommunity = undefined) {
...
+
   // show submit for review button conditions extracted to be reused
   const _showSubmitReviewButton =
-    communityIsSelected &&
-    !isReviewForSelectedCommunityDeclinedOrExpired &&
-    !hasStatus(record, [DepositStatus.PUBLISHED, DepositStatus.NEW_VERSION_DRAFT]);
+    (communityIsSelected &&
+      !isReviewForSelectedCommunityDeclinedOrExpired &&
+      !hasStatus(record, [DepositStatus.PUBLISHED, DepositStatus.NEW_VERSION_DRAFT])) ||
+    isRecordNewVersionDraftWithReview;
```
palkerecsenyi: If the review for a new version is declined, you cannot resubmit the same new version draft. You have to discard the draft and create a new one.

palkerecsenyi: Rename to `isNewRecordVersionDraftWithReview`

invenio_rdm_records/records/systemfields/draft_status.py
```diff
@@ -60,7 +60,12 @@ def __get__(self, record, owner=None):
         if record is None:
             return self  # returns the field itself.
 
-        review = getattr(record.parent, "review")
+        review = getattr(record, "review")
+        if review is None:
+            # The `review` field of `RDMParent` is deprecated and unused for new records.
+            # We are keeping this for backward-compatibility with existing records.
+            review = getattr(record.parent, "review")
```
palkerecsenyi: Move this into `if is_published:` to decide whether to get from the parent or the record.

invenio_rdm_records/records/systemfields/draft_status.py
palkerecsenyi, line 35: This needs to be conditional on whether it's a new version or a new record.

invenio_rdm_records/requests/community_submission.py
```diff
@@ -59,24 +60,31 @@ def execute(self, identity, uow, **kwargs):
         # The curator (receiver) should still have access, via the community
         # The creator (uploader) should also still have access, because
         # they're the uploader
-        draft.parent.review = None
+        # This is kept for backward compatibility with records that were submitted
+        # before the `review` field was moved to `RDMDraft`.
+        if draft.review is not None:
+            draft.review = None
+        if draft.parent.review is not None:
+            draft.parent.review = None
```
palkerecsenyi, line 69: Don't need the if statements

invenio_rdm_records/requests/community_submission.py
```diff
@@ -59,24 +60,31 @@ def execute(self, identity, uow, **kwargs):
...
-        draft.parent.communities.add(
-            community, request=self.request, default=is_default
-        )
-
-        if getattr(community, "parent", None):
-            draft.parent.communities.add(community.parent, request=self.request)
+        # Only add to the parent record's community list if it hasn't already been added.
+        # E.g. if this request is for a new version of an existing published record, the parent record
+        # will already belong to the community, so we don't want to add it again.
+        if str(community.id) not in draft.parent.communities.ids:
```
palkerecsenyi: Simplify: if the draft status is new_version_draft, we don't add a community (make that assumption). Otherwise we do add.

invenio_rdm_records/requests/community_submission.py
```diff
@@ -59,24 +60,31 @@ def execute(self, identity, uow, **kwargs):
...
+            draft.parent.communities.add(community, default=is_default)
+            if getattr(community, "parent", None):
+                draft.parent.communities.add(community.parent)
 
         uow.register(
             ParentRecordCommitOp(draft.parent, indexer_context=dict(service=service))
         )
+        uow.register(RecordCommitOp(draft, indexer=service.indexer))
```
palkerecsenyi: Most of the time this will be unnecessary so maybe make it conditional

invenio_rdm_records/requests/community_submission.py
```diff
@@ -109,10 +117,17 @@ def execute(self, identity, uow):
         # TODO: this shouldn't be required BUT because of the caching mechanism
         # in the review systemfield, the review should be set with the updated
         # request object
-        draft.parent.review = self.request
-        uow.register(
-            ParentRecordCommitOp(draft.parent, indexer_context=dict(service=service))
-        )
+        if draft.review is not None:
```
palkerecsenyi: Flip the conditions

invenio_rdm_records/requests/community_submission.py
```diff
@@ -130,11 +145,19 @@ def execute(self, identity, uow):
         # Remove draft from request
         # Same reasoning as in 'decline'
         draft = self.request.topic.resolve()
-        draft.parent.review = None
-        uow.register(
-            ParentRecordCommitOp(draft.parent, indexer_context=dict(service=service))
-        )
         super().execute(identity, uow)
+
+        if draft.review is not None:
```
palkerecsenyi: Flip

invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -450,9 +450,10 @@ def mask_removed_by(self, obj, orig, **kwargs):
     @pre_dump
     def add_communities_permissions_and_roles(self, obj, **kwargs):
         """Inject current user's permission to community receiver."""
-        receiver = (
-            obj.get("expanded", {}).get("parent", {}).get("review", {}).get("receiver")
-        )
+        receiver = obj.get("expanded", {}).get("review", {}).get("receiver") or obj.get(
```
palkerecsenyi: Flip

invenio_rdm_records/services/access/service.py
```diff
@@ -118,11 +118,11 @@ def expandable_fields(self):
     # Update parent request on access changes
     #
 
-    def _update_parent_request(self, parent, uow):
+    def _update_record_request(self, record, uow):
         """Update the parent record request."""
```
palkerecsenyi: Update docstring

invenio_rdm_records/services/components/review.py
```diff
@@ -20,13 +24,15 @@ class ReviewComponent(ServiceComponent):
 
     def create(self, identity, data=None, record=None, **kwargs):
         """Create the review if requested."""
-        data = data.get("parent", {}).get("review")
+        record_review = data.get("review")
+        parent_review = data.get("parent", {}).get("review")
+        data = record_review or parent_review
```
palkerecsenyi: Only need parent here, the record.review should not be psossible

invenio_rdm_records/services/components/review.py
```diff
@@ -57,3 +61,30 @@ def publish(self, identity, draft=None, record=None, **kwargs):
...
+
+    def new_version(self, identity, draft=None, record=None):
+        """
+        Create a review for a new record version if required for the identity.
+
+        If the identity does not have the `skip_review_for_new_version` permission, each new version of existing published records
+        they create must go through a review by the parent record's default community.
+        """
+        assert draft is not None and record is not None
+        record = current_rdm_records_service.read(identity, id_=record.get("id"))
```
palkerecsenyi: Double check if we really need to read again

invenio_rdm_records/services/review/service.py
```diff
@@ -87,9 +91,9 @@ def create(self, identity, data, record, uow=None):
             uow=uow,
         )
 
-        # Set the request on the record and commit the record
-        record.parent.review = request_item._request
```
palkerecsenyi, line 91: Change this: assign to the parent for brand-new records and to the record for new versions

invenio_rdm_records/services/review/service.py
```diff
@@ -61,7 +65,7 @@ def _validate_request_type(self, request_type):
     @unit_of_work()
     def create(self, identity, data, record, uow=None):
```
palkerecsenyi: add kwarg `assign_request_to_record=False` and clarify in the docstring

invenio_rdm_records/services/review/service.py
```diff
@@ -140,18 +147,24 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
-        if draft.parent.review.is_open:
+        if review.is_open:
             raise ReviewStateError(_("An open review cannot be deleted."))
 
         # Keep the request when not open or not closed so that the user can see
         # the request's events. The request is deleted only when in `draft`
         # status
-        if not (draft.parent.review.is_closed or draft.parent.review.is_open):
-            current_requests_service.delete(identity, draft.parent.review.id, uow=uow)
+        if not review.is_closed:
```
palkerecsenyi: Fix the condition (`is_open`)

invenio_rdm_records/services/review/service.py
```diff
@@ -140,18 +147,24 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
+            current_requests_service.delete(identity, review.id, uow=uow)
+
         # Unset on record
-        draft.parent.review = None
-        uow.register(ParentRecordCommitOp(draft.parent))
-        uow.register(RecordIndexOp(draft, indexer=self.indexer))
+        if draft.review is not None:
+            draft.review = None
+            uow.register(RecordCommitOp(draft, indexer=self.indexer))
```
palkerecsenyi: Move outside of if statemenet

invenio_rdm_records/services/config.py
```diff
@@ -589,6 +590,9 @@ class RDMRecordServiceConfig(RecordServiceConfig, ConfiguratorMixin):
     quota_increase_policy = FromConfig(
         "RDM_QUOTA_INCREASE_POLICY", default=QuotaIncreasePolicyEvaluator
     )
+    version_review_policy = FromConfig(
```
palkerecsenyi: rename, new version policy

invenio_rdm_records/records/api.py
```diff
@@ -285,6 +289,26 @@ class CommonFieldsMixin:
...
+    def get_own_or_parent_review(self):
+        """
+        Returns the review on the parent, or on the draft/record if it doesn't exist.
+
+        We store the review on the parent record for new unpublished records, and on the record for new versions of
+        already published records.
+
+        If the review does not exist on the parent or the record, None is returned.
+        """
+        return self.parent.review if self.parent.review is not None else self.review
```
OliverGeneser: Hmm this does silently prioritize a parent review over a record review. Maybe we should add a check that raises a ReviewStateError in case that both a parent and record review exists? (Migrations can fail...)

invenio_rdm_records/services/review/service.py
```diff
@@ -59,12 +63,25 @@ def _validate_request_type(self, request_type):
...
             raise ReviewExistsError(_("A review already exists for this record"))
         # Validate that record has not been published.
-        if record.is_published or record.versions.index > 1:
+        # Additionally, either `assign_request_to_record` must be true, or the record must be
+        # the first version.
+        if record.is_published or (
+            record.versions.index != 1 and not assign_request_to_record
+        ):
             raise ReviewStateError(
                 _("You cannot create a review for an already published record.")
```
OliverGeneser, line 84: Now that you added the check for non-first versions, the error message is probably still technically correct since those only exist for already-published records? However I think it can be read a bit unclear as the draft itself is unpublished. Maybe something like "You cannot create a review for this record version." would be a bit clearer? Or maybe it's just me 😆

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
```diff
@@ -155,8 +165,11 @@ export function computeDepositState(record, selectedCommunity = undefined) {
   // Serialize selectedCommunity
   const _selectedCommunity = getSelectedCommunityMetadata(record, selectedCommunity);
   const communityIsSelected = !_isEmpty(_selectedCommunity);
+  // The ID of the default community of the record's parent (if any)
+  const defaultCommunityId = record.parent?.communities?.default;
 
-  const draftReview = record?.parent?.review;
+  const alreadyPublished = isAlreadyPublished(record);
+  const draftReview = alreadyPublished ? record?.review : record?.parent?.review;
```
OliverGeneser: Hmm I think the naming is a bit confusing because isAlreadyPublished depends on record.versions.index !== 1, but "already published" and "new version" are not identical concepts, right? A new version draft can itself be unpublished while still belonging to a published parent?

invenio_rdm_records/services/review/service.py
```diff
@@ -140,18 +186,23 @@ def delete(self, identity, id_, revision_id=None, uow=None):
...
         # Unset on record
-        draft.parent.review = None
-        uow.register(ParentRecordCommitOp(draft.parent))
-        uow.register(RecordIndexOp(draft, indexer=self.indexer))
+        if draft.review is not None:
+            draft.review = None
+        elif draft.parent.review is not None:
+            draft.parent.review = None
+            uow.register(ParentRecordCommitOp(draft.parent))
+        uow.register(RecordCommitOp(draft, indexer=self.indexer))
```
OliverGeneser, line 213: Small question. When removing draft.parent.review is committing the parent enough or should the parent also be explicitly reindexed? Just wondering in case any indexed/expanded fields depend on the parent review state.

OliverGeneser (commented):
I think that there is a lot of record.versions.index == 1, maybe a helper or utility function (record.is_first_version/is_new_version)  

I think disablePublishButton sounds a bit like a generic UI state, but it's actually the specific business logic of blocking while we wait for a review, so maybe renaming to something like publishBlockedByReview?

invenio_rdm_records/services/review/service.py
palkerecsenyi: `ParentRecordCommitOp` already does the indexing for the parent record, so an explicit reindex is not needed. However, it only does the reindex if the `indexer_context` kwarg is set, so I need to fix that!

invenio_rdm_records/services/review/service.py
palkerecsenyi: This is a good point, i will separate it into 2 error message depending on the case

invenio_rdm_records/records/api.py
palkerecsenyi: In theory it should never be possible for both to exist based on the way the review component is constructed. However you are right that it is better to be safe than sorry so I will add a check here

invenio_rdm_records/services/review/service.py
```diff
@@ -59,15 +63,32 @@ def _validate_request_type(self, request_type):
...
             raise ReviewExistsError(_("A review already exists for this record"))
         # Validate that record has not been published.
-        if record.is_published or record.versions.index > 1:
+        if record.is_published:
             raise ReviewStateError(
                 _("You cannot create a review for an already published record.")
             )
+        # Additionally, either `allow_new_record_version_review` must be true, or the record must be
+        # the first version.
+        if record.versions.index > 1 and not allow_new_record_version_review:
```
ntarocco, line 88: Can we check the service config `new_version_review_policy` instead of passing a param? It is unclear to me why we need to add this param.

ntarocco (commented):
In general, it LGTM, good job. I have a general question: can we instead get rid of the review at the parent level, and:
have one review on the first record's version
OR
have reviews at each record's versions.

With this change, it feels like that the parent review becomes superfluous.

invenio_rdm_records/records/mappings/os-v1/rdmrecords/drafts/draft-v6.0.0.json
```diff
@@ -1621,6 +1621,49 @@
...
+            "dynamic": "true"
+          },
+          "topic": {
+            "type": "object",
+            "dynamic": "true"
+          },
+          "receiver": {
+            "type": "object",
+            "dynamic": "true"
+          },
```
slint, line 1658: minor/comment: I think for all the dynamic fields, we need to have some sort of dynamic template, otherwise we might get weird inferred types, and not be able to do queries like `review.user:12345`, but instead need `review.user.keyword:12345`. We probably haven't done that for the `parent.review` field, but no time better than the present I guess.

(Note: I'm leaving the comment here, but applies to all mappings files of course)

invenio_rdm_records/services/review/service.py
palkerecsenyi: The `new_version_review_policy` is only evaluated in the [review component](https://github.com/palkerecsenyi/invenio-rdm-records/blob/f811806da1f21582a50f94bb32de88a4b6f87621/invenio_rdm_records/services/components/review.py#L79) when creating a new record version draft.

This kwarg has been added here to ensure that without explicitly setting `allow_new_record_version_review=True`, the behaviour of review creation remains exactly the same as before, with the same error if you're trying to create a review for a non-first version of a record. In case an instance is using this method directly for whatever reason, we don't want to make it a breaking change.

invenio_rdm_records/requests/community_submission.py
```diff
@@ -59,24 +60,38 @@ def execute(self, identity, uow, **kwargs):
         # The curator (receiver) should still have access, via the community
         # The creator (uploader) should also still have access, because
         # they're the uploader
-        draft.parent.review = None
+        parent_needs_commit = False
```
slint, line 63: Nice trick to manage this :)

invenio_rdm_records/requests/community_submission.py
```diff
@@ -59,24 +60,38 @@ def execute(self, identity, uow, **kwargs):
...
+        if draft.review is not None:
+            draft.review = None
+            uow.register(RecordCommitOp(draft, indexer=service.indexer))
```
slint:
```suggestion
            uow.register(RecordCommitOp(draft, indexer=service.draft_indexer))
```

Big bamboozle, and I'm sure we're doing the same mistake in other places...

slint (commented):
Overall LGTM, some minor correctness things, and agree also with @ntarocco comment on avoiding a new param on the review service methods.

invenio_rdm_records/records/mappings/os-v1/rdmrecords/drafts/draft-v6.0.0.json
palkerecsenyi: This sounds like a good idea but maybe out of scope for this PR. Is it okay if I shelve it in a new issue?

PR #2334, opened 2026-05-11
fix(vcs): extract license ID when returning

palkerecsenyi:
* The user might not have permission to read the license vocabulary directly, resulting in a PermissionDeniedError when extracting the repository metadata.

* Since the list of licenses is non-sensitive (and is queried already on the deposit page), it is safe to use the system identity for this query.

### DIFF
--- a/invenio_rdm_records/services/vcs/release.py
+++ b/invenio_rdm_records/services/vcs/release.py
@@ -151,7 +151,7 @@ class RDMVCSRelease(VCSRelease):
             license_vocab = current_vocabularies_service.read(
                 identity=self.user_identity, id_=("licenses", license_spdx_id.lower())
             )
-            return {"id": license_vocab.pid.id}
+            return {"id": license_vocab.id}
         except (PIDDoesNotExistError, NoResultFound):
             self.add_warning(
                 f"The repository's license '{license_spdx_id}' is not recognised and has been added as a custom license. "


zubeydecivelek (approved):

invenio_rdm_records/services/vcs/release.py
```diff
@@ -149,7 +149,7 @@ def get_license_pid(self) -> dict | None:
         try:
             # Try reading from the vocab list; it may not exist since the VCS IDs do not map 1-to-1 with the vocabulary IDs
             license_vocab = current_vocabularies_service.read(
-                identity=self.user_identity, id_=("licenses", license_spdx_id.lower())
+                identity=system_identity, id_=("licenses", license_spdx_id.lower())
```
kpsherva: system user shouldn't be needed for this part, licenses should be public (did you check the permission policy for the license vocabulary)

the prior self.user_identity looks a bit fishy to me, I guess you were sure that this class always gets instantiated with a correctly assigned identity object right?

palkerecsenyi:
I have added the `any_user` role to `self.user_identity`, which was the cause of the error since it was the missing permission. Please see here: https://github.com/inveniosoftware/invenio-vcs/pull/11/changes#diff-7e21816d389c34ae9ac0285516fb2e3af328a4f48c283c90e15927fc4a4e0ecaR98.

This PR still contains a small bug fix that's related to the license, so we can still merge it.

kpsherva (approved):

PR #2339, opened 2026-05-12
feat(notifications): add NOTIFICATIONS_SHOW_EMOJIS config option

### LINKED ISSUE #2240: Notifications: Add an option to remove icons/emojis from Mail templates

Hi, our institution thinks it would be more "professional" if the mails InvenioRDM sends would not contain icons like 📥 (https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.submit.jinja#L19C14-L19C16) or ✅ and so on.

I would propose to introduce a new config var to be able to disable the emojis in the mail templates.

- introduce new config var for use in invenio.cfg: `NOTIFICATIONS_SHOW_EMOJIS = False`
- Add in all mail template: `{% set show_emojis = config.get('NOTIFICATIONS_SHOW_EMOJIS', True) %}`
- use something like:

```diff
{%- block subject -%}
+      {%- set emoji = "✅ " if show_emojis else "" -%}
-       {{ _("✅ Community inclusion accepted for 
+       {{ emoji }}{{ _("Community inclusion accepted for '{record_title}'").format(record_title=record_title) }}
{%- endblock subject -%}
```
My question is if this would be of interest and if I should open a PR or if you wnat to keep the emojis by all costs.

Thanks!

karkraeg:
## Summary

Closes #2240.

Introduces `NOTIFICATIONS_SHOW_EMOJIS` (default: `True`) so operators can disable emoji prefixes in notification email subject lines by setting `NOTIFICATIONS_SHOW_EMOJIS = False` in `invenio.cfg`.

- New config variable in `config.py` (default `True`, fully backwards-compatible)
- All 17 notification templates updated: emoji extracted from `_()` translation string and rendered conditionally via `{%- set emoji = "…" if show_emojis else "" -%}`
- Emojis removed from translatable strings as a side effect (i18n improvement)

Templates without emojis in their subjects (`record-deletion.decline`, all `repository-release.*`) are untouched.

## Test plan

- [ ] Send a notification with default config → subject still contains emoji
- [ ] Set `NOTIFICATIONS_SHOW_EMOJIS = False` → all notification subjects are emoji-free
- [ ] Existing notification tests still pass

### DIFF
--- a/invenio_rdm_records/config.py
+++ b/invenio_rdm_records/config.py
@@ -958,3 +958,6 @@ RDM_RECORDS_CONTAINER_EXTENSIONS = [".zip"]
 Experimental, this config can later be removed."""
 
 RDM_RECORD_FILE_EXTRACTORS = [ZipExtractor()]
+
+NOTIFICATIONS_SHOW_EMOJIS = True
+"""Whether to show emojis in notification email subject lines."""
--- a/invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.accept.jinja
+++ b/invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.accept.jinja
@@ -21,10 +21,11 @@
     )
 %}
 {% set account_settings_link = invenio_url_for("invenio_notifications_settings.index") %}
-
+{% set show_emojis = config.get('NOTIFICATIONS_SHOW_EMOJIS', True) %}
 
 {%- block subject -%}
-       {{ _("✅ Community inclusion accepted for '{record_title}'").format(record_title=record_title) }}
+{%- set emoji = "✅ " if show_emojis else "" -%}
+       {{ emoji }}{{ _("Community inclusion accepted for '{record_title}'").format(record_title=record_title) }}
 {%- endblock subject -%}
 
 {%- block html_body -%}
--- a/invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.cancel.jinja
+++ b/invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.cancel.jinja
@@ -21,9 +21,11 @@
     )
 %}
 {% set account_settings_link = invenio_url_for("invenio_notifications_settings.index") %}
+{% set show_emojis = config.get('NOTIFICATIONS_SHOW_EMOJIS', True) %}
 
 {%- block subject -%}
-       {{ _("❌ Community inclusion cancelled for '{record_title}'").format(record_title=record_title) }}
+{%- set emoji = "❌ " if show_emojis else "" -%}
+       {{ emoji }}{{ _("Community inclusion cancelled for '{record_title}'").format(record_title=record_title) }}
 {%- endblock subject -%}
 
 {%- block html_body -%}
--- a/invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.decline.jinja
+++ b/invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.decline.jinja
@@ -21,9 +21,11 @@
     )
 %}
 {% set account_settings_link = invenio_url_for("invenio_notifications_settings.index") %}
+{% set show_emojis = config.get('NOTIFICATIONS_SHOW_EMOJIS', True) %}
 
 {%- block subject -%}
-       {{ _("⛔️ Community inclusion declined for '{record_title}'").format(record_title=record_title) }}
+{%- set emoji = "⛔️ " if show_emojis else "" -%}
+       {{ emoji }}{{ _("Community inclusion declined for '{record_title}'").format(record_title=record_title) }}
 {%- endblock subject -%}
 
 {%- block html_body -%}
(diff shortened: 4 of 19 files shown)


invenio_rdm_records/translations/ar/LC_MESSAGES/messages.po
```diff
@@ -1208,8 +1208,8 @@ msgstr "فتح في تبويبة جديدة"
```
Samk13, line 1208: I understand that this is included while you are testing, but Is it really necessary to update all translation catalogs in this PR, or can those changes be left to the normal translation update workflow?

Since the change is driven by removing emojis from the source msgids, the catalogs can be regenerated through the normal translation workflow in the next pull and updated by translators later.

invenio_rdm_records/config.py
```diff
@@ -958,3 +958,6 @@ def lock_edit_published_files(service, identity, record=None, draft=None):
 Experimental, this config can later be removed."""
 
 RDM_RECORD_FILE_EXTRACTORS = [ZipExtractor()]
+
+NOTIFICATIONS_SHOW_EMOJIS = True
```
Samk13: Should the configuration key be more specific to avoid being overly generic. see [config initialization](https://github.com/inveniosoftware/invenio-rdm-records/blob/master/invenio_rdm_records/ext.py#L123-L132) ?
```suggestion
RDM_NOTIFICATIONS_SHOW_EMOJIS = True
```

Samk13 (commented):
Thanks for this, the change makes sense overall. I left a few comments and questions for consideration.

invenio_rdm_records/translations/ar/LC_MESSAGES/messages.po
karkraeg: i was actually following copilots suggestion in https://github.com/inveniosoftware/invenio-rdm-records/pull/2339#discussion_r3228386627 but sure it would be better not to have this in this PR!

invenio_rdm_records/config.py
karkraeg: sounds very reasonable, I'll apply

invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.accept.jinja
```diff
@@ -21,10 +21,11 @@
     )
 %}
 {% set account_settings_link = invenio_url_for("invenio_notifications_settings.index") %}
-
+{% set show_emojis = config.get('NOTIFICATIONS_SHOW_EMOJIS', True) %}
```
Samk13, line 24: Since the emoji toggle logic is repeated across many templates, 
did you consider exposing a small Jinja helper for it, or do you prefer keeping the logic explicit in each template? Not blocking, just wondering about the tradeoff.

For example:

```python
 def notification_emoji(symbol):
     return f"{symbol} " if current_app.config.get("NOTIFICATIONS_SHOW_EMOJIS", True) else ""
# Register it as a Jinja global:
app.jinja_env.globals["notification_emoji"] = notification_emoji
 ```

 Then in the template:

 ```jinja
 {{ notification_emoji("✅") }}{{ _("Community inclusion accepted for ...") }}
 ```

Not blocking, but it could reduce duplication across notification templates.

Comment (karkraeg) [reply invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.accept.jinja:24]: Hi Sam, tbh I didn't think of doing it this way. While this would be cleaner I also think that keeping it explicit is a bit more accessible for developers since it doesn't introduce another layer of abstraction. But I'll go either way, no problem to change it!

File: invenio_rdm_records/translations/ar/LC_MESSAGES/messages.po (hunk 1 above)
Comment (karkraeg) [reply invenio_rdm_records/translations/ar/LC_MESSAGES/messages.po:1208]: removed them 👍

File: invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.accept.jinja (hunk 3 above)
Comment (Samk13) [reply invenio_rdm_records/templates/semantic-ui/invenio_notifications/community-submission.accept.jinja:24]: Thanks for the context, maybe it would be useful to get one more opinion before deciding.
Perhaps @fenekku or @utnapischtim can weigh in, and we can move forward with whichever direction has broader support.

File: invenio_rdm_records/config.py (hunk 2 above)
Comment (Samk13) [reply invenio_rdm_records/config.py]: Looks like you missed this one :)

**Comment by karkraeg** [comment]
Hi, are there any open issues here or would it be possible to merge this?

**Review by utnapischtim — APPROVED** [review]

**Review by Samk13 — APPROVED** [review]
LGTM, thanks for this!
Yes, this can be merged. Backporting to v14 would be difficult since it changes translation strings, so I’d target this for the v14.1 release instead.

File: invenio_rdm_records/config.py [hunk 4]
```diff
@@ -958,3 +958,6 @@ def lock_edit_published_files(service, identity, record=None, draft=None):
...
+"""Whether to show emojis in notification email subject lines."""
```
Samk13, line 963: TODO(docs): this config need to be added for v14.1 release notes.

PR #2342, opened 2026-05-18
fix: restoring schema field in RDMdraft

Ducica:
* after undeleting draft (when clicking on edit of published record)
* schema field is missing in the draft

### Description

When you have a published record and then you want to edit it. The restored draft (undeleted), does not contain $schema field at this stage. In current RDM this is not visible, but only by chance. For us, we use this field to determine the type of record (metadata) it is and then apply appropriate serialization to the record based on this and it crashes at this point. 

As $schema is RDM related thing, my logic was to put this inside of RDMdraft. Alternative approach would be to use a service component with edit method that would inject this into the draft, similarly how it is done for access field, though I felt like this was more in line with https://github.com/inveniosoftware/invenio-drafts-resources/blob/0bd2cc6dd2a333c8a2db39eb6c3e17128ebeeefc/invenio_drafts_resources/records/api.py#L231-L240 what is already done at Draft level here. 

Please let us know if we can merge this. Thanks a lot.

Dusan S

CESNET

### DIFF
--- a/invenio_rdm_records/records/api.py
+++ b/invenio_rdm_records/records/api.py
@@ -2,6 +2,7 @@
 #
 # Copyright (C) 2020-2024 CERN.
 # Copyright (C) 2021-2023 TU Wien.
+# Copyright (C) 2026 CESNET i.a.l.e.
 #
 # Invenio-RDM-Records is free software; you can redistribute it and/or modify
 # it under the terms of the MIT License; see LICENSE file for more details.
--- a/invenio_rdm_records/records/api.py
+++ b/invenio_rdm_records/records/api.py
@@ -422,6 +422,17 @@ class RDMDraft(CommonFieldsMixin, Draft):
 
     status = DraftStatus()
 
+    @classmethod
+    def edit(cls, record):
+        """Restore $schema for soft-deleted records when editing a record."""
+        # ConstantField.pre_init skips data=None (soft-deleted) records, so
+        # $schema is missing after undelete — restore it from the field. $schema
+        # field is RDM related and therefore it is not handled in edit method of
+        # the base Draft class https://github.com/inveniosoftware/invenio-drafts-resources/blob/0bd2cc6dd2a333c8a2db39eb6c3e17128ebeeefc/invenio_drafts_resources/records/api.py#L231
+        draft = super().edit(record)
+        draft.setdefault("$schema", cls.schema.value)
+        return draft
+
 
 RDMFileDraft.record_cls = RDMDraft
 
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/records/api.py
```diff
@@ -422,6 +422,15 @@ class RDMDraft(CommonFieldsMixin, Draft):
 
     status = DraftStatus()
 
+    @classmethod
+    def edit(cls, record):
+        """Restore $schema for soft-deleted records when editing a record."""
+        # ConstantField.pre_init skips data=None (soft-deleted) records, so
+        # $schema is missing after undelete — restore it from the field.
```
utnapischtim: i would add here or in the commit message an explanation why an override of the base function is necessary.

probably better directly in the code. so something like. `$schema is defined in rdm-records, therefore it can't be handled in the parent drafts-resources class` maybe you improve the sentence a little bit

invenio_rdm_records/records/api.py
```diff
@@ -477,9 +486,11 @@ class RDMRecord(CommonFieldsMixin, Record):
         store=False,
         dump=True,
         # Don't dump files if record is public and files restricted.
-        dump_entries=lambda record: not (
-            record.access.protection.record == Visibility.PUBLIC.value
-            and record.access.protection.files == Visibility.RESTRICTED.value
+        dump_entries=lambda record: (
```
utnapischtim: nit: this is only a reformating leftover or?

Ducica: yes sorry, I seem to have a different formatter, I will revert this to original form

invenio_rdm_records/records/api.py
Ducica: Done thanks

utnapischtim (approved):

ntarocco (commented):
Thank you for the contribution! 
Given that we are touching the very core of InvenioRDM data model, may I please ask if you could a test(s) for this change?
It will be very much appreciated! Thank you!

Ducica:
> Thank you for the contribution! Given that we are touching the very core of InvenioRDM data model, may I please ask if you could a test(s) for this change? It will be very much appreciated! Thank you!

Sure Nikola, thank you guys for your time. I've added a test. If there are any other concerns, I'd be happy to resolve. Regards, Dusan

Ducica:
Hello, I would just ask please to let us know by the end of the week, if you would be willing to merge this, so that we would know to go in alternative directions if not. Thank you!

PR #2353, opened 2026-06-05
fix(CreatibutorsField): Mount CreatibutorsModal only when opened

sakshamarora1:
Partly addresses: https://github.com/inveniosoftware/invenio-app-rdm/issues/3006

  * This change fixes the Creator/Contirbutor section in the deposit form where every CreatibutorsFieldItem unconditionally rendered a CreatibutorsModal, and inside each modal the Formik instance with enableReinitialize and a Yup schema instantiated at class-definition time for each author/creator, which led to every stroke triggering a redraw and re-mount for all the modals

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsFieldItem.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsFieldItem.js
@@ -9,14 +9,36 @@
 
 import { i18next } from "@translations/invenio_rdm_records/i18next";
 import _get from "lodash/get";
-import React from "react";
+import React, { useRef, useState, useEffect, useMemo } from "react";
 import { useDrag, useDrop } from "react-dnd";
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsFieldItem.js
```diff
@@ -26,15 +48,23 @@ export const CreatibutorsFieldItem = ({
   addLabel,
   editLabel,
   initialCreatibutor,
-  displayName,
```
kpsherva, line 29: I think we should leave this as a prop, because people might have already overridden this component and removing it will require a major release, since you are changing the function interface. I would leave it as a param, and in case the value is given - us the value, otherwise use `creatibutorNameDisplay(initialCreatibutor),` like you do in L63

sakshamarora1: Yeah, you're right. Thanks! Updated!

kpsherva (approved):

PR #2359, opened 2026-06-10
feat(Creatibutors): add inline scrolling and file importing in deposit

### LINKED ISSUE #3006: Large author lists
Records with many thousands of authors have some usability issues.

The landing page gets very long:

Image

The requests view gets very long:
Image

It takes a long time to save a draft.

The deposit form gets very long. It would also be super useful to be able to search for specific authors or affiliations.
Image

This bug is annoying https://github.com/inveniosoftware/invenio-app-rdm/issues/2999

sakshamarora1:
Closes: https://github.com/inveniosoftware/invenio-app-rdm/issues/3006

  * When the author or contributor count exceeds RDM_CREATIBUTORS_INLINE_LIST_THRESHOLD, the deposit form shows an inline panel with search and becomes scrollable.
  * Large lists are rendered in batches of RDM_CREATIBUTORS_RENDER_BATCH_SIZE to keep the UI responsive.
  * Added FileImportReviewModal with JSON parsing (RDM API format)

https://github.com/user-attachments/assets/701afc95-597f-491f-a38d-f9aee4dc190d

Screenshot 2026-06-10 at 16 04 41

Screenshot 2026-06-10 at 16 05 19


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsInlinePanel.js
```diff
@@ -0,0 +1,277 @@
...
+import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import PropTypes from "prop-types";
+import { Button, Input } from "semantic-ui-react";
+import { i18next } from "@translations/invenio_rdm_records/i18next";
+import { CreatibutorsList, getCreatibutorDisplayName } from "./CreatibutorsList";
+
+const getCreatibutorSchemaLabel = (schema) =>
+  schema === "contributors" ? i18next.t("contributors") : i18next.t("authors");
+
+const getScrollEdges = (el) => {
```
kpsherva: if I understand correctly the responsibility of this function, I think it would be easier to handle it with CSS (and less code to maintain, especially because I can see there is state management for this)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsInlinePanel.js
```diff
@@ -0,0 +1,277 @@
...
+
+  const updateScrollState = useCallback(() => {
+    const el = scrollRef.current;
+    if (!el) return;
+    const { canScrollUp, canScrollDown } = getScrollEdges(el);
+    setCanScrollUp(canScrollUp);
+    setCanScrollDown(canScrollDown);
+  }, []);
+
+  const clearHighlight = useCallback((index) => {
```
kpsherva: the highlighting could be handled with an animated CSS class, without a need to keep the indices in memory and performing all the operations in the JS. It will also save us some lines of code

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
```diff
@@ -158,17 +212,14 @@ class CreatibutorsFieldForm extends Component {
 }
 
 export class CreatibutorsFieldComponent extends Component {
-  render() {
-    const { fieldPath } = this.props;
+  // Stable class-instance function so FieldArray never sees a new `component`
```
kpsherva, line 267: what does this mean in practice? I have a bit of a hard time understanding

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/FileImportReviewModal.js
```diff
@@ -0,0 +1,407 @@
...
+  }
+]`;
+
+const parseCreatibutorFile = (text) => {
+  // Only allow a file that is a JSON array of objects at the top-level
+  let parsed;
+  try {
+    parsed = JSON.parse(text);
+  } catch {
+    return [];
```
kpsherva: I think we should display some errors in case it happens, instead of empty array? 
I saw you did handle errors during the demo so this is a bit confusing

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/FileImportReviewModal.js
```diff
@@ -0,0 +1,407 @@
...
+
+    return (
+      <>
+        <Segment
+          placeholder
+          textAlign="center"
+          className="mt-0"
+          onDragOver={handleDragOver}
+          onDrop={handleDrop}
+          style={{ cursor: loading ? "not-allowed" : "pointer" }}
```
kpsherva: 👀 inline styling detected

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/FileImportReviewModal.js
```diff
@@ -0,0 +1,407 @@
...
+            primary
+            type="button"
+            loading={loading}
+            disabled={loading}
+            onClick={this.triggerFileInput}
+          >
+            <Icon name="folder open outline" />
+            {i18next.t("Choose file")}
+          </Button>
+          <div style={{ marginTop: "1em", color: "#888" }}>
```
kpsherva: 👀

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/FileImportReviewModal.js
```diff
@@ -0,0 +1,407 @@
...
+          </p>
+          <pre className="creatibutors-import-example">
+            {CREATIBUTOR_IMPORT_EXAMPLE}
+          </pre>
+        </Message>
+      </>
+    );
+  }
+
+  renderReviewPhase() {
```
kpsherva: this is a prime candidate to split into a separate components, to give them split responsibilities like you try to do by dividing in methods - if you split, it will simplify the code

invenio_rdm_records/config.py
```diff
@@ -872,6 +872,21 @@ def lock_edit_published_files(service, identity, record=None, draft=None):
...
 
+RDM_CREATIBUTORS_INLINE_LIST_THRESHOLD = 10
+"""When a author/contributor list exceeds this count, expanded list UI is enabled.
+
+On the record landing page this shows a searchable "show all" modal instead of
+the full inline list. In the deposit form this enables inline search and
+scrolling for authors and contributors.
+"""
+
+RDM_CREATIBUTORS_RENDER_BATCH_SIZE = 20
```
kpsherva: I think we should handle these as props (customisation will be possible by overridable's parametrize, if needed) otherwise we are using backend config variables to setup frontend, and this will cause our config to bloat more in the long term

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/FileImportReviewModal.js
sakshamarora1: The error handling in the demo is for missing fields based on the common yup schema

This is a different case where the json doesn't match the rdm json at all.
But yes, if the whole json is completely different, i can show some helo text.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
sakshamarora1: Basically a new function was being created on every render. Formik compares the component prop by reference, so it was causing unneeded re-mounting if even it this didn't change. So defining it as a class property keeps the reference stable.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/FileImportReviewModal.js
sakshamarora1: Removed! Thanks!

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/FileImportReviewModal.js
sakshamarora1: You're right! I've simplified it now.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsInlinePanel.js
sakshamarora1: I have moved CSS specific stuff, but to keep track of first and last element to show/hide the scroll button and the control gradient for it, I have kept some JS code which I couldn't find a way to take care of with CSS

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/FileImportReviewModal.js
sakshamarora1: Now it will prompt the user
Screenshot 2026-06-25 at 16 22 30

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsInlinePanel.js
```diff
@@ -0,0 +1,295 @@
...
+import { Button, Input } from "semantic-ui-react";
+import { i18next } from "@translations/invenio_rdm_records/i18next";
+import { CreatibutorsList, getCreatibutorDisplayName } from "./CreatibutorsList";
+
+// Normalise a string for diacritic insensitive search: decompose into base chars +
+// combining marks, strip the marks, then lower-case while still matching the literal character.
+const normalizeSearch = (str) =>
+  str
+    .normalize("NFD")
+    .replace(/[\u0300-\u036f]/g, "")
```
kpsherva, line 17: what is this character range representing?

sakshamarora1: The decomposed diacritic characters after normalizing
https://www.compart.com/en/unicode/block/U+0300

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsInlinePanel.js
```diff
@@ -0,0 +1,295 @@
...
+    .toLowerCase();
+
+const getCreatibutorSchemaLabel = (schema) =>
+  schema === "contributors" ? i18next.t("contributors") : i18next.t("authors");
```
kpsherva:
```suggestion
  schema === "contributors" ? i18next.t("contributors") : i18next.t("creators/authors");
```
otherwise it is inconsistent with the deposit form, unless it is already like that in the landing page then please ignore this comment

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsInlinePanel.js
```diff
@@ -0,0 +1,295 @@
...
+
+export const CreatibutorsInlinePanel = React.memo(function CreatibutorsInlinePanel({
```
kpsherva, line 20: this component has many responsibilities, is there a reasonable split we could do? otherwise it is also difficult to review

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsFieldItem.js
```diff
@@ -37,6 +37,12 @@ export const getCreatibutorDisplayName = (value) => {
   return `${name}${affiliation}`;
 };
 
+const renderRole = (role, roleOptions) => {
+  if (!role) return null;
+  const friendlyRole = roleOptions.find(({ value }) => value === role)?.text ?? role;
```
kpsherva, line 26: aren't roles names translated? will .text return only english ?
and ... why friendly role 😅 ?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsInlinePanel.js
sakshamarora1: This component is only taking care of the scrolling and highlighting, then the rendering of the actual list is taken care of by `CreatibutorsList`, I tried to split the reponsibilities, I don't see anything more that could be split into

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsInlinePanel.js
sakshamarora1: It is already like that
image

kpsherva: so shouldn't you change the code above to `Authors/Creators`? for consistency

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsInlinePanel.js
kpsherva: there is a lot of prop drilling, is there any way to organise it in a way to decouple this? because it seems to me that every time you are highlighting one element, you are re-rendering the whole list as the child component. Can the highlighting be done on the item itself directly, instead of wrapping the whole list in a parent which is searching for the highlighted element? 
otherwise you are coupling the "servicing" of the list elements with the whole list, while it should be more "individual"

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsFieldItem.js
sakshamarora1: Just moved it outside of the component to reuse, [it already existed](https://github.com/inveniosoftware/invenio-rdm-records/pull/2359/files#diff-dd26385497bc2f4d0602f560ea9ef44f4034b167d972a6bc9137bd3772a24c1aL114).

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsInlinePanel.js
sakshamarora1: I have updated the code to use context so that we don't do the prop drilling. And the components only take care of what they need to without the extra props making it confusing.

On the highlighting, using React.memo already so the re-render only happens for the new/updated list element. Moving it to the item will overcomplicate it because each item will need to handle it instead.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsDisplay/CreatibutorsList.js
```diff
@@ -0,0 +1,149 @@
...
+      // Reset the guard so a future render can schedule the next batch.
+      batchScheduled.current = false;
+    };
+  }, [useBatchRender, renderLimit, entries.length, batchSize]);
+
+  // Extend on append so new rows at the end can be mounted.
+  if (entries.length > prevEntriesLenRef.current) {
+    const newLimit = Math.max(renderLimit, entries.length);
+    if (newLimit !== renderLimit) {
+      setRenderLimit(newLimit);
```
palkerecsenyi: seems a little unusual to have a non-pure operation during the render of a component. should this not be inside an effect or something?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
```diff
@@ -19,18 +18,82 @@ import {
...
 import Overridable from "react-overridable";
 
 class CreatibutorsFieldForm extends Component {
-  handleOnContributorChange = (selectedCreatibutor) => {
+  // Refs hold the latest Formik array helpers. Formik recreates these functions
+  // on every render, but we need stable references so the list doesn't re-render
+  // when unrelated fields change.
+  _removeRef = { current: null };
+  _replaceRef = { current: null };
+  _moveRef = { current: null };
```
palkerecsenyi: should these not be created using `React.createRef()`? i've never seen the ref-shaped object be manually constructed

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
```diff
@@ -19,18 +18,82 @@ import {
...
+    if (
+      !this._lastItemContextValue ||
+      this._lastItemContextValue.roleOptions !== roleOptions ||
+      this._lastItemContextValue.schema !== schema ||
+      this._lastItemContextValue.autocompleteNames !== autocompleteNames ||
+      this._lastItemContextValue.addLabel !== addLabel ||
+      this._lastItemContextValue.editLabel !== editLabel ||
+      this._lastItemContextValue.serializeSuggestions !== serializeSuggestions ||
+      this._lastItemContextValue.serializeCreatibutor !== serializeCreatibutor ||
+      this._lastItemContextValue.deserializeCreatibutor !== deserializeCreatibutor
```
palkerecsenyi: aren't some of these values objects/arrays/functions?

palkerecsenyi: and anyway what's wrong with assigning `this._lastItemContextValue` if none of the values have changed? It is not part of the state so a change to it won't inherently cause a rerender right?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
```diff
@@ -53,6 +116,11 @@ class CreatibutorsFieldForm extends Component {
       deserializeCreatibutor,
     } = this.props;
 
+    // Keep refs pointing at the latest helpers without changing their identity.
+    this._removeRef.current = formikArrayRemove;
+    this._replaceRef.current = formikArrayReplace;
+    this._moveRef.current = formikArrayMove;
```
palkerecsenyi: why do these functions change value?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
```diff
@@ -81,6 +149,12 @@ class CreatibutorsFieldForm extends Component {
       };
     }
 
+    const isContributors = schema === "contributors";
+    const totalCount = creatibutorsList.length;
+
+    // Display label for the inline filter placeholder.
+    const type = isContributors ? "contributors" : "authors/creators";
```
palkerecsenyi, line 167: i18n?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsFileModal.js
```diff
@@ -0,0 +1,517 @@
...
+          onChange={onFileChange}
+        />
+      </Segment>
+
+      <Message info size="small" className="text-left mt-15">
+        <Message.Header>{i18next.t("File format")}</Message.Header>
+        <p>
+          {i18next.t("Upload a JSON file containing a list of {{type}}.", { type })}
+        </p>
+        <pre className="creatibutors-import-example">{CREATIBUTOR_IMPORT_EXAMPLE}</pre>
```
palkerecsenyi, line 123: Do we have somewhere in the docs we could link the user to for more info on the format?

palkerecsenyi: Or we can shelve this as an issue and add it later

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
sakshamarora1: Because it's used to hold stable references to Formik's array helper functions, not DOM elements or components, so a plain object works without adding the React ref overhead.
I'll correct the shape to remove confusion. Thanks!

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
sakshamarora1: [Formik's helper functions](https://formik.org/docs/api/fieldarray#fieldarray-helpers) instances on each render change their identity, triggering updates in components even if the data remains the same.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsField.js
sakshamarora1: It's taken care of when it's injected in the title, using it here and there as well was causing weird encoding issues.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/CreatibutorsField/CreatibutorsFileModal.js
sakshamarora1: It will/should be a separate issue from this

palkerecsenyi: Created: https://github.com/CERNDocumentServer/cds-rdm/issues/868

palkerecsenyi (approved):
lgtm

kpsherva (approved):

PR #2363, opened 2026-06-12
fix(get_record_requests): allow anonymous users to access landing page

sakshamarora1:


invenio_rdm_records/services/communities/service.py
```diff
@@ -500,7 +500,10 @@ def get_record_requests(self, identity, record):
 
         Output: {<Community-UUID>: <Request-UUID>}
         """
-        self.require_permission(identity, "review", record=record._record)
+        can_review = self.check_permission(identity, "review", record=record._record)
+        if not can_review:
+            # Instead of raising PermissionDeniedError, return an empty dictionary because anonymous users do not have permissions to search requests
```
fenekku, line 505: minor: I would have thought the linter would flag those

```suggestion
            # Instead of raising PermissionDeniedError, return an empty
            # dictionary because anonymous users do not have permissions to
            # search requests
```

fenekku (approved):
Thanks! I think keeping the `type(identity) == AnonymousIdentity` is fine for now mostly because we don't have the bandwidth to investigate the ramifications and original context for that. In the future, it would be nice to verify, because I agree it seems redundant.

PR #2367, opened 2026-06-19
fix(performance): replace expensive `deepcopy()` calls with lighter alternatives

max-moser:
Experiments with the profiler have shown that deep copying of data on read operations can have a significant impact on performance in some cases.
Since we don't generally fetch data from records and immediately manipulate it without further consideration, we can be a bit more aggressive here.

Edit: Some comparison between the different functions for a record's data of 5k creators and some other fields:
```
In [4]: %timeit -r 20 -n 10 deepcopy(dict(rec))
11.3 ms ± 2.47 ms per loop (mean ± std. dev. of 20 runs, 10 loops each)

In [5]: %timeit -r 20 -n 10 simple_deepcopy(dict(rec))
4.13 ms ± 1.1 ms per loop (mean ± std. dev. of 20 runs, 10 loops each)

In [6]: %timeit -r 20 -n 10 very_simple_deepcopy(dict(rec))
18.8 μs ± 3.41 μs per loop (mean ± std. dev. of 20 runs, 10 loops each)
```

### DIFF
--- a/invenio_rdm_records/records/dumpers/search.py
+++ b/invenio_rdm_records/records/dumpers/search.py
@@ -1,0 +1,25 @@
+# SPDX-FileCopyrightText: 2026 TU Wien.
+# SPDX-License-Identifier: MIT
+
+"""Faster search dumper that runs a faster deep-ish copy of the data before dumping."""
+
+from invenio_records.dumpers import SearchDumper as BaseSearchDumper
+
+from ...utils import very_simple_deepcopy
+
+
+class SearchDumper(BaseSearchDumper):
+    """Search dumper that creates a faster deep-ish copy of the data before dumping.
+
+    In this version of the search dumper, we replace ``copy.deepcopy()`` with a simpler
+    copy deep-ish copy mechanism that still satisfies our expectations, for performance
+    gains.
+    """
+
+    def _copy_record(self, record):
+        """Create a simpler version of a deepcopy for the record."""
+        return very_simple_deepcopy(dict(record))
+
+    def _copy_data(self, data):
+        """Create a simpler version of a deepcopy for the record."""
+        return very_simple_deepcopy(data)
--- a/invenio_rdm_records/records/models.py
+++ b/invenio_rdm_records/records/models.py
@@ -1,5 +1,6 @@
 # SPDX-FileCopyrightText: 2020-2026 CERN.
 # SPDX-FileCopyrightText: 2026 Graz University of Technology.
+# SPDX-FileCopyrightText: 2026 TU Wien.
 # SPDX-License-Identifier: MIT
 
 """Record and draft database models."""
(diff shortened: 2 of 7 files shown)


invenio_rdm_records/records/models.py
```diff
@@ -24,6 +24,16 @@
 from .systemfields.deletion_status import RecordDeletionStatusEnum
 
 
+class FastReadRecordMetadataBase(RecordMetadataBase):
```
ntarocco: Much faster indeed, thank you for finding this out! Necessary meme:
Image

Shall we release this after v14 (and v14.1), to be on the safe side and have time for testing?

max-moser: Yes, I think this is a reasonable way to move forward!

invenio_rdm_records/records/dumpers/search.py
```diff
@@ -0,0 +1,66 @@
...
+        return [*data]
+    elif isinstance(data, tuple):
+        return tuple(*data)
+    return data
+
+
+class FastSearchDumper(SearchDumper):
+    """Search dumper that creates a faster deep-ish copy of the data before dumping."""
+
+    def _copy_record(self, record):
```
max-moser, line 19: This requires https://github.com/inveniosoftware/invenio-records/pull/356

invenio_rdm_records/records/dumpers/search.py
```diff
@@ -0,0 +1,66 @@
...
+    if isinstance(data, dict):
+        return {k: simpledeepcopy(v) for k, v in data.items()}
+    elif isinstance(data, list):
+        return [*data]
+    elif isinstance(data, tuple):
+        return tuple(*data)
+    return data
+
+
+class FastSearchDumper(SearchDumper):
```
utnapischtim: nit: i would not call is `Fast`. this is a relative naming. in a year there could be the necessity to use a new `Fast2` because that is even faster as this one. 

i would keep `SearchDumper`

invenio_rdm_records/records/models.py
utnapischtim: nit: i would not use `Fast` as the prefix and i am unsure why we need `Read` in the naming?

for me `Fast` is a subjective assessment. Reusing the old name is better. We improve the current implementation.

max-moser: The "fast read" was in the naming to signal that only the read operations are affected by changes to make them faster.
The docstring should reflect this sufficiently though, so the original naming is fine as well.

ntarocco:
What is the status of this PR?

max-moser:
Along with the other PRs listed in https://github.com/inveniosoftware/invenio-app-rdm/issues/3494, it's waiting for reviews

utnapischtim:
have you tried to remove the copy completely and use the "original" data? @ntarocco should we discuss this "extreme" change?

ptamarit (approved):

ptamarit:
> have you tried to remove the copy completely and use the "original" data? @ntarocco should we discuss this "extreme" change?

@utnapischtim I checked with colleagues.
* The main thing is probably to generally guarantee encapsulation.
* Another potential reason is due to the fact that we load objects from the DB via SQLAlchemy. Without a copy, there would be a risk that developers could mutate objects without realizing that these objects will then be considered dirty by SQLAlchemy, and these changes will be flushed and committed to the DB.

utnapischtim:
>     * Another potential reason is due to the fact that we load objects from the DB via SQLAlchemy. Without a copy, there
> would be a risk that developers could mutate objects without realizing that these objects will then be considered dirty by
> SQLAlchemy, and these changes will be flushed and committed to the DB.

as i know, this is the main reason why deepcopy was used in the first place, but can the flush really happen? if we have the scenario that a landing page should be created the request triggers to load from the database, then it gets used to serialize it and then it is served to the client. `get_record` has a [no_autoflush](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L347) which should not flush back.

max-moser:
It's been a hot minute since I've opened this PR, but...

I have not tried removing the copy mechanism altogether and straight-up reusing the same reference.
Already, when replacing the `deepcopy()` with a shallow(er) copy, there was one system field (looks like the tombstone one, actually in this PR here) that needed a slight adaptation to avoid working on the same references.
Without this adaptation, one or two test cases were failing.

I think there's not a whole lot of spots in InvenioRDM where we perform manipulation of records or record-like objects that aren't going to be stored in the DB anyway, but I'd be hesitant to remove the copy mechanism altogether (at least without a whole lot of testing).

The performance hit from `deepcopy` mainly came from nested fields like creators (imagine 5k of them), where all kinds of properties (IIRC, also Python internal ones that don't affect the JSON) got copied.
I don't think that removing the copy mechanism altogether brings significant performance boosts when compared to shallow-ish copies.

ptamarit:
Thanks for your comments!
Based on what I read, I will go ahead and merge this; we can reopen the "no copy at all" route at a later stage if deemed necessary.

PR #2368, opened 2026-06-19
iiif: fix: logging without reserved filename key

ptamarit:
The following code:

```python
from flask import current_app
current_app.logger.exception("This is message", extra={"filename": "test.txt"})
```

Fails with:

```
KeyError: "Attempt to overwrite 'filename' in LogRecord"
```

This is due to the fact that `filename` is a field of `LogRecord`, so it cannot be passed as an `extra`.

This pull request fixes this logging statement by using a another key.

### DIFF
--- a/invenio_rdm_records/services/iiif/storage.py
+++ b/invenio_rdm_records/services/iiif/storage.py
@@ -141,7 +141,9 @@ class LocalTilesStorage(TilesStorage):
                 "Failed to delete tiles for record.",
                 extra={
                     "record_id": record["id"],
-                    "filename": filename,
+                    # Not using the key `filename` as it is a field of `LogRecord`
+                    # and using it would raise a KeyError exception.
+                    "filename_": filename,
                 },
             )
             return False


slint (approved):

invenio_rdm_records/services/iiif/storage.py
```diff
@@ -141,7 +141,9 @@ def delete(self, record, filename):
                 "Failed to delete tiles for record.",
                 extra={
                     "record_id": record["id"],
-                    "filename": filename,
+                    # Not using the key `filename` as it is a field of `LogRecord`
```
ntarocco, line 144: What is the class LogRecord?

ntarocco (approved):

invenio_rdm_records/services/iiif/storage.py
ptamarit: It's an internal Python logging class.
When calling `current_app.logger.exception`, it [calls a `makeRecord` method](https://github.com/python/cpython/blob/v3.14.6/Lib/logging/__init__.py#L1663-L1664) which creates an instance of `LogRecord`.

PR #2369, opened 2026-06-22
fix: fixed isParentAlreadyPublished not working on /new

Ducica:

### Description

When you go to create a new record, you select a community, and press save, review is not called. It is because in the "empty" record in the beginning the check in current code is incorrect and so you just save draft. Only after you save for the second time is review called. 

So you think you selected community, if you refresh the page, right after the first save, the community is gone.

### DIFF
--- a/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
+++ b/invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
@@ -69,7 +69,7 @@ function hasStatus(record, statuses = []) {
 // within its parent.
 // This essentially corresponds to whether the parent record has been published.
 function isParentAlreadyPublished(record) {
-  return hasStatus(record, [DepositStatus.PUBLISHED]) || record.versions?.index !== 1;
+  return hasStatus(record, [DepositStatus.PUBLISHED]) || record.versions?.index > 1;
 }
 
 function getSelectedCommunityMetadata(record, selectedCommunity) {


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/deposit.js
```diff
@@ -69,7 +69,9 @@ function hasStatus(record, statuses = []) {
 // within its parent.
 // This essentially corresponds to whether the parent record has been published.
 function isParentAlreadyPublished(record) {
-  return hasStatus(record, [DepositStatus.PUBLISHED]) || record.versions?.index !== 1;
+  return (
+    hasStatus(record, [DepositStatus.PUBLISHED]) || (record.versions?.index ?? 1) !== 1
```
kpsherva: I understand this is dealing with the unpublished subsequent versions of the record (not 1st version)
Can this cause any side effects if we always assume the first version if `record.versions?.index` is undefined or null? 
there is ongoing work to add reviews for each version probably it should be cross-checked if it has any impact @palkerecsenyi @zzacharo

zzacharo: @Ducica what about if you change the line to 
```suggestion
    hasStatus(record, [DepositStatus.PUBLISHED]) || (record.versions?.index > 1)
```

Ducica: Hey, imho, same functionality but done. Thanks

zzacharo: Hi @Ducica I agree technically they do the same thing, but it seems now the code is a bit easier to read :) Thanks a lot!

zzacharo (approved):
@Ducica the tests of python 3.9 are failing, otherr than that it looks good thanks!

Ducica:
> @Ducica the tests of python 3.9 are failing, otherr than that it looks good thanks!

You are welcome, the issue is https://pypi.org/project/commonmeta-py/ released a new version and it is incompatible with python 3.9 according to my research. The tests also fail on main https://github.com/inveniosoftware/invenio-rdm-records/actions/runs/28432786926/job/84251302247

So I think they forgote to put that python should be >=3.10. In other words, I think we would need to pin this dependency to previous version for CI to pass

tmorrell:
Martin is going to do a new commonmeta release tomorrow to fix the issue. So you could merge now with the failed test or wait.

PR #2375, opened 2026-06-30
fix(entity_resolvers): Fix ghost_record for drafts on submission

### LINKED ISSUE #2358: requests: Show record version in community submission request
Since per-record-version requests were added in #2325, a community submission request can correspond to a specific version of a record, not only the first version. Right now, it is not clear in the UI which version it corresponds to.

The only way to see is by going to the "Record" tab and seeing the new version at the bottom of the version list. This is a little bit hidden.

Image

**We should add a more prominent version display**, maybe in the list of metadata fields of the request? I.e. at the bottom here?

Image

sakshamarora1:
For new draft requests, the UI was failing with a bug because for drafts, it returns the ghost record with "id" only, so a check was required to confirm if the version is needed for the ui serielizer or not

fixes: https://github.com/inveniosoftware/invenio-rdm-records/issues/2358
related: https://github.com/inveniosoftware/invenio-rdm-records/pull/2365

### DIFF
--- a/invenio_rdm_records/resources/serializers/ui/schema.py
+++ b/invenio_rdm_records/resources/serializers/ui/schema.py
@@ -97,7 +97,11 @@ def record_version(obj):
     field_data = obj.get("metadata", {}).get("version")
 
     if not field_data:
-        return f"v{obj['versions']['index']}"
+        field_data = obj.get("versions", {}).get("index")
+        if field_data:
+            return f"v{field_data}"
+        # If both metadata.version and versions.index are not available, return None as it is not published yet
+        return None
 
     return field_data
 


invenio_rdm_records/resources/serializers/ui/schema.py
```diff
@@ -94,6 +94,10 @@ def apply_idx(affiliation):
 
 def record_version(obj):
     """Return record's version."""
+    # For drafts, the ghost_record() method is called.
+    if obj.get("is_draft", False):
```
kpsherva: why do we even reach this point with drafts? Draft version is not yet published, so I would like to know which flow leads to here

sakshamarora1: The requests UI serializer with the recent change to show the version of the record (the related issue in description)
https://github.com/inveniosoftware/invenio-rdm-records/blob/87075d4acf5507c21a5ee572f86c946f4d72cce2/invenio_rdm_records/requests/entity_resolvers.py#L64-L70

zzacharo (approved):

PR #2380, opened 2026-07-08
fix(iiif): add handlers for more possible IIIF errors

max-moser:
We get a good number of requests that result in HTTP 500 responses even though they really don't need to.
`Flask-IIIF` is already doing a decent job at raising semantically meaningful errors (including messages [1]), we're just not handling them here.
This PR adds error handlers to basically forward the error messages that `Flask-IIIF` gives us, to the user.

Requires https://github.com/inveniosoftware/flask-iiif/pull/110 for accessing `e.requested_format` and `e.supported_formats` for translations.

[1] `Flask-IIIF` doesn't have translations yet though.

### DIFF
--- a/invenio_rdm_records/resources/iiif.py
+++ b/invenio_rdm_records/resources/iiif.py
@@ -15,7 +15,11 @@ import marshmallow as ma
 import requests
 from flask import Response, current_app, g, request, send_file
 from flask_cors import cross_origin
-from flask_iiif.errors import MultimediaImageNotFound
+from flask_iiif.errors import (
+    MultimediaError,
+    MultimediaImageFormatError,
+    MultimediaImageNotFound,
+)
 from flask_resources import (
     HTTPJSONException,
     JSONSerializer,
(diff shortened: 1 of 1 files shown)


invenio_rdm_records/resources/iiif.py
```diff
@@ -110,6 +114,19 @@ class IIIFResourceConfig(ResourceConfig, ConfiguratorMixin):
                 code=404, description=_("The requested image cannot be found.")
             )
         ),
+        MultimediaImageFormatError: create_error_handler(
+            lambda e: HTTPJSONException(
+                code=400,
+                description=_(
+                    "The requested format '%(format)s' is not supported. Please try one of: %(supported_formats)s",
+                    format=e.requested_format,
```
tmorrell: Ok, I see the variable here now. Now my question is why can't you use the description from the default MultimediaImageFormatError. Or make them consistent?

tmorrell (commented):
Generally supportive of the improvements, just wonder about having duplicate/different errors between the modules

invenio_rdm_records/resources/iiif.py
max-moser: This here is because `Flask-IIIF` doesn't have support for translations (yet), and adding that there is a bit out of scope for now, I'd argue.

Once translations are added there, this can be removed here again.

tmorrell: I think adding translations to Flask-IIF wound't be too bad and might be a better solution. Otherwise translators are going to have to duplicate their work. But I think at a minimum the description should be the same so someone in the future knows they can be removed.

max-moser: Fair points.
Even though I have basically no clue about translations, I trust your opinion on the difficulty of adding support for them in a package.

However, this still feels like a task for another day [1], so I went with the minimal route here :sweat_smile: 
Aligned the text values, and threw in a comment for good measure.

[1] I volunteer to do that soon, just have to check a couple things out for that. I'd argue it's actually a good opportunity to fill a knowledge hole for me.

tmorrell (approved):
Excellent, I approve the plan!

Issue #2386, opened 2026-07-10
External DOIs are blocked when `crossref_prefix` is empty


zubeydecivelek:
@mfenner We received a support ticket from a user who is unable to add an external DOI to a record. The following validation error is displayed:

> The prefix `''` is managed by CDS. Please supply an external DOI or select “No” to have a DOI generated for you.

The issue appears to have been introduced in this [commit](https://github.com/inveniosoftware/invenio-rdm-records/commit/e66b71a3267741b223f4939e539555a23e49620f) 

In the DOI validation logic, `crossref_prefix` is included among the managed prefixes [here](https://github.com/inveniosoftware/invenio-rdm-records/commit/e66b71a3267741b223f4939e539555a23e49620f#diff-638be557f5d538de83a96ebf1b51f30fe0f21f8b5e005ac849b4ca32f3470fa6R577-R584).

Its default value is defined as an empty string [here](https://github.com/inveniosoftware/invenio-rdm-records/commit/e66b71a3267741b223f4939e539555a23e49620f#diff-638be557f5d538de83a96ebf1b51f30fe0f21f8b5e005ac849b4ca32f3470fa6R720-R721)

If this value is not overrided, the empty prefix matches every DOI. As a result, all external DOIs are treated as managed DOIs and rejected.

mfenner:
Thanks for the report, will look into this.

mfenner:
@zubeydecivelek can you help me explain the context when this error occured? Three questions: what is the prefix of the DOI where you see this error? What happens if you set CROSSREF_PREFIX to a real prefix that will never occur, e.g. '10.5555' (a prefix used by Crossref for testing)? And what is the setting for DATACITE_PREFIX? The reason that CROSSREF_PREFIX defaults to '' is that this is also the default for DATACITE_PREFIX and I tried to align the Crossref functionality as much as possible. Setting CROSSREF_PREFIX to '' (and probably also DATACITE_PREFIX) might not be good default values, as they might disable external DOIs.

zubeydecivelek:
@mfenner, here is the additional context:

* **What is the prefix of the DOI where the error occurs?**
  We discovered the issue while trying to submit an external DOI with the prefix `10.11588`. However, when `CROSSREF_PREFIX` is set to an empty string (`""`), the validation blocks external DOIs with any prefix, because the empty string matches all DOI values.

* **What happens when `CROSSREF_PREFIX` is set to a prefix that will never occur?**
[...]

PR #2389, opened 2026-07-14
fix: handle empty CROSSREF_PREFIX and/or DATACITE_PREFIX addresses #2386

mfenner:

### Description

Fixes issue #2386 where an empty CROSSREF_PREFIX or DATACITE_PREFIX disables external DOIs.

### Checklist

- [ ] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/community/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).

**Reminder**

### DIFF
--- a/invenio_rdm_records/services/pids/providers/external.py
+++ b/invenio_rdm_records/services/pids/providers/external.py
@@ -23,7 +23,9 @@ class BlockedPrefixes:
 
     @property
     def prefixes(self):
-        """Get list of blocked prefixes. Can be string, list or dict in config."""
+        """Get list of blocked prefixes. Can be string, list or dict in config.
+
+        Do not include empty strings."""
         _prefixes = []
         for name in self._config_names:
             val = current_app.config[name]
@@ -36,7 +38,7 @@ class BlockedPrefixes:
         return [prefix for prefix in _prefixes + self._prefixes if prefix]
 
     def __call__(self, record, identifier, provider, errors):
-        """Validator call."""
+        """Validator call. Validate the identifier as non-empty and against the blocked prefixes."""
         for p in self.prefixes:
             if p and identifier.startswith(p):
                 errors.append(
--- a/invenio_rdm_records/services/pids/providers/external.py
+++ b/invenio_rdm_records/services/pids/providers/external.py
@@ -33,12 +33,12 @@ class BlockedPrefixes:
                 _prefixes += list(val.values())
             else:
                 _prefixes += val
-        return _prefixes + self._prefixes
+        return [prefix for prefix in _prefixes + self._prefixes if prefix]
 
     def __call__(self, record, identifier, provider, errors):
         """Validator call."""
         for p in self.prefixes:
-            if identifier.startswith(p):
+            if p and identifier.startswith(p):
                 errors.append(
                     _(
                         "The prefix '%(prefix)s' is managed by %(sitename)s. Please supply an external DOI or select 'No' to have a DOI generated for you.",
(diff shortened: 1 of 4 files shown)


tmorrell:
Could you add the CrossRef 5.5 schema changes in the tests in this PR too?

mfenner:
@tmorrell the Crossref 5.5 schema changes are handled in the commonmeta-py dependency (>= 0.256). I can update the Crossref serializer test to show the use of schema 5.5, and one of the new 5.5 features (support for resource_type poster).

invenio_rdm_records/services/pids/providers/external.py
```diff
@@ -33,12 +33,12 @@ def prefixes(self):
                 _prefixes += list(val.values())
             else:
                 _prefixes += val
-        return _prefixes + self._prefixes
+        return [prefix for prefix in _prefixes + self._prefixes if prefix]
```
tmorrell, line 39: I'm struggling a bit to understand the logic here. Why does the addition of self._prefixes depend on the prefix variable?

tmorrell (commented):
Most of the fixes are great, I'm just a bit confused by the one prefixes definition.

invenio_rdm_records/services/pids/providers/external.py
mfenner: only include prefix if not empty to address the issue with CROSSREF_PREFIX = '' disabling external DOIs.

tmorrell: Isn't that the line 41 change? Maybe there just needs to be a comment. self._prefixes would be the default, and then _prefixes are from the config? This whole block is just confusing, which is not your fault but the added code makes it even more difficult to understand.

mfenner: Line 41 addresses the bug as reported. Line 36 removes '' from self._prefixes which might cause unexpeced issues in the future.

pyproject.toml
max-moser, line 1: most of the changes here look like the doings of a code formatter; assuming so, which tool did you use here @mfenner ?

for the `setup.cfg` -> `pyproject.toml` migration, i've used [tombi](https://tombi-toml.github.io/tombi/) for formatting.
@utnapischtim has suggested adding `tombi` as a required linter step for PRs
@fenekku has voiced sentiments against using more tools for such tasks; keyword: tool fatigue

personally, i don't have a strong opinion on this one; `tombi` was literally just the first google search result and made my life easier for the `hatchling` migration

---

my perspective: this doesn't have to be solved now [1], but it looks like this topic is going to come up sooner or later naturally.
maybe a strong candidate for a future maintainers call.

[1] i'm fine with any changes here right now; there's not exactly a rich history behind this file so far that would be obscured with formatting changes anyway.

invenio_rdm_records/services/pids/providers/external.py
max-moser: basically this here makes sure that `blocked_prefixes.prefixes` only gives out non-empty strings in its list; this might be called from wherever.
L41 is extra defensive and strictly speaking not necessary with this line here anymore, because due to this line here, the list from `self.prefixes` will never contain a falsy value anyway.

so the extra check in L41 could be removed for the tiniest bit of performance gain?

either way is fine by me, but i agree that a comment might make reading this piece a bit easier for future developers

max-moser:
```suggestion
        # Avoid empty values that could come from being passed to the constructor,
        # or from the app config (e.g. `app.config["CROSSREF_PREFIX"] = ""`)
        return [prefix for prefix in _prefixes + self._prefixes if prefix]
```

max-moser (approved):
The changes look reasonable to me; my comments aren't necessarily blockers.

invenio_rdm_records/services/pids/providers/external.py
mfenner: @max-moser @tmorrell Added two comments explaining the two new checks.

PR #2391, opened 2026-07-16
fix(tests): chase changes between crossref schema v5.4.0 -> v5.5.0

max-moser:
Looks like the CrossRef serialization now gives us output according to a new schema version.
Ping @mfenner


tmorrell:
Fix is included in https://github.com/inveniosoftware/invenio-rdm-records/pull/2389. If you could give an additional review to that we can get it merged

max-moser:
Thanks, didn't see that!
Closing in favor of that one.

PR #2409, opened 2026-07-27
feat(review): review each rec. version class

ntarocco:
This new class will be used as example in the documentation.

### DIFF
--- a/invenio_rdm_records/services/review/policy.py
+++ b/invenio_rdm_records/services/review/policy.py
@@ -25,3 +25,16 @@ class NewRecordVersionReviewPolicy:
         a community review.
         """
         return False
+
+
+class AllRecordVersionsReviewPolicy:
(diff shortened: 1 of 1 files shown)


ntarocco:
to be ported to `master` after merging it

tmorrell (approved):
Makes sense to me. I'd consider baking in the community-specific version too, though it's also fine if we just leave that in the docs.

palkerecsenyi (approved):
This looks good to me and I agree with adding it, since the default behaviour is still kept. Would you like us to mention this class in the documentation?

zzacharo (approved):

ntarocco:
> This looks good to me and I agree with adding it, since the default behaviour is still kept. Would you like us to mention this class in the documentation?

It is already done, thank you!

PR #2416, opened 2026-07-30
fix(quota): don't cast record id to int

slint:
* The topic record id is a PID value, which is a string for any instance
  using the default alphanumeric provider. Only repositories that mint
  numeric recids (e.g. Zenodo) could accept a quota increase request.

### DIFF
--- a/invenio_rdm_records/requests/quota_increase.py
+++ b/invenio_rdm_records/requests/quota_increase.py
@@ -38,8 +38,8 @@ class AcceptAction(actions.AcceptAction):
 
     def execute(self, identity, uow, **kwargs):
         """Apply the quota increase."""
-        DRAFT = int(self.request.get("topic", {}).get("record"))
-        QUOTA_SIZE = int(self.request.get("payload", {}).get("quota_size"))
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/requests/quota_increase.py
```diff
@@ -38,7 +38,7 @@ class AcceptAction(actions.AcceptAction):
 
     def execute(self, identity, uow, **kwargs):
         """Apply the quota increase."""
-        DRAFT = int(self.request.get("topic", {}).get("record"))
+        DRAFT = self.request.get("topic", {}).get("record")
```
ntarocco: Do we want to/Can we have a `None` value here?
```suggestion
        DRAFT = self.request["topic"]["record"]
```

Same below.

slint: Agree, didn't want to touch functionality, but we should be more "assertive" when accessing values that have already been validated on the datamodel (or marshmallow schema).

I'll change it.

utnapischtim (approved):

PR #2418, opened 2026-07-30
feat(quota): implement policy evaluation for allowing users

ntarocco:
- refactor the request policy evaluators to enable the evaluation if a user is allowed only, before evaluating the policies themselves
- allow controlling if a feature is enabled or not depending on the logged in user
- closes https://github.com/inveniosoftware/invenio-app-rdm/issues/3534

### DIFF
--- a/invenio_rdm_records/services/request_policies.py
+++ b/invenio_rdm_records/services/request_policies.py
@@ -22,12 +22,12 @@ class BasePolicy:
     description: str
     tombstone_description: str
 
-    def is_allowed(self, identity, record):
-        """Whether the identity is allowed to delete the record."""
+    def is_allowed(self, identity, record=None):
+        """Whether the identity can perform an action, optionally on a given record."""
         raise NotImplementedError
 
     def evaluate(self, identity, record):
-        """Whether the record meets the conditions to be deleted."""
+        """Whether the record meets the conditions to perform the action."""
         raise NotImplementedError
 
     @property
@@ -51,8 +51,10 @@ class GracePeriodPolicy(BasePolicy):
             "Record owners can delete their records within {grace_period} days of publishing."
         ).format(grace_period=grace_period.days)
 
-    def is_allowed(self, identity, record):
-        """Whether the identity is allowed to delete the record."""
+    def is_allowed(self, identity, record=None):
+        """Whether the identity is allowed to perform the action."""
+        if record is None:
+            return False
         is_record_owner = identity.user.id == record.parent.access.owned_by.owner_id
         return is_record_owner
 
@@ -79,8 +81,10 @@ class RequestDeletionPolicy(BasePolicy):
             "published for more than {grace_period} days."
         ).format(grace_period=grace_period.days)
 
-    def is_allowed(self, identity, record):
+    def is_allowed(self, identity, record=None):
         """Whether the identity is allowed to delete the record."""
+        if record is None:
+            return False
         is_record_owner = identity.user.id == record.parent.access.owned_by.owner_id
         return is_record_owner
 
@@ -101,8 +105,10 @@ class FileModificationGracePeriodPolicy(BasePolicy):
             "You can edit the files of your records within {grace_period} days of publishing."
         ).format(grace_period=grace_period.days)
 
-    def is_allowed(self, identity, record):
+    def is_allowed(self, identity, record=None):
         """Whether the identity is allowed to modify files."""
+        if record is None:
+            return False
         is_record_owner = identity.user.id == record.parent.access.owned_by.owner_id
         return is_record_owner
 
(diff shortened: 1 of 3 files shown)


invenio_rdm_records/services/request_policies.py
```diff
@@ -22,12 +22,12 @@ class BasePolicy:
     description: str
     tombstone_description: str
 
-    def is_allowed(self, identity, record):
-        """Whether the identity is allowed to delete the record."""
+    def is_allowed(self, identity, record=None):
```
ntarocco, line 25: This is a breaking change

invenio_rdm_records/services/request_policies.py
```diff
@@ -195,16 +202,31 @@ class PolicyEvaluator:
...
+        """Result object."""
 
         enabled: bool
-        valid_user: bool = False  # so we can show the button as disabled
+        valid_user: bool = False
         allowed: bool = False
         policy: Optional[BasePolicy] = None
 
+    @classmethod
+    def evaluate_allowed(cls, enabled, policy_config, identity):
```
ntarocco, line 213: This is the real change: it evaluates the `is_allowed`, without evaluating the policy.
This method can be used to toggle a feature depending on the attributes of the current logged in user, for example if the user is verified or not.

slint (approved):
LGTM, and agree that given the feature is unreleased, we should be ok with the breaking change.

invenio_rdm_records/services/request_policies.py
max-moser:
```suggestion
    def evaluate_allowed(cls, enabled_config_name, policy_config_name, identity):
```
`enabled` sounds a *lot* like a boolean flag, which confused me for a minute here
a type hint might also help with reducing the confusion, but that's not a requirement

on another note, regarding the API design:
wouldn't it make sense to have the `QuotaIncreaseAdminPolicy` have knowledge about its own config items?
is there a use case where we supply the same policy class with different config names, and if so maybe we should use subclassing in such cases instead?

invenio_rdm_records/views.py
```diff
@@ -163,9 +161,8 @@ def _format_storage(data):
 @login_required
 def storage_settings():
     """User storage page."""
-    if not current_app.config.get(
-        "RDM_IMMEDIATE_QUOTA_INCREASE_ENABLED", False
-    ) or not getattr(current_user, "verified_at", None):
+    policy = current_app.config.get("RDM_QUOTA_INCREASE_POLICY")
+    if not policy or not policy.evaluate_identity(g.identity).valid_user:
```
max-moser, line 165: i suppose the `verified_at` check is now responsibility of the actual policy to implement?
improvement for flexibility.

max-moser (commented):
since the PR is targeting the `maint-32.x` branch, i suppose this is just a backport?
in that case, it might be too late about my remarks regarding API design and naming 😅

ntarocco:
> since the PR is targeting the `maint-32.x` branch, i suppose this is just a backport? in that case, it might be too late about my remarks regarding API design and naming 😅

It is not a backport. I have targeted first maint-32.x to be sure v14 is fixed, as it is the priority.

invenio_rdm_records/views.py
ntarocco: Yes indeed

ntarocco:
@max-moser regarding https://github.com/inveniosoftware/invenio-rdm-records/pull/2418#discussion_r3682827117

I wanted to apply the min. amount of changes, including the args names.

> wouldn't it make sense to have the QuotaIncreaseAdminPolicy have knowledge about its own config items?

Yes, I had the same thought! And after starting the refactoring, I noticed that there are some subclasses that they have more/different config items. Again, I wanted to keep changes to the min. given that we need to release.

A larger internal refactoring can be implemented later if we want. If you don't mind, I would keep things as they are.

max-moser (approved):
Sure, I've created an issue for a follow-up refactoring here: https://github.com/inveniosoftware/invenio-rdm-records/issues/2420

invenio_rdm_records/services/request_policies.py
```diff
@@ -195,16 +202,31 @@ class PolicyEvaluator:
...
+        """Evaluate the identity is allowed to perform the action."""
+        result = cls.Result(current_app.config[enabled])
+        if not result.enabled:
+            return result
+
+        policies = current_app.config[policy_config]
+
+        for policy in policies:
+            if policy.is_allowed(identity):
```
utnapischtim, line 222: nit: you could do `if policy.is_allowed(identity, record=None)` to be backwards compatible. to remove the `=None` necessity in the `is_allowed` method definition.

utnapischtim (approved):

PR #2424, opened 2026-08-04
fix(uppy): handle upload failures in the files list

corovcam:

### Description

Uploads that failed could get stuck in the files list table, with no way of removing their entry, and were left behind in Uppy's state:

- Deleting an entry of an upload that failed to be initialized crashed on its missing links, so the entry could never be cleared.
- Uploads aborted from another browser tab kept reporting an error that the user could not act upon, leaving the Dashboard in a failed state that had to be reset manually.
- Removing an entry of an upload that failed to be finalized left its file in Uppy's state, re-uploading it whenever any other file was added.

Files removed through the files list table are now removed from Uppy's state as well (aborting their ongoing requests), while deletions of files that no longer exist on the backend are treated as successful.

Failures of uploads that were aborted on the backend are told apart from the other failures by checking whether their file still exists, and are discarded silently instead of being reported. Finalization failures of single-part uploads are newly reported to the user too, consistently with the multi-part ones.

#### Verified scenarios

Manually verified against a local instance with S3 (MinIO) storage, on top of the unit tests:

| Scenario | Expected | Result |
| --- | --- | --- |
| Single-part and multi-part upload | Commits, leaves the Dashboard, entry shows 100% and its checksum | ✅ |
| Upload initialization fails | Error reported, entry clearable without a crash, Dashboard resets | ✅ |
| Finalization fails (multi-part and single-part) | Error reported, entry clearable, Dashboard resets, cleared file is not re-uploaded with the next one | ✅ |
| Upload aborted from another tab | No error reported, requests aborted, entry removed, Dashboard resets | ✅ |
| Removing a file that failed to be finalized | Deleted both in Invenio and in the storage (its S3 multipart upload gets aborted) | ✅ |

> [!NOTE]
> Running the JS tests installs `node_modules` into `invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/`. Since instance assets are symlinked to this directory, the installed copies of `react`/`react-dom` are picked up by the bundler and break the deposit form with an "Invalid hook call" error. Remove the directory before rebuilding the assets.

### Checklist

- [x] I've marked [translation strings](https://inveniordm.docs.cern.ch/community/translations/i18n/).
- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/community/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).

**Frontend**

- [x] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/community/code/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/community/code/best-practices/react/) guidelines.
- [x] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/) guidelines.
- [x] I've followed the [user interface](https://inveniordm.docs.cern.ch/community/code/best-practices/ui/) guidelines.

**Reminder**


invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositFilesService.js
```diff
@@ -219,6 +223,10 @@ export class RDMDepositFilesService extends DepositFilesService {
     await this._startNextUpload();
   };
 
+  get = async (fileLinks) => {
```
mirekys: just `get` is confusing (unclear if that means get file contents, or metadata, or upload status,...), better declare the same `getFile` or maybe even better `getFileMetadata`

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
```diff
@@ -108,30 +124,129 @@ export class RDMUppyUploaderPlugin extends AwsS3Multipart {
...
+   * all of its ongoing requests and drops its files list entry
+   * (see `#onFileRemoved`).
+   */
+  #discardRemotelyAbortedUpload = (file) => {
+    this.#remotelyAbortedFiles.add(file.id);
+
+    if (this.uppy.getFile(file.id)) {
+      this.uppy.removeFile(file.id);
+    }
+    this.uppy.hideInfo();
```
mirekys: No guarantee this hides the correct toast when multiple displayed. I'd remove, toasts disappear after 10s either way...

mirekys:
When aborting failed upload (file count limit set to 2), there's now a visual flicker glitch (quota reached overlay shows briefly, suggesting aborted file goes through successful state for a brief period?):
https://github.com/user-attachments/assets/a0acf8a3-a120-47a9-a7cd-1573bf989426

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/files.js
mirekys, line 104: THis also deserves `isDeletedUpload` guard

mirekys: else a commit response racing a delete resurrects a ghost row that can never be deleted.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
```diff
@@ -108,30 +124,129 @@ export class RDMUppyUploaderPlugin extends AwsS3Multipart {
...
+    return !(await this.opts.checkFileExists(currentFile));
+  };
+
+  /**
+   * Discards uploads that were aborted on the backend, letting all the other
+   * failures be reported to the user. Returns the error to be re-thrown by the caller.
+   */
+  #handleUploadFailure = async (file, error) => {
+    if (await this.#isUploadAbortedRemotely(file)) {
+      this.#discardRemotelyAbortedUpload(file);
```
mirekys: Rather than making the upload magically disappear to user, i'd report an error simply saying the upload has been aborted

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
```diff
@@ -108,30 +124,129 @@ export class RDMUppyUploaderPlugin extends AwsS3Multipart {
...
+    try {
+      await this.completeMultipartUpload(file);
+      // Multi-part uploads are already committed when reported as successful,
+      // so only single-part uploads are removed from Uppy state here.
+      this.uppy.removeFile(file.id);
+    } catch (error) {
+      // Unlike for multi-part uploads, finalization of single-part uploads
+      // happens outside of Uppy's upload flow, so failures have to be
+      // reported to it explicitly.
+      this.uppy.emit("upload-error", this.uppy.getFile(file.id) ?? file, error);
```
mirekys, line 267: this error reporting  i would hoist to `this.#handleUploadFailure`

mirekys: whole try catch block seems kinda redundant then

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
```diff
@@ -28,6 +38,8 @@ const defaultOptions = {
 
 export class RDMUppyUploaderPlugin extends AwsS3Multipart {
   #maxMultipartParts = 10_000;
+  // Uploads that no longer exist on the backend, so their failures are not reported
```
mirekys: They should be reported, as mentioned in previous comment (toast notification saying something like upload aborted - pls check uppy or invenio translation bundle for any existing message)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
```diff
@@ -12,6 +12,15 @@ import { humanReadableBytes } from "react-invenio-forms";
 
 import { FileSizeError, InvalidPartNumberError, SignedUrlExpiredError } from "./error";
 
+/**
+ * Reasons passed to `uppy.removeFile()` to distinguish who initiated the removal.
+ */
+export const FileRemovalReason = {
+  // File was removed through the files list table, which deletes
+  // the file on the backend on its own.
+  filesListEntryDeleted: "files-list-entry-deleted",
```
mirekys: maybe use better naming - deleted by backend (Files list calls backend api to remove it, other REST API clients could potentially do the same thing...)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
```diff
@@ -108,30 +124,129 @@ export class RDMUppyUploaderPlugin extends AwsS3Multipart {
     this.opts.setUploadProgress(file, percentage);
   };
 
-  #onFileRemoved = (file) => {
-    if (!file.progress.uploadComplete) {
+  #onFileAdded = (file) => {
```
mirekys: This callback won't be needed (see above)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
```diff
@@ -108,30 +124,129 @@ export class RDMUppyUploaderPlugin extends AwsS3Multipart {
...
+  #onFileRemoved = (file, reason) => {
+    const isDeletedByFilesList = reason === FileRemovalReason.filesListEntryDeleted;
+    // Uploads that failed to be committed have all their bytes uploaded,
+    // but are still to be cleaned up (both in Invenio and in the storage).
+    const isUnfinishedUpload = !file.progress.uploadComplete || Boolean(file.error);
+
+    if (!isDeletedByFilesList && isUnfinishedUpload) {
       file.links = file.meta.links;
-      this.opts.abortUpload(file);
+      Promise.resolve(this.opts.abortUpload(file)).catch((error) =>
```
mirekys: async/await? To keep consistent with the rest

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
```diff
@@ -108,30 +124,129 @@ export class RDMUppyUploaderPlugin extends AwsS3Multipart {
...
+  /**
+   * Silently discards an upload that no longer exists on the backend, e.g. when
+   * its file was deleted from another browser tab.
+   *
+   * Such an upload can neither be resumed nor retried and the user has already
+   * asked for it to be gone, so no error is reported. Removing the file aborts
+   * all of its ongoing requests and drops its files list entry
+   * (see `#onFileRemoved`).
+   */
+  #discardRemotelyAbortedUpload = (file) => {
```
mirekys: Rather than making the upload magically disappear to user, i'd report an error simply saying the upload has been aborted (see previous comments)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
```diff
@@ -108,30 +124,129 @@ export class RDMUppyUploaderPlugin extends AwsS3Multipart {
...
+      return false;
+    }
+    return !(await this.opts.checkFileExists(currentFile));
+  };
+
+  /**
+   * Discards uploads that were aborted on the backend, letting all the other
+   * failures be reported to the user. Returns the error to be re-thrown by the caller.
+   */
+  #handleUploadFailure = async (file, error) => {
```
mirekys: This callback kinda duplicates `#onUploadError` callback, i'd merge into one

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/UppyUploader.js
```diff
@@ -70,6 +70,15 @@ const createDuplicateFileChecker = (uppy, filesList) => {
   };
 };
 
+const findUppyFile = (uppy, filename) =>
```
mirekys: nit: findUppyFileByName

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/actions/files.js
```diff
@@ -137,30 +143,65 @@ export const finalizeUpload = (commitFileUrl, file) => {
...
+    const fileLinks = getFileLinks(file);
+
+    if (!fileLinks?.self) {
+      // Upload was never initialized, so nothing exists on the backend
+      return false;
+    }
+
+    try {
+      await config.service.files.get(fileLinks);
+      return true;
```
mirekys: check service call result rather than assuming success

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/actions/files.js
```diff
@@ -137,30 +143,65 @@ export const finalizeUpload = (commitFileUrl, file) => {
...
+    } catch (error) {
+      if (isFileGone(error)) {
+        return false;
+      }
+      // Any other failure (e.g. a network issue) is inconclusive,
+      // thus the file is assumed to still exist.
+      console.error("Error fetching file", error, file);
+      return true;
```
mirekys: Rather then potentially falsely reporting true, rethrow the error

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositFilesService.js
corovcam: Renamed to `getFileMetadata`

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
corovcam: True, it always drops the head of the queue. Removed. Aborted uploads now get their own message instead, and the generic one Uppy queued is taken back by matching its text, so only one toast is shown.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/files.js
corovcam: I moved the guard in front of the `switch` and listed the action types it applies to

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
corovcam: Used `uppy.info(i18next.t("Upload of {{file}} was aborted."), "error")`. I didn't find any existing translation string. Cleanup and clearing `state.error` stay as they were.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
corovcam: Used `uppy.info(i18next.t("Upload of {{file}} was aborted."), "error")`. I didn't find any existing translation string. Cleanup and clearing `state.error` stay as they were.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
corovcam: Used `uppy.info(i18next.t("Upload of {{file}} was aborted."), "error")`. I didn't find any existing translation string. Cleanup and clearing `state.error` stay as they were.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
corovcam: I removed the `handleUploadFailure` and its try/catch.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
corovcam: Renamed to `deletedOnBackend`

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
corovcam: Gone, along with the `#remotelyAbortedFiles`

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
corovcam: Done

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
corovcam: Merged, `#onUploadError` is the only function now

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/UppyUploader.js
corovcam: Renamed

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/actions/files.js
corovcam: Now returns `Boolean(fileMetadata?.key)`, so an unexpected 2xx body counts as "not there"

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/actions/files.js
corovcam: Changed. `checkFileExists` now rethrows when it doesn't know. The calling function logs it and treats the failure as any other upload error

corovcam:
> When aborting failed upload (file count limit set to 2), there's now a visual flicker glitch (quota reached overlay shows briefly, suggesting aborted file goes through successful state for a brief period?): https://github.com/user-attachments/assets/a0acf8a3-a120-47a9-a7cd-1573bf989426

The thing is, the aborted file's FilesList entry only goes away once the DELETE returns, and in between the overlay sees an empty uploader with a still-full quota. Fixed by counting deletions in the files reducer and keeping the overlay hidden while any are pending. It's counted in the `deleteFile` action.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/files.js
```diff
@@ -31,10 +32,30 @@ export const UploadState = {
...
   // Filename needs to be normalised due to encoding differences between client and server.
   const remoteFileName = action.payload?.filename?.normalize() ?? "";
+  // Deletions in flight, tracked to tell an emptied uploader apart from one
+  // whose files are still being deleted.
+  const pendingDeletions = state.pendingDeletions ?? 0;
+
+  // Late updates of an already deleted upload (e.g. a commit response racing
+  // its deletion) must not resurrect its files list entry, as such an entry
+  // could never be deleted again.
+  if (ENTRY_UPDATE_ACTIONS.includes(action.type) && !state.entries?.[remoteFileName]) {
```
mirekys, line 55: For `FILE_UPLOAD_INITIALIZED`, there might be no entry yet and it's correct to add it. WOuldn't this break its functionality? Please double check this for both single-part/multi-part flow with both Uppy & legacy simple uploaders

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/UppyUploader.js
```diff
@@ -257,6 +269,19 @@ export const UppyUploaderComponent = ({
     });
   }, [uppy, filesList, filesSize, quota, decimalSizeDisplay]);
 
+  const deleteFilesListEntry = React.useCallback(
+    async (file) => {
+      const uppyFile = findUppyFileByName(uppy, file.name);
+      if (uppyFile) {
+        // Aborts any ongoing requests of the file and keeps the Dashboard in sync,
+        // deletion of the file itself is done below.
+        uppy.removeFile(uppyFile.id, FileRemovalReason.deletedOnBackend);
```
Samk13, line 278: Question: could this leave Uppy and the files list out of sync if the backend deletion fails? 
We remove/abort the Uppy file before `deleteFile(file)` succeeds, while FILE_DELETE_FAILED keeps the Redux entry.

Should we consider removing the Uppy file only after backend deletion succeeds, or separate aborting active requests from removing the Uppy state?

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
```diff
@@ -108,30 +121,151 @@ export class RDMUppyUploaderPlugin extends AwsS3Multipart {
...
     }
   };
 
-  #removeFileOnSuccess = (file) => {
+  /**
+   * Handles all upload failures, whether reported by Uppy's upload flow or
+   * emitted by this plugin (see `#completeSinglePartUpload`).
+   */
+  #onUploadError = async (file) => {
+    if (await this.#isUploadAbortedRemotely(file)) {
```
Samk13, line 149: Could this async existence check race with a retry/reinitialization?
e.g: An older failed attempt could return "gone" after a new attempt has started and then remove the current Uppy file.

Maybe we should verify that the file/link identity is still the same before calling `#discardAbortedUpload()`, and add a retry-race test?
e.g:
```js
if (file && (await this.#isUploadAbortedRemotely(file))) {
  this.#discardAbortedUpload(file);
}
```

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/actions/files.js
```diff
@@ -137,30 +144,68 @@ export const finalizeUpload = (commitFileUrl, file) => {
...
+        filename: file.name,
+      },
+    };
+
+    dispatch({ type: FILE_DELETE_STARTED });
+
+    if (!fileLinks?.self) {
+      // Upload initialization failed, so there is nothing to delete on the
+      // backend and the file entry can be removed from the state right away.
+      dispatch(deletedSuccess);
```
Samk13, line 193: What happens when deleting a file that is still queued in `RDMDepositFilesService` and has not been initialized yet?
Could this remove the Redux entry without cancelling the queued upload, allowing it to initialize and upload later while the reducer ignores its subsequent upload actions?

Samk13 (commented):
Thanks for addressing this! We had a report of this issue before, but we weren't able to reproduce it at the time. 
I added a few comments/questions on the changes.

Samk13:
Related indirectly to inveniosoftware/invenio-app-rdm#3278 
this seems to cover some of the frontend lifecycle behavior around uploads disappearing remotely, although #3278 itself was fixed on the S3 backend side.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/reducers/files.js
corovcam: Both places that dispatch `FILE_UPLOAD_INITIALIZED` dispatch `FILE_UPLOAD_ADDED` for the same file first (`initializeFileUpload` for Uppy, `upload()` for the legacy uploader), so the entry is there unless it was deleted in between (which is fixed now, see https://github.com/inveniosoftware/invenio-rdm-records/pull/2424#discussion_r3795255951

Also tried it live: Single-/multi-part + legacy - all commit with a checksum. The legacy uploader has no multi-part flow, it always does local transfers.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/UppyUploader.js
corovcam: The order is intentional: `removeFile()` aborts the file's in-flight requests, so the DELETE can't race a part upload or a commit that would recreate the file. Uppy has no way to abort without removing, so I couldn't make it more separated. On failure the entry stays in the files list and can be deleted again. I made it so it now shows "{file} could not be deleted."

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/fields/UppyUploader/RDMUppyUploaderPlugin.js
corovcam: Agreed, thanks. The guard is on the file re-read from Uppy's state after the check (the handler's `file` is the event payload, always passes the guard). A retry clears `file.error`, so a retried or removed file is left alone.

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/state/actions/files.js
corovcam: Yes, you are right, I tested this scenario. Deleting a file before its upload was initialized only removed the entry and the upload went ahead without UI changes.
But it is a bit bigger than just the queue. Fixed: `deleteFile` now cancels the upload when the file has no links yet, and the service handles both cases: a queued upload is dropped, an initializing one is discarded when its initialization finishes. Running uploads show Cancel instead of the trash, so they're not affected.

corovcam:
> Related indirectly to [inveniosoftware/invenio-app-rdm#3278](https://github.com/inveniosoftware/invenio-app-rdm/issues/3278) this seems to cover some of the frontend lifecycle behavior around uploads disappearing remotely, although #3278 itself was fixed on the S3 backend side.

Thanks for the link. This PR only solves the frontend side once an upload has disappeared, it doesn't depend on the backend fix.

corovcam:
For #2456, this PR also covers it.

tmorrell:
Thanks! Could you make sure all the JS tests pass?

corovcam:
> Thanks! Could you make sure all the JS tests pass?

Yes, I pushed the fix for test `actions/files.test.js` in commit https://github.com/inveniosoftware/invenio-rdm-records/pull/2424/changes/df3fc1018e02c13cd6eb40ac69a9383a4a0e7268. Along with new tests in subsequent commit. Tested locally. Unfortunately, I cannot run CI workflow manually here.

Samk13 (approved):

tmorrell:
I want to do a full local test for this, since it makes a bunch of changes. I'm also thinking we might want this to only be available on 14.1 (and obviously v15)

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositFilesService.js
```diff
@@ -96,6 +126,14 @@ export class DepositFilesService {
     throw new Error("Not implemented.");
   }
 
+  /**
+   * Cancels an upload that hasn't started yet, returning whether it was found.
+   * A no-op by default, as services without an upload queue have nothing to cancel.
+   */
+  cancelQueuedUpload(fileName) {
```
mirekys: Where is this used? If needed, should just throw new Error("Not implemented."); as other abstract methods, leave implementation to specific class

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositFilesService.js
```diff
@@ -36,12 +37,41 @@ class UploaderQueue {
...
+      this.cancelled.add(startingUpload);
+      return true;
+    }
+
+    return false;
+  }
+
+  // One-shot check, the cancellation is forgotten once read.
+  isCancelled(file) {
+    return this.cancelled.delete(file);
```
mirekys: check like this shouldn't have any unexpected side effects

invenio_rdm_records/assets/semantic-ui/js/invenio_rdm_records/src/deposit/api/DepositFilesService.js
```diff
@@ -163,10 +205,29 @@ export class RDMDepositFilesService extends DepositFilesService {
     this._startNextUpload();
   };
 
+  // Deletes the backend entry of an upload cancelled during its initialization.
+  _discardCancelledUpload = async (file, fileLinks) => {
+    // Free the slot first, the cleanup must not hold up the queued uploads.
+    this.uploaderQueue.markCompleted(file);
```
mirekys: What if delete fails on 215? This marks it as complete, though the entry becomes not deleted and not complete

PR #2427, opened 2026-08-07
fix(quota): exclude deleted records from remaining_storage

mairasalazar:
:heart: Thank you for your review!

### Description

`remaining_storage()` queries RDMRecordQuota to calculate the available remaining storage for a draft and user. The query includes quota for records even if they have no published versions (only deleted).

This commit filters out the cases where a record has no published versions, so the remaining storage reflects only active records.

### Checklist

**Reminder**

### DIFF
--- a/invenio_rdm_records/alembic/c2d3e4f5a6b7_add_parent_id_index.py
+++ b/invenio_rdm_records/alembic/c2d3e4f5a6b7_add_parent_id_index.py
@@ -1,0 +1,42 @@
+# SPDX-FileCopyrightText: 2026 CERN.
+# SPDX-License-Identifier: MIT
+
+"""Add parent_id index."""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "c2d3e4f5a6b7"
+down_revision = "1780576627"
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    op.create_index(
+        op.f("ix_rdm_records_metadata_parent_id"),
+        "rdm_records_metadata",
+        ["parent_id"],
+        unique=False,
+    )
+
+    op.create_index(
+        op.f("ix_rdm_drafts_metadata_parent_id"),
+        "rdm_drafts_metadata",
+        ["parent_id"],
+        unique=False,
+    )
+
+
+def downgrade():
+    """Downgrade database."""
+    op.drop_index(
+        op.f("ix_rdm_records_metadata_parent_id"),
+        table_name="rdm_records_metadata",
+    )
+
+    op.drop_index(
+        op.f("ix_rdm_drafts_metadata_parent_id"),
+        table_name="rdm_drafts_metadata",
+    )
--- a/invenio_rdm_records/services/storage/service.py
+++ b/invenio_rdm_records/services/storage/service.py
@@ -12,13 +12,16 @@ from invenio_accounts.models import User
 from invenio_db import db
 from invenio_files_rest.models import Bucket
 from invenio_search.engine import dsl
-from sqlalchemy import func
+from sqlalchemy import func, or_
 
 from invenio_rdm_records.records.models import (
     RDMDraftMetadata,
     RDMRecordMetadata,
     RDMRecordQuota,
     RDMUserQuota,
+)
+from invenio_rdm_records.records.systemfields.deletion_status import (
+    RecordDeletionStatusEnum,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


invenio_rdm_records/services/storage/service.py
```diff
@@ -107,13 +110,19 @@ def max_additional_quota_value(self, user_id, record=None):
 
     def remaining_storage(self, user_id, record):
         """Remaining storage for this draft and user."""
+        has_active_version = exists().where(
+            RDMRecordMetadata.parent_id == RDMRecordQuota.parent_id,
+            RDMRecordMetadata.deletion_status
+            == RecordDeletionStatusEnum.PUBLISHED.value,
```
slint, line 116: major: two bamboozles here:

- `RDMRecordMetadata.parent_id` doesn't have an index AFAIK... funnily in Zenodo we do have an index though (see SQL result from the prod DB below). We might need to add this to InvenioRDM (which also means an Alembic recipe 😬).
- You still need to filter in this `EXISTS (...)` clause by  `RDMRecordMetadata.parent_id == record.parent.id`, or else this checks if **any** row in this table satisfies the condition of having a published record (which most often is the case 😅).

```sql
zenodo_read@dbod-zenodo-prod.cern.ch:zenodo> \d rdm_records_metadata;
+-----------------+-----------------------------+-----------+
| Column          | Type                        | Modifiers |
|-----------------+-----------------------------+-----------|
| created         | timestamp without time zone |  not null |
| updated         | timestamp without time zone |  not null |
| id              | uuid                        |  not null |
| json            | jsonb                       |           |
| version_id      | integer                     |  not null |
| index           | integer                     |           |
| bucket_id       | uuid                        |           |
| media_bucket_id | uuid                        |           |
| deletion_status | character varying(1)        |  not null |
| parent_id       | uuid                        |           |
+-----------------+-----------------------------+-----------+
Indexes:
    "pk_rdm_records_metadata" PRIMARY KEY, btree (id)
    "idx_rdm_records_metadata_bucket_id" btree (bucket_id)
    "idx_rdm_records_metadata_parent_id" btree (parent_id)
    "ix_rdm_records_metadata_bucket_id" btree (bucket_id)
    "ix_rdm_records_metadata_media_bucket_id" btree (media_bucket_id)
Foreign-key constraints:
    "fk_rdm_records_metadata_bucket_id_files_bucket" FOREIGN KEY (bucket_id) REFERENCES files_bucket(id)
    "fk_rdm_records_metadata_media_bucket_id_files_bucket" FOREIGN KEY (media_bucket_id) REFERENCES files_bucket(id)
    "fk_rdm_records_metadata_parent_id_rdm_parents_metadata" FOREIGN KEY (parent_id) REFERENCES rdm_parents_metadata(id) ON DELETE RESTRICT
Referenced by:
    TABLE "rdm_records_files" CONSTRAINT "fk_rdm_records_files_record_id_rdm_records_metadata" FOREIGN KEY (record_id) REFERENCES rdm_records_metadata(id) ON DELETE RESTRICT
    TABLE "rdm_records_media_files" CONSTRAINT "fk_rdm_records_media_files_record_id_rdm_records_metadata" FOREIGN KEY (record_id) REFERENCES rdm_records_metadata(id) ON DELETE RESTRICT
    TABLE "rdm_versions_state" CONSTRAINT "fk_rdm_versions_state_latest_id_rdm_records_metadata" FOREIGN KEY (latest_id) REFERENCES rdm_records_metadata(id)
```

slint (commented):
bit of a blocker on the query 😮‍💨

tmorrell:
I'm also not sure that we want to exclude draft records from the quota calculation. Excluding specifically deleted records is fine, but this seems to be excluding everything that isn't published.

slint:
> I'm also not sure that we want to exclude draft records from the quota calculation. Excluding specifically deleted records is fine, but this seems to be excluding everything that isn't published.

We discussed briefly IRL and it seems that the logic should be to filter for the "negation", i.e. explicit reasons for why to exclude some quota increase from the total. The main case is deleted records (all of their versions) without an unpublished new version draft (which is an edge case to cover for).

invenio_rdm_records/services/storage/service.py
mairasalazar: I pushed a new version but actually the first filter didn't change much from my previous version. I think that, since the outer query is already filtered by user_id, it only iterates over that user's quota rows. Then for each of those outer quota rows (since it is a correlated query), the subquery checks rdm_records_metadata for rows with that same parent_id and looks at if at least one of those records is published or there is a draft. Maybe I'm missing something though 🤔 

I ended up not filtering out the negation because those queries became more verbose and more complex (using many negations or using group by + having + count)

invenio_rdm_records/alembic/c2d3e4f5a6b7_add_parent_id_index.py
```diff
@@ -0,0 +1,42 @@
...
+    """Upgrade database."""
+    op.create_index(
+        op.f("ix_rdm_records_metadata_parent_id"),
+        "rdm_records_metadata",
+        ["parent_id"],
+        unique=False,
+    )
+
+    op.create_index(
+        op.f("ix_rdm_drafts_metadata_parent_id"),
```
mairasalazar, line 25: Also added an index for drafts since in Zenodo we also have it

invenio_rdm_records/services/storage/service.py
slint: Ok, I did a refresher on subqueries/correlations and it makes sense now 😅

I feel that we need tests though for this one, because it took me a hot minute to understand the logic, and seeing the test fixtures + desired state would probably help.

PR #2435, opened 2026-08-20
fix(signposting): fix custom licenses links percent-encoding

ptamarit:
How to reproduce the problem:
* Add a custom license to a record with the link `https://customlicense.org/licenses/testé/1.0/?testé=testé` 
* Save the record
* The rendering of the landing page will fail with `TypeError: http header must be encodable in latin1` in `invenio_app_rdm_records.record_detail`

### DIFF
--- a/invenio_rdm_records/resources/serializers/signposting/schema.py
+++ b/invenio_rdm_records/resources/serializers/signposting/schema.py
@@ -6,6 +6,7 @@
 """Signposting schemas."""
 
 import re
+import urllib
 
 import idutils
 from invenio_base import invenio_url_for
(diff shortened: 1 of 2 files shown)


invenio_rdm_records/resources/serializers/signposting/schema.py
```diff
@@ -139,8 +140,16 @@ def extract_link(right):
...
                 return right["props"].get("url")
 
+        def percent_encode(url):
+            url_parts = urllib.parse.urlsplit(url)
+            url_parts = url_parts._replace(path=urllib.parse.quote(url_parts.path))
+            url_parts = url_parts._replace(
+                query=urllib.parse.urlencode(urllib.parse.parse_qsl(url_parts.query))
+            )
+            return url_parts.geturl()
+
```
ptamarit, line 153: I couldn't find a simpler way to percent-encode the URL as per the [link header specs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Link#uri-reference), but maybe there is a better way to do it?

fenekku:
That slipped throught the cracks. I will take a look this week.

invenio_rdm_records/resources/serializers/signposting/schema.py
fenekku: After investigation, I went for this approach that doesn't rely on private `_replace` and is a bit clearer I find (through importing the urllib.parse functions we need directly to reduce visual clutter).

Overall makes sense.

fenekku (approved):
Thanks for finding that. I refactored the code to avoid using a private method and reduce the `urllib.parse` repetition.

I will let you merge and backport!

ptamarit:
Backported to InvenioRDM v14 (invenio-rdm-records v32.x) in #2465

PR #2436, opened 2026-08-20
fix(pids): keep concept DOI hidden after last deletion

slint:
> **AI disclosure:** I wrote the one-line fix to `else:` the part that always ran before. But for updating the tests I a) made sure they fail after my change, b) had Codex update them, and c) went through the changed test cases to see that they still make sense.

* Avoid queuing a parent PID metadata update after the final published version is deleted because DataCite treats the update as a publish and reverses the hide operation.

### DIFF
--- a/tests/services/pids/test_pids_tasks.py
+++ b/tests/services/pids/test_pids_tasks.py
@@ -498,46 +498,18 @@ def test_invalidate_pid(
     assert parent_pid.status == PIDStatus.REGISTERED
 
     tombstone_info = {"note": "no specific reason, tbh"}
+    hide_call_count = mock_datacite_client.api.hide_doi.call_count
+    update_call_count = mock_datacite_client.api.update_doi.call_count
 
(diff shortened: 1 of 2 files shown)


tests/services/pids/test_pids_tasks.py
```diff
@@ -498,46 +498,18 @@ def test_invalidate_pid(
...
-                    "doi": parent_doi,
-                    "publicationYear": "2020",
-                    "publisher": {"name": "Acme Inc"},
-                },
-                doi=parent_doi,
-                url=f"https://127.0.0.1:5000/doi/{parent_doi}",
-            ),
-        ]
-    )
+    assert mock_datacite_client.api.hide_doi.call_args_list[hide_call_count:] == [
```
slint, line 508: After deleting the last vesion, we should only have two hide calls for the last version DOI and the parent DOI

kpsherva (approved):

PR #2464, opened 2026-09-18
fix(github): resolve badge DOI in 1 query

ciurana-life:
* `RDMGithubRelease.badge_value` was doing 110 SQL statements per badge before, now 1.

### Description

`/badge/<github_id>.svg` was doing to many sql queries to get the object, this endpoint has become problematic because of this. 

Now we use [this call](https://invenio-records.readthedocs.io/en/latest/_modules/invenio_records/api.html#Record.get_record) instead.

Can be tested in zenodo setting up a Github release with:
https://github.com/zenodo/zenodo-rdm/pull/1455

### Checklist

- [x] I've identified the [copyright holder(s)](https://inveniordm.docs.cern.ch/community/copyright-policy/) and updated copyright headers for touched files (>15 lines contributions).

**Frontend**

- [ ] I've followed the [CSS/JS](https://inveniordm.docs.cern.ch/community/code/best-practices/css-js/) and [React](https://inveniordm.docs.cern.ch/community/code/best-practices/react/) guidelines.
- [ ] I've followed the [web accessibility](https://inveniordm.docs.cern.ch/community/code/best-practices/accessibility/) guidelines.
- [ ] I've followed the [user interface](https://inveniordm.docs.cern.ch/community/code/best-practices/ui/) guidelines.

**Reminder**


invenio_rdm_records/services/github/release.py
```diff
@@ -240,8 +241,23 @@ def badge_title(self):
         if current_app.config.get("DATACITE_ENABLED"):
             return "DOI"
 
+    def resolve_record_obj(self):
```
ptamarit: Since this method is very similar to `resolve_record`, I would be tempted to place it just next to it.
I would also add an underscore prefix ~~and mention in the docstring that no access rights checks are performed~~.

Edit: actually we use `system_identity` for `resolve_record`, so there are also no access rights checks.

tests/services/test_github_release.py
```diff
@@ -0,0 +1,70 @@
...
+    db.session.add_all([repo, release])
+    db.session.commit()
+    # Session refresh here prevents the `count_queries` from
+    # counting this as part of the test.
+    db.session.refresh(release)
+    return release
+
+
+def test_bade_value_query_count(db, github_release, published_record):
+    """Regression test: `basge_value` stays at one query."""
```
ptamarit: Little typo on `badge` here.

tests/services/test_github_release.py
```diff
@@ -0,0 +1,70 @@
...
+    doi = published_record["pids"]["doi"]["identifier"]
+    with count_queries(db) as statements:
```
ptamarit, line 62: Very nice, I don't think we do any assertions on number of statements anywhere else.

invenio_rdm_records/services/github/release.py
```diff
@@ -240,8 +241,23 @@ def badge_title(self):
...
+            return None
+        try:
+            return current_rdm_records_service.record_cls.get_record(
+                self.release_object.record_id
+            )
+        except NoResultFound:
+            return None
+
     @property
     def badge_value(self):
```
ptamarit, line 256: I think we could apply this change to the `badge_value` method in the `vcs` directory now.

ptamarit (approved):

invenio_rdm_records/services/vcs/release.py
```diff
@@ -185,6 +185,17 @@ def resolve_record(self):
             db.session.commit()
             return published_record
 
+    def _resolve_record_obj(self):
```
palkerecsenyi, line 188: Why not change the existing `resolve_record` method instead of adding a new one? It feels like this is duplicating the work of `resolve_record`, which is already linked to from Invenio VCS itself. The idea is that it should be a generic method that could in theory be implemented by a non-RDM consumer of Invenio VCS.

Maybe I am misunderstanding this of course and there is a good reason for it so no worries if so, just wanted to check

utnapischtim: i wanted to write the same question, but the difference i see is that `resolve_record` uses `read` which returns a `ResultItem` which build the links too.  `get_record` doesn't do that. the `links` are used e.g. [here](https://github.com/inveniosoftware/invenio-rdm-records/blob/7cf78b308e9545ac1df981ed9ef1944453c64349/invenio_rdm_records/services/github/release.py#L233)

ptamarit: Indeed, the "full" object is needed for `record_url` and `serialize_record`, but retrieving the "full" object is very expensive (one example in our instance does 50+ DB connections, and take a bit more than 0.5 seconds to complete).

This PR proposes to resolve a lighter version of the object for the `badge_value`, where we just need the DOI.

The new method is underscore-prefixed because it's an implementation detail (chosing to not rely on the full `resolve_record` method).

palkerecsenyi (approved):

